PR #4, opened 2014-06-21
LOG4J2-673 – plugin preloading fails in shaded jar files

michaelsembwever:
Support for plugin preloading through the standard javax.annotation.processing tool was adding in LOG4J2-595

But the plugin processor always creates and stores the processed "Plugin" annotated classes into the same "Log4j2Plugins.dat" file. This works fine when the classpath consists of individual jar files, but fails when shaded jar files are used.

Add a ResourceTransformer that can be used in conjunction with the maven-shaded-plugin to merge multiple Log4j2Plugins.dat into one.


jvz:
Shit, I closed this by accident. Meant to close #5.

PR #9, opened 2015-04-06
[LOG4J2-991] Fixed an issue that the async root logger configuration was...

ryanrupp:
... not defaulting include location to false.

This could happen if async loggers were configured without the use of the logger context system property as the logic from LoggerConfig.includeLocation was being used instead of AsyncLoggerConfig.includeLocation


ryanrupp:
Fixed by 3cee912e394f6b6eca7d9913b60884027c92fd1b

PR #11, opened 2015-04-22
LOG4J2-997: Fix addFilter() and removeFilter() in AbstractFilterable.class

mayt:
Accessing the filter after addition/removal should follow the following logic:
* If there are no Filter, return null
* If there is one Filter, return the Filter
* If there are more than one Filters, return a CompositeFilter of the Filters


remkop:
https://issues.apache.org/jira/browse/LOG4J2-997 Has been addressed some time ago, closing this PR.

PR #12, opened 2015-07-21
LOG4J2-524: autodelete older files while rollover 

nbartels:
This pull request adds the possibility to configure a `maxAge` in days. The original work is done by @arun0009.

If no `maxAge` is given (or `maxAge <= 0`),  the `DefaultRolloverStrategy` works as before. 

ATM I have some problems writing a test, because the timespan is at least one day :smile: Hopefully I get some input to provide one.


nbartels:
I have to take this back, because there are some major issues with the log-file detection. I'm atm not able to find the old files corresponding to the `filePattern`.

benji:
See also https://issues.apache.org/jira/browse/LOG4J2-435 where some strategies to find the old log files are discussed.

PR #19, opened 2015-11-03
LOG4J2-649: Purging dynamically created routing sub-appenders

frendos:
1. Added PurgePolicy interface
2. Added IdlePurgePolicy which is purging sub-appenders that were not in use last X time units
3. Added ability to access sub-appenders, so custom purging controlled
outside log4j2 also possible.


frendos:
Changes merged to master

PR #21, opened 2016-01-06
Correct Marker example

britter:
- Add MarkerManager to list of imports
- use `setParents` methods to set parent marker


jvz:
Applied to master, fix is in for the 2.6 site update.

PR #22, opened 2016-01-11
how to log client ip in tcp socket server

jarvan4dev:
Hello developers

I am trying migrate log4j1.x to 2, and I have a socketAppender, in log4j1.x the socketServer is easily override, but in the 2nd one, the TcpSocketServer is not so, so many class members are protected or private, so it is really difficult. Oh, I am sorry, I forgot to come up with my need, I need log the client ip or hostname via socket server, just like this:

2016-01-12 02:09:09,234 [thread-exec-8] 120.24.176.190 com.zhb.service.StuedentService.getIns(19)-age: 19, name:Eason

Could you give me a hand, thanks so much.


remkop:
Web applications usually put the username and client IP address in the ThreadContext map, and in the pattern layout use %X{username} etc to log the values. 
You'll need to set these values at the beginning of each request and remove them when you finished handling the request. 

The log4j 1 equivalent was MDC (or NDC). 

If this is not what you need please explain more about your use case. 

Remko

jvz:
Not exactly sure what happened to this pull request.

jvz:
I'm going to close this due to inactivity in fixing whatever is wrong here. Go ahead and reopen this with an update if this is still relevant.

PR #25, opened 2016-03-01
LOG4J2-1301 added optionalMonitorFiles attribute for automatic reconf…

alanachtenberg:
…iguration

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/xml/XmlConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/xml/XmlConfiguration.java
@@ -16,10 +16,7 @@
  */
 package org.apache.logging.log4j.core.config.xml;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -67,7 +64,7 @@ public class XmlConfiguration extends AbstractConfiguration implements Reconfigu
             "http://apache.org/xml/features/xinclude/fixup-language";
     private static final String XINCLUDE_FIXUP_BASE_URIS =
             "http://apache.org/xml/features/xinclude/fixup-base-uris";
-    private static final String[] VERBOSE_CLASSES = new String[] {ResolverUtil.class.getName()};
+    private static final String[] VERBOSE_CLASSES = new String[]{ResolverUtil.class.getName()};
     private static final String LOG4J_XSD = "Log4j-config.xsd";
 
     private final List<Status> status = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


garydgregory:
I'm not sure I agree with this approach. Give me a few days to look around.

alanachtenberg:
I agree it would be nice to automatically pick up xincluded files; It would good if it was a "just works" solution for the end user without having to select files to monitor, but that solution would not be very portable to the other file formats. If specifying a list of files is deemed acceptable this would be easy to reproduce for json and probably YAML(I have no experience with it).

Also being able to specify the files gives the user alot more power. In my case what if I only want to monitor the log4j-xinclude-loggers.xml(changed often) for changes and not the appenders(changed rarely, usually with a new release if at all) to reduce overhead with the checks. Secondly a user could enable and disable monitoring of files at anytime since the main configuration will always be monitored. Lastly I think there could be some good use cases for monitoring non log4j2 files for changes (Especially when you consider extensions or plugins of log4j2) . For example an application that uses a single generic configuration/properties file(client accessible) for application wide configuration could set some properties to be used in the logging system initialization.

I am a happy camper either way if i can get reconfiguration based on the xincluded files :)

Thanks for the feedback and let me know if you find anything with regards to a "java 8 way". I looked into where the xinclude is processed and it seemed it would be difficult to retrieve the files without hacking up the xml code. However, I have relatively little experience and you might see a better solution than me.

-Alan

garydgregory:
We are on Java 7 for now so we cannot use Java 8 features unless we do so in a way is not used on Java 7.

The bigger picture item is: If we invented a Log4j include mechanism that works across all of our configuration formats (XML, JSON, YAML), then we would not need a new watchFiles attribute for this used case because we would know which files to watch. We could even enable/disable watching on a per-include basis:

<Include monitorIntervalSeconds="60">file://...</Include>

This monitorIntervalSeconds would override the value set on the Configuration element.

Sorry, I am getting side tracked... I need to bring this up on the dev ML.

alanachtenberg:
On the retrieval of xinclude files for the automatic watching route.

I was trying out this neat little library for xml . https://github.com/jOOQ/jOOX 

From their wiki.

>  jOOX stands for Java Object Oriented XML. It is a simple wrapper for the org.w3c.dom package, to allow for fluent XML document creation and manipulation where DOM is required but too verbose. jOOX only wraps the underlying document and can be used to enhance DOM, not as an alternative.

Usage as follows works in my simple test program.
`Document document = $(xmlFile).document();
            Match includeElements  = $(document).find("include");
            for (Element element : includeElements){
                String includeFilePath = element.getAttribute("href");
            }`

I can retrieve the include elements because the builder they use does not have xinclude enabled by default. Therefore the include elements are not processed by the builder. Which made me realize if we wanted to retrieve the included files we could parse the file for them before enabling xinclude. It would require parsing the file twice(once to get the file names, and once for actual configuration) which is not ideal however.

mikaelstaldal:
Linking: https://issues.apache.org/jira/browse/LOG4J2-1301

rgoers:
In light of LOG4J2-494 and LOG4J2-913 this PR now has conflicts that would have to be resolved. This functionality should be achievable through the mechanisms implemented in those, although it doesn't specifically apply to XInclude.

rgoers:
Closing this since it has been replaced with Composite Configurations.

PR #26, opened 2016-03-03
LOG4J2-1227 - Fix unit test (error in commit 33ee4bfd0)

olivierlemasle:
Thank you for having merged my pull request #20.
But there was an error in unit tests, sorry!


jvz:
Thanks for the fix.

PR #28, opened 2016-04-21
[LOG4J2-621] Pattern support to drop elements from front in logger name

lkamal:
To drop the parts from logger or caller's name starting from the front, I have implemented a new class named org.apache.logging.log4j.core.pattern.NameAbbreviator.DropFirstElementsAbbreviator which will analyse the name and drop from the start of the name using "." separator. If a path does not contain enough path elements to drop, nothing will be dropped. 

I have implemented test cases for the existing feature where the parts are retained from the end of the names, and also implemented test cases for the new feature where the parts are dropped from the front of the names.


garydgregory:
It seems to me that overloading the current %c pattern would be more elegant. For example, today you can do:

%c{3} 
input: org.apache.commons.Foo
output: apache.commons.Foo

But you could also implement using negative numbers:

%c{-3} 
input: org.apache.commons.Foo
output: Foo

Thoughts?

lkamal:
Hi @garydgregory 

I am sorry, I did not understood your suggestion correctly. Let me try to elaborate a bit on the approach I took, so that you could point me in correct direction further. 

Exactly, current implementation supports below.
```
%c{3} 
input: org.apache.commons.Foo
output: apache.commons.Foo
```

Above is done via the LoggerPatternConverter class which internally uses the NameAbbreviator.getAbbreviator(final String pattern) to get an instance of a suitable NameAbbreviator (in this case MaxElementAbbreviator).

And %c{-3} is not yet supported.

So I tried to implement this feature by updating the NameAbbreviator.getAbbreviator(final String pattern) method to return another new instance of NameAbbreviator named DropFirstElementsAbbreviator which supports negative numbers in the "pattern" parameter.

With that the below is supported.
```
%c{-3} 
input: org.apache.commons.Foo
output: Foo
```

Could you please provide some thoughts based on this?

Thanks a lot.

lkamal:
Hi @garydgregory 

I looked at that further and I think you were suggesting to use the MaxElementAbbreviator itself to support both positive and negative numbers; so I will be updating in that approach. 

Please review that as well and provide feedback.

Thanks

lkamal:
Hi @garydgregory 

I have updated the code [#28] so that both NameAbbreviator.getAbbreviator(final String pattern) and NameAbbreviator.MaxElementAbbreviator support both negative and positive integers.

Also verified that the test cases are still passing 100%.

Thanks

garydgregory:
Great. Would you mind updating the site docs as well? I'll take a look at it all tonight hopefully.

lkamal:
Hi @garydgregory 

I have updated the site doc to reflect this change.

Thanks

lkamal:
Updated site content looks as below.

log4j2-621_site_update

garydgregory:
Thank you,

I should be able to this tomorrow.

Gary

On Wed, Apr 27, 2016 at 11:23 AM, Kamal Mettananda <notifications@github.com
> wrote:

> Updated site content looks as below.
>
> [image: log4j2-621_site_update]
> <https://cloud.githubusercontent.com/assets/1687033/14863352/fd6a3b86-0cd2-11e6-937a-0e1cf6f2d120.JPG>
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/logging-log4j2/pull/28#issuecomment-215181472>
>



-- 
E-Mail: garydgregory@gmail.com | ggregory@apache.org
Java Persistence with Hibernate, Second Edition
<http://www.manning.com/bauer3/>
JUnit in Action, Second Edition <http://www.manning.com/tahchiev/>
Spring Batch in Action <http://www.manning.com/templier/>
Blog: http://garygregory.wordpress.com
Home: http://garygregory.com/
Tweet! http://twitter.com/GaryGregory

PR #32, opened 2016-06-25
LOG4J2-1448 - added host parameter to FlumeAppender

kelaban:


kelaban:
This patch changes the interface of FlumeAppender which would affect anyone who may configure it in code. Is there away to overload the `@PluginFactory` annotation? Another possibility would be to switch FlumeAppender to have a Builder implementation and keep the old static `createAppender`, removing `@PluginFactory`, and adding a deprecation warning for any programatic configuration. Any preference or rule against changing the interface?

jvz:
What I do when switching to a Builder is mark the factory method deprecated, remove the annotations from it, and add annotations all to the builder instead. Either way, the builder class is used before the factory method is, so just having a builder class is enough to update the config. I do recommend going the deprecation route, however.

kelaban:
Updated PR based on suggestions from @jvz

garydgregory:
@jvz, do you have time to validate the new version of the patch?

jvz:
Yeah I can look at it this weekend unless someone else looks at it today.

PR #33, opened 2016-07-07
GELF Custom Level Compatibility

julienCrine:
For send custom level to GELF.


julienCrine:
GELF Custom Level Compatibility .
For send custom level to GELF.

PR #34, opened 2016-07-23
LOG4J2-1341 GC-free HighlightConverter and StyleConverter

rzschech:
GC-Free HighlightConverter and StyleConverter

### DIFF
--- a/log4j-core/src/test/resources/log4j-style.xml
+++ b/log4j-core/src/test/resources/log4j-style.xml
@@ -20,7 +20,7 @@
   <Appenders>
     <List name="List">
        <PatternLayout>
-         <Pattern>%d %highlight{%p} %style{%logger}{bright,cyan} %C{1.} %msg%n</Pattern>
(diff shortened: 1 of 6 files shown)


remkop:
Thanks! I will take a look at this in the next few days.

log4j-core/src/test/resources/log4j-style.xml
```diff
@@ -20,7 +20,7 @@
   <Appenders>
     <List name="List">
        <PatternLayout>
-         <Pattern>%d %highlight{%p} %style{%logger}{bright,cyan} %C{1.} %msg%n</Pattern>
+         <Pattern>%d %highlight{%p} %style{%logger}{bright,cyan}%style{}{bright,cyan} %C{1.} %msg%n</Pattern>
```
rzschech: This change is to verify that the no text appended within the style case.

garydgregory:
We should make sure that these test fixtures are not being overloaded. One config file and test should ideally test one thing IMO.

PR #35, opened 2016-07-23
LOG4J2-1488 Add support for period in ISO8601 format

rzschech:
Add support for period in ISO8601 format as per the documentation: "millisecond separator may be either a comma ',' or a period '.'"

Also Java's `java.time.Instant.toString()` and JavaScript's `new Date().toISOString()` use '.' as well.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
@@ -75,6 +75,11 @@ public class FixedDateFormat {
         ISO8601_BASIC("yyyyMMdd'T'HHmmss,SSS", "yyyyMMdd'T'", 2, ' ', 0, ',', 1),
 
         /**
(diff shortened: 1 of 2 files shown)


garydgregory:
How about a unit tests? ;-)

remkop:
PR looks good at first glance. Unit test would be nice. 

Can you link this pull request to another Jira? 
LOG4J2-1341 is the ticket for GC-free pattern converters which is not related to FixedDateFormat. Feel free to create a new Jira ticket if no matching one exists.

rzschech:
This change is related to GC-free patterns as it allows DatePatternConverter to support a period in the ISO8601 format. Without this change the DatePatterConverter successfully formats using "yyyy-MM-dd'T'HH:mm:ss.SSS" but it doesn't do it using the GC-free FixedDateFormat.

If you wish though I can create a more specific Jira.

rzschech:
Opps didn't mean to "Comment and close"

remkop:
If you don't mind, more specific Jira tickets would be good: in our release notes we always include a detailed breakdown of what was changed and separate Jiras help with that.

rzschech:
I've created a Jira, added a unit test and updated the pull request.

PR #36, opened 2016-07-23
LOG4J2-1489 Add time zone support to FixedDateFormat

rzschech:
Add time zone support to FixedDateFormat


remkop:
Can you link this pull request to another Jira? 
LOG4J2-1341 is the ticket for GC-free pattern converters which is not related to FixedDateFormat. Feel free to create a new Jira ticket if no matching one exists.

remkop:
Is this related to [LOG4J2-1246](https://issues.apache.org/jira/browse/LOG4J2-1246)?

rzschech:
This change is related to GC-free patterns as it allows DatePatternConverter to support a timezone parameter and remain GC-free. Before this change the DatePatternConverter falls back to the non-GC-free FastDateFormat if there is a timezone parameter.

This is not related to LOG4J2-1246 as it is not outputting the timezone information. It is just outputting the date/time relative to the timezone parameter rather than the default timezone.

rzschech:
I've created a Jira LOG4J2-1489 and updated the pull request.

PR #41, opened 2016-08-26
Documentation bug for purgeAscending method.

challarao:
For "@param highIndex  Log file associated with high index will be deleted if needed. "
That is incorrect it seems, that applies to purgeDescending.

That message should be under lowIndex with message "@param lowIndex. Log file associated with low index will be deleted if needed. "

Because purgeAscending deletes the one with the low index if required, if I'm not wrong.


challarao:
Thanks for accepting pull request.

PR #42, opened 2016-09-02
[LOG4J2-1559] Check proper value in LevelRangeFilter

andreyp1986:
https://issues.apache.org/jira/browse/LOG4J2-1559

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/filter/LevelRangeFilter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/filter/LevelRangeFilter.java
@@ -61,7 +61,7 @@ public final class LevelRangeFilter extends AbstractFilter {
             @PluginAttribute("onMismatch") final Result mismatch) {
             // @formatter:on
         final Level actualMinLevel = minLevel == null ? Level.ERROR : minLevel;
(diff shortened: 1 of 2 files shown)


garydgregory:
Thank you for your report.
Can you also provide a unit test the demonstrates the problem? IOW, it should fail without your patch.

andreyp1986:
NP, I'll add a test in org.apache.logging.log4j.core.filter.LevelRangeFilterTest.
Is it possible to somehow edit current pull request to add unit test commit?

jvz:
If you want to keep it in a single commit, make your changes, then do `git commit --amend` and `git push --force`. There's also a way to do that using `git rebase -i`, but I find that overkill for this sort of thing.

andreyp1986:
BTW, LevelRangeFilter default range values are wrong now.
```
        final Level actualMinLevel = minLevel == null ? Level.ERROR : minLevel;
        final Level actualMaxLevel = minLevel == null ? Level.ERROR : maxLevel;
```
According to log4j 1.2, default range is full range.
https://logging.apache.org/log4j/extras/apidocs/org/apache/log4j/filter/LevelRangeFilter.html
IE, better range should be OFF->ALL.

andreyp1986:
Pushed a unit test.

PR #43, opened 2016-09-22
Adds targetNamespace in log4j-config.xsd

shubhankar1100:
Adds targetNamespace in log4j-config.xsd , As there is no targetNamespace for log4j2-config which lead to failure of xml validation.


garydgregory:
Thank you for the PR! Merged in Git master.

PR #44, opened 2016-09-27
Update appenders.xml

shubham391:
Fix "filePattern" description


garydgregory:
Thank you for your report. Diff applied to Git master.

PR #46, opened 2016-10-05
add configurable syslog layout

doumdoum:
Implements configurable syslog message as sketched in:
https://issues.apache.org/jira/browse/LOG4J2-969.


garydgregory:
If the new layout implements a builder instead of a factory method, this will avoid deprecating the factory method in the future if anything needs to change in the layout. At least, that how I would do it.

doumdoum:
OK on your remarks. 
I've simplified implementation: everything is done now in SyslogLayout.
A child Layout (such as PatternLayout) can be  specified, thus removing the need to specify layout properties in Factory.

doumdoum:
ping ?

cbarbaza:
Hi,

This feature is really important for me, what is the status of the pull request ?

Thanks.

garydgregory:
Hi,

Does it apply cleanly on top of Git master now? Does a full build pass all tests?

Thank you,
Gary [away from PC]

garydgregory:
@cbarbaza, @doumdoum,

Are you available to update this patch?

Gary

rgoers:
LOG4J2-969 was closed. Is this patch still necessary?

mikaelstaldal:
Can this PR be closed @garydgregory ?

garydgregory:
Sure, since https://issues.apache.org/jira/browse/LOG4J2-969 is closed.

PR #48, opened 2016-11-11
LOG4J2-1692: Add putAll() / pushAll() methods to CloseableThreadContext

GregDThomas:
This provides a patch to implement CloseableThreadContext.putAll() and CloseableThreadContext.pushAll()

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
@@ -62,8 +62,8 @@ public class CloseableThreadContext {
     }
 
     /**
-     * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
-     * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original values when
+     * Populates the Thread Context Map with the supplied key/value pair. Any existing key in the
+     * {@link ThreadContext} will be replaced with the supplied value, and restored back to their original value when
      * the instance is closed.
      *
      * @param key   The  key to be added
@@ -72,6 +72,31 @@ public class CloseableThreadContext {
      */
     public static CloseableThreadContext.Instance put(final String key, final String value) {
         return new CloseableThreadContext.Instance().put(key, value);
+    }
+
+    /**
+     * Populates the Thread Context Stack with the supplied stack. The information will be popped off when
+     * the instance is closed.
+     *
+     * @param values The stack of values to be added
+     * @return a new instance that will back out the changes when closed.
+     * @since 2.7.1
+     */
+    public static CloseableThreadContext.Instance pushAll(final ThreadContext.ContextStack stack) {
+        return new CloseableThreadContext.Instance().pushAll(stack);
+    }
+
+    /**
+     * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
+     * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original value when
+     * the instance is closed.
+     *
+     * @param values The map of key/value pairs to be added
+     * @return a new instance that will back out the changes when closed.
+     * @since 2.7.1
+     */
+    public static CloseableThreadContext.Instance putAll(final Map<String, String> values) {
+        return new CloseableThreadContext.Instance().putAll(values);
     }
 
     public static class Instance implements AutoCloseable {
@@ -110,13 +135,13 @@ public class CloseableThreadContext {
         }
 
         /**
-         * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
-         * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original values when
+         * Populates the Thread Context Map with the supplied key/value pair. Any existing key in the
+         * {@link ThreadContext} will be replaced with the supplied value, and restored back to their original value when
          * the instance is closed.
          *
          * @param key   The  key to be added
          * @param value The value to be added
-         * @return the instance that will back out the changes when closed.
+         * @return a new instance that will back out the changes when closed.
          */
         public Instance put(final String key, final String value) {
             // If there are no existing values, a null will be stored as an old value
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -74,6 +74,31 @@ private CloseableThreadContext() {
         return new CloseableThreadContext.Instance().put(key, value);
     }
 
+    /**
+     * Populates the Thread Context Stack with the supplied stack. The information will be popped off when
+     * the instance is closed.
+     *
+     * @param values The stack of values to be added
+     * @return a new instance that will back out the changes when closed.
+     * @since 2.7.1
```
mikaelstaldal: The next version will be 2.8.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -74,6 +74,31 @@ private CloseableThreadContext() {
...
+    }
+
+    /**
+     * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
+     * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original value when
+     * the instance is closed.
+     *
+     * @param values The map of key/value pairs to be added
+     * @return a new instance that will back out the changes when closed.
+     * @since 2.7.1
```
mikaelstaldal: The next version will be 2.8.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -128,6 +153,37 @@ public Instance put(final String key, final String value) {
         }
 
         /**
+         * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
+         * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original value when
+         * the instance is closed.
+         *
+         * @param values The map of key/value pairs to be added
+         * @return a new instance that will back out the changes when closed.
+         * @since 2.7.1
```
mikaelstaldal: The next version will be 2.8.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -128,6 +153,37 @@ public Instance put(final String key, final String value) {
...
+            return this;
+        }
+
+        /**
+         * Populates the Thread Context Stack with the supplied stack. The information will be popped off when
+         * the instance is closed.
+         *
+         * @param values The stack of values to be added
+         * @return a new instance that will back out the changes when closed.
+         * @since 2.7.1
```
mikaelstaldal: The next version will be 2.8.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -128,6 +153,37 @@ public Instance put(final String key, final String value) {
...
         /**
+         * Populates the Thread Context Map with the supplied key/value pairs. Any existing keys in the
+         * {@link ThreadContext} will be replaced with the supplied values, and restored back to their original value when
+         * the instance is closed.
+         *
+         * @param values The map of key/value pairs to be added
+         * @return a new instance that will back out the changes when closed.
+         * @since 2.7.1
+         */
+        public Instance putAll(final Map<String, String> values) {
```
mikaelstaldal: You should use `ThreadContext.putAll()` here.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -74,6 +74,31 @@ private CloseableThreadContext() {
...
 
+    /**
+     * Populates the Thread Context Stack with the supplied stack. The information will be popped off when
+     * the instance is closed.
+     *
+     * @param values The stack of values to be added
+     * @return a new instance that will back out the changes when closed.
+     * @since 2.7.1
+     */
+    public static CloseableThreadContext.Instance pushAll(final ThreadContext.ContextStack stack) {
```
mikaelstaldal: It would be more useful to have the parameter as a `List<String>` since it will be a bit tricky to construct a `ThreadContext.ContextStack`.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
mikaelstaldal: Also consider using `ThreadContext.removeAll()` that was recently added as part of [LOG4J2-1689](https://issues.apache.org/jira/browse/LOG4J2-1689).

src/site/xdoc/manual/thread-context.xml
```diff
@@ -26,164 +26,183 @@
...
-            Stack and then be identified by their order in the Stack or by the data itself. Since key/value
-            pairs are more flexible, the Thread Context Map is recommended when data items may be added during
-            the processing of the request or when there are more than one or two items.
-          </p>
-           <p>To uniquely stamp each request using the Thread Context Stack, the user pushes contextual information
-             on to the Stack.
-           </p>
-            <pre class="prettyprint linenums">
-ThreadContext.push(UUID.randomUUID().toString()); // Add the fishtag;
+        <section name="Log4j 2 API">
```
mikaelstaldal: Can you revert the formatting changes? It makes it hard to review the actual changes.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
GregDThomas: Aha, took the version from the SNAPSHOT, will update this and others.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
GregDThomas: I don't think that's sensible here;  the `CloseableThreadContext` will backup any existing values and restore them (or remove them, if no existing values) when closed. So even if we do use `ThreadContext.putAll()` it will still be necessary to iterate over the `Map`.

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
GregDThomas: I was anticipating that the usage will be simply `ThreadContext.getImmutableStack()` but as you say `List` is probably easier (and more closely aligns with the `Map` used for `putAll(...)`).

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
mikaelstaldal: Using `ThreadContext.putAll()` still makes sense since it will do an atomic update of `ThreadContext`.

mikaelstaldal: The atomicity does not cover map and stack together, only the context map. So to correct myself: `ThreadContext.putAll()` will do an atomic update of the context map.

It is also about performance since the context map is often implemented in a copy-on-write fashion.

GregDThomas: OK, I'm not personally convinced, but will make the appropriate changes.

mikaelstaldal:
Looks good.

PR #49, opened 2016-12-10
LOG4J2-1731 SslSocketManager should respect connectTimeoutMillis

chrisribble:


chrisribble:
It looks like this test failed, but all tests - including this one - passed on my machine.

TlsSyslogAppenderTest>SyslogAppenderTest.testUDPStructuredAppender:86->SyslogAppenderTestBase.sendAndCheckStructuredMessage:84->SyslogAppenderTestBase.checkTheNumberOfSentAndReceivedMessages:103 The number of received messages should be equal with the number of sent messages expected:<1> but was:<0>

This is my first time contributing to log4j2, so I'm not really sure what the next step is.

PR #50, opened 2016-12-13
[LOG4J2-1620] Add sync.send configuration property to Kafka Appender …

thedutchy:
…to support asynchronous sends

This change is backwards compatible, as the property is set to "true" by default. The implementation mirrors the implementation of the KafkaLog4jAppender in the Apache Kafka project (which is however based on log4j instead of log4j2).

The asynchronous send is relevant, as it significantly improves throughput to the Kafka broker. On my machine, setting the property to false, lead to a throughput improvement of approximately 50x.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaManager.java
@@ -19,11 +19,13 @@ package org.apache.logging.log4j.core.appender.mom.kafka;
 
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.kafka.clients.producer.Producer;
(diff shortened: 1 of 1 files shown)


mikaelstaldal:
The documentation in `src/site/xdoc/manual/appenders.xml` needs to be updated.

mikaelstaldal:
This PR is about LOG4J2-1733, right?

thedutchy:
Yes, it is. I don't know why I entered 1620.

thedutchy:
After going through the documentation again, I decided the proper place for the parameter is not in the properties. Instead it should go in the appender parameters. I will close this pull request and create a new one, this way I can also correct the PR to 1733.

garydgregory:
I am in the process of updating this appender to the builder pattern. You
might want to wait until I am done.

Gary

thedutchy:
Thanks for the message, I will wait until you're finished. Should I simply checkin to master after you're done?

Vincent

garydgregory:
This patch does not feel right to me because the configuration properties array is used to carry this new setting. These properties are meant to be used only for Kafka properties https://kafka.apache.org/documentation.html#producerconfigs

IMO, what needs to happen is convert the appender to using a builder and then add a new setting. 

The build changes are now in Git master. See Apache JIRA LOG4J2-1739.

PR #51, opened 2016-12-14
[LOG4J2-1733] Add syncSend configuration property to Kafka Appender

thedutchy:
… to support asynchronous sends

This change is backwards compatible, as the property is set to "true" by default. The implementation mirrors the implementation of the KafkaLog4jAppender in the Apache Kafka project (which is however based on log4j instead of log4j2).

The asynchronous send is relevant, as it significantly improves throughput to the Kafka broker. On my machine, setting the property to false, lead to a throughput improvement of approximately 50x.


garydgregory:
Thank you for your contribution. Keep 'em coming! ;-)

PR #53, opened 2016-12-26
Adds xmlns in schema and some tags

shubhankar1100:
Adds xmlns in schema and adds elements like Appender, console, PatternLayout and remove  mandatory elements type from Properties and ThresholdFilters as they are not mandatory always.

### DIFF
--- a/log4j-core/src/main/resources/Log4j-config.xsd
+++ b/log4j-core/src/main/resources/Log4j-config.xsd
@@ -16,7 +16,7 @@
  limitations under the License.
 
 -->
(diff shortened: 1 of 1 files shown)


shubhankar1100:
Hi, 

Can anyone review this patch?

garydgregory:
Did you do a full build with this patch like "mvn clean install" ?

garydgregory:
In Git master. Local build OK with 'mvn clean install'. This will be in out upcoming 2.8 release.

Thank you for your contribution. Keep them coming! :-)

shubhankar1100:
Yes, I only did "mvn install" not maven clean install

PR #57, opened 2017-02-02
[LOG4J2-1799]: Error determining the current charset

edwgiz:
https://issues.apache.org/jira/browse/LOG4J2-1799

It would be nice to add this fix to 2.8.1 release.


edwgiz:
How to restart the Travis check? There's just a log length exceeding because a lot of Maven downloads

jvz:
I don't see any way to replay it from Travis. You could try doing a `git commit --amend` and `git push --force`. That will change the commit id which should trigger a new build.

edwgiz:
Ok, thanks a lot

PR #59, opened 2017-02-16
Fix rollover to work when filePattern contains no directory components

xkr47:
Prior to this patch, the following config threw NPE at `parent.mkdirs()` because `parent` was null when using e.g.:
```xml
        <RollingRandomAccessFile name="FILE"
                                 fileName="foo.log"
                                 filePattern="foo-%d{yyyy-MM-dd}-%i.log.gz"
                                 immediateFlush="false">
```
while this config did not:
```xml
        <RollingRandomAccessFile name="FILE"
                                 fileName="foo.log"
                                 filePattern="./foo-%d{yyyy-MM-dd}-%i.log.gz"
                                 immediateFlush="false">
```
e.g. by adding `./` in front of the filePattern value.


garydgregory:
Fixed in Git master.

xkr47:
Thanks, will try it out.

mikaelstaldal:
Can this PR be closed?

garydgregory:
@mikaelstaldal Yes, this can be closed.

PR #62, opened 2017-03-08
(LOG4J2-1864) Support capped collection for MongoDB Log-Provider

codescale:
MongoDB supports sth. called [capped collections](https://docs.mongodb.com/manual/core/capped-collections/). If the nosql-mongodb-appender supports this feature, the mongodb-collection could never "overflow" and stick to a defined maximum size.

What do you think? :) We'd love to have this feature.

### DIFF
--- a/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
+++ b/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
@@ -21,33 +21,39 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.mongodb.DB;
-import com.mongodb.MongoClient;
-import com.mongodb.MongoCredential;
-import com.mongodb.ServerAddress;
-import com.mongodb.WriteConcern;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.Core;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
-import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderFactory;
 import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.config.plugins.convert.TypeConverters;
+import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
 import org.apache.logging.log4j.core.config.plugins.validation.constraints.ValidHost;
 import org.apache.logging.log4j.core.config.plugins.validation.constraints.ValidPort;
+import org.apache.logging.log4j.core.filter.AbstractFilterable;
 import org.apache.logging.log4j.core.util.NameUtil;
 import org.apache.logging.log4j.nosql.appender.NoSqlProvider;
 import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.util.LoaderUtil;
 import org.apache.logging.log4j.util.Strings;
 
+import com.mongodb.DB;
+import com.mongodb.MongoClient;
+import com.mongodb.MongoCredential;
+import com.mongodb.ServerAddress;
+import com.mongodb.WriteConcern;
+
 /**
  * The MongoDB implementation of {@link NoSqlProvider}.
  */
-@Plugin(name = "MongoDb", category = "Core", printObject = true)
+@Plugin(name = "MongoDb", category = Core.CATEGORY_NAME, printObject = true)
 public final class MongoDbProvider implements NoSqlProvider<MongoDbConnection> {
 
     private static final WriteConcern DEFAULT_WRITE_CONCERN = WriteConcern.ACKNOWLEDGED;
     private static final Logger LOGGER = StatusLogger.getLogger();
     private static final int DEFAULT_PORT = 27017;
+    private static final int DEFAULT_COLLECTION_SIZE = 536870912;
 
     private final String collectionName;
     private final DB database;
@@ -80,8 +86,6 @@ public final class MongoDbProvider implements NoSqlProvider<MongoDbConnection> {
      * Factory method for creating a MongoDB provider within the plugin manager.
      *
      * @param collectionName The name of the MongoDB collection to which log events should be written.
-     * @param isCapped If the MongoDB collection shall be a capped collection.
-     * @param collectionSize The size of the MongoDB collection in case it's a capped collection.
      * @param writeConcernConstant The {@link WriteConcern} constant to control writing details, defaults to
      *                             {@link WriteConcern#ACKNOWLEDGED}.
      * @param writeConcernConstantClassName The name of a class containing the aforementioned static WriteConcern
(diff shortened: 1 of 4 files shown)


log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -53,18 +53,22 @@
     private final DB database;
     private final String description;
     private final WriteConcern writeConcern;
+    private final Boolean isCapped;
```
jvz: Could you stick to the primitive types instead? Unless these are nullable, but it doesn't appear so.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -97,6 +103,8 @@ public String toString() {
     @PluginFactory
     public static MongoDbProvider createNoSqlProvider(
             @PluginAttribute("collectionName") final String collectionName,
+            @PluginAttribute(value = "capped", defaultBoolean = false) final Boolean isCapped,
```
jvz: In order to preserve backwards compatibility with users who may be configuring things programmatically, we've been converting these factory methods into builders over time so that new attributes can be added. Do you think you could add a builder and deprecate this instead?

codescale:
Hi @jvz, Thanks that you're interested in this feature.
I added a `feature request` to [jira](https://issues.apache.org/jira/browse/LOG4J2-1864). And I will take a look at your remarks as soon as possible. Cheers

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
codescale: Hi @jvz, I changed the `Boolean` member to be a primitive type. But I don't know what kind of Builder you mean for the attributes? Does this builder exist for the MongoDbProvider? Cheers Matt

jvz: Take a look at CassandraAppender for how to write a plugin builder class. Any manual type conversion in the factory method can be generally updated to use the proper types as conversions for most common types is supported (and new types can always be added as appropriate; see InetAddress et al. added for CassandraAppender).

codescale: I took a look at the CassandraAppender. It's not a big deal to adjust the MongoDbProvider to match this style. But the API is also broken. Or do you intend to keep the old public methods? Shouldn't this even be another issues and pull-request?

jvz: What we've been doing with old factory methods is removing the annotations from them and marking them deprecated. That way it doesn't break custom configurations, but only the new builder is used for normal configuration (although builders are used before factory methods regardless).

codescale:
I updated the `MongoDbProvider` how I understood you want it. But it's not working. Maybe you know why @jvz ?

jvz:
It looks correct based on a quick look, so I'm not sure what exactly is incorrect. I would like to note that the builders generally use the parsed type rather than a string, though I can help clean that up afterward.

codescale:
Fixed the problem. I extended from the wrong `Builder`. Please review and let me know if it's ok. Thx.

mikaelstaldal:
Ping @jvz

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -21,50 +21,60 @@
...
  * The MongoDB implementation of {@link NoSqlProvider}.
  */
-@Plugin(name = "MongoDb", category = "Core", printObject = true)
+@Plugin(name = "MongoDb", category = Core.CATEGORY_NAME, printObject = true)
 public final class MongoDbProvider implements NoSqlProvider<MongoDbConnection> {
 
     private static final WriteConcern DEFAULT_WRITE_CONCERN = WriteConcern.ACKNOWLEDGED;
     private static final Logger LOGGER = StatusLogger.getLogger();
     private static final int DEFAULT_PORT = 27017;
+    private static final int DEFAULT_COLLECTION_SIZE = 536870912;
```
jvz: Magic number?

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -93,130 +103,249 @@ public String toString() {
...
+	public static class Builder<B extends Builder<B>> extends AbstractFilterable.Builder<B>
+			implements org.apache.logging.log4j.core.util.Builder<MongoDbProvider> {
+
+		@PluginBuilderAttribute
+		@ValidHost
+		private String server = "localhost";
+
+		@PluginBuilderAttribute
+		@ValidPort
+		private String port = "" + DEFAULT_PORT;
```
jvz: You can make this an int.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -93,130 +103,249 @@ public String toString() {
...
+
+		public B setWriteConcernConstant(String writeConcernConstant) {
+			this.writeConcernConstant = writeConcernConstant;
+			return asBuilder();
+		}
+        
+		@Override
+		public MongoDbProvider build() {
+	        DB database;
+	        String description;
```
jvz: It's not a huge deal performance-wise, but you could use a StringBuilder here for the more advanced concatenation. I doubt javac can optimize it like it does with other string concatenations.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
```diff
@@ -93,130 +103,249 @@ public String toString() {
...
+	        } else if (Strings.isNotEmpty(databaseName)) {
+	            final List<MongoCredential> credentials = new ArrayList<>();
+	            description = "database=" + databaseName;
+	            if (Strings.isNotEmpty(userName) && Strings.isNotEmpty(password)) {
+	                description += ", username=" + userName + ", passwordHash="
+	                        + NameUtil.md5(password + MongoDbProvider.class.getName());
+	                credentials.add(MongoCredential.createCredential(userName, databaseName, password.toCharArray()));
+	            }
+	            try {
+	                final int portInt = TypeConverters.convert(port, int.class, DEFAULT_PORT);
```
jvz: As mentioned elsewhere, if you make the field an int, you don't need to do this.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
codescale: Again ... no changes from my side.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
codescale: https://github.com/codescale/logging-log4j2/blob/master/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java#L141

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java
codescale: These changes have been introduced to my pull-request by a rebase from master.
This change exists there already. see https://github.com/codescale/logging-log4j2/blob/master/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/mongodb/MongoDbProvider.java#L56

jvz: Oh, someone else added that in a separate PR, I see.

jvz:
Alright, everything looks fine to merge then. @mikaelstaldal you can merge this if you have time.

garydgregory:
Don't forget to update the docs please.

codescale:
I must admit I don't know where/how I can help you to keep the docs up to date. If I'm able to help, please point me to the right location :) thanks.

jvz:
https://github.com/apache/logging-log4j2/blob/master/src/site/xdoc/manual/appenders.xml is the appenders manual page which should be what you're looking for. Just follow the patterns for how to markup the docs.

mikaelstaldal:
@jvz I have no experience with our MongoDB appender, and only limited experience with MongoDB in general. Maybe better that you, or someone else with more experience, merge this.

jvz:
I'm not too experienced with Mongo either because I prefer not losing data, but no worries, I'll take care of merging this now that I've got some time.

codescale:
Thanks @jvz for extending the documentation :+1: 
https://github.com/apache/logging-log4j2/commit/af2dd30e16cece20b4ae475da851c8538bfa8879

PR #66, opened 2017-03-16
Fix the default value of RandomAccessFileAppender.Builder append field

wangyuntao:
See [LOG4J2-1853](https://issues.apache.org/jira/browse/LOG4J2-1853)


wangyuntao:
Fixed by this [commit](https://github.com/apache/logging-log4j2/commit/03e9743e8607f27e6f54d9009404d6c2c5228259).

PR #67, opened 2017-03-20
LOG4J2-1854 - Add new parameter to include tcp Gelf message delimiter

xjodoin:


log4j-core/src/main/java/org/apache/logging/log4j/core/layout/GelfLayout.java
```diff
@@ -237,11 +241,11 @@ public B setAdditionalFields(KeyValuePair[] additionalFields) {
     @Deprecated
     public GelfLayout(final String host, final KeyValuePair[] additionalFields, final CompressionType compressionType,
                       final int compressionThreshold, final boolean includeStacktrace) {
-        this(null, host, additionalFields, compressionType, compressionThreshold, includeStacktrace, true);
+        this(null, host, additionalFields, compressionType, compressionThreshold, includeStacktrace, true, false);
```
mikaelstaldal: Please do not change this method since that will break binary compatibility. Users will have to use builder to use the new feature.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/GelfLayout.java
```diff
@@ -135,14 +136,17 @@ public DeflaterOutputStream createDeflaterOutputStream(final OutputStream os) th
         @PluginBuilderAttribute
         private boolean includeThreadContext = true;
 
+        @PluginBuilderAttribute
+        private boolean includeTcpMessageDelimiter = false;
```
mikaelstaldal: Maybe name this as `includeNullMessageDelimiter` to indicate what it does, rather than what is can be used for?

mikaelstaldal:
I have fixed this in master branch, this PR can be closed.

PR #68, opened 2017-03-23
[LOG4J2-1855] Add an optional random delay in TimeBasedTriggeringPolicy

m-anthony:
https://issues.apache.org/jira/browse/LOG4J2-1855


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/TimeBasedTriggeringPolicy.java
```diff
@@ -84,15 +91,18 @@ public boolean isTriggeringEvent(final LogEvent event) {
...
      * @param modulate If true the time will be rounded to occur on a boundary aligned with the increment.
+     * @param maxRandomDelay If non-zero the rolling will be delayed by a random amount of time, up to the specified value (in seconds)  
      * @return a TimeBasedTriggeringPolicy.
      */
     @PluginFactory
     public static TimeBasedTriggeringPolicy createPolicy(
             @PluginAttribute("interval") final String interval,
-            @PluginAttribute("modulate") final String modulate) {
+            @PluginAttribute("modulate") final String modulate,
+            @PluginAttribute("maxRandomDelay") final int maxRandomDelay) {
```
jvz: Instead of adding a new param to the factory method, do you think you could convert this to a builder and deprecate this method? This way, users who have used the factory method programmatically will not have an API change break their config, and this way we can also use the type conversion support to simplify the builder/factory as it is.

garydgregory: +1

m-anthony: OK I will do that :+1:

m-anthony: Done

PR #69, opened 2017-04-01
[LOG4J2-1862] Correct errors in the comments of Order and OrderComparator

wangyuntao:
See [LOG4J2-1862](https://issues.apache.org/jira/browse/LOG4J2-1862)


wangyuntao:
See [LOG4J2-1862](https://issues.apache.org/jira/browse/LOG4J2-1862)

PR #70, opened 2017-04-08
Fix a bug in StringBuilderEncoder.encode()

leventov:
Headroom for char sequence encoded data is overestimated.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
@@ -285,7 +285,7 @@ public class OutputStreamManager extends AbstractManager implements ByteBufferDe
     protected synchronized void flushBuffer(final ByteBuffer buf) {
         buf.flip();
         if (buf.limit() > 0) {
-            writeToDestination(buf.array(), 0, buf.limit());
+            writeToDestination(buf.array(), buf.arrayOffset(), buf.remaining());
         }
         buf.clear();
     }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
@@ -26,6 +26,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.logging.log4j.core.Layout;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.layout.ByteBufferDestination;
+import org.apache.logging.log4j.core.layout.ByteBufferDestinationHelper;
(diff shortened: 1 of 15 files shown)


leventov:
And, well, it doesn't fix the problem entirely, because `getByteBuffer()` is not called in the context of synchronized on the ByteBufferDestination, and after this call the remaining space may be reduced. I'm working on a fix.

leventov:
The intended fix is to add the following method to the `ByteBufferDestination` interface:

```java
    /**
     * Writes the given data to this ByteBufferDestination entirely. Call of this method should *not* be protected
     * with synchronized on this ByteBufferDestination instance.
     */
    public void write(ByteBuffer data);
```

Implementations of this method should synchronize internally, if needed.

Then `StringBuilderEncoder.encode()` only estimates if the data fits the thread-local byte buffer, without querying `destination.getByteBuffer().remaining()`, and then calls `destination.write(temp)` with the buffer with encoded data.

remkop:
Roman, what problem are you trying to solve? Is it related to this [twitter thread](https://twitter.com/leventov/status/845418721254486020)?

leventov:
@remkop there is a race condition that could lead to tearing encoded events apart of writing a truncated event close to the boundaries of Manager's ByteBuffer. I. e. in case of MemoryMappedFileManager, close to 32 MB boundaries.

It is related to that Twitter thread in the sense that I spotted this problem while working on more concurrency-friendly implementation of MemoryMappedFileManager. A pull-request will follow after this PR.

I've pushed the full solution to the race condition in StringBuilderEncoder.encode() now; please review. It adds `ByteBufferDestination.write(ByteBuffer)` and `ByteBufferDestination.write(byte[], int, int)` methods which should care about synchronization themselves. This is also essential for lock-free MemoryMappedFileManager, because exactly those methods are going to be implemented lock-free. `getByteBuffer()` + `drain()` are designed so that synchronization couldn't be avoided. But the expectation is that most of encoded events fit the thread-local buffers, and on of `write()` methods could be called instead of writing to `destination.getByteBuffer()`.

leventov:
Also this PR alone improves concurrency, because it delays locking on the `ByteBufferDestination` object as far as possible. The encoding work should now mostly be done outside of synchronized block on the `ByteBufferDestination`.

remkop:
OK I will try to look at this on the weekend because I'm swamped at work this week. Can I ask you to raise a Jira for this? (The text of your last 2 comments are fine as description.)

leventov:
@remkop I reviewed the existing code again and found that there is actually no race condition or bugs, if only one assumption is true: `ByteBufferDestination.drain()` returns ByteBuffer of capacity which is the same as the former `ByteBufferDestination.getByteBuffer()`'s capacity.

However this refactoring is still needed, because it allows lock-free MemoryMappedFileManager. I've "rebraded" this PR as #71. The Jira issue with description is https://issues.apache.org/jira/browse/LOG4J2-1874.

PR #73, opened 2017-04-19
Add a new LuceneAppender which writes logging events to a lucene index library.

liyuj:
Add a new LuceneAppender which writes logging events to a lucene index library.The log4j2.xml configuration is as follows:
&lt;Lucene name="lucene" ignoreExceptions="true" target="/target/lucene/index" expiryTime=“1296000”&gt;
    &lt;PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/&gt;
	  &lt;IndexField name="time" pattern="%d{UNIX_MILLIS}" type="LongField"/&gt;
	  &lt;IndexField name="level" pattern="%-5level" /&gt;
	  &lt;IndexField name="content" pattern="%d{HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/&gt;
&lt;/Lucene&gt;

this appender relies on the Lucene 5.5.4 version.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/lucene/LuceneAppender.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/lucene/LuceneAppender.java
@@ -1,0 +1,295 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


garydgregory:
Thank you for your patch! Can you please redo this work as a patch to provide a new module called "log4j-lucene"? This does not belong in our Core module and we do not have a home for this kind of work yet.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,295 @@
...
+				}
+			}
+		}
+	}
+
+	/**
+     * Factory method for creating a lucene appender within the plugin manager.
+     *
+     */
+	@PluginFactory
```
garydgregory: Please use a builder instead of a factory method. This will allow us to avoid breaking binary compatibility in the future when new parameters are needed.

garydgregory:
Can you please provide some JUnit tests? There are no tests ATM.

Thank you,
Gary

jvz:
This vaguely sounds like a log4j-nosql appender. I'd consider a similar one for Elasticsearch to fall under that module.

liyuj:
If so, I think it is more reasonable to put it in the log4j-nosql module.
We rearrange the code and submit it again.

liyuj:
If so, I think it is more reasonable to put it in the log4j-nosql module.
We rearrange the code and submit it again.

PR #74, opened 2017-04-24
Add a new LuceneAppender which writes logging events to a lucene index library.

liyuj:
Add a new LuceneAppender which writes logging events to a lucene index library.The log4j2.xml configuration is as follows:
&lt;Lucene name="lucene" ignoreExceptions="true" target="/target/lucene/index" expiryTime=“1296000”&gt;
&lt;PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/&gt;
&lt;IndexField name="time" pattern="%d{UNIX_MILLIS}" type="LongField"/&gt;
&lt;IndexField name="level" pattern="%-5level" /&gt;
&lt;IndexField name="content" pattern="%d{HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/&gt;
&lt;/Lucene&gt;

this appender relies on the Lucene 5.5.4 version.
this patch adds the corresponding test cases.

### DIFF
--- a/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
+++ b/log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
@@ -1,0 +1,301 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.log4j.nosql.appender.lucene;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Paths;
+import java.util.Calendar;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.logging.log4j.core.Appender;
+import org.apache.logging.log4j.core.Filter;
+import org.apache.logging.log4j.core.Layout;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.appender.AbstractAppender;
+import org.apache.logging.log4j.core.appender.AppenderLoggingException;
+import org.apache.logging.log4j.core.config.Node;
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderFactory;
+import org.apache.logging.log4j.core.config.plugins.PluginElement;
+import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
+import org.apache.logging.log4j.util.Strings;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.document.Field;
+import org.apache.lucene.document.LongField;
+import org.apache.lucene.document.StringField;
+import org.apache.lucene.document.TextField;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.search.NumericRangeQuery;
+import org.apache.lucene.store.FSDirectory;
+
+/**
+ * This Appender writes logging events to a lucene index library. It takes a list of
+ * {@link IndexField} with which determines which fields are written to the index library.
+ * <Lucene name="lucene" ignoreExceptions="true" target="/target/lucene/index">
+ *    <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/>
+ *    
+ *	  <IndexField name="time" pattern="%d{UNIX_MILLIS}" type="LongField"/>
+ *	  <IndexField name="level" pattern="%-5level" />
+ *	  <IndexField name="content" pattern="%d{HH:mm:ss.SSS} %-5level %class{36} %L %M - %msg%xEx%n"/>
+ * </Lucene>
+ */
+@Plugin(name = "Lucene", category = Node.CATEGORY, elementType = Appender.ELEMENT_TYPE, printObject = true)
+public class LuceneAppender extends AbstractAppender {
+	/**
+	 * index directory
+	 */
+	private final String target;
+	/**
+	 * Index expiration time (seconds)
+	 */
+	private final Integer expiryTime;
+	/**
+	 * IndexField array.
+	 */
+	private final LuceneIndexField[] indexFields;
+	/**
+	 * Periodically clear the index and submit the IndexWriter thread pool
+	 */
+	private final ScheduledExecutorService scheduledExecutor = Executors.newScheduledThreadPool(2);
+	/**
+	 * IndexWriter corresponding to each index directory.
+	 */
+	private static final ConcurrentHashMap<String, IndexWriter> writerMap = new ConcurrentHashMap<String, IndexWriter>();
+	
+	static{
+		Runtime.getRuntime().addShutdownHook(new Thread() {
+			@Override
+			public void run() {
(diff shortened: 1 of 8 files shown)


garydgregory:
Why not use the current version of Lucene? Is the version you are using and the latest (6.5.0) compatible?

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+import org.apache.lucene.document.TextField;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.search.NumericRangeQuery;
+import org.apache.lucene.store.FSDirectory;
+
+/**
+ * This Appender writes logging events to a lucene index library. It takes a list of
+ * {@link IndexField} with which determines which fields are written to the index library.
+ * <Lucene name="lucene" ignoreExceptions="true" target="/target/lucene/index">
```
jvz: Might want to surround this with something like `<pre>{@code ... }</pre>` or however it'd be best formatted. XML doesn't work well in javadocs ;)

jvz: Also, these sorts of examples work better in the manual page. The javadocs are more useful for users programmatically configuring things, so they don't necessarily need XML examples here.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+	 * Periodically clear the index and submit the IndexWriter thread pool
+	 */
+	private final ScheduledExecutorService scheduledExecutor = Executors.newScheduledThreadPool(2);
+	/**
+	 * IndexWriter corresponding to each index directory.
+	 */
+	private static final ConcurrentHashMap<String, IndexWriter> writerMap = new ConcurrentHashMap<String, IndexWriter>();
+	
+	static{
+		Runtime.getRuntime().addShutdownHook(new Thread() {
```
jvz: Not a fan of adding a shutdown hook here when it's already managed by the stop() method.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+	 */
+	private final Integer expiryTime;
+	/**
+	 * IndexField array.
+	 */
+	private final LuceneIndexField[] indexFields;
+	/**
+	 * Periodically clear the index and submit the IndexWriter thread pool
+	 */
+	private final ScheduledExecutorService scheduledExecutor = Executors.newScheduledThreadPool(2);
```
jvz: I think we have some built in log4j mechanisms for scheduling things (see for example the rolling file appender plugins). Might be more robust here to make that all configurable, but I'm open to suggestions.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+	/**
+	 * IndexWriter corresponding to each index directory.
+	 */
+	private static final ConcurrentHashMap<String, IndexWriter> writerMap = new ConcurrentHashMap<String, IndexWriter>();
```
jvz: I can't say that this is enforced everywhere, but the general pattern we like to follow is that shared state for appenders are handled by a manager class instead of static fields. Helps with certain deployment scenarios like WARs, reconfigurations, and multiple appenders using the same config.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+	 */
+	private void registerClearTimer() {
+		Calendar calendar = Calendar.getInstance();
+		long curMillis = calendar.getTimeInMillis();
+		calendar.add(Calendar.DAY_OF_MONTH, 1);
+		calendar.set(Calendar.HOUR_OF_DAY, 0);
+		calendar.set(Calendar.MINUTE, 0);
+		calendar.set(Calendar.SECOND, 0);
+		calendar.set(Calendar.MILLISECOND, 0);
+		long difMinutes = (calendar.getTimeInMillis() - curMillis) / (1000 * 60);
```
jvz: You could use `TimeUnit.MILLIS.toMinutes(x)` here to be more expressive.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+	public void append(LogEvent event) {
+		if (null != indexFields && indexFields.length > 0) {
+			IndexWriter indexWriter = initIndexWriter();
+			if(null != indexWriter){
+				Document doc = new Document();
+				doc.add(new LongField("timestamp", event.getTimeMillis(), Field.Store.YES));
+				try {
+					for (LuceneIndexField field : indexFields) {
+						String value = field.getLayout().toSerializable(event);
+						if (Strings.isEmpty(value) || value.matches("[$]\\{.+\\}")) {
```
jvz: Generally like to precompile regular expressions into Pattern objects. That's actually one case where a static field makes sense come to think of it. ;)

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+							return;
+						}
+						value = value.trim();
+						String type = field.getType();
+						if (Strings.isNotEmpty(type)) {
+							Class<?> clazz = Class.forName("org.apache.lucene.document." + type);
```
jvz: Try to stick to the `LoaderUtil` methods for loading classes unless you have a ClassLoader already (in which case there might be another method in LoaderUtil or Loader for that)

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+							}
+						} else {
+							doc.add(new TextField(field.getName(), value, Field.Store.YES));
+						}
+					}
+					indexWriter.addDocument(doc);
+				} catch (Exception e) {
+					LOGGER.error(e.getMessage(), e);
+					if (!ignoreExceptions()) {
+						throw new AppenderLoggingException(e);
```
jvz: This is automatically handled by a wrapper appender that all appenders are constructed within. You can simplify this.

log4j-nosql/src/main/java/org/apache/logging/log4j/nosql/appender/lucene/LuceneAppender.java
```diff
@@ -0,0 +1,301 @@
...
+
+    public static class Builder<B extends Builder<B>> extends AbstractAppender.Builder<B>
+        implements org.apache.logging.log4j.core.util.Builder<LuceneAppender> {
+
+        @PluginElement("IndexField")
+        @Required(message = "No IndexField provided")
+        private LuceneIndexField[] indexField;        
+
+        @PluginBuilderAttribute        
+        private Integer expiryTime;
```
jvz: Is the default supposed to be null?

liyuj:
Lucene 6 and later versions require Java 8 or later.

garydgregory:
Good to know! Thank you. I wonder if we should then create this under a "lucene5" package so that we can have a "lucene6" package in a different module if needed.

mikaelstaldal:
I suggest we place this in a new module "log4j-lucene" instead of in log4j-nosql. We probably want to get rid of log4j-nosql, so we shouldn't put new stuff in it.

garydgregory:
And maybe even log4j-lucene5...

jvz:
If Lucene's API is anything like Elasticsearch, then they don't bother changing the package name between major versions and there are hacky ways to support multiple versions.

garydgregory:
I was thinking that if you can use Java 8 then you depend on log4j-lucene6 and if you can't then log4j-lucene5.

jvz:
That makes sense. In the case of Elasticsearch (1.x, 2.x, and 5.x), you can support all versions using the same code in certain scenarios (usually simple things). If Lucene follows the usual Apache idea of renaming their packages between major releases, then it'd be necessary to have both modules anyways.

liyuj:
Lucene5 API and lucene6 are incompatible.
I think the future is divided into multiple modules is more reasonable, such as log4j-lucene5 and log4j-lucene6.

garydgregory:
How about updating the patch to provide a new module log4j-lucene5?

liyuj:
We found a bug that is being corrected and then resubmitted in the form of a new module.

PR #78, opened 2017-05-16
Fix writeObject in FormattedMessage.

bsokolovu:
To pass `FormattedMessage` completely we should pass string values from `argArray` together with `argArray.length`.
For now there is such bug:  on `readObject` it's `InputStream` brings `length` of `stringArgs` but there is no values in `InputStream` to read. So if `length > 0` there will be `EOFException` while reading `null` value from `InputStream`.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/FormattedMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/FormattedMessage.java
@@ -256,7 +256,9 @@ public class FormattedMessage implements Message {
         stringArgs = new String[argArray.length];
         int i = 0;
         for (final Object obj : argArray) {
-            stringArgs[i] = obj.toString();
+            final String string = String.valueOf(obj);
+            stringArgs[i] = string;
+            out.writeUTF(string);
             ++i;
         }
     }
--- a/log4j-api/src/test/java/org/apache/logging/log4j/message/FormattedMessageTest.java
+++ b/log4j-api/src/test/java/org/apache/logging/log4j/message/FormattedMessageTest.java
@@ -16,13 +16,19 @@
  */
 package org.apache.logging.log4j.message;
 
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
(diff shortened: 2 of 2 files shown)


jvz:
Code looks good. Can you add a unit test? Also, does the serialVersionUID need to be changed, or is this a safe serialization change?

bsokolovu:
I don't see reasons to change serialVersionUID. It's not new version of entity.
I'd like to add unit test but I can't build even `master` branch because of some odd error on execution `java9-compile`. It looks like `javac: invalid flag: --release`. If you'll help me to get rid of it to run tests I'll add case to unit test.

> don't mind the travis check

Could you please explain what do you mean?

jvz:
Do you have java 9 installed? Make sure to follow the [setup instructions](https://github.com/apache/logging-log4j2/blob/master/BUILDING.md). As for the travis check, the thing below the comments. Don't worry that it says failed because it doesn't work at the moment anyways.

remkop:
The serialVersionUID can remain unchanged if
* the new version of the software can deserialize "old" serialized objects (backwards compatible)
* the "old" version of the software can deserialize "new" serialized objects (forward compatible)

Sometimes it's okay to keep the serialVersionUID unchanged if the change is backwards compatible but not forward compatible. That allows new versions of Log4j2 to read old serialized logs. 

This needs to be clearly documented.

bsokolovu:
I've amended commit with unit test.
And once more about serialVersionUID - it's not a new version of entity. It's fix for serialization of an old version. If you revert my fix but keep the test it will fail.
So there is no reason to update serialVersionUID.

jvz:
PR looks great! Ready to merge. Thanks for the clarifications regarding serialization.

bsokolovu:
Hi gatekeepers! 
Are you going to merge it and use it? Or you have new comments? Or you just forgot about it? ))

jvz:
I've been pretty busy lately and haven't gotten to do anything on this project in a little while. However, as mentioned, this PR can be merged by anyone else. We don't have plans to release anytime soon, so don't worry about missing out! ;)

PR #79, opened 2017-05-25
LOG4J2-1901: Fix travis-ci build

PascalSchumacher:
remove apt-get, use trusty dist

not sure this will be enough

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@
 language: java
 # sudo enabled because sudo vms have more memory: https://docs.travis-ci.com/user/ci-environment/
 sudo: true
+# trusty required for oraclejdk9
+dist: trusty
 
 jdk:
   - oraclejdk7
@@ -20,11 +22,6 @@ jdk:
 env:
   global:
     JAVA_OPTS=-Xmx4g
-
-addons:
-  apt:
-    packages:
-      - oracle-java9-installer
 
 install:
   - ./mvnw --toolchains=./.travis-toolchains.xml install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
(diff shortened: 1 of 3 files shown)


mikaelstaldal:
Now it fails with the socket option test that has been fixed. Have you synced with upstream?

PascalSchumacher:
Yes. The problem is that these tests also fail on java 7 when run on travis.

PascalSchumacher:
Added assumes to skip these tests on travis-ci.

Now a test of og4j-api-scala_2.11 fails:

[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.228 s <<< FAILURE! - in org.apache.logging.log4j.scala.LoggerTest
[ERROR] org.apache.logging.log4j.scala.LoggerTest  Time elapsed: 0.55 s  <<< ERROR!
java.lang.NoClassDefFoundError: org/apache/logging/log4j/scala/Logger$
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:365)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:272)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:236)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:159)
Caused by: java.lang.ClassNotFoundException: org.apache.logging.log4j.scala.Logger$
	at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:365)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:272)
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:236)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:159)

mikaelstaldal:
That's strange. However, the Scala modules will soon be remove from this repository (as they live in the logging-log4j-scala repo). So maybe we should not spend time on fixing that problem in this repo.

jvz:
Not finding the `Logger$` class sounds like the Scala code isn't being compiled properly. That's the companion object class name.

mikaelstaldal:
It may be due to macros. When using Scala macros, you need to compile the macro definitions separately from whatever using the macros. If src/main and src/test are compiled together, it will not work.

But why is it differently from Jenkins? Isn't both Jenkins and Travis CI just running the Maven build? Or does Travis CI tries to be "smart" and actually do something stupid?

PR #81, opened 2017-05-28
[LOG4J2-1923] Allow having the stacktrace as a string in JSONLayout, XMLLayout and YAMLLayout

imriz:


mikaelstaldal:
Can you file an JIRA issue about this here: https://issues.apache.org/jira/browse/LOG4J2/

and mention the JIRA issue here?

It would be good with some motivation why this would be a good idea, such as example use case.

imriz:
@mikaelstaldal I've opened the JIRA: LOG4J2-1923

mikaelstaldal:
It would be nice to have the same functionality in XmlLayout and YamlLayout. It should be easy to add that since they use the same Jackson based framework.

imriz:
Sure. I will check if I can extend it to xml and yaml.

mikaelstaldal:
Documentation in `layouts.xml.vm` should be updated.

imriz:
@mikaelstaldal Hi :) I've extended this to YAML and XML + updated `layouts.xml.vm`

mikaelstaldal:
To avoid breaking binary compatibility, we should not modify public/protected constructors/methods of JsonLayout, XmlLayout, YamlLayout. Add the new flag to the Builder only.

mikaelstaldal:
It would be nice to unit test this new feature for XmlLayout and YamlLayout also.

PR #83, opened 2017-06-22
[LOG4J2-1699] Refactor PosixFileAttributeViewAction with Builder. Fix posix tests on jenkins.

phymbert:
Refactor PosixFileAttributeViewAction with Builder rather than method factory. Split posix tests inner static classes to 2 differents test class files because it doesnt run on jenkins. Add a static method in FileUtils to check if posix file attribute view is supported in order to log if posix attributes are defined
but the filesystem doesnt support it. Fix when user has only one group in the test case.

Signed-off-by: phymbert <pierrick.hymbert@gmail.com>


phymbert:
Merged in 9d32793b12e2fe9b7a77519e59f5027942db4917

PR #84, opened 2017-06-22
[LOG4J2-1950] Fix docker build with jdk9 requirements

phymbert:
Revert docker image base from openjdk:7-alpine to openjdk:7-jdk (debian) since yet alpine does not include jdk9.
Install jdk9 and specify toolchains file in the mvn build

Signed-off-by: phymbert <pierrick.hymbert@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,18 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM openjdk:7-alpine
+# FROM openjdk:7-alpine
+# Reverted to debian yet alpine does not include jdk9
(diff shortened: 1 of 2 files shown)


garydgregory:
Curious, why not Maven 3.5.0?

jvz:
Can certainly be updated for that since it's not installed via apt-get.

phymbert:
I tried to minimise changes, I would prefer to install mvn as a package rather to freeze the version. I may do it tom if you are fine with it.

phymbert:
Updated to 3.5.0, please check

PR #86, opened 2017-06-23
[LOG4J2-1766] Temporary archived file pattern during compression

phymbert:
Rebased patch onto master, added documentation


phymbert:
Hi @garydgregory, understood, little question, is it preferable to deprecate existing constructor and create another one or set after instantiate new parameter in the builder ?

phymbert:
Added Builder API for rollover strategies, I introduce new constructors for this additional parameter. Please check.

jvz:
Depends if the constructors are private or not. For public ones where having a default value for the new constructor doesn't make much sense, deprecation is the way to go. When a sensible default is possible, I wouldn't bother deprecating the old one.

PR #91, opened 2017-07-04
failing test - irregular log4j2.json nodes order

romanso123:
Failing test, showing that log4j2.json config file nodes specific order can fail log4j2 framework functionality.

Test description:
 "properties" node in resources/log4j2.json config file is the last node (not regular json nodes order)
 due to this, log4j2 framework doesn't act normally
 this log msg is not printed correctly - "${layout}" is printed.
 if we set "properties" node to be before "appenders" node, everything will work as expected.
 why would json-node order matter? the nodes order in a JSON shouldn't matter ...

### DIFF
--- a/failing-test/pom.xml
+++ b/failing-test/pom.xml
@@ -1,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>log4j</artifactId>
+        <groupId>org.apache.logging.log4j</groupId>
+        <version>2.9-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>failing-test</artifactId>
+
+    <dependencies>
+        <dependency>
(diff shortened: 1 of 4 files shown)


jochenw:
Quoting Ralph Goers:

This is not a bug but is intentional. Log4j processes configuration files from beginning to end. It does not scan the file twice as would be necessary if no order was required. Properties MUST come first so that they can be used by the configuration that follows. Because of that we will generate an error if the properties are encountered after some other element is found.

romanso123:
Yea, saw that.
But still... in general, a JSON is not supposed to have order..
In my opinion this design is flawed.
Anyway you decide what to do with it.

Thanks.

waymirec:
After a bunch of interesting reading it appears that ES2015 says that Object property names are to be ordered in the following way:  integer-like Strings in ascending order, non-integer-like Strings in creation order, Symbols in creation order. 

Since Log4j doesn't use integer-like Strings as property names we can reasonably expect our JSON to be in the order it was specified (which is exactly what we get with Jackson).

rgoers:
After a bunch of interesting reading it appears that ES2015 says that Object property names are to be ordered in the following way: integer-like Strings in ascending order, non-integer-like Strings in creation order, Symbols in creation order.

Since Log4j doesn't use integer-like Strings as property names we can reasonably expect our JSON to be in the order it was specified (which is exactly what we get with Jackson).

mikaelstaldal:
For https://issues.apache.org/jira/browse/LOG4J2-1933

rgoers:
Closing this since it only provides a failing unit test, doesn't provide code to fix it and the JSON in the test doesn't match our documentation.

PR #93, opened 2017-07-17
modification to resolver to discover jar assests without the manipula…

ta924:
…tion of asset paths.

This PR is to address issue https://issues.apache.org/jira/browse/LOG4J2-1852.  If the resource submitted for scanning is of Jar protocol it utilizes a JarURLConnection to scan for classes matching the test, it will no longer manipulate the URI for Jars.  In the event the protocol is file it will proceed with the existing approach.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
@@ -112,7 +112,7 @@ public class ResolverUtil {
 
     /**
      * Returns the matching resources.
-     * 
+     *
      * @return A Set of URIs that match the criteria.
      */
     public Set<URI> getResources() {
@@ -342,12 +342,9 @@ public class ResolverUtil {
      *        the URL Connection to be examined for classes
      */
     private void loadImplementationsInJar(final ResolverUtil.Test test, final String parent, final URLConnection connection) {
-
-        JarFile jarFile = null;
-        try {
-            if (connection != null && connection instanceof JarURLConnection) {
-                JarURLConnection jarURLConnection = (JarURLConnection) connection;
-                jarFile = jarURLConnection.getJarFile();
+        if (connection != null && connection instanceof JarURLConnection) {
+            JarURLConnection jarURLConnection = (JarURLConnection) connection;
+            try (JarFile jarFile = jarURLConnection.getJarFile()) {
                 final Enumeration<JarEntry> entries = jarFile.entries();
 
                 while (entries.hasMoreElements()) {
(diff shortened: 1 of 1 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
```diff
@@ -210,11 +210,13 @@ public void findInPackage(final Test test, String packageName) {
                 } else if (BUNDLE_RESOURCE.equals(url.getProtocol())) {
                     loadImplementationsInBundle(test, packageName);
                 } else {
-                    final File file = new File(urlPath);
-                    if (file.isDirectory()) {
-                        loadImplementationsInDirectory(test, packageName, file);
-                    } else {
-                        loadImplementationsInJar(test, packageName, file);
+                    if (JAR.equals(url.getProtocol())){
+                        this.loadImplementationsInJar(test, packageName, url.openConnection());
```
garydgregory: Thank you for the patch.  Who closes this connection?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
```diff
@@ -329,6 +331,49 @@ private void loadImplementationsInJar(final Test test, final String parent, fina
...
+                        this.addIfMatching(test, name);
+                    }
+                }
+            }
+        }catch (IOException e){
+            LOGGER.error("Could not search JAR file '{}' for classes matching criteria {}, file not found", jarFile,
+                    test, e);
+        }finally {
+            if (jarFile != null){
+                try {
```
garydgregory: Is it possible to simplify with a try-with-resources block?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
garydgregory: Nevermind, the openConnection() does not allocate an actual connection.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/ResolverUtil.java
ta924: Yes agreed this can be cleaned up with try-with-resources block.  I will make those modifications.

ta924:
I will evaluate the test logic in place to see if it is an option to create a sample jar of the same structure.

ta924:
Submitted code change to introduce try-with-resources block, thank you for the catch.

garydgregory:
Hi. Were you able to create a sample jar for testing?

ta924:
Hi. Sorry for the delay had to prep for a few presentations the last couple days. I will shoot to have one ready by the end of the week.

garydgregory:
Hi @ta924  Any news?

ta924:
Sorry Gary had some things pop up on the work front that have eaten up most of my time. I will get you an example as soon as things free up.

ta924:
After some additional research I found that an internal library was excluding all log4jplugins.dat files to ensure that only specific packages were scanned.  This was accomplished by a custom classloader.  This forced the PluginManager to to attempt resolution based off relocated package names from the shade plugin.  This modification made scanning and resolution possible for the custom wrappers.  I noticed that Spring Boot includes a log4jplugins.dat file, which was getting loaded by default while trying to create a sample app.  That being said this becomes much more of a one off situation and doesn't fit the normal operating patterns of log4j.  I think the approach is a good fall back in the event log4jplugins.dat file is not present and provided a resolution mechanism for core plugins.
Again sorry for the late turn around on correspondence on the PR.

rgoers:
A modified version of this patch has been applied.

PR #95, opened 2017-07-21
Kpm1985

kpm1985:
I have cleaned up a few files and removed empty comments. The default buffer size in certain implementations was 8102, which I believe was a typo because 8192 is a power of two. Today was my first PR and it went to Apache, thank you guys!


kpm1985:
While reading through the PDF user guide, I have noticed on page 88 the default buffer size is indeed intended to be 8192. If this PR is dirty and needs to be cleaned up please let me know (I'm new to git). Thank you.

garydgregory:
Thank you for your patch. Applied to Git master.

PR #101, opened 2017-07-28
LOG4J2-Fix type warnings and remove deprecated methods where appropriate

kpm1985:
Unit tests pass. Maven builds successfully.


garydgregory:
Thank you for the patch.

In the future, please do not use redundant type arguments. 

For example:

HashSet<String> foo = new HashSet<String>();

can simply be:

HashSet<String> foo = new HashSet<>()

PR #103, opened 2017-08-01
LOG4J2-1997 Remove deprecated methods and close resource leaks.

kpm1985:
This closes the jira.


garydgregory:
The resources are still leaking if the tests fail midway through a method. How about using try-with-resources clauses?

PR #104, opened 2017-08-01
LOG4J-1997 Cleanup of deprecated methods, warnings and use of try-with-resource to

kpm1985:
remove resource leaks.

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/appender/HangingAppender.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/appender/HangingAppender.java
@@ -90,4 +90,8 @@ public class HangingAppender extends AbstractAppender {
         setStopped();
         return true;
     }
+
+	public static long getSerialversionuid() {
+		return serialVersionUID;
+	}
 }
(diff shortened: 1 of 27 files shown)


log4j-core/src/test/java/org/apache/logging/log4j/core/appender/HangingAppender.java
```diff
@@ -90,4 +90,8 @@ public boolean stop(final long timeout, final TimeUnit timeUnit) {
         setStopped();
         return true;
     }
+
+	public static long getSerialversionuid() {
```
mikaelstaldal: What's the point of this?

garydgregory: None that I can think of...

kpm1985:
Remove eclipse warning for the unused field unless you think the getter is
a big deal

mikaelstaldal:
Such warning is bogus, try to disable it rather than adding an unnecessary method to the code base.

garydgregory:
Agreed.

kpm1985:
Done

kpm1985:
Please let me know if you have any other feedback, I have removed the useless getter method.

kpm1985:
I've also made some more changes to reflect consistent non wildcard imports, spacing of imports, etc that makes the core test package look a little cleaner.

kpm1985:
I will create a new patch and pull request that addresses the minimal set of changes to close LOG4J2-1997. Although the changes are fairly trivial, I realize that 27 files might be too large. Thank you for feedback on my previous PR.

PR #105, opened 2017-08-03
Marker examples should not use deprecate flow APIs

fabriziocucci:


garydgregory:
Our base requirement is still Java 7. So we cannot just yet take this whole patch unless we split the examples into Java 7 and Java 8. I would like to take the changes for the entry APIs though.

fabriziocucci:
@garydgregory sorry, I completely missed the requirements section:

> Log4j 2.4 and greater requires Java 7, versions 2.0-alpha1 to 2.3 required Java 6.

I will revert all Java 8 stuff and keep the entry API changes as you suggested.

Out of curiosity, is there any plan to ditch Java 7 support?

fabriziocucci:
Done!

jvz:
No real plan to upgrade to Java 8 for at least another year or so I believe. Too many people still depending on 7 as a baseline. It would be neat if there were some creative way to support multiple versions of Java similar to how Babel works for JS, though.

garydgregory:
Fixed in Git master and tracked with https://issues.apache.org/jira/browse/LOG4J2-2029

fabriziocucci:
I guess this one can be closed now!

PR #107, opened 2017-08-14
Mark FileRenameAction as successful when using alternative ways to mo…

benji:
If the primary strategy for moving a file (Files.move) fails, then the code tries other strategies but never returned successfully.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/FileRenameAction.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/FileRenameAction.java
@@ -129,9 +129,9 @@ public class FileRenameAction extends AbstractAction {
                         try {
                             Files.copy(Paths.get(source.getAbsolutePath()), Paths.get(destination.getAbsolutePath()),
                                     StandardCopyOption.REPLACE_EXISTING);
-                            result = true;
                             try {
                                 Files.delete(Paths.get(source.getAbsolutePath()));
+                                result = true;
                                 LOGGER.trace("Renamed file {} to {} using copy and delete",
                                         source.getAbsolutePath(), destination.getAbsolutePath());
                             } catch (final IOException exDelete) {
@@ -139,6 +139,7 @@ public class FileRenameAction extends AbstractAction {
                                         exDelete.getClass().getName(), exDelete.getMessage());
                                 try {
                                     new PrintWriter(source.getAbsolutePath()).close();
+                                    result = true;
                                     LOGGER.trace("Renamed file {} to {} with copy and truncation",
                                             source.getAbsolutePath(), destination.getAbsolutePath());
                                 } catch (final IOException exOwerwrite) {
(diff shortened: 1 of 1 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/FileRenameAction.java
```diff
@@ -124,11 +124,12 @@ public static boolean execute(final File source, final File destination, final b
                 } catch (final IOException exMove) {
                     LOGGER.error("Unable to move file {} to {}: {} {}", source.getAbsolutePath(),
                             destination.getAbsolutePath(), exMove.getClass().getName(), exMove.getMessage());
-                    final boolean result = source.renameTo(destination);
+                    boolean result = source.renameTo(destination);
                     if (!result) {
                         try {
                             Files.copy(Paths.get(source.getAbsolutePath()), Paths.get(destination.getAbsolutePath()),
                                     StandardCopyOption.REPLACE_EXISTING);
+                            result = true;
```
garydgregory: Shouldn't this "result = true" be done _after_ this method successfully deletes the file?

benji: Maybe, I considered it but in my case I'd rather compress the destination file even if the source file has not been deleted successfully. I can change that if you want to only succeed when the delete has been successful has well.

benji:
Changes are ready, let me know.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/FileRenameAction.java
```diff
@@ -124,20 +124,22 @@ public static boolean execute(final File source, final File destination, final b
...
                                 Files.delete(Paths.get(source.getAbsolutePath()));
+                                result = true;
                                 LOGGER.trace("Renamed file {} to {} using copy and delete",
                                         source.getAbsolutePath(), destination.getAbsolutePath());
                             } catch (final IOException exDelete) {
                                 LOGGER.error("Unable to delete file {}: {} {}", source.getAbsolutePath(),
                                         exDelete.getClass().getName(), exDelete.getMessage());
                                 try {
                                     new PrintWriter(source.getAbsolutePath()).close();
+                                    result = true;
```
garydgregory: I would not set result to true here since the rename is incomplete at this point, as the logging statement above says "Unable to delete file". The result should be atomic IMO: true, the file was renamed. false: not done, or incomplete.

mikaelstaldal:
What is the status of this PR @garydgregory ?

garydgregory:
Status is WIP it seems since there was no reply to my Aug 15 comment.

benji:
No changes have been made:
https://github.com/benji/logging-log4j2-LOG4J2-2012/commit/f42b27fa5e2b5b3092960cb6dcabd8f8723b8b8c
Sorry I should have updated this thread.

rgoers:
Closing this. Patch was already applied.

PR #109, opened 2017-09-01
Fix stack locator calc location on JDK 9

jasontedor:
This commit fixes a bug in StackLocator#calcLocation on JDK 9. The particular issue here is that on JDK 9, all stack trace locations report as line 71 of StackLocatorUtil. This is due to a bug in the JDK 9 implementation of StackLocator. The bug is that instead of dropping the top frames of the stack until the first frame that matches the fully-qualified class name of the logger, the implementation would drop all frames from the top that match the fully-qualified class name of the logger. Of course, at this point in the stack trace, there would be none. The fix is to reverse the condition, that we drop all frames until we reach a frame matching the fully-qualified class name of the logger, and then drop all frames matching the fully-qualified class name of the logger. This commit also adds a test that was broken before this change and now passes.

### DIFF
--- a/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
+++ b/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
@@ -68,8 +68,12 @@ public class StackLocator {
     }
 
     public StackTraceElement calcLocation(final String fqcnOfLogger) {
-        return stackWalker.walk(s -> s.dropWhile(f -> f.getClassName().equals(fqcnOfLogger)).
-                dropWhile(f -> f.getClassName().equals(fqcnOfLogger)).findFirst()).get().toStackTraceElement();
+        return stackWalker.walk(
+                s -> s.dropWhile(f -> !f.getClassName().equals(fqcnOfLogger)) // drop the top frames until we reach the logger
+                        .dropWhile(f -> f.getClassName().equals(fqcnOfLogger)) // drop the logger frames
+                        .findFirst())
+                .get()
+                .toStackTraceElement();
     }
 
     public StackTraceElement getStackTraceElement(final int depth) {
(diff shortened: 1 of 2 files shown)


jasontedor:
This is for: https://issues.apache.org/jira/browse/LOG4J2-2028

rgoers:
Finding the frame prior to the first frame that matches the class name of the logger is likely to return a frame that still points at the logger class because multiple methods in the logger may be called.
What the original code is trying to do is find the first stack frame that matches the fully-qualified class name of the logger and then skip all the stack frames for that logger.  If it isn't doing that correctly then it obviously needs to be fixed.

jasontedor:
@rgoers Thanks for the feedback, that makes sense! I will do this:
 - make a change that will make that work
 - add a comment to the code that explains why the code is doing what it's going to be doing
 - update the test to account for that situation

rgoers:
@jasontedor Thanks for contributing to Log4j - we appreciate all the help we can get!

jasontedor:
You're welcome @rgoers, it is my pleasure. I pushed a commit as we discussed.

MartyIX:
@jasontedor 

> This is due to a bug in the JDK 9 implementation of StackLocator.

Do you happen to have a bug report link?

Thank you :)

jasontedor:
@MartyIX The bug report is linked [above](https://github.com/apache/logging-log4j2/pull/109#issuecomment-326626895). The bug manifests as that the source line on any logging event will appear to come from stack locator itself, which is clearly a mistake. This also impacts the use of the `%l` pattern. This bug only manifests on JDK 9.

jasontedor:
Thanks @rgoers.

MartyIX:
@jasontedor Does Log4j2 2.9.1 (with JDK 9) work correctly for you? All my log messages contain `calcLocation` method name. Am I the only one?

When I switch to JDK 8, it works as expected

rgoers:
I just tried running one of my webapps on tomcat under Java 9. Although I didn't have much success due to Spring (that is a different story) I do see the location information correctly.

jasontedor:
I can confirm that as well, both the unit tests we have that rely on this are passing now, and the application is logging the correct location when `%l` is used as a pattern.

PR #110, opened 2017-09-12
JsonLayout support for custom key-value pairs being inserted into JSON

mdvorak:
_Note: This is not ready PR, there are no unit tests for new feature - please review before i'll invest more time into it._

* Changed `AbstractJacksonLayout.convertMutableToLog4jEvent` to `protected Object wrapLogEvent(LogEvent event)`
* Added Extras as `KeyValuePair[]` to `JsonLayout`
* When extras as specified, `LogEvent` is wrapped in new `LogEventWithAdditionalFields` class, which merges `LogEvent` serialized data with provided extras

This is generally wanted feature, for example see here:
https://github.com/ggrandes/log4j2-simplejson
https://github.com/majikthys/log4j2-logstash-jsonevent-layout

How it works:
```xml
<JsonLayout>
  <KeyValuePair key="hostName" value="${hostName}"/>
</JsonLayout>
```
Inserts into each generated JSON `hostName` key:
```json
{
"timeMillis":1505228896690,
"thread":"main",
"level":"INFO",
"loggerName":"org.springframework....",
"message":"Refreshing ...",
"endOfBatch":false,
"loggerFqcn":"org....",
"contextMap":{},
"threadId":1,
"threadPriority":5,
"hostName":"wp21257b"
}
```

This is accomplished by tricks with Jackon json serializer:
```java
public static class LogEventWithAdditionalFields {
    // ....
    @JsonUnwrapped // Deserializes actual LogEvent into root object
    public Object getLogEvent() { return logEvent; }

    @JsonAnyGetter // Adds everything from the map into root object
    public Map<String, Object> getAdditionalFields() { return additionalFields; }
}
```

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout.java
@@ -40,10 +40,10 @@ abstract class AbstractJacksonLayout extends AbstractStringLayout {
 
         @PluginBuilderAttribute
         private boolean eventEol;
-        
+
         @PluginBuilderAttribute
         private boolean compact;
-        
+
         @PluginBuilderAttribute
         private boolean complete;
 
@@ -199,7 +199,7 @@ abstract class AbstractJacksonLayout extends AbstractStringLayout {
         }
     }
 
-    private static LogEvent convertMutableToLog4jEvent(final LogEvent event) {
+    protected Object wrapLogEvent(final LogEvent event) {
         // TODO Jackson-based layouts have certain filters set up for Log4jLogEvent.
         // TODO Need to set up the same filters for MutableLogEvent but don't know how...
         // This is a workaround.
@@ -210,7 +210,7 @@ abstract class AbstractJacksonLayout extends AbstractStringLayout {
 
     public void toSerializable(final LogEvent event, final Writer writer)
             throws JsonGenerationException, JsonMappingException, IOException {
-        objectWriter.writeValue(writer, convertMutableToLog4jEvent(event));
+        objectWriter.writeValue(writer, wrapLogEvent(event));
         writer.write(eol);
         if (includeNullDelimiter) {
             writer.write('\0');
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout.java
@@ -19,17 +19,27 @@ package org.apache.logging.log4j.core.layout;
 import java.io.IOException;
 import java.io.Writer;
 import java.nio.charset.Charset;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginElement;
 import org.apache.logging.log4j.core.impl.MutableLogEvent;
+import org.apache.logging.log4j.core.jackson.XmlConstants;
+import org.apache.logging.log4j.core.lookup.StrSubstitutor;
+import org.apache.logging.log4j.core.util.KeyValuePair;
 import org.apache.logging.log4j.core.util.StringBuilderWriter;
 import org.apache.logging.log4j.util.Strings;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import com.fasterxml.jackson.annotation.JsonRootName;
+import com.fasterxml.jackson.annotation.JsonUnwrapped;
 import com.fasterxml.jackson.core.JsonGenerationException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectWriter;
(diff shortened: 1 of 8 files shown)


garydgregory:
I wonder if the KeyValuePair[] should be pulled up from GelfLayout up to its superclass AbstractStringLayout. This would let other layouts like AbstractJacksonLayout use it since it also subclasses AbstractStringLayout.

mdvorak:
@garydgregory You are right, there is already `AdditionalField`, i'll rename it.
There is however problem with pulling it up to `AbstractStringLayout`, since its used by many layouts where would be problematic (or impossible) to use it. Thats why I chose to implement it directly to JsonLayout.

garydgregory:
You're right, we do not want to say that all subclasses can or should implement extra fields.

garydgregory:
Note that in the GelfLayout, the field is called "AdditionalField". It seems like a good idea to use the same name. Also, don't forget to update /src/site/xdoc/manual/layouts.xml.vm

garydgregory:
You should not change the createLayout() method which is already deprecated. Adding new features is one of the main reasons we switched to the builder pattern.

mdvorak:
Rolled back change of createLayout in eeebf4e, you are indeed correct about that one.

mikaelstaldal:
Linking to JIRA issue: https://issues.apache.org/jira/browse/LOG4J2-1694

mikaelstaldal:
It would be nice to add this to XmlLayout and YamlLayout also.

mikaelstaldal:
Have a look at the unfinished branch https://github.com/apache/logging-log4j2/tree/LOG4J2-1694 you can probably pick suitable unit tests from there.

mdvorak:
@mikaelstaldal Thanks, I looked for existing issue but somehow missed this one. I'll merge the branches and we'll see whether it will work for all cases. I did not know how to make it work for XML, did the ObjectMapper.withAttributes work in XML? Thanks

mdvorak:
@mikaelstaldal I've looked at `GelfLayout`, and it resolves `${}` expressions during serialization. Is this really nescessary? Looks like pretty big overhead to me. I've tested only XML config, but expressions are resolved automatically when parsed. Won't there be a possible conflict? I don't see so far.

mdvorak:
@mikaelstaldal I've merged your tests, thanks for that! That saved some time.

Anyway, I've pulled up my changes in `JsonLayout` to `AbstractJacksonLayout` - annotations, despite being named `@Json*` works both for XML and YAML as well.

Please take a look, it looks promising, hopefully it won't break anything (I don't fully trust the `@JsonUnwrapped` annotation, but seems to be working atm).

mdvorak:
Link to JIRA issue:
https://issues.apache.org/jira/browse/LOG4J2-1694

garydgregory:
Hi,

Thank you for the PR. Did you run 'mvn clean install' successfully?

mdvorak:
Partially, log4j-core completed successfully including all tests, whole project failed somewhere on Cassandra with some JNI exception. I did not look further into that yet. Does not seem related.

mikaelstaldal:
@mdvorak Resolving `${}` expressions during serialization is necessary to support dynamic lookups which depend on the current log event, such as [ContextMapLookup](https://logging.apache.org/log4j/2.x/manual/lookups.html#ContextMapLookup).

mdvorak:
@mikaelstaldal Yeah i figured after digging deeper. Implemented it with correct support.

Only thing that bugs me is that dynamically created LinkedHashMap - thats wasteful and expensive (I know Jackson alone is not exactly garbage-free, but still).

I'm thinking about using `ThreadLocal`s, but I understand there are both potentional memory leaks, and Log4j alone has special policies about that (`log4j2.enable.threadlocals`), and I don't have atm time to dive into that. Any ideas? Or do you consider it fine as it is?

mikaelstaldal:
@mdvorak I think that we should not care so much about garbage and micro optimizations in the Jackson based layouts, since they are far from garbage-free or optimal anyway.

Keep it as is, and don't bother with ThreadLocals.

If we want to make JsonLayout garbage-free, we would have to stop using Jackson and do something similar to GelfLayout.

mikaelstaldal:
@mdvorak The code looks fine to me. But we will wait a while before merging, since we plan to do a 2.9.1 bugfix release soon. This will go into 2.10.0.

mdvorak:
This is build log, it hangs after that.. But i'm not sure why I'm seeing this exception.
https://gist.github.com/anonymous/9909ea92d10ac4ebe36f27a0610e7c2d

mdvorak:
@mikaelstaldal is there any estimation on 2.10 release date? Obviously I need this :)

garydgregory:
We are in the process of releasing 2.9.1, anything beyond that does not have a hard set date yet.

mikaelstaldal:
2.9.1 is now released, and the next release will (most likely) be 2.10.0 which will include this. We have no date set for that release, but historically there has been at least three months between minor releases of Log4j 2.x, and 2.9.0 was released one month ago.

dankirkd:
This is a needed feature.  Can a release go out soon that will include it?

dankirkd:
Does this support variable replacement.

i.e. in PatternLayout one can do something like this:

`<Pattern>%d{ISO8601}{UTC} %d{Z}{UTC} [CID:%X{conversation-id}] [%t] %-5p %c{1.} - %m%n</Pattern>`

But it seems that 

`<KeyValuePair key="CID" value="%X{conversation-id}"/>`

or even

`<KeyValuePair key="CID" value="${conversation-id}"/>`

doesn't work.

PR #111, opened 2017-09-29
LOG4J2-1216: Nested pattern layout options broken

zsomborklara:
Change is for the most part as suggested by the original reported of the issue in the Jira, I only added a unit test. I made one small modification:
```
if (depth > 0) {          // option not closed, continue with pattern on opening "{"
    return begin-1;
}
```
Turned to:
```
if (depth > 0) {          // option not closed, continue with pattern after closing bracket
   return pattern.indexOf('}', start) + 1;
}
```
Because the original version was breaking PatternParserTest#testCustomPattern.

Also it seems there are 3 unit test in core that are broken on head.


garydgregory:
Reference: https://issues.apache.org/jira/browse/LOG4J2-1216

mikaelstaldal:
Is this PR still relevant? @garydgregory ?

PR #112, opened 2017-09-30
LOG4J2-2062 Added property to KafkaAppender to send a Key to Kafka

flowcont:


mikaelstaldal:
@flowcont I think it would make more sense to have the value for the key as an attribute on the Kafka appender element, and not as a nested property, since the nested properties are passed directly to the [Kafka client library](http://kafka.apache.org/documentation.html#producerconfigs).

So you can specify it like this:

```
  <Appenders>
    <Kafka name="Kafka" topic="log-test" key="myKey">
      <PatternLayout pattern="%date %message"/>
      <Property name="bootstrap.servers">localhost:9092</Property>
    </Kafka>
  </Appenders>
```

mikaelstaldal:
It would also be useful to enable interpolation with [lookups](https://logging.apache.org/log4j/2.x/manual/lookups.html) for the key, so that you could e.g. use some value from the thread context:

```
    <Kafka name="Kafka" topic="log-test" key="$${ctx:key}">
```

See e.g. [here](https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/HttpURLConnectionManager.java#L95) about how to do that.

mikaelstaldal:
And don't forget to update documentation in `appenders.xml`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaManager.java
```diff
@@ -47,18 +47,24 @@
...
         this.syncSend = syncSend;
         config.setProperty("key.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer");
         config.setProperty("value.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer");
         config.setProperty("batch.size", "0");
         for (final Property property : properties) {
+
             config.setProperty(property.getName(), property.getValue());
         }
+
+        this.key = (key != null ) ? key.getBytes() : null ;
```
mikaelstaldal: `key.getBytes()` will use some platform default charset. It's not good to rely on that, better to specify charset explicitly.

log4j-core/src/test/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaAppenderTest.java
```diff
@@ -135,6 +135,22 @@ public void testAsyncAppend() throws Exception {
...
+        final Appender appender = ctx.getRequiredAppender("KafkaAppenderWithKey");
+        final LogEvent logEvent = createLogEvent();
+        appender.append(logEvent);
+        final List<ProducerRecord<byte[], byte[]>> history = kafka.history();
+        assertEquals(1, history.size());
+        final ProducerRecord<byte[], byte[]> item = history.get(0);
+        assertNotNull(item);
+        assertEquals(TOPIC_NAME, item.topic());
+        String msgKey = item.key().toString();
+        byte[] keyValue = "key".getBytes();
```
mikaelstaldal: Specify same charset here also.

src/site/xdoc/manual/appenders.xml
```diff
@@ -1735,6 +1735,11 @@ public class JpaLogEntity extends AbstractLogEventWrapperEntity {
               <td>The Kafka topic to use. Required.</td>
             </tr>
             <tr>
+              <td>key</td>
+              <td>String</td>
+              <td>The key that will be sent to Kafka with every message.</td>
```
mikaelstaldal: Should probably mention that it's optional, and what default is.

src/site/xdoc/manual/appenders.xml
```diff
@@ -1735,6 +1735,11 @@ public class JpaLogEntity extends AbstractLogEventWrapperEntity {
...
+            </tr>
+            <tr>
```
mikaelstaldal: We should add a comment on the `properties` section to not specify `key.serializer` (as well as not specify `value.serializer`).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaManager.java
Abrasha: For example: `java.nio.charset.StandardCharsets.UTF_8`

mikaelstaldal:
:+1:

PR #113, opened 2017-10-11
LOG4J2-2105: ClassLoaderContextSelector more resilient to multiple classloaders

cakofony:
When log4j classes are shared between ClassLoader instances, we
can avoid creating additional contexts.


mikaelstaldal:
It would be nice to have a JIRA issue for this.

cakofony:
updated

cakofony:
It's possible that this is undesirable if we expect containers to provide log4j libraries to child contexts.

cakofony:
After reading the docs [here](https://logging.apache.org/log4j/2.x/manual/appenders.html#FileAppender) I think this change would break things. Closing.

PR #115, opened 2017-10-11
[LOG4J2-2036] CompositeConfiguration supports Reconfiguration

cakofony:
Regression was introduced by the fix for LOG4J2-1912.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/composite/CompositeConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/composite/CompositeConfiguration.java
@@ -151,15 +151,15 @@ public class CompositeConfiguration extends AbstractConfiguration implements Rec
             final ConfigurationSource source = config.getConfigurationSource();
             final URI sourceURI = source.getURI();
             Configuration currentConfig;
-            if (sourceURI == null) {
-                LOGGER.warn("Unable to determine URI for configuration {}, changes to it will be ignored",
-                        config.getName());
+            if (sourceURI != null) {
                 currentConfig = factory.getConfiguration(getLoggerContext(), config.getName(), sourceURI);
                 if (currentConfig == null) {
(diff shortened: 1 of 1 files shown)


cakofony:
@garydgregory issue appears to have been introduced here: 5b4f3db4f5e7b24862c147b790806bca0b2d1892

cakofony:
@garydgregory The fix you provided also seems broken, after a couple reloads I'm seeing:

```

--
at org.apache.logging.log4j.core.config.composite.DefaultMergeStrategy.mergConfigurations(DefaultMergeStrategy.java:139)
at org.apache.logging.log4j.core.config.composite.CompositeConfiguration.setup(CompositeConfiguration.java:122)
at org.apache.logging.log4j.core.config.AbstractConfiguration.initialize(AbstractConfiguration.java:235)
at org.apache.logging.log4j.core.config.AbstractConfiguration.start(AbstractConfiguration.java:249)
at org.apache.logging.log4j.core.LoggerContext.setConfiguration(LoggerContext.java:545)
at org.apache.logging.log4j.core.LoggerContext.onChange(LoggerContext.java:667)
at org.apache.logging.log4j.core.config.ConfiguratonFileWatcher$ReconfigurationRunnable.run(ConfiguratonFileWatcher.java:68)
```

Also, if factory.getConfiguration returns null, currentConfig is left null instead of currentConfig.

garydgregory:
Avoiding the NPE is easy but that might not solve the actual underlying problem. Do you have a unit test for us? Or solution patch?

cakofony:
@garydgregory The change in this PR does not cause the failure, the change that was merged instead introduced the bug.

I have a repo here which I've used to reproduce the issue: https://github.com/cakofony/log4j2-composite-config-repro

I'm not sure the best way to add such a test in this codebase (it gets a bit funky to write out a file, then modify it at runtime, and clean up). I can take a look at putting something together though

PR #119, opened 2017-10-22
LOG4J2-2086 Sort Import Statments with Maven

kpm1985:
I've started a discussion on JIRA here:
https://issues.apache.org/jira/browse/LOG4J2-2086

It is not possible to  include this in the root pom afaik because it complains about modules, but it works fine in the others. A committer can play around with this like I have its really cool!

### DIFF
--- a/log4j-api/pom.xml
+++ b/log4j-api/pom.xml
@@ -221,6 +221,26 @@
         <artifactId>maven-deploy-plugin</artifactId>
         <version>2.8.2</version>
       </plugin>
+      <plugin>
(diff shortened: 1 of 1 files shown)


garydgregory:
Most IDEs do this already just fine.

kpm1985:
@garydgregory This check forces ordering of import statements regardless if the developer uses that functionality of the IDE or not, and it does it at build time. Most people are lazy, this covers everyone. In addition, it sorts all of the old code out too, preventing any possible future need to hear about import ordering. Its fantastic.

PR #120, opened 2017-10-26
LOG4J2-2091: Log4j respects the configured "log4j2.is.webapp" property

cakofony:
Some applications run embedded servlet containers sharing
a single logging context from the root application.


jvz:
This PR looks good, though could you associate this with a JIRA ticket? Also, a unit test would be great to help avoid this bug from resurfacing in the future. Thanks!

cakofony:
Hi @jvz, thanks for the quick reply!

I've added a jira ticket, but I'm not sure the best way to add real tests for this given how this is configured as static state. Any chance you could point me in the right direction?

jvz:
I think you could add a simple test in log4j-web that asserts `Constants.IS_WEB_APP` if there isn't one already.

cakofony:
Updated, thanks for the feedback!

jvz:
Looks great, thanks! Anyone else on the team is welcome to merge this as I won't be around an appropriate computer until tomorrow at the earliest.

PR #121, opened 2017-10-30
(doc) Remove wrong xml tag "<Log4j1XmlLayout>"

dr0i:
This is a trivial correction of the documentation, removing an old log4j1-tag from the log4j2-config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ sudo: true
 dist: trusty
 
 jdk:
-  - oraclejdk7
+  - oraclejdk8
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ sudo: true
 dist: trusty
 
 jdk:
-  - oraclejdk8
+  - oraclejdk9
 
 env:
   global:
--- a/src/site/xdoc/manual/migration.xml
+++ b/src/site/xdoc/manual/migration.xml
@@ -189,9 +189,7 @@
             <pre class="prettyprint linenums"><![CDATA[<?xml version="1.0" encoding="UTF-8"?>
 <Configuration>
   <Appenders>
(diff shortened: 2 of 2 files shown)


rgoers:
I'm not sure that this is really a problem. The Log4j-1.2-api provides a Log4j1XmlLayout to be compatible with Log4j 1.x.

mikaelstaldal:
This is not a problem, the documentation is correct as it is.

dr0i:
I see. It's just that if I try to maven-build with the Log4j1XmlLayout, I got:

`2017-11-02 08:56:11,394 main ERROR File contains an invalid element or attribute "Log4j1XmlLayout"`

But it seems to log ok. Also, it could be that I am doing something wrong because I got these WARN  messages just before that:

```
   log4j:WARN Please initialize the log4j system properly.
   log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
```

remkop:
The `log4j:WARN` messages are generated by a log4j-1.2.x.jar that is somewhere in the classpath, not by Log4j 2.

dr0i:
@remkop Right. Appreciate your help (also on SO :) ). I tidied up a bit and got rid of these messages. The `main ERROR File contains an invalid element ` stays, but as the logging _is_ working it's ignorable.

mikaelstaldal:
@dr0i I would suggest you post to the [log4j-user mailing list](http://logging.apache.org/log4j/2.x/mail-lists.html) for support.

Do you have a configuration with a Log4j1XmlLayout in it? If so you need to have log4j-1.2-api on your classpath, since that layout is there (not in log4j-core as most other layouts).

dr0i:
@mikaelstaldal Yes, I have `log4j-1.2-api` on the classpath. Anyway, I don't need the `Log4j1XmlLayout` - I was just wondering why a docu-snippet wasn't working right out of the box.

PR #124, opened 2017-11-06
LOG4J2-2101: MapMessage correctly handles non-string map values

cakofony:
Previously formatting non-string values worked, however filtering,
which uess the "get" method would fail due to MapMessage attempting
to cast values to String.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/MapMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/MapMessage.java
@@ -200,7 +200,7 @@ public class MapMessage<M extends MapMessage<M, V>, V> implements MultiformatMes
      * @param map The Map to add.
      */
     public void putAll(final Map<String, String> map) {
-        for (final Map.Entry<String, ?> entry : map.entrySet()) {
+        for (final Map.Entry<String, String> entry : map.entrySet()) {
             data.putValue(entry.getKey(), entry.getValue());
         }
     }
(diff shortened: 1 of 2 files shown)


log4j-api/src/test/java/org/apache/logging/log4j/message/MapMessageTest.java
```diff
@@ -99,4 +102,29 @@ public void testMutableByDesign() { // LOG4J2-763
...
+    @Test
+    public void testRemoveNonStringValue() {
+        final String key = "Key";
+        final MapMessage<?, Object> msg = new MapMessage<>()
+                .with(key, 1L);
+        assertEquals("1", msg.remove(key));
+    }
+
+    @Test
+    public void testJSONFormatNonStringValue() {
```
mikaelstaldal: Would be nice to test XML format of non-String value as well.

cakofony: Updated. Good call, turns out it was broken!

log4j-api/src/main/java/org/apache/logging/log4j/message/MapMessage.java
```diff
@@ -200,7 +200,7 @@ public void put(final String key, final String value) {
      * @param map The Map to add.
      */
     public void putAll(final Map<String, String> map) {
```
cakofony: perhaps this should take `Map<String,  V>` instead. I'd prefer to make that change separately though

cakofony:
I would like to update MapMessage to take advantage of `PatternFormatter.recursiveDeepToString` for more efficient value handling, but I think that makes sense in a follow up commit.

I've got several MapMessage changes in review at the moment, so it's difficult to push other fixes without stomping on my own feet!

PR #125, opened 2017-11-06
LOG4J2-2102: MapMessage Json encoding escapes key and value strings

cakofony:


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/EncodingPatternConverter.java
```diff
@@ -86,100 +87,4 @@ public void format(final LogEvent event, final StringBuilder toAppendTo) {
...
-        };
-
-        /**
-         * Escapes text using a standardized format from a given starting point to the end of the string.
-         *
-         * @param toAppendTo string buffer to escape
-         * @param start      where to start escaping from
-         */
-        abstract void escape(final StringBuilder toAppendTo, final int start);
-    }
```
cakofony: Alternatively I can pull the json implementation into a shared utility class elsewhere.

mikaelstaldal:
We should do the same for the XML format, you can use the new XML format in `EncodingPatternConverter` for that (pull from master). Good to place that in a shared utility class.

cakofony:
Updated, thanks!

cakofony:
Feel free to close this in favor of #129 if you prefer to iterate there.

PR #127, opened 2017-11-09
LOG4J2-2109: Added "log4j.formatMsgNoLookups" global configuration

cakofony:
This allows applications to globally disable message pattern
from looking up replacements without specifying "%m{nolookups}".


jvz:
Could you rename this option to `log4j2.formatMsgNoLookups` or something similar? We have a new naming scheme in 2.10.0+ which follows `log4j2.camelCasePropertyName`.

cakofony:
Updated, thanks for the info @jvz!

cakofony:
There wasn't any test coverage for lookups in `MessagePatternConverterTest` so I've added a couple.

mikaelstaldal:
Please update documentation in `configuration.xml.vm`.

cakofony:
Sorry about that, updated

PR #128, opened 2017-11-09
LOG4J2-2110 adding a ServletContext parameter to allow the contextIni…

thomasbeauvais:
Please refer to https://issues.apache.org/jira/browse/LOG4J2-2110


jvz:
Could you reformat this using spaces instead of tabs?

rgoers:
I'd be hesitant to apply this without understanding why contextInitialized isn't being called.

thomasbeauvais:
I have fixed the whitespace. 

What can I do to help you figure out why it wasn't being initialized? A colleague and myself spent a couple hours debugging to find this. It very strange behavior and seems to have been there for quite sometime. We have added this exact fix, and now everything is started and stopped properly each time.

Out problem was that we are starting up a few of resources. Without this fix, we noticed leaks on redeployment. The stop was never call since the {{servletContext}} in the {{Log4jServletContextListener}} is always {{null}} no matter what you do.

rgoers:
The associated Jira was closed several months ago.

PR #131, opened 2017-11-14
[LOG4J2-2118] Modify MutableLogEvent to preserve message format string.

zslayton:
Addresses [LOG4J2-2118](https://issues.apache.org/jira/browse/LOG4J2-2118)

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MutableLogEvent.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MutableLogEvent.java
@@ -56,6 +56,7 @@ public class MutableLogEvent implements LogEvent, ReusableMessage {
     private String loggerName;
     private Message message;
     private StringBuilder messageText;
+    private String messageFormat;
     private Object[] parameters;
     private Throwable thrown;
     private ThrowableProxy thrownProxy;
@@ -124,6 +125,7 @@ public class MutableLogEvent implements LogEvent, ReusableMessage {
         level = null;
         loggerName = null;
         message = null;
+        messageFormat = null;
         thrown = null;
         thrownProxy = null;
         source = null;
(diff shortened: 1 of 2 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MutableLogEvent.java
```diff
@@ -295,7 +301,7 @@ public Message memento() {
             return message;
         }
         final Object[] params = parameters == null ? new Object[0] : Arrays.copyOf(parameters, parameterCount);
-        return new ParameterizedMessage(messageText.toString(), params);
+        return new ParameterizedMessage(getFormat(), params);
```
zslayton: The original version of this code appears to cause nested pattern interpretation. For example, if you created a message with the pattern `"{}"` and the first parameter was `"json: [{}]"`, calling `memento()` on this event would cause the fully formatted message (`json: [{}]`) to be treated as the format for the `ParameterizedMessage` being constructed here. Because it contains braces, the message would attempt to pull from the parameter list to format it.

mikaelstaldal:
Is this PR still relevant? @remkop ?

uncleyo:
Facing the same issue. It stands in our way of having nice structured logging, so definitely +1

cakofony:
@uncleyo @zslayton This should be resolved as of [this](https://github.com/apache/logging-log4j2/commit/967bb0f71155aa00acadbd95362a90832bed69f6) commit (included in 2.11.0).

[LOG4J2-2307](https://github.com/apache/logging-log4j2/commit/0738cf970c111a07a7307549b8c16af4d44af8a4) may also be relevant for scenarios where a memento is created (AsyncAppender, for instance)

rgoers:
Closing due to the comments by @carterkozak 2 years ago and no response since.

PR #132, opened 2017-11-15
ParameterFormatter.deepToString checks more special types

cakofony:
This change allows us to avoid StringBuilder usage for primitives.


cakofony:
Added byte (to the StringBuilders helper as well).

I avoided BigDecimal/BigInteger because they're non-final, it's possible they may be implemented to support StringBuilderFormattable.

PR #134, opened 2017-11-16
Remove unnecessary threadlocal in MdcPatternConverter

cakofony:
We can write directly to the supplied StringBuilder.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MdcPatternConverter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MdcPatternConverter.java
@@ -20,7 +20,6 @@ import org.apache.logging.log4j.util.PerformanceSensitive;
 import org.apache.logging.log4j.util.ReadOnlyStringMap;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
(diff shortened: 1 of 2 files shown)


mikaelstaldal:
@remkop what do you think? You added this `ThreadLocal`.

remkop:
The proposed change would result in a trailing comma after the last key-value pair. For that reason I would say no. 

If there is a way to avoid the temporary buffer without changing the behavior then that would be good. So I agree with the general sentiment but I don’t see how to accomplish it (yet).

In its current form we cannot accept the pull request.

cakofony:
@remkop MdcPatternConverter.java:126 takes care of that case, I added additional testing to ensure it doesn't break moving forward either.

remkop:
My bad, I missed that. :-/
Let me take another proper look.

cakofony:
No worries, thanks!

remkop:
Ok. Looks good. Tests all pass.

Ralph started the release earlier today so I don't want to interrupt that, but I'll push the merge when the release is done.

remkop:
I created https://issues.apache.org/jira/browse/LOG4J2-2127 for this PR.

remkop:
Fixed in master. Please verify and close the https://issues.apache.org/jira/browse/LOG4J2-2127 ticket.

Thanks for the PR! Keep 'em coming!

PR #137, opened 2017-12-06
PatternFormatter avoids identityToString until collisions detected

cakofony:
This implementation uses an IdentityHashMap backed Set to avoid
building a string and hashing it unnecessarily.


cakofony:
I have not yet

PR #138, opened 2017-12-10
POC: declarative runtime null-checks

denis-zhdanov:
Hello,

I'd like to contribute to log4j by replacing explicit *Objects.requireNonNull()* checks by parameter's *NotNull* annotation and *javac* plugin which enhances resulting bytecode by the corresponding runtime check.  

Example - we mark *AbstractAppender*'s *name* parameter by a *NotNull* annotation - [link](https://github.com/denis-zhdanov/logging-log4j2/commit/69e2f0cc20950b18bbf1e93661e09189913b5a18#diff-0460d1e99cf5e9ce995fd26d05eabe6e) and resulting bytecode looks as below:  

```
javap -c ./log4j-core/target/classes/org/apache/logging/log4j/core/appender/AbstractAppender.class

...
protected org.apache.logging.log4j.core.appender.AbstractAppender(java.lang.String, org.apache.logging.log4j.core.Filter, org.apache.logging.log4j.core.Layout<? extends java.io.Serializable>, boolean);
    Code:
        ...
      18: ifnonnull     31
      21: new           #6                  // class java/lang/NullPointerException
      24: dup
      25: ldc           #7                  // String name
      27: invokespecial #8                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      30: athrow
```

I.e. the suggestion is to replace all *Objects.requireNonNull()* calls in the project by corresponding annotations on method arguments. The checks are inserted by the [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin (I'm its author).  

Please let me know if you're interested in that, I'm ok to provide a PR for the full project source code then.

### DIFF
--- a/log4j-core/pom.xml
+++ b/log4j-core/pom.xml
@@ -339,6 +339,16 @@
       <artifactId>HdrHistogram</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+      <scope>provided</scope><!-- make the jar eligible for compilation only -->
+    </dependency>
+    <dependency>
+      <groupId>tech.harmonysoft</groupId>
+      <artifactId>traute-javac</artifactId>
+      <scope>provided</scope><!-- make the jar eligible for compilation only -->
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AbstractAppender.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AbstractAppender.java
@@ -33,6 +33,7 @@ import org.apache.logging.log4j.core.config.plugins.validation.constraints.Requi
 import org.apache.logging.log4j.core.filter.AbstractFilterable;
 import org.apache.logging.log4j.core.layout.PatternLayout;
 import org.apache.logging.log4j.core.util.Integers;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Abstract base class for Appenders. Although Appenders do not have to extend this class, doing so will simplify their
@@ -146,10 +147,10 @@ public abstract class AbstractAppender extends AbstractFilterable implements App
      * @param ignoreExceptions If true, exceptions will be logged and suppressed. If false errors will be logged and
      *            then passed to the application.
      */
-    protected AbstractAppender(final String name, final Filter filter, final Layout<? extends Serializable> layout,
-            final boolean ignoreExceptions) {
+    protected AbstractAppender(@NotNull final String name, final Filter filter,
+                               final Layout<? extends Serializable> layout, final boolean ignoreExceptions) {
         super(filter);
-        this.name = Objects.requireNonNull(name, "name");
+        this.name = name;
         this.layout = layout;
         this.ignoreExceptions = ignoreExceptions;
     }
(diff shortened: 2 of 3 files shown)


garydgregory:
I am -1: If we want something like this, we should use Project Lombok which is well known and feature rich https://projectlombok.org/features/NonNull

mikaelstaldal:
I do like the general idea in principle. And I do like the work made by Jetbrains in this area.

But I don't think it would be wise to depend on Jetbrains specific annotations in Log4j. And probably not on a compiler plugin [requiring Java 8](https://github.com/denis-zhdanov/traute/blob/master/core/javac/README.md#5-limitations) as long as Log4j support Java 7.

denis-zhdanov:
Hey guys, 

Thanks for the feedback, a couple of notes:
* as [mentioned in the Traute documentation](http://traute.oss.harmonysoft.tech/#3-alternatives), Lombok has a couple of limitations - it supports only Lombok *NonNull* annotation and it's a *APT*-based solution, i.e. it generates additional data (\*.class files) - *javac* plugin is more natural in this area. Regarding the quality - yes, Lombok is more mature, but it also was young and it was necessary to find host projects for it :) *Traute* already has an [extensive test suit](https://github.com/denis-zhdanov/traute/tree/master/core/test/src/test/java/tech/harmonysoft/oss/traute/test/suite) and I'm fully on it to quickly react on any new enhancement requests
* as *java* is backward compatible, it's possible to compile the sources by *javac 1.8* with target version *1.7* (that is current log4j2 configuration, implied by the maven-compiler-plugin 3.7.0, so, there is no problem here)
* there is no requirement to use jetbrains annotations here - *Traute* [supports](http://traute.oss.harmonysoft.tech/core/javac/#7-settings) a number of *NotNull* annotations by default and it's always possible to configure custom annotations to use (btw, jsr-305's *Nonnull* is also supported by default)

rgoers:
All release builds are done with the Java 7 compiler.

denis-zhdanov:
Wow, just curious, why not using javac8?

rgoers:
To be 100% certain that we didn't reference anything from Java 8 by accident.

denis-zhdanov:
But *javac -target 1.7* provides that level of confidence, doesn't it?

garydgregory:
javac -target 1.7 just tells the compiler to generate 1.7 byte codes; it does nothing to avoid calling Java 8, 9 or 10 APIs.

denis-zhdanov:
Sorry, re-checked *javac* documentation and the right argument is *-source*.  

E.g. here is what I get on attempt to compile a source below:  

```java
package org;

import java.util.Arrays;

public class Xxx {
        Arrays.stream(args).forEach(System.out::println);
    }
}
```

```
javac -source 1.7 ./org/Xxx.java
./org/Xxx.java:7: error: method references are not supported in -source 1.7
        Arrays.stream(args).forEach(System.out::println);
                                                ^
  (use -source 8 or higher to enable method references)
1 error
```

remkop:
@denis-zhdanov The 1.8 javac compiler will recognize and reject new language constructs like lambdas, but it cannot check at the API (library) level. 

For example `Files.readAllLines(Path path)` is a new method added to the `Files` class in Java 8. If Log4j accidentally contained code that called that method, javac 8, would compile fine which is not good...

denis-zhdanov:
Makes sense, thanks.

This PR can be closed then as *Traute* is a *javac* plugin and that *API* is available since *javac8*

PR #141, opened 2018-01-16
LOG4J2-2190 support writing JSON object in message field

franzwong:
If we can make message as a Json object, user does not need to deserialize twice (first the whole log event, second the message field).
Also, @JsonSerialize annotation defined in LogEventJsonMixIn prevents user to override the serializer.


mikaelstaldal:
You should update documentation in `layouts.xml.vm`.

franzwong:
`layouts.xml.vm` is updated

AmlanAlok:
> `layouts.xml.vm` is updated

Can someone please share the documentation link here.

PR #142, opened 2018-01-17
[LOG4J2-2197] Document default property value support

fabdouglas:
Complete the documentation about the default environment and system properties syntaxes :
Updated pages are : 
- https://logging.apache.org/log4j/2.x/manual/lookups.html
- https://logging.apache.org/log4j/2.x/manual/configuration.html


remkop:
Looks like this was merged in. If you’re happy with the result, can you close both the JIRA and the pull request?

PR #146, opened 2018-02-04
fixes capitalization of log4j2.ContextDataInjector property name in Documentation

jhovell:
See here:

https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ContextDataInjectorFactory.java#L65


jvz:
It's supposed to be lower case now with the new property naming scheme in use. It would be more useful to update existing code/docs to use the normalized property names instead IMO.

rgoers:
Closing this per Matt's comments.

PR #147, opened 2018-02-06
LOG4J2-2032 Curly braces in parameters should not be treated as placeholders

kshchepanovskyi:
https://issues.apache.org/jira/browse/LOG4J2-2032

When logging event parameter contains two consecutive opening and
closing curly braces - for example {} in the middle of parameter,
they should not be treated as a placeholder.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/MementoMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/MementoMessage.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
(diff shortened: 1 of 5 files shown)


kshchepanovskyi:
Travis build fails with

```
$ ./mvnw --toolchains=./.travis-toolchains.xml install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
Error: JAVA_HOME is not defined correctly.
  We cannot execute /usr/lib/jvm/java-7-oracle/bin/java
The command "./mvnw --toolchains=./.travis-toolchains.xml install -DskipTests=true -Dmaven.javadoc.skip=true -B -V" failed and exited with 1 during .
```

Definitely not related to changes I did.

cakofony:
Thank you for your contribution @kshchepanovskyi! I think this is fixed by a very similar commit for a slightly different permutation of this issue [here](https://github.com/apache/logging-log4j2/commit/0738cf970c111a07a7307549b8c16af4d44af8a4).

I've committed your test and fix for windows.

kshchepanovskyi:
Thanks!

PR #148, opened 2018-02-13
Reusable LogEvents pass along the original format string

cakofony:
This allows custom layouts to group logged messages parameterized
values without creating new single-use messages for each event.


log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
```diff
@@ -55,7 +55,7 @@ public void formatTo(final StringBuilder buffer) {
      */
     @Override
     public String getFormat() {
-        return getFormattedMessage();
+        return obj instanceof String ? (String) obj : null;
```
cakofony: I think `return "{}"` makes more sense here. Thoughts?

cakofony:
My goal is to write a custom layout which logs `Message in a {}` and parameters `["bottle"]` instead of `Message in a bottle`.

This allows us to perform analysis on all parameter values sent to a particular log statement, as well as potentially differentiating logged parameters for redaction.

I'd be happy to chat at length if folks have questions or concerns!

rgoers:
I don't understand why any of these changes are required. If you want access to those in your Layout just call getFormat() and getParameters() instead of getFormattedMessage().

cakofony:
@rgoers That's exactly what I'd like to do, and it works if threadlocals are disabled.

However if threadlocals are enabled (the default for non-webapps), ReusableParameterizedMessage -> MutableLogEvent fails to retain the original format string. This change fixes that case.

regarding `RFC5424Layout`, `StructuredDataMessage` doesn't implement `ReusableMessage` and is annotated with `AsynchronouslyFormattable`, so the original message object is passed along.

cakofony:
The `MutableLogEventTest.testInitFromReusableCopiesFormatString` provides an example of the failure behavior when run against master without this change.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableSimpleMessage.java
```diff
@@ -43,7 +43,7 @@ public String getFormattedMessage() {
 
     @Override
     public String getFormat() {
-        return getFormattedMessage();
+        return charSequence instanceof String ? (String) charSequence : null;
```
remkop: This is wrong. If charSequence is a StringBuilder or CharBuffer, the method would return null after the proposed change.

remkop:
Please ignore my previous comment.
On closer inspection, I realized that `MutableLogEvent::getFormat` and `RingBufferLogEvent::getFormat` currently already return `null` unconditionally, and return `this` from their `getMessage` method, so current custom user code already has to deal with `null` values from `Message::getFormat`.

I'm generally okay with the proposed changes.

I'm not sure that it is a good idea to make the decision on whether to return `null` or not hinge on whether returning a non-null String would create garbage. it may be more consistent to always return `null` from `Message` implementations that do not have embedded curly braces. But the proposed implementation is also ok.

cakofony:
> it may be more consistent to always return null from Message implementations that do not have embedded curly braces. But the proposed implementation is also ok.

I'd be happy to implement that change. My preference is to wait until after this fix merges to implement the refactor.

PR #149, opened 2018-02-14
ReusableMessage implementations no longer allocate for 0-len params

cakofony:


garydgregory:
Thank you for the PR. That's a lot of duplicated code (relatively). Can the empty array and new expression be refactored in a common site?

cakofony:
Yep, updated.

cakofony:
This isn't as important to me now that we have #150 
I can rebase this PR if you think it's still helpful, otherwise feel free to close it out!

PR #150, opened 2018-02-14
LOG4J2-2253: Added ReusableMessage.forEachParameter

cakofony:
This method allows us to iterate over parameters in a
ReusableMessage without array creation.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterConsumer.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterConsumer.java
@@ -1,0 +1,26 @@
+package org.apache.logging.log4j.message;
+
+/**
+ * An operation that accepts two input arguments and returns no result.
+ *
+ * <p>
+ * The third parameter lets callers pass in a stateful object to be modified with the key-value pairs,
+ * so the ParameterConsumer implementation itself can be stateless and potentially reusable.
+ * </p>
+ *
+ * @param <S> state data
+ * @see ReusableMessage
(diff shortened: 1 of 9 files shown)


cakofony:
another option here is to update #149 to hold `ThreadLocal<Object[]>` for 1-3 parameters where we can copy chunks of the parameters array rather than allocating a new one.

That way we don't have to expand the API.

Thoughts?

remkop:
I would like to add this method in a way that doesn’t break backwards compatibility. This can be done by putting the method in a new interface and let all `ReusableMessage` implementation classes also implement this new interface. 

At the usage site, you’d check that the message is an instance of the new interface, and do a type cast so you can call the `forEach` method.

cakofony:
Good call, updated.

Curious of your thoughts on this approach vs reusable threadlocal arrays to copy parameters into for Message.getParameters. That approach would have to copy a few references around, but the cost should be fairly small, and wouldn't necessarily work for messages with more than a few parameters, but those aren't particularly common anyhow. I suppose it's also more dangerous if consumers attempt to hold onto the returned parameter array across thread boundaries.

remkop:
I haven't looked at #149 in detail yet, but wouldn't the ThreadLocal prevent the async logger background thread from seeing the parameter values?

cakofony:
The ThreadLocals would be constant sized arrays only used for calls to getParameters in order to avoid creating a new array.

Though now that I think through that idea it would have the potential for memory leaks unless the arrays are nulled out.

I think I've convinced myself that this PR is the way to go :-)

PR #152, opened 2018-02-24
exposes the location part of the Message to loggers

shawjef3:
This exposes the location part of the Message to loggers, so that if the source location is known at compile time, it can be added to the log with no runtime cost. An example use is the scala loggers which are implemented with macros, and therefore have that information available at no cost to the developer, either.


lynda8spin:
repository_explore_github_marketplace_ci_cta_shown

rgoers:
Closing this as this can now be accomplished by using the builder methods in the Logger class.

PR #153, opened 2018-02-25
LOG4J2-2270 Strings::join([null]) should return EMPTY

mcoolive:
- The javadoc of Strings::join states: 'Null objects or empty strings
within the iteration are represented by empty strings'.
But when call with an array of one element that is null, it returns
"null", where I expect EMPTY ("").
- Add unit tests on Strings::join. One would fail without the change


mcoolive:
Gary Gregory implemented a different fix than the one proposed in the PR (unit tests were copied/pasted). Thanks for the reactivity!
See [commit b562b07](https://github.com/apache/logging-log4j2/commit/b562b07177efb0f4b7f3bf5b55c2b123414f6743).
So this PR can be closed.

garydgregory:
We like contributions! :-)

PR #156, opened 2018-02-27
Update ParameterConsumer index type from short to int

cakofony:
Originally ParameterVisitable was built for ReusableMessage which
provides a "short getParameterCount()" method. Now that the
interface isn't bound to ReusableMessage it might as well use
integer.


cakofony:
based on feedback from @remkop

cakofony:
Backport to `release-2.x` branch is here: #157

PR #161, opened 2018-04-04
LOG4J2-2306: Compile with Error Prone

cakofony:


remkop:
Changes look fine to me.

PR #165, opened 2018-04-14
LOG4J2-2269: ReusableLogEventFactory.release clears MutableLogEvent

cakofony:
Fixes a memory leak where parameter references were held beyond
MutableLogEvent usage.


cakofony:
merged

PR #166, opened 2018-04-16
LOG4J2-2318: AsyncQueueFullMessageUtil logs to StatusLogger

cakofony:
We no longer mutate messages when logging them out of order.


cakofony:
merged

PR #167, opened 2018-04-16
(doc) fix links to java docs, the correct path is log4j-web.

anton-balaniuc:
Java doc links contain incorrect package name _log4j-core_, however the right package name is _log4j-web_.


rgoers:
Log4j-2390 created to track this.

PR #168, opened 2018-04-17
Mongodb2 test rule I/O isolated to the target directory

cakofony:
Access to user home directory may cause test flakes.


jvz:
Does this help fix the jenkins/travis build problem? I'm not exactly sure if it's building with the new settings or not here.

jvz:
From what I've seen, you'll want to mess around with the `~/.embedmongo` directory that gets created.

garydgregory:
The painful part about putting anything in the target folder is that it gets deleted on each build. Downloading Mongo on each build is something we want to avoid IMO. I think we just need to make the user folder work...

cakofony:
Makes sense, closing this one out for now. Will come back to the flake when I have a moment.

PR #170, opened 2018-04-20
[LOG4J2-2322] Custom async ContextSelectors disable location by default

cakofony:
AsyncLoggerContextSelector disables "includeLocation" on loggers
unless specified. Custom ContextSelector implementations will
trigger this logic as well.


cakofony:
merged

PR #172, opened 2018-05-02
Fix for https://issues.apache.org/jira/browse/LOG4J2-2331

mikebski:
Fix for https://issues.apache.org/jira/browse/LOG4J2-2331 - use logger string formatting instead of concatenation


rgoers:
This PR was merged in May but apparently neglected to use the format to trigger github to close it.

PR #173, opened 2018-05-03
[LOG4J2-2332] Unresolved variables in any file appender

Lemongrass3110:
Proposed fix for LOG4J2-2332


jvz:
Hey, first of, thanks for the contribution! Personally, I'm excited to see someone noticed the validation API in the first place. :D

Anyways, I certainly like this idea. In fact, this actually solves a long standing annoyance I've had with the same issue in the past.

garydgregory:

I am not a fan of the name "@NoUnresolvedVariables" because of the two negatives "No" and the "Un" in "Unresolved". Can we come up with a simpler name? Maybe "@ValidateVariables". 

I also wonder if the annotation should have an optional log level so that not ALL validation failures end up as an ERROR event.

jvz:
Maybe `@RequiresVariables` or something like that?

And all the validation annotations so far have all been used to err a plugin factory which seems like a config error to me. As a user, I'd like to see errors related to where I messed up most of all.

garydgregory:
```@RequiresVariables``` is not quite right either IMO, ```@RequiresResolvedVariables``` might be better.

rgoers:
Why not just an attribute of the existing plugin?

Lemongrass3110:
@jvz I debugged my way through and stumbled over it while doing so. I actually wondered why some `@Required` tags are not there for some of the file loggers "file" attribute, while they are there in other cases.

Personally I do not really care for whatever name you guys decide on, it is just the functionality I need. My point of view is just that it sucks to run a java program with priviledges that allow you to create directories and end up creating directories all over the place where you do not actually want them with really weird names.

Regarding backwards compatibility I do not really know if this is a reason to take into account since this is as you said yourself actually a configuration issue that goes around unnoticed for now and should actually be reported/validated before calling mkdirs on it. It might just bring up configuration files that were actually broken all time long and should have been fixed anyway. But those are just my few cents on this topic.

If you decide to make this non-default behaviour:
Is there a way I can make it default behavior without using a configuration file? Because I supply my customers with a framework that uses log4j2 as it's logging utility and they can supply configuration files I do not have any control over and I want to prevent them messing up their directories unintentionally.

I would have been happy enough if I could have intercepted the mkdirs call somehow and cancel it, since I am validating the appenders target directory after the context initialization, but it is too late at this point already.

rgoers:
This shouldn't need to be a system property. it can be an attribute of the configuration element.

garydgregory:
We should consider if the presence of unresolved variables should:
- Log a DEBUG, WARN, ERROR, or FATAL messages
- Fail the the whole config

rgoers:
@garydgregory Agreed. That was what I was questioning in my first comment.

Lemongrass3110:
I think it should at least be error level, since during context initialization by default only StatusLogger is used and only level error or higher is reported to stderr(correct me if I am wrong please).

As for the rest I will leave it uncommented since you guys probably know better how you handled similar problems in the past.

rgoers:
@Lemongrass3110 @jvz @garydgregory After all this time I am not sure where this stands. It now has conflicts that need to be resolved. The PR is over 2 years old and we either need to make the changes needed to merge it or close it.

Lemongrass3110:
> 
> 
> @Lemongrass3110 @jvz @garydgregory After all this time I am not sure where this stands. It now has conflicts that need to be resolved. The PR is over 2 years old and we either need to make the changes needed to merge it or close it.

@rgoers the Log4J team has to decide on how they want to resolve this, as you can read here:
> 
> 
> We should consider if the presence of unresolved variables should:
> 
>     * Log a DEBUG, WARN, ERROR, or FATAL message.
> 
>     * Fail fast or log all unresolved variables.
> 
>     * Fail the the whole config.

I cannot force any decisions onto you.

Lemongrass3110:
@rgoers I tried to resolve the conflicts as good as possible.
Do you want me to move the new classes from `org.apache.logging.log4j.core.config.plugins.validation.*` to `org.apache.logging.log4j.plugins.validation.*` as well?

Can you maybe check the Travis CI status? I saw that master was also failing right now, so I am not really sure, if the problem is because of my changes or simply related to master.

rgoers:
Leaving this open as we may want to look at it again after I get the master build straightened out.

ppkarwasz:

Lemongrass3110:
@ppkarwasz sorry but after almost 5 years and no decision from your team I will not waste any time on this anymore.

PR #174, opened 2018-05-03
[LOG4J2-2333] Handle errors thrown in disruptor ExceptionHandler

cakofony:
Previously this would kill the background thread and stop processing
events.


cakofony:
merged

PR #176, opened 2018-06-06
[LOG4J2-1721] Allow composite configuration for context parameter

fluxroot:
For non-JNDI, allow comma-separated configuration locations (composite configuration syntax) for "log4jConfiguration" context parameter. This pull request implements LOG4J2-1721.


fluxroot:
That was my initial thought as well. However, there have been a couple of issues which made me copy the functionality from Configurator.initialize() over to Log4jWebInitializerImpl.getConfigURIs():

1. log4j-web is an optional dependency and only useful in a servlet container environment. Common code would have to go into log4j-core.
2. Parsing config location in Log4jWebInitializerImpl.getConfigURI() is more complicated. In addition to resolving file URIs, we have to search WEB-INF and also consider servlet context name to find the matching config.

The solution in Java 8 would be to pass a closure to resolve the config location. For Java 7 I guess there's no elegant solution for this problem (or maybe I'm missing something :)). So for now I would leave this fix as is.

PR #177, opened 2018-06-07
Log4j 1.2 api tests do not leak MDC values.

cakofony:


garydgregory:
I do not think this patch should include ignoring Mongo tests.

cakofony:
> I do not think this patch should include ignoring Mongo tests.

Agreed, I've only merged the first commit. I'm curious of your thoughts on the mongo tests, I'd like to disable them until we can fix the flake rate, but I'm not sure when I'll have time to dig into the root cause.

garydgregory:
I would not '@Ignore' the mongo tests. You could use a JUnit Assume that skips the tests if an sys prop is set or something like that. 

Note that the Jenkins builds are fine for now:
- 2.x https://builds.apache.org/view/L/view/Logging/job/Log4j%202%202.x/
- 3.x https://builds.apache.org/view/L/view/Logging/job/Log4j%202%203.x/

The tests pass on my machine but I understand that it can be a pain if they do not on your set up.

cakofony:
They pass about four of every five runs on my machine, but flake in the same way as the travis build every now and then.
I can reproduce consistently by executing `mvn clean && mvn test -pl log4j-mongodb2` in a loop 10 times.

PR #178, opened 2018-06-07
Use Travis CI configuration from master

fluxroot:


jvz:
The travis config has been updated again on master and release-2.x. I don't think this is necessary anymore.

PR #179, opened 2018-06-11
AsyncLoggerConfig optimization

cakofony:
AsyncLoggerConfig behaves exactly like a standard LoggerConfig
when no appenders are present. This potentially avoids unnecessarily
adding events to the background thread when an AsyncLoggerConfig
is used to set level without adding appenders directly.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
@@ -565,6 +565,10 @@ public class LoggerConfig extends AbstractFilterable {
         return Boolean.parseBoolean(includeLocationConfigValue);
     }
 
+    protected final boolean hasAppenders() {
+        return !appenders.isEmpty();
+    }
+
     /**
      * The root Logger.
      */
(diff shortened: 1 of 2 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -565,6 +565,10 @@ protected static boolean includeLocation(final String includeLocationConfigValue
         return Boolean.parseBoolean(includeLocationConfigValue);
     }
 
+    protected final boolean hasAppenders() {
+        return !appenders.isEmpty();
```
remkop: Does this call need to be thread-safe? (It does not seem to be at the moment. Not sure if that is a problem...)

cakofony: The `isEmpty()` implementation is threadsafe `appenderArray.get().length == 0`.

There's a race between this call and iterating over appenders, however in the worst case (no appenders when this is checked, then an appender is added) an even may be logged synchronously, rather than added to the disruptor.

In most cases (on applications I support) LoggerConfig blocks are used to set level, so we may traverse 5-10 configs before reaching one with appenders.

remkop:
Thanks for the clarification. No objections from me!

cakofony:
Thanks for the review!

PR #181, opened 2018-06-12
[LOG4J2-2352] Reusable event message mementos are memoized

cakofony:
Fixed RingBufferLogEvent.memento which returned a
ParameterizedMessage rather than a MememntoMessage.


cakofony:
merged

PR #182, opened 2018-06-12
Apply ThreadContextRule to Log4j1XmlLayoutTest

cakofony:
NDC value is leaking into the test intermittently causing a
comparison to fail.


cakofony:
merged

PR #183, opened 2018-06-22
[LOG4J2-2355] Avoid NullPointerException in PropertiesUtil.reload()

henrik242:
SystemPropertiesPropertySource.forEach(..) uses Property.getProperty(..)
to resolve values.  If that value is a non-String, the value will be
null.  Since `literal` is a ConcurrentHashMap, a put(..) with null value
will yield a NullPointerException.

This is especially hard to debug in the case of e.g. StatusLogger,
which initializes PropertiesUtil as a static variable.  The class is
unable to load, throws a NoClassDefFoundError, and hides the
NullPointerException.

Here's what I got when I had a Property value that was a File:

Caused by: java.lang.NoClassDefFoundError: Could not initialize class org.apache.logging.log4j.util.PropertiesUtil
	at org.apache.logging.log4j.status.StatusLogger.<clinit>(StatusLogger.java:78)
	at org.apache.logging.log4j.LogManager.<clinit>(LogManager.java:60)

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -330,13 +330,17 @@ public final class PropertiesUtil {
                 source.forEach(new BiConsumer<String, String>() {
                     @Override
                     public void accept(final String key, final String value) {
-                        literal.put(key, value);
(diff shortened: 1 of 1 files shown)


jvz:
Oh wow I missed the merge done elsewhere.

cakofony:
I should have commented here -- I've taken care of it and added a test.

henrik242:
Merged, see https://issues.apache.org/jira/browse/LOG4J2-2355

PR #187, opened 2018-07-03
LOG4J2-2312 LOG4J2-2341 Fix jackson layout with async loggers

cakofony:
AsyncLoggerContextSelector RingBufferLogEvents were not properly
handled by the jackson layout. Currently the jackson layout
implementation requires Log4jLogEvent instances, so we must
convert all other LogEvent implementations into these.

### DIFF
--- a/log4j-layout-jackson-json/pom.xml
+++ b/log4j-layout-jackson-json/pom.xml
@@ -55,6 +55,11 @@
       <type>test-jar</type>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.lmax</groupId>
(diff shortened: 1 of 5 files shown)


log4j-layout-jackson-json/pom.xml
```diff
@@ -55,6 +55,11 @@
       <type>test-jar</type>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>com.lmax</groupId>
+      <artifactId>disruptor</artifactId>
+      <scope>test</scope>
+    </dependency>
```
cakofony: Required to construct a RingBufferLogEvent

PR #188, opened 2018-07-03
LOG4J2-2365 Make NameAbbreviator correctly abbreviate first fragments

undefz:


undefz:
@rgoers Could you please review this pull request?
I believe your original logic was broken when implementing LOG4J-1321

PR #189, opened 2018-07-09
[LOG4J2-2368] Recursive logging doesn't clobber cached StringBuidlers

cakofony:


log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractStringLayout.java
```diff
@@ -114,6 +115,10 @@ public B setHeaderSerializer(final Serializer headerSerializer) {
      * @return a {@code StringBuilder}
      */
     protected static StringBuilder getStringBuilder() {
+        if (AbstractLogger.getRecursionDepth() > 1) { // LOG4J2-2368
```
garydgregory: Curious and somewhat unrelated: Why isn't the recursion depth a Integer or AtomicInteger?

cakofony: My guess is Integer would require boxing/unboxing on set/access and AtomicInteger incurs a cost for thread safety. The array is a cheap MutableInt. @remkop could likely provide a definitive answer.

remkop: `AbstractLogger.recursionDepthHolder` is of type `ThreadLocal<int[]>` because that was the simplest data structure that met all the requirements:
* It needs to be thread-local, because when multiple threads are logging concurrently each thread has its own recursion depth
* Incrementing and decrementing the counter happens at least once for each log message so should be fast and not allocate any objects

An alternative would have been to use `ThreadLocal<AtomicInteger>`. This would allow the `AbstractLogger` code to look like `getRecursionDepthHolder().incrementAndGet()` instead of `getRecursionDepthHolder()[0]++`. 

That's perhaps a bit nicer, but I thought it looked strange to hold a thread-safe atomic data structure inside a thread-local. `ThreadLocal<int[]>` felt simpler.

PR #195, opened 2018-07-17
LOG4J2-2389 fix the CacheEntry map in ThrowableProxy#toExtendedStackTrace to be put and gotten with same key

liuwenchn:
- fix the CacheEntry map in method `toExtendedStackTrace` to be put and gotten with same key 
  - `stackTraceElement.toString()` returns a string representation of this stack trace element, just as `MyClass.mash(MyClass.java)`, which is not equal to `className`
- benchmark result:
  - https://github.com/liuwenchn/log4j2-throwableproxy-benchmark/

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
@@ -738,7 +738,7 @@ public class ThrowableProxy implements Serializable {
                     final CacheEntry entry = this.toCacheEntry(stackTraceElement,
                         this.loadClass(lastLoader, className), false);
                     extClassInfo = entry.element;
-                    map.put(stackTraceElement.toString(), entry);
+                    map.put(className, entry);
                     if (entry.loader != null) {
                         lastLoader = entry.loader;
                     }


liuwenchn:
hi  garydgregory, 
      thanks for your reply. We  have thought about what you said. 
 - The main ideas and standpoints are as follows: 
    - Each item in our map corresponds to a class, so the total number of items in the map must not exceed the total number of classes loaded.
    - And then for each item, we have a very small amount of data in there, extendedClassInfo, just a few fields.
    - A class, which has everything in memory that contains all the information about the class. Each item in our map takes up much less memory than its corresponding class.
    - In my benchmarking test [result,](https://github.com/liuwenchn/log4j2-throwableproxy-benchmark/) when make the cache global, about 2 times further improvement than the version of just fixing bug in ThrowableProxy accessing the cache.
- Weighing the pros and cons, we we think that the map should be global

rgoers:
This makes no sense to me. 
1. The cache is meant to keep a record of the cache entries in a single set of chained exceptions. 
1. This logic attempts to optimize the performance of logging exceptions, but exceptions are supposed to be infrequent so optimization isn't important.
1. This will cache stack entries from any exception that happens from anywhere in the program. While there may be a finite limit to that this could get extremely large.

I am +1 on fixing the bug and -1 on the global cache.

xnslong:
@rgoers Thanks for the reply and I see the problem. There are 2 problems:

1. If multiple ClassLoader has loaded classes with the same name, they will interfere each other.
2. We can't clear it from the cache if the class is unloaded.

But I also checked the benchmark. After fixing the bug, the performance has been improved a lot if there are duplicated classes in the same stack. And it will improve further more for all kinds of exceptions if the cache is global, it's very attractive. So I just think if there is a way to solve these problems.

I found the cache map is a map with `String` type key indicating a class. So why not make the key type as `Class`? The `Class` objects will be different if they are loaded by different `ClassLoader`s even with the same qualified name. And as for the `unloading`, we may introduce the `WeakReference` in the key. When a class is unloaded and collected by GC, the reference will receive a notice. We can clear the information from the cache map then.

And after all, I still agree only to fix the bug for this time. If there should be such further optimizations, we can do it later after close study.

Yours.

cakofony:
I would also prefer to avoid a global cache anywhere we can avoid it, however I have noticed when my stacks start encountering failures (temporary network loss, for example) I've seen quite a lot of threads stuck in ThrowableProxy.

While I would prefer not to optimize performance for exceptional cases, I put together a simple benchmark locally which uses GC-free logging with both patterns `%m%n%xEx%n` and `%m%n%ex%n`. The former uses extended traces, the latter does not, which avoids creating a ThrowableProxy instance since I'm using synchronous logging.
The former is logging 2,100 events/second while the latter is logging 161,000 events/second.

I'm worried that we may be creating more work for the application than we expect.

rgoers:
Closing since ThrowableProxy was completely refactored and the change I see here has been made to the source.

PR #196, opened 2018-07-21
Fix LOG4J2-2388 / Blocking thread when logging message in an interrup…

failled:
…ted thread

- Fix unfinite loop in org.apache.logging.log4j.flume.appender.FlumePersistentManager.send when the thread is interrupted (Note : The loop has been completely removed since it does not seems to serve any purpose)
- Preserve Thread interruption status on InterruptedException
- Add unit test for this case


failled:
PR has been merged

PR #201, opened 2018-08-06
workaround flaky "Operation not permitted" failures in mongo tests

bjlaub:
mongo tests intermittenly fail with an "Operation not permitted"
IOException when trying to execute the mongod extracted by
de.flapdoodle.embed.process; this seems to be related to surefire
running tests in parallel. Work around that here by forcing serial
execution of the mongo tests.

### DIFF
--- a/log4j-mongodb2/pom.xml
+++ b/log4j-mongodb2/pom.xml
@@ -89,6 +89,14 @@
           </instructions>
         </configuration>
       </plugin>
+      <!-- workaround flaky "Operation not permitted" failures when running tests in parallel -->
+      <plugin>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <forkCount>1</forkCount>
+          <reuseForks>false</reuseForks>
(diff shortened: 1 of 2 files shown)


garydgregory:
Is this needed for module log4j-mongodb3 as well?

garydgregory:
Also, I thought {{forkCount=1}} was the default?

cakofony:
We do the same thing in log4j-core [here](https://github.com/apache/logging-log4j2/blob/master/log4j-core/pom.xml#L384-L385)

Is this module inheriting from the root pom.xml? https://github.com/apache/logging-log4j2/blob/master/pom.xml#L1101-L1112

cakofony:
@garydgregory I'm going to merge this commit later today given that this is the first time in recent memory that travis has passed tests. Please let me know if you'd prefer that I hold off.

garydgregory:
Go for it.

cakofony:
Merged, thanks!

PR #203, opened 2018-08-08
[LOG4J2-2363] pass object parameter to reusable events

bjlaub:
make ReusableObjectMessage always pass its object as the first parameter

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
@@ -94,22 +94,31 @@ public class ReusableObjectMessage implements ReusableMessage, ParameterVisitabl
     }
 
     /**
-     * This message does not have any parameters, so this method returns the specified array.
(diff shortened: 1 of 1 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
```diff
@@ -100,16 +100,17 @@ public Throwable getThrowable() {
      */
     @Override
     public Object[] swapParameters(final Object[] emptyReplacement) {
+        emptyReplacement[0] = obj;
```
cakofony: I can't imagine that `emptyReplacement` should ever be empty, but it might be worth  doing something along these lines:

```java
if (emptyReplacement.length == 0) {
    Object[] params = new Object[10]; // Default reusable parameter buffer size
    params[0] = obj;
    return params;
}
emptyReplacement[0] = obj;
return emptyReplacement;
```

bjlaub: @cakofony fixed

PR #205, opened 2018-08-08
[LOG4J2-2399] support for async formatting on reusable messages

bjlaub:
Adds support for deferring message content formatting on certain
ReusableMessage implementations to a background thread when async
logging is enabled. This change adds the ability for log events to
retain an instance of a formatter for a message so that the message
content can be formatted later, after the message has already been
recycled.

This is an effort to optimize performance for use cases where formatting
on the logging thread is unnecessary or undesired.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
@@ -30,7 +30,7 @@ import org.apache.logging.log4j.util.StringBuilders;
  * @since 2.6
  */
 @PerformanceSensitive("allocation")
-public class ReusableParameterizedMessage implements ReusableMessage, ParameterVisitable, Clearable {
+public class ReusableParameterizedMessage implements ReusableMessage, ParameterVisitable, Clearable, MessageContentFormatterProvider {
 
     private static final int MIN_BUILDER_SIZE = 512;
     private static final int MAX_PARMS = 10;
@@ -39,8 +39,6 @@ public class ReusableParameterizedMessage implements ReusableMessage, ParameterV
 
     private String messagePattern;
     private int argCount;
-    private int usedCount;
-    private final int[] indices = new int[256];
     private transient Object[] varargs;
     private transient Object[] params = new Object[MAX_PARMS];
     private transient Throwable throwable;
@@ -125,25 +123,28 @@ public class ReusableParameterizedMessage implements ReusableMessage, ParameterV
         this.varargs = null;
         this.messagePattern = messagePattern;
         this.argCount = argCount;
-        final int placeholderCount = count(messagePattern, indices);
-        initThrowable(paramArray, argCount, placeholderCount);
-        this.usedCount = Math.min(placeholderCount, argCount);
+
+        int usedParams = count(messagePattern, null);
+        if (usedParams < argCount && paramArray[argCount - 1] instanceof Throwable) {
+            this.throwable = (Throwable) paramArray[argCount - 1];
+        } else {
+            this.throwable = null;
+        }
     }
 
     private static int count(final String messagePattern, final int[] indices) {
+        if (indices == null) {
+            // if indices is null, we are counting only and don't care about
+            // retaining position information for fast substitution
+            // this is only used to determine if there is a Throwable in the params list
+            return ParameterFormatter.countArgumentPlaceholders(messagePattern);
+        }
+
         try {
             // try the fast path first
             return ParameterFormatter.countArgumentPlaceholders2(messagePattern, indices);
         } catch (final Exception ex) { // fallback if more than int[] length (256) parameter placeholders
             return ParameterFormatter.countArgumentPlaceholders(messagePattern);
-        }
-    }
-
-    private void initThrowable(final Object[] params, final int argCount, final int usedParams) {
-        if (usedParams < argCount && params[argCount - 1] instanceof Throwable) {
-            this.throwable = (Throwable) params[argCount - 1];
-        } else {
-            this.throwable = null;
         }
     }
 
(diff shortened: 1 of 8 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
```diff
@@ -125,12 +123,23 @@ private void init(final String messagePattern, final int argCount, final Object[
...
-        final int placeholderCount = count(messagePattern, indices);
-        initThrowable(paramArray, argCount, placeholderCount);
-        this.usedCount = Math.min(placeholderCount, argCount);
+
+        int usedParams = count(messagePattern, null);
+        if (usedParams < argCount && paramArray[argCount - 1] instanceof Throwable) {
+            this.throwable = (Throwable) paramArray[argCount - 1];
+        } else {
+            this.throwable = null;
+        }
```
cakofony: I wonder if we should check `argCount > 0 && paramArray[argCount - 1] instanceof Throwable` prior to invoking count? Could be worth a benchmark, either way one iteration over a format string isn't very expensive.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
```diff
@@ -355,4 +352,35 @@ public void clear() { // LOG4J2-1583
...
+                indices.set(result);
+            }
+            return ReusableParameterizedMessage.count(messagePattern, result);
+        }
+
+        @Override
+        public void formatTo(String formatString, Object[] parameters, int parameterCount, StringBuilder buffer) {
+            int placeholderCount = computeIndices(formatString);
+            int usedCount = Math.min(placeholderCount, parameterCount);
+            int[] computedIndices = indices.get();
```
cakofony: Can we avoid the second `ThreadLocal.get` here? Might be easiest if we do the logic from `computeIndices` in this method

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
```diff
@@ -355,4 +352,35 @@ public void clear() { // LOG4J2-1583
...
         throwable = null;
     }
+
+    @Override
+    public MessageContentFormatter getMessageContentFormatter() {
+        return formatter;
+    }
+
+    private static final MessageContentFormatter formatter = new MessageContentFormatter() {
+        private final ThreadLocal<int[]> indices = new ThreadLocal<>();
```
cakofony: What happens to these values if nested logging occurs in ParameterFormatter.formatMessage2? See [LOG4J2-1583](https://issues.apache.org/jira/browse/LOG4J2-1583) referenced on `ReusableParameterizedMessage.reserved`. We should write a test for this case.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MutableLogEvent.java
```diff
@@ -204,15 +206,19 @@ public void setLoggerName(final String loggerName) {
...
         }
         return message;
     }
 
     public void setMessage(final Message msg) {
         if (msg instanceof ReusableMessage) {
             final ReusableMessage reusable = (ReusableMessage) msg;
-            reusable.formatTo(getMessageTextForWriting());
+            if (Constants.FORMAT_MESSAGES_IN_BACKGROUND && msg instanceof MessageContentFormatterProvider) {
+                this.messageContentFormatter = ((MessageContentFormatterProvider) msg).getMessageContentFormatter();
```
cakofony: I don't think this quite works for the AsyncLogger case. Using mixed sync/async loggers via configuration, message data flows:

1. ReusableMessage.init(<data>)
2. MutableLogEvent.initFrom(<data, including the reusable message>)
3. MutableLogEvent may be logged to synchronous appenders, then data is copied to the background thread for asynchronous format/append operations. This is done by initializing another MutableLogEvent (on the disruptor ringbuffer) from the existing MutableLogEvent, see AsyncLoggerConfigDisruptor:174.

We can fix this by making MutableLogEvent and RingBufferLogEvent both implement MessageContentFormatterProvider, and check `messageContentFormatter` for null in which case  `reusable.formatTo(getMessageTextForWriting());` should be invoked.

remkop:
@bjlaub @cakofony Please see my comment on the [jira ticket](
https://issues.apache.org/jira/browse/LOG4J2-2399?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=16573738#comment-16573738) about the need for a benchmark that demonstrates the expected performance gain.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
bjlaub: indeed, this fails `NestedLoggingFromToStringTest`. Will take a look

cakofony:
Sorry, I didn't mean to close this out!

The PR number was referenced from a ticket for a fix that was merged. Sorry about that!

bjlaub:
> Sorry, I didn't mean to close this out!

see #208

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
bjlaub: see https://github.com/apache/logging-log4j2/pull/208/commits/e015b4aab968a2aa20dc9a40082d98cfe63528cb in #208

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
bjlaub: see https://github.com/apache/logging-log4j2/pull/208/commits/fc8ed21311d3a428b5a673a1b289ea2624e60cdb in #208

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MutableLogEvent.java
bjlaub: see https://github.com/apache/logging-log4j2/pull/208/commits/fc8ed21311d3a428b5a673a1b289ea2624e60cdb in #208

PR #206, opened 2018-08-08
[LOG4J2-2400] Initial implementation of RedisAppender.

nataliemeurer:
This PR implements a dedicated RedisAppender in log4j2. More details can be found in the log4j2 ticket here: https://issues.apache.org/jira/projects/LOG4J2/issues/LOG4J2-2400?filter=allopenissues

### DIFF
--- a/log4j-redis/pom.xml
+++ b/log4j-redis/pom.xml
@@ -1,0 +1,176 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- ~ Licensed to the Apache Software Foundation (ASF) under one or more ~ contributor license agreements. See the NOTICE 
+  file distributed with ~ this work for additional information regarding copyright ownership. ~ The ASF licenses this file 
+  to You under the Apache License, Version 2.0 ~ (the "License"); you may not use this file except in compliance with ~ the 
+  License. You may obtain a copy of the License at ~ ~ http://www.apache.org/licenses/LICENSE-2.0 ~ ~ Unless required by applicable 
+  law or agreed to in writing, software ~ distributed under the License is distributed on an "AS IS" BASIS, ~ WITHOUT WARRANTIES 
+  OR CONDITIONS OF ANY KIND, either express or implied. ~ See the License for the specific language governing permissions and 
+  ~ limitations under the License. -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>org.apache.logging.log4j</groupId>
+    <artifactId>log4j</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>log4j-redis</artifactId>
+  <name>Apache Log4j Redis</name>
+  <description>
+    Apache Log4j Redis.
+  </description>
+  <properties>
+    <log4jParentDir>${basedir}/..</log4jParentDir>
+    <docLabel>Log4j Redis Documentation</docLabel>
+    <projectDir>/log4j-redis</projectDir>
+    <module.name>org.apache.logging.log4j.redis</module.name>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+    </dependency>
+    <!-- Used for Redis appender -->
+    <dependency>
+      <groupId>redis.clients</groupId>
+      <artifactId>jedis</artifactId>
+      <version>2.9.0</version>
+    </dependency>
+    <!-- Test Dependencies -->
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+      <type>test-jar</type>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <type>test-jar</type>
+    </dependency>
+    <!-- Mocking framework for use with JUnit -->
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- Required for AsyncLoggers -->
+    <dependency>
+      <groupId>com.lmax</groupId>
+      <artifactId>disruptor</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <Fragment-Host>org.apache.logging.log4j.core.appender.mom.redis</Fragment-Host>
+            <Export-Package>*</Export-Package>
+          </instructions>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+  <reporting>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-changes-plugin</artifactId>
+        <version>${changes.plugin.version}</version>
+        <reportSets>
+          <reportSet>
+            <reports>
+              <report>changes-report</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+        <configuration>
+          <issueLinkTemplate>%URL%/show_bug.cgi?id=%ISSUE%</issueLinkTemplate>
+          <useJql>true</useJql>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>${checkstyle.plugin.version}</version>
+        <configuration>
+          <!--<propertiesLocation>${vfs.parent.dir}/checkstyle.properties</propertiesLocation> -->
+          <configLocation>${log4jParentDir}/checkstyle.xml</configLocation>
+          <suppressionsLocation>${log4jParentDir}/checkstyle-suppressions.xml</suppressionsLocation>
+          <enableRulesSummary>false</enableRulesSummary>
+          <propertyExpansion>basedir=${basedir}</propertyExpansion>
+          <propertyExpansion>licensedir=${log4jParentDir}/checkstyle-header.txt</propertyExpansion>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <version>${javadoc.plugin.version}</version>
+        <configuration>
+          <bottom><![CDATA[<p align="center">Copyright &#169; {inceptionYear}-{currentYear} {organizationName}. All Rights Reserved.<br />
+            Apache Logging, Apache Log4j, Log4j, Apache, the Apache feather logo, the Apache Logging project logo,
+            and the Apache Log4j logo are trademarks of The Apache Software Foundation.</p>]]></bottom>
+          <!-- module link generation is completely broken in the javadoc plugin for a multi-module non-aggregating project -->
+          <detectOfflineLinks>false</detectOfflineLinks>
+          <linksource>true</linksource>
+        </configuration>
+        <reportSets>
+          <reportSet>
+            <id>non-aggregate</id>
+            <reports>
+              <report>javadoc</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+      <plugin>
+        <groupId>com.github.spotbugs</groupId>
+        <artifactId>spotbugs-maven-plugin</artifactId>
+        <configuration>
+          <fork>true</fork>
+          <jvmArgs>-Duser.language=en</jvmArgs>
+          <threshold>Normal</threshold>
+          <effort>Default</effort>
+          <excludeFilterFile>${log4jParentDir}/spotbugs-exclude-filter.xml</excludeFilterFile>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jxr-plugin</artifactId>
+        <version>${jxr.plugin.version}</version>
+        <reportSets>
+          <reportSet>
+            <id>non-aggregate</id>
+            <reports>
+              <report>jxr</report>
+            </reports>
+          </reportSet>
+          <reportSet>
+            <id>aggregate</id>
+            <reports>
+              <report>aggregate</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-pmd-plugin</artifactId>
+        <version>${pmd.plugin.version}</version>
+        <configuration>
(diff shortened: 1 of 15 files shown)


log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,202 @@
...
+        int len = (header != null ? header.length : 0) + body.length;
+        byte[] data = new byte[len];
+        if (header != null) {
+            System.arraycopy(header, 0, data, 0, header.length);
+            System.arraycopy(body, 0, data, header.length, body.length);
+        } else {
+            System.arraycopy(body, 0, data, 0, body.length);
+        }
+        manager.send(data);
+    }
```
cakofony: Should this support batching based on `event.isEndOfBatch()`?

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,202 @@
...
+
+        @PluginAttribute(value = "host")
+        @Required(message = "No Redis hostname provided")
+        private String host;
+
+        @PluginAttribute(value = "port")
+        private int port;
+
+        @PluginAttribute(value = "ssl")
+        private boolean ssl = false;
```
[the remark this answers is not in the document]
garydgregory: +1

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        int getPort() {
+            return port;
+        }
+
+        public B withKeys(final String key) {
+            this.keys = new String[]{key};
+            return asBuilder();
+        }
```
garydgregory: I think we should use "set" as the setter prefix instead of "with".

kmeurer: @garydgregory I tend to prefer set as well, but the AbstractAppender and other common appenders (i.e. FileAppender, Rolling RandomAccessFileAppender, etc) all seem to use with___. Of course happy to still use set if you prefer.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisManager.java
```diff
@@ -0,0 +1,104 @@
...
+        } else {
+            this.poolConfiguration = poolConfiguration;
+        }
+    }
+
+    JedisPool createPool(String host, int port, SslConfiguration sslConfiguration) {
+        if (sslConfiguration != null) {
+            return new JedisPool(
+                    poolConfiguration,
+                    URI.create(host + ":" + String.valueOf(port)),
```
aaabramov: What is the purpose of explicitly stringifying `port `?

log4j-redis/src/test/java/org/apache/logging/log4j/redis/appender/RedisManagerTest.java
```diff
@@ -0,0 +1,97 @@
+package org.apache.logging.log4j.redis.appender;
```
rgoers: Missing license header.

log4j-redis/src/test/java/org/apache/logging/log4j/redis/appender/RedisAppenderTest.java
```diff
@@ -0,0 +1,159 @@
+/*
```
rgoers: Incorrect license header.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisManager.java
```diff
@@ -0,0 +1,104 @@
+package org.apache.logging.log4j.redis.appender;
```
rgoers: Missing license header

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,243 @@
...
+
+        @PluginBuilderAttribute("host")
+        @Required(message = "No Redis hostname provided")
+        private String host;
+
+        @PluginBuilderAttribute("keys")
+        private String keys = "";
+
+        @PluginBuilderAttribute("port")
+        private int port = 6379;
```
garydgregory: Can default values be picked up from Jedis?

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisManager.java
```diff
@@ -0,0 +1,104 @@
...
+import redis.clients.jedis.Jedis;
+import redis.clients.jedis.JedisPool;
+import redis.clients.jedis.JedisPoolConfig;
+import redis.clients.jedis.exceptions.JedisConnectionException;
+
+import java.net.URI;
+import java.util.Queue;
+import java.util.concurrent.TimeUnit;
+
+class RedisManager extends AbstractManager {
```
garydgregory: Some basic class-level Javadoc would help maintenance.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/LoggingJedisPoolConfiguration.java
```diff
@@ -0,0 +1,127 @@
+package org.apache.logging.log4j.redis.appender;
+
+import org.apache.logging.log4j.core.Core;
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderFactory;
+import redis.clients.jedis.JedisPoolConfig;
+
+@Plugin(name = "PoolConfiguration", category = Core.CATEGORY_NAME, printObject = true)
+public class LoggingJedisPoolConfiguration extends JedisPoolConfig {
```
garydgregory: Some basic class-level Javadoc would help maintenance.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,243 @@
...
+     */
+    @PluginBuilderFactory
+    public static <B extends Builder<B>> B newBuilder() {
+        return new Builder<B>().asBuilder();
+    }
+
+    @Override
+    public void append(final LogEvent event) {
+        final Layout<? extends Serializable> layout = getLayout();
+        if (event.getLoggerName() != null && AbstractLogger.getRecursionDepth() > 1) {
```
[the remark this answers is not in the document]
kmeurer: Unfortunately, I don't know. I just noticed that we included a check on recursion in the Kafka Appender and thought it made sense to include similar checks here. I don't have an explicit reason to have this here otherwise.

log4j-redis/src/test/java/org/apache/logging/log4j/redis/appender/RedisAppenderTest.java
kmeurer: Whoops -- that's embarrassing!

log4j-redis/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+    </dependency>
+    <!-- Used for Redis appender -->
+    <dependency>
+      <groupId>redis.clients</groupId>
+      <artifactId>jedis</artifactId>
+      <version>2.9.0</version>
```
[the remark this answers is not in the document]
kmeurer: It's definitely not perfect, and it's actually quite strange that they haven't released given the recent commits in the history: https://github.com/xetorthio/jedis/commits/master

It works fine for now, but I'll file an issue to see what's up.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
kmeurer: Jedis doesn't expose the default value for port nor offer good constructors to exclude it (i.e. you can't exclude port and provide an SSLConfig). Redis is unlikely to change their default port, though, so this seems pretty safe.

garydgregory: A minor improvement might be to refactor all of these initialization values into constants; I'll leave that up to you. But which ever way it goes, it would be nice to Javadoc the port and say what you provided in your comment. That would help for maintenance I think. Thank you!

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/LoggingJedisPoolConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.logging.log4j.core.config.plugins.Plugin;
+import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderFactory;
+import redis.clients.jedis.JedisPoolConfig;
+
+/**
+ * Plugin configuration for {@link redis.clients.jedis.JedisPool} objects, allowing end users to set pool configuration
+ * if desired. If not set, will default to JedisPool defaults.
+ */
+@Plugin(name = "PoolConfiguration", category = Core.CATEGORY_NAME, printObject = true)
```
cakofony: `PoolConfiguration` is fairly generic, thoughts on something along the lines of `RedisConnectionPool`?

It may be worthwhile to name our classes in a way that references Redis rather than the Jedis library, that way we can change the implementation in the future without breaking library consumers.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,244 @@
...
+            LOGGER.warn("Recursive logging from [{}] for appender [{}].", event.getLoggerName(), getName());
+        } else if (layout instanceof StringLayout) {
+            logQueue.add(((StringLayout)layout).toSerializable(event));
+            if (shouldFlushLogQueue(event.isEndOfBatch())) {
```
cakofony: There's a race here. Appenders may be used concurrently by multiple threads, if two events are logged at the same time, one invocation of `logQueue.add` will throw.

There's a helpful table on this page: https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html

I think we want to use offer, if that fails, we can fall back to flushing and try again.

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,244 @@
...
+                    error("Unable to write to Redis in appender [" + getName() + "]", event, e);
+                }
+            }
+        } else {
+            throw new AppenderLoggingException("The Redis appender only supports StringLayouts.");
+        }
+    }
+
+    private boolean shouldFlushLogQueue(boolean endOfBatch) {
+        return immediateFlush || endOfBatch || logQueue.size() >= queueCapacity;
```
cakofony: I think we always want to use the result of `BlockingQueue.offer` instead of `logQueue.size() >= queueCapacity`

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private void tryFlushQueue() {
+        manager.sendBulk(logQueue);
```
cakofony: We may want to use [BlockingQueue.drainTo](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html#drainTo(java.util.Collection)) in this call

kmeurer: I had initially implemented it that way, but my thinking on using `.poll()` in the inner method is that we can abort on a single failed log and keep the queue intact as opposed to dumping to a set and having to refill the queue with those logs that didn't throw exceptions and haven't yet been sent.

Is there something I'm missing that would make drainTo preferred?

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
kmeurer: +1 that's much cleaner

log4j-redis/src/main/java/org/apache/logging/log4j/redis/appender/RedisAppender.java
cakofony: The client should ideally use bulk operations to reduce overhead and increase throughput. We can use drainTo in order to get the bag of values for which to make bulk operations.

A load spike will produce events faster than we can persist them. Without bulk operations, many threads are likely to fill the queue and individually attempt to flush, forcing us to do even more work overall.

kmeurer: Gotcha -- I'm on the same page there. I thought the jedis.rpush API didn't support that, but it looks like the second argument is `String...`. I'll use drainTo instead, though this could increase our risk of dropping logs more broadly because a failure could mean a whole queue loss (Redis is pretty forgiving in what you send it though).

log4j-redis/pom.xml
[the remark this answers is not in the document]
jvz: Here's another more recent Redis library that I've heard good things about (but haven't tried using directly; I've mainly used this via Spring in some demo code): https://lettuce.io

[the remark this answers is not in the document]
kmeurer: thanks @jvz. @garydgregory Do you have a preference based on your reading of the documentation? For our purposes (simple appending to a redis queue), I suspect either would work.

kmeurer:
@jvz @garydgregory I ran some quick benchmarks of Jedis vs. Redis against a local Redis docker container. I rpushed 10000 values across 1000 runs. Here were the results:

```Jedis: 
Max: 27 ms
Min: 2 ms
Mean: 3.36 ms

Lettuce Synchronous: 
Max: 62 ms
Min: 4 ms
Mean: 5.567 ms

Lettuce Asynchronous:
Max: 23 ms
Min: 4 ms
Mean: 5.519 ms```

garydgregory:
The simplest solution is to call this module log4j-redis-jedis. Then we or someone can add log4j-redis-lettuce later.

kmeurer:
@garydgregory Changed the module names. Let me know if anything else occurs to you before we merge here.

If Lettuce gains traction and this starts to feel like an outdated implementation, I'd be happy to take another pass here or add a separate log4j-redis-lettuce module.

kmeurer:
I went ahead and reverted the commit for now. Happy to leave the decision to you both, @garydgregory and @rgoers, but I tend to agree that the module should probably be implementation-agnostic.

jvz:
Point of comparison again, Spring: https://projects.spring.io/spring-data-redis/

Realistically, we could have a generic log4j-redis module that could contain implementations using both libraries (as optional dependencies), or it could be done as submodules of that module while the log4j-redis module can provide abstractions or a default provider at least.

rgoers:
@jvz Exactly. But I am fine if refactoring to support that is done later should anyone want to do it.

kmeurer:
@rgoers Should we go ahead and merge this PR to refactor later in that case?

rgoers:
I am probably fine with that.

kmeurer:
Sounds good. I'll leave it to you or @cakofony to merge this in. Do you have a sense for when we might get it into a downloadable release?

cakofony:
@kmeurer I would recommend that you create a forked build to run with in production for a bit before cutting an official release here. That will likely tease out a few issues without being tied to the log4j release cycle.

jvz:
Considering the ability to finally break APIs, yeah, I don't see 3.0 being released anytime soon. I'd imagine we'll have a development period to re-stabilize the 3.0 API first.

kmeurer:
That sounds good to me. I can definitely run a separate build in the interim.

PR #210, opened 2018-08-14
[LOG4J2-2391] Lazily initialize ThrowableProxy data

carterkozak:
Goal is to avoid the most expensive components until render
time, allowing patterns using non-extended traces to avoid
the cost of class loading.

This implementation is fairly naive in that initialization
will deeply initialize all causes and suppressed causes
when any action requires initialization, rather than
lazily initializing only what is required.

Benchmarking results with this change show that the
"complexLog4j2ThrowableAsync" performance matches
"complexLog4j2Throwable" due to avoiding ThrowableProxy
initialization.

```
Benchmark                                                            Mode  Cnt       Score       Error  Units
FileAppenderThrowableBenchmark.complexLog4j1                        thrpt    3    8537.030 ±  5312.479  ops/s
FileAppenderThrowableBenchmark.complexLog4j2ExtendedThrowable       thrpt    3    2534.330 ±   600.769  ops/s
FileAppenderThrowableBenchmark.complexLog4j2ExtendedThrowableAsync  thrpt    3    2488.063 ±   838.325  ops/s
FileAppenderThrowableBenchmark.complexLog4j2Throwable               thrpt    3   13206.010 ±   320.408  ops/s
FileAppenderThrowableBenchmark.complexLog4j2ThrowableAsync          thrpt    3   12795.284 ±  2677.436  ops/s
FileAppenderThrowableBenchmark.complexLogbackFile                   thrpt    3   11377.932 ± 25787.767  ops/s
FileAppenderThrowableBenchmark.julFile                              thrpt    3   54271.234 ± 42052.793  ops/s
FileAppenderThrowableBenchmark.log4j1                               thrpt    3   75694.883 ± 13310.643  ops/s
FileAppenderThrowableBenchmark.log4j2ExtendedThrowable              thrpt    3   46216.007 ±  1046.523  ops/s
FileAppenderThrowableBenchmark.log4j2Throwable                      thrpt    3  110604.119 ± 33981.960  ops/s
FileAppenderThrowableBenchmark.logbackFile                          thrpt    3  112429.311 ± 95759.913  ops/s
```

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
@@ -258,7 +258,7 @@ public final class LoaderUtil {
         return newCheckedInstanceOf(className, clazz);
     }
 
-    private static boolean isIgnoreTccl() {
+    public static boolean isIgnoreTccl() {
         // we need to lazily initialize this, but concurrent access is not an issue
         if (ignoreTCCL == null) {
             final String ignoreTccl = PropertiesUtil.getProperties().getStringProperty(IGNORE_TCCL_PROPERTY, null);
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
@@ -16,6 +16,7 @@
  */
 package org.apache.logging.log4j.core.impl;
 
+import java.io.IOException;
 import java.io.Serializable;
 import java.net.URL;
 import java.security.CodeSource;
@@ -83,11 +84,11 @@ public class ThrowableProxy implements Serializable {
 
     private static final long serialVersionUID = -2752771578252251910L;
 
-    private final ThrowableProxy causeProxy;
+    private ThrowableProxy causeProxy;
 
     private int commonElementCount;
 
-    private final ExtendedStackTraceElement[] extendedStackTrace;
+    private ExtendedStackTraceElement[] extendedStackTrace;
 
     private final String localizedMessage;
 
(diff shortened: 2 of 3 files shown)


garydgregory:
Is this level of complication really worth it? Can you translate you findings in plain English SVP?

rgoers:
As I said on the mailing list, I would like to simplify ThrowableProxy by moving all the rendering logic to a new Class, ThrowableProxyRenderer.  I don't know how it got this way but somewhere along the line we have over-complicated this class to the point that it is hard to figure out what it is doing.

cakofony:
@rgoers Absolutely, this is not ready to merge.

@garydgregory The idea is that loading classes in order to add extended stack trace information can be incredibly expensive depending on the application. On java 8 using the default ClassLoader (network classloaders may be much slower), the benchmarks show a performance drop from 13,000 ops/s to 2,500 ops/s depending on whether we use the throwable pattern, or extended throwable pattern, which uses a ThrowableProxy. For synchronous logging, we can simply select the `%ex` pattern instead of `%xEx` (default) and avoid initializing a ThrowableProxy instance, avoiding class loading.

Upon further inspection, AsyncLoggerContextSelector (unlike AsyncLoggerConfig) does not initialize ThrowableProxy until it is is requested, which always occurs on the background thread. This is broken in a slightly different way. The performance is ideal because it avoids creating a stack trace as well as class loading, however when class loading is required, the context classloader from where the throwable was logged has already been lost, and the stack trace only contains frames from the disruptor background thread rather than those from the stack where the exception was caught and logged.

garydgregory:
Clearly we should fix bugs. It seems normal that using layout patterns that query for more data will be slower... Maybe that fact should be better documented in the pattern layout: Displaying extended stack traces is slower than normal stack traces... for those who care... ;-)

rgoers:
Yes, the performance is important when it clearing is underperforming.

rgoers:
@carterkozak What is the status of this PR?

carterkozak:
Sorry, this has gotten stale and I need to come back to the problem when I have some time. I'm not confident this is the best solution as it only defers initialization instead of reducing the total cost, so it's only helpful for the AsyncLoggerConfig codepath, not AsyncLoggerContextSelector (fully async) or synchronous modes. I'll take a look at updating the async config code to defer throwableproxy initialization in the same way as fully async so they're more closely comparable.

rgoers:
Any progress? Should this be closed in lieu of a different solution?

PR #212, opened 2018-08-21
[LOG4J2-2391] Refactor as much logic as possible out of ThrowableProxy

cakofony:
This change isolates initialization logic into a new
ThrowableProxyInitializationUtils class, and rendering logic
into ThrowableProxyRenderUtils.

Both new utility classes are package private. ThrowableProxy public
API has not changed.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxy.java
@@ -81,7 +81,7 @@ public class ThrowableProxy implements Serializable {
         this.causeProxy = null;
         this.message = null;
         this.localizedMessage = null;
-        this.suppressedProxies = ThrowableProxyInitializationUtils.EMPTY_THROWABLE_PROXY_ARRAY;
+        this.suppressedProxies = ThrowableProxyHelper.EMPTY_THROWABLE_PROXY_ARRAY;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


rgoers:
I like it.

cakofony:
I like `ThrowableProxyRenderer`though I don't think `ThrowableProxyCache` is quite right. The initialization logic includes a cache to avoid loading the same class many times, but that's an optimization.

Maybe `ThrowableProxyInitialization`? I'm open to suggestions. Renaming it in the future shouldn't be dangerous since it's package private.

Apologies for the slow reply, this week has been busier than usual!

jvz:
The Initializer maybe? And good point, it can definitely be renamed later if necessary!

And no worries on the slow reply. I'm just finally getting back into the swing of things myself!

rgoers:
ThrowableProxyRenderer was the same name I was going to use so I think we are good on that one. As for the other I would be ok with ThrowableProxyHelper, ThrowableProxyUtil. I am not as thrilled with ThrowableProxyInitization or ThrowableProxyInitializer as we may desire to add code to it not related to initialization.

jvz:
Hmm, maybe it could go as a util class then for now, particularly since it's package private.

cakofony:
Merged. I haven't added a change note because this change modifies neither API nor implementation, but I'd be more than happy to write one if anyone here prefers it.

rgoers:
I don't think a note is needed for the refactoring unless it is going to close the Jira issue. I'd like to look at the new code and the performance tests before doing that. Hopefully there is more work we can do to improve this.

PR #213, opened 2018-08-24
LOG4J2-2025: JUL bridge handler (incl. unit test and site doc)

thiesw:
Solution for LOG4J2-2025 (as alternative to JUL-LogManager).
Rest see JavaDoc and site-doc (index.md)

This runs with log4j 2.11, so maybe can be even release with 2.11.2 (and not 3.0).

### DIFF
--- a/log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
+++ b/log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
@@ -16,9 +16,18 @@
  */
 package org.apache.logging.log4j.jul;
 
-// note: NO import of Logger, LogManager etc. to prevent conflicts JUL/log4j
+//note: NO import of Logger, Level, LogManager to prevent conflicts JUL/log4j
+import java.beans.PropertyChangeEvent;
+import java.beans.PropertyChangeListener;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 import java.util.logging.LogRecord;
 
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
 import org.apache.logging.log4j.spi.ExtendedLogger;
 import org.apache.logging.log4j.status.StatusLogger;
 
@@ -32,10 +41,11 @@ import org.apache.logging.log4j.status.StatusLogger;
  * Installation/usage:<ul>
  * <li> Declaratively inside JUL's <code>logging.properties</code>:<br>
  *    <code>handlers = org.apache.logging.log4j.jul.Log4jBridgeHandler</code><br>
- *    (Note: in a webapp running on Tomcat, you may create a <code>WEB-INF/classes/logging.properties</code>
- *     file to configure JUL for this webapp only: configured handlers and log levels affect your webapp only!
- *     This file is then the <i>complete</i> JUL configuration, so JUL's defaults (e.g. log level INFO) apply
- *     for stuff not explicitly defined therein.)
+ *    (and typically also: Â  <code>org.apache.logging.log4j.jul.Log4jBridgeHandler.propagateLevels = true</code> )<br>
+ *    Note: in a webapp running on Tomcat, you may create a <code>WEB-INF/classes/logging.properties</code>
+ *    file to configure JUL for this webapp only: configured handlers and log levels affect your webapp only!
+ *    This file is then the <i>complete</i> JUL configuration, so JUL's defaults (e.g. log level INFO) apply
+ *    for stuff not explicitly defined therein.
  * <li> Programmatically by calling <code>install()</code> method,
  *    e.g. inside ServletContextListener static-class-init. or contextInitialized()
  * </ul>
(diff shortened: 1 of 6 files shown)


log4j-jul/src/site/markdown/index.md
```diff
@@ -74,3 +74,29 @@ Java Level | Log4j Level
...
+
+
+# Log4j JDK Logging Bridge
+
+The LogManager is not always useable because you have to set a JVM wide effective system
+property - e.g. in web servers this is not possible if you are not the administrator.
+
+The [`Log4jBridgeHandler`](apidocs/org/apache/logging/log4j/jul/Log4jBridgeHandler.html) is an
+alternative that can be declaratively used via `logging.properties`.
+It is less performant than the LogManager but still okay to use.
```
rgoers: This should also note that logging methods are subject to filtering by BOTH java.util.logging and log4j 2.

log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+ *      I.e. logging.properties and log4j2.xml have some redundancies.
+ * <li> Only JUL log events that are allowed according to the JUL log level get to this handler and thus to log4j.
+ *      If you set <code>.level = SEVERE</code> only error logs will be seen by this handler and thus log4j
+ *      - even if the corresponding log4j log level is ALL.<br>
+ *      On the other side, you should NOT set <code>.level = FINER  or  FINEST</code> if the log4j level is higher.
+ *      In this case a lot of JUL log events would be generated, sent via this bridge to log4j and thrown away by the latter.
+ * </ul>
+ *
+ * @author Thies Wellpott (twapache@online.de)
+ * @author authors of original org.slf4j.bridge.SLF4JBridgeHandler (ideas and some basis from there)
```
rgoers: Author tags should be removed. Log4j doesn't use them.

log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
```diff
@@ -0,0 +1,198 @@
...
+ */
+// TODO: @author authors of original ch.qos.logback.classic.jul.LevelChangePropagator (ideas and some basis from there)
```
jvz: Logback is EPL/LGPLv2.1. Not sure if the licensing is compatible here.

jvz:
This overall seems reasonable, but I'm concerned about licensing. If any code is copied from Logback, then it's not allowed inside an Apache project. The EPL and LGPL, while neat, are both weak copyleft licenses which can only be used in _optional_ dependencies in Apache projects.

We would either need this contribution to be fully ALv2 (or compatible), or this would need to remain a 3rd party component unfortunately.

thiesw:
Documentation is enhanced and I removed the author tags.
Regarding logback and license: I did not think about this, but your are right. Currently the level propagation is not implemented, so *nothing* from logback is used (this the outcommented author tag for it); it is just a hint for later. Even if this comes, there will be no problem because the implementation is quite trivial and has to be adjusted to log4j's internals, so jut the idea will be from logback's implementation.

thiesw:
Some questions to the log4j experts regarding the automatic log level propagation:

My idea is to have an interace `LoggerListener` or maybe better named `LoggerConfigListener` that is called by log4j's implementation at the relevant places:
- `configApplied()`:  initialisation of log4j and readout of configuration (log4j2.xml ...) is finished and applied, so one get get the current info via `((org.apache.logging.log4j.core.LoggerContext) LogManager.getContext()).getLoggers()` Maybe the LoggerContext may be given as parameter. Method name could also be `initFinished()`.

Following methods are only called after initial creation, i.e. after calling above method:
- `added(Logger)`:  the given logger has been added, called after its configuration (log level) is set
- `remove(Logger)`:  the given logger will be removed, so JUL logger level can be set to null - if removal is possible in log4j at all
- `setLevel(Logger)`:  log level has been set/changed, thus no extra parameter because `logger.getLevel()` may be used

There must be a method (in LogManager?) to set/add this listener. E.g. the init. code of Log4jBridgeHandler will call it.

I have not found any possibility to hook into the relevant places of log4j's implementation to allow log level propagation. So this is a generic solution. What do you think?

thiesw:
A rainy and cold weekend in northern Germany helped to complete this.
From my point, the implementation is finished - but do your review, please.

rgoers:
I just looked at the statistics and 13% of last month's downloads were for 2.3 so I guess a fair number of people are still stuck at Java 6. Although I have no problem providing support for bugs we would have to discuss whether we want to add new features. It has been so long since this was released I am not even sure how much of the release process has changed, if at all. 

I personally have no problem providing a link to your project. However, you should not be using the org.apache.logging package name as that may confuse users into thinking it is supported by Apache Logging Service. Of course, if we agree to incorporate it then no change is required it is integrated into Log4j.

thiesw:
Okay Ralph, nice to get this number (my Java6 project is now already migrated to Java8, but I have this code lying around).
If you want to create a "2.3.1" release with my stuff included, fine. I will create a pull request then.
But if this is to complicated because the release process "has gone", no problem - log4j is an open source volunteer project.
I will adjust the package of my addon-project then to prevent wrong support thinking.
Simply tell me your (or log4j's team) intent.

PR #214, opened 2018-08-28
LOG4J2-1002 - ThrowablePatternConverter should preserve EOF

quaff:
throwable.printStackTrace() will produce EOF, but w.toString().split(Strings.LINE_SEPARATOR) drop it, buffer should append newline after loop.


rgoers:
I tried applying this and it caused unit test failures in log4j-taglib.

PR #215, opened 2018-08-30
LOG4J2-2427 log4j2 filter for logs with no markers

rimal:
Related JIRA issue: https://issues.apache.org/jira/browse/LOG4J2-2427


rgoers:
I was about to merge this but then realized I would have to write the documentation for it. Is it possible you could update the PR to add the documentation to the filters section?

rimal:
@rgoers I have added the documentation. Now the check is giving a build error. Not sure why, because I have not changed any code. Not able to check on local as the build requires java 9 and now java 9 download is not available. Can you please help me out here.

rimal:
> @rgoers I have added the documentation. Now the check is giving a build error. Not sure why, because I have not changed any code. Not able to check on local as the build requires java 9 and now java 9 download is not available. Can you please help me out here.

Merging with latest master resolved the issues. 
@rgoers Let me know if I need to make any other change in the PR.

garydgregory:
Hi All,

I think NullFilterMarker would be a better class name.

Gary

rgoers:
I have to disagree. NullFilterMarker sounds like it somehow is a marker that represents a null filter, whatever that is, whereas NoMarkerFilter clearly represents a filter for when no marker is present.

carterkozak:
Agreed, I could see a case for `NullMarkerFilter` over `NullFilterMarker`, but I think the current name `NoMarkerFilter` makes the functionality very clear.

PR #218, opened 2018-10-11
[LOG4J2-2475] More efficient parameter reset for reusable messages

cakofony:
Pushing to GH for feedback.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEvent.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEvent.java
@@ -135,8 +135,9 @@ public class RingBufferLogEvent implements LogEvent, ReusableMessage, CharSequen
             reusable.formatTo(getMessageTextForWriting());
             messageFormat = reusable.getFormat();
             if (parameters != null) {
-                parameters = reusable.swapParameters(parameters);
+                // Must read the parameter count first, swapParameters mutates the target message
                 parameterCount = reusable.getParameterCount();
+                parameters = reusable.swapParameters(parameters);
             }
         } else {
             this.message = InternalAsyncUtil.makeMessageImmutable(msg);
@@ -271,6 +272,7 @@ public class RingBufferLogEvent implements LogEvent, ReusableMessage, CharSequen
     public Object[] swapParameters(final Object[] emptyReplacement) {
         final Object[] result = this.parameters;
         this.parameters = emptyReplacement;
+        this.parameterCount = 0;
         return result;
     }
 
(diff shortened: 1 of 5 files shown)


remkop:
I'm just not sure that this is a good idea.
It seems to make the code more brittle: if the code changes in the future this may break something.
It seems safer to just clear the whole array.

That said, I can be convinced otherwise if there is a clear advantage. Can you put this in perspective with some performance numbers? How much can be gained by this change?

cakofony:
In most cases, I don't think this will make much difference because setting all 10 spots in a the reusable parameter buffer is incredibly fast.

This aims to solve problems that arise when a larger parameter array is passed to a logger, since that array will be reused and passed between MutableLogEvent (or RingBufferLogEvent) and ReusableParameterizedMessage. I don't want to discard parameter arrays above a given threshold because there are scenarios where we would want to allow them to be reused, but we also don't want to iterate over each index and set it to null if we know there's only a single parameter.

I've put together a simple benchmark to illustrate performance implications. It's fairly contrived, close to the worst case scenario.

Without this change:
```
Benchmark                                      Mode  Cnt     Score    Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  1706.636 ± 89.416  ops/s
```

With this change:
```
Benchmark                                      Mode  Cnt        Score         Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  5472588.826 ± 1029286.946  ops/s
```

remkop:
I cannot imagine any program logging one million parameters so this benchmark is perhaps a bit academic. :-)
In a realistic program a "large" parameter array is probably more like 100 elements (and I would not like to maintain a program that has a logging statement with 100 parameters).

Can we measure again with an extreme case of 1000 parameters?

I suspect that with an array of 1000 (or even 10,000) elements we won't see much difference between nulling out only the first 10 or the full array, in which case it is more robust to simply null out the whole array.

garydgregory:
One "realistic" example would be to think of logging against a database
table. Some tables have lots of columns. How many, well, a thousand is
possible I suppose.

Gary

cakofony:
n=1000
```
Benchmark                                      Mode  Cnt        Score       Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  1406346.075 ± 82926.012  ops/s
```

n=100
```
Benchmark                                      Mode  Cnt        Score        Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  4332639.602 ± 232499.236  ops/s
```

n=10
```
Benchmark                                      Mode  Cnt        Score         Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  5480582.409 ± 1238380.733  ops/s
```

And another run with this change
```
Benchmark                                      Mode  Cnt        Score        Error  Units
ManyParametersInSetupBenchmark.fewParameters  thrpt    3  5814710.334 ± 899317.421  ops/s
```

cakofony:
No worries, thank you for your feedback as always!

We might want to consider updating `ReusableMessageFactory.newMessage(final String message, final Object... params)` to return a non-reusable ParameterizedMessage instance. It seems like a bug that logging an object array may result in that array being mutated the next time an event is logged on the same thread. It's unfortunate that we cannot differentiate between an array passed as a value, and an array created by a varargs invocation.

Thoughts?

cakofony:
Pushed the above idea to #220

PR #219, opened 2018-10-15
LOG4J2-2478 Return the computed variables on each benchmark to avoid DCE

DiegoEliasCosta:
As mentioned in [JIRA](https://issues.apache.org/jira/browse/LOG4J2-2478), this patch simply returns the computation on every benchmark that uses the static _consume_ method. We do that to prevent dead-code elimination, which is only partially addressed in the current implementation.

Another patch that could simplify the benchmark code, and get considerably faster throughputs, would be to simply rely on _Blackhole.consume_ methods to sink the computation. We then, would not need the _consume_ method we have in the class.


remkop:
These are all good changes and we should merge this PR.
Not sure when I can do it but perhaps someone else gets there first.

cakofony:
I can take care of this :-)

PR #220, opened 2018-10-16
[WIP][LOG4J2-2475] ReusableMessageFactory uses non-reusable messages in some cases

cakofony:
Large parameter arrays passed to the logger will result in a
ParameterizedMessage instance being created rather than
adding the varargs to a ReusableParameterizedMessage instance.
This avoids potential mutation of the input array when it
is reused later.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
@@ -107,7 +107,36 @@ public final class ReusableMessageFactory implements MessageFactory2, Serializab
      */
     @Override
     public Message newMessage(final String message, final Object... params) {
-        return getParameterized().set(message, params);
+        if (params == null) {
+            return newMessage(message);
+        }
+        switch (params.length) {
+            case 0:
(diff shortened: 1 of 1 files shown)


remkop:
Good point that we cannot distinguish between parameter arrays created by the user (which we shouldn’t reuse) and parameter arrays created by the compiler because of a varargs invocation. 

Would it be an idea to clone large (size greater than 10) parameter arrays? (And use this cloned array in the reusable message.)

That would ensure we won’t modify user-specified arrays.

cakofony:
I think cloning large arrays will mitigate any advantages of using ReusableParameterizedMessage over ParameterizedMessage.

If we use ParameterizedMessage for large parameter arrays, we can also simplify ReusableParameterizedMessage to remove the `varargs` field, and make the code easier to read.
This also solves the issue illustrated by my earlier commit where large arrays are referenced by ReusableParameterizedMessage/MutableLogEvent/RingBufferLogEvent until another logger is called with a parameter array.

cakofony:
I suppose `InternalAsyncUtil.makeMessageImmutable` invokes `Message.getFormattedMessage`, which is more expensive than creating a copy of the input parameters.

PR #221, opened 2018-10-18
[LOG4J2-2134] Stackoverflow at AwaitCompletionReliabilityStrategy.

garydgregory:
Looking for feedback and if this is the right away to patch this.


remkop:
What was the issue? Is there a stack trace? Or a link?

garydgregory:
See https://issues.apache.org/jira/browse/LOG4J2-2134

rbarbey:
With the latest version 2.11.1 I'm running into this issue as well. Will there be a patch release which includes this fix? Thank you!

PR #222, opened 2018-10-21
LOG4J2-2482 BasicContextSelector cannot be used in a OSGI application.

rgansevles:
Removed optional dependencies from log4j-api to log4j-core, use the thread context classloader to load core classes.


rgansevles:
Pull-request was applied

PR #223, opened 2018-10-22
[LOG4J2-2476] missing classes for 1.2 compatibility

AlBundy33:
missing classes in 1.2-api

These are needed to avoid compile-errors/NoClassDefFoundErrors with libraries that try to configure log4j programatically.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/ConsoleAppender.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/ConsoleAppender.java
@@ -1,0 +1,52 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.log4j;
+
+import org.apache.log4j.spi.LoggingEvent;
+
+/**
+ * Console-appender.
+ */
+public class ConsoleAppender extends AppenderSkeleton
+{
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void close()
+  {
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public boolean requiresLayout()
+  {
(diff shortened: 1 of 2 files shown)


AlBundy33:
Classes where manually added.

abysas:
It is a great pity, that added ConsoleAppender does not match the original class from 1.2 - it does not contain the public constructors that existed there: `ConsoleAppender(Layout layout, String target)` and `ConsoleAppender(Layout layout)`.

garydgregory:
@abysas, 

We strongly discourage using constructors when builders are available: `org.apache.logging.log4j.core.appender.ConsoleAppender.newBuilder()` . Builders are the best way for us to maintain API compatibility moving forward.

Gary

AlBundy33:
But sometimes old libraries are using the constructors.

So for compatibility reasons it would be nice to have the old constructors too.

rgoers:
I would go even further than that. Log4j 1.2 had many architectural problems and was very limiting. So when Log4j 2 was designed maintaining compatibility wasn't very important. All Log4j components, such as Appenders, Layouts and Filters are designed to be plugins that are constructed via a builder or a factory. Therefore the constructor intentionally is usually private. 

Furthermore, when people try to port from Log4j 1 to Log4j 2 they often just try to retrofit the Log4j 1 code instead of thinking about the problem they are trying to solve. Log4j 2 has many more features than Log4j 1 and frequently different ways should be used to solve the problem. While the Log4j 1 adapter module is provided to allow applications that cannot be modified to continue to work when Log4j 2 is the actual framework being used, most of the things it does configuration wise are just no-ops similar to what this patch does.  

Also, note that Log4j 2 doesn't using the same package names as Log4j 1 so even if the ConsoleAppender's constructor matched exactly it still wouldn't be of any use to a old library without modifying that library's code.

rgoers:
And now I suddenly suspect we are talking about apples and oranges. Looking at the patch for this PR I see that it doesn't have the constructors you are mentioning. Yes, to be compatible with Log4j 1.x the Log4j 1.x adapter classes should match the signatures they had in Log4j 1.

abysas:
@rgoers, thank you for confirming about the compatibility mismatch. I was so happy seeing the v2.11.2 release notes about added `ConsoleAppender`. The [migration guide ](https://logging.apache.org/log4j/2.x/manual/migration.html) from Log4j 1.x clearly indicates that applications that want to successfully use `log4j-1.2-api.jar` bridge to replace Log4j, the application:

> ... must not access methods and classes internal to the Log4j 1.x implementation such as **Appenders**, LoggerRepository or Category's callAppenders method.

Based on this clear limitation, I didn't tried to suggest a replacement for `ConsoleAppender` in Log4j2 bridge. We have a 3rd party application, which use Log4j and instantiates `ConsoleAppender` with specific `Layout` supplied in constructor. To support this application along with Log4j2 bridge, in my project I added custom `ConsoleAppender` having the same package as original one of v1.2. 

And to my dismay the added `ConsoleAppender` in v2.11.2 is incompatible with the original class - it breaks the workaround, as added incompatible  `ConsoleAppender` clashes with v1.2 compatible  `ConsoleAppender` I've added.

It becomes a blocker to us to upgrade to the Log4j2 v2.11.2+ version.

PR #224, opened 2018-10-24
[LOG4J2-2403] Allow zero padding the counter of a RollingFileAppender	

hupfdule:
This patch introduces the possibility to use zero-padding for the counter in the `RollingFileAppender` and therefore fixes the issue [LOG4J2-2403.](https://issues.apache.org/jira/browse/LOG4J2-2403)

Incidentally the functionality of zero-padding can be used for any pattern, not only the `filePattern` of the `RollingFileAppender`. While that may not make much sense for data like the logger category, it can be used e.g. for the sequence number in log entries.

The patch contains the actual code changes as well as some tests and changes to the corresponding documentation.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/AbstractRolloverStrategy.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/AbstractRolloverStrategy.java
@@ -46,6 +46,8 @@ public abstract class AbstractRolloverStrategy implements RolloverStrategy {
      */
     protected static final Logger LOGGER = StatusLogger.getLogger();
 
+    public static final Pattern PATTERN_COUNTER= Pattern.compile(".*%((?<ZEROPAD>0)?(?<PADDING>\\d+))?i.*");
+
     protected final StrSubstitutor strSubstitutor;
 
     protected AbstractRolloverStrategy(final StrSubstitutor strSubstitutor) {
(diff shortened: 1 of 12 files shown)


hupfdule:
Why is there still no comment on this PR?

It fixes the problem described in [LOG4J-2403](https://issues.apache.org/jira/browse/LOG4J2-2403) and was also mentioned in [this Stackoverflow question](https://stackoverflow.com/questions/35073491/log4j2-left-pad-filename-with-zeroes). Two log4j developers answered on that stackoverflow question requesting such a PR. 

An still this PR is being ignored!

rgoers:
Because we are all volunteers and do this in our spare time. It will get looked at.

Ralph

garydgregory:
I like the idea but I cannot take the time to review ATM. Thank you for your contribution :-)

hupfdule:
Thank you for taking the time to review this PR!

I have prepared an [alternative PR](https://github.com/apache/logging-log4j2/pull/280) that does this change only in the IntegerPatternConverter using the format `%i{03}`.

However this imposes an inconsistency with the justification that can already be specified and always uses spaces for padding as described in the [PatternLayout section of the Layout manual](https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout).

Therefore I think the approach of this orginal PR is better.

rgoers:
After looking at the doc I agree this is the better option. Can you also create a PR for the release-2.x branch?

hupfdule:
Done: [#281](https://github.com/apache/logging-log4j2/pull/281)

However, the documentation changes are not included since the asciidoc manual doesn't exist in that branch.

PR #240, opened 2018-11-16
Fix LOG4J2-2510 NullPointerException during shutdown

Blindfold:
This should fix LOG4J2-2510

NullPointerEception gets thrown while executing shutdown hook in configuration with RollingRandomAccessFile appender and DirectWriteRolloverStrategy strategy. It happens when no messages logged during the session and thus no log file created.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
@@ -156,7 +156,9 @@ public class RollingRandomAccessFileManager extends RollingFileManager {
         flush();
         if (randomAccessFile != null) {
             try {
-                randomAccessFile.close();
+                if (randomAccessFile != null) {
+                    randomAccessFile.close();
+                }
                 return true;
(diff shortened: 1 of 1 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
```diff
@@ -156,7 +156,9 @@ public synchronized boolean closeOutputStream() {
         flush();
         if (randomAccessFile != null) {
             try {
-                randomAccessFile.close();
+                if (randomAccessFile != null) {
+                    randomAccessFile.close();
+                }
```
cakofony: I think this was recently fixed on line 157, see https://issues.apache.org/jira/browse/LOG4J2-2418

Blindfold: True. My bad, couldn't find that one. Closing.

cakofony: No worries, thank you for taking the time to submit a change upstream!

PR #242, opened 2018-12-02
New iso date time formats

garydgregory:


garydgregory:
Oops, need a PR vs. release-2.x, not master.

PR #244, opened 2018-12-03
New iso date time formats

garydgregory:


log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
```diff
@@ -183,16 +218,27 @@ static FixedFormat lookupIgnoringNanos(final String pattern) {
...
+                indexEnd = indexEnd < 0 ? pattern.length() : indexEnd; 
+                for (int i = indexStart + 1; i < indexEnd; i++) {
                     if (pattern.charAt(i) != SECOND_FRACTION_PATTERN) {
-                        return -1;
+                        return EMPTY_RANGE;
                     }
                 }
             }
-            return index;
+            return new int [] {indexStart, indexEnd};
```
remkop: This allocates a new `int[]` object every time this method is called. Can the `nanoRange` method be split up into `nanoStart` and `nanoEnd` instead?

garydgregory: Hi @remkop 

Thank you for your review.

Are you sure this really matters? This code only kicks in once per thread or when a configuration is created, not for each log event, so creating the array should not create garbage for each event.

WDYT?

Gary

remkop: Yes from a performance perspective it does not matter, you are right.
I do think that the code would look slightly cleaner like this:

```
final int nanoStart = nanoStart(pattern);
final int nanoEnd = nanoEnd(pattern, nanoStart);
```
But not a major problem as it is now.

garydgregory: Hm, I get your point, but the code duplication is not nice in this version:
```
        /**
         * @return int[0] start index inclusive; int[1] end index exclusive
         */
        private static int nanoRangeStart(final String pattern) {
            final int indexStart = pattern.indexOf(SECOND_FRACTION_PATTERN);
            int indexEnd = -1;
            if (indexStart >= 0) {
                indexEnd = pattern.indexOf('Z', indexStart);
                indexEnd = indexEnd < 0 ? pattern.indexOf('X', indexStart) : indexEnd;
                indexEnd = indexEnd < 0 ? pattern.length() : indexEnd; 
                for (int i = indexStart + 1; i < indexEnd; i++) {
                    if (pattern.charAt(i) != SECOND_FRACTION_PATTERN) {
                        return -1;
                    }
                }
            }
            return indexStart;
        }

        /**
         * @return int[0] start index inclusive; int[1] end index exclusive
         */
        private static int nanoRangeEnd(final String pattern) {
            final int indexStart = pattern.indexOf(SECOND_FRACTION_PATTERN);
            int indexEnd = -1;
            if (indexStart >= 0) {
                indexEnd = pattern.indexOf('Z', indexStart);
                indexEnd = indexEnd < 0 ? pattern.indexOf('X', indexStart) : indexEnd;
                indexEnd = indexEnd < 0 ? pattern.length() : indexEnd; 
                for (int i = indexStart + 1; i < indexEnd; i++) {
                    if (pattern.charAt(i) != SECOND_FRACTION_PATTERN) {
                        return -1;
                    }
                }
            }
            return indexEnd;
        }
```

Maybe you can take a shot at it after I commit if you really care? ;-)

rgoers:
@garydgregory What is happening with this PR?

garydgregory:
Wel, it looks like the PR is/was ready to go now it's stale.

rgoers:
I don't know what that means. Are you planning on addressing the issues or should it be closed?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
jaystyles: @Dependabot merge

PR #246, opened 2018-12-19
[LOG4J2-2523]  adding support for request related lookup  web module

rmannibucau:


rgoers:
This code can only be applied to 3.0 as it contains Lambdas.

PR #247, opened 2018-12-30
LOG4J2-2527: ListAppender getters return immutable snapshots

cakofony:
Provides a bit more thread safety to ListAppender. The class
is marked not-thread-safe, however this functionality is
important for testing, particularly when asynchronous logging is
configured.

Previously we returned unmodifiable wrappers around the internal
state, which can be dangerous to iterate over while events are
actively logged.

State is tracked in synchronized lists, which should be safer
than synchronizing on individual methods.


cakofony:
> Why not use a CopyOnWriteArrayList?

Two reasons:
1. I expect more appends than gets. This implementation is basically a CopyOnRead.
2. I find it confusing that the unmodifiable list we return is updated when new events are logged, while we use it, regardless of the list implementation we use I'd prefer for the returned values to be immutable snapshots. The benefit of a CoW list is that iterator usage is safe, but because we copy the collection this doesn't help us.

cakofony:
Merged

PR #249, opened 2019-01-10
Fix MapMessage -> StringMapMessage API breaks

cakofony:
Note: This updates MapPatternConverter and MapMessageLookup to use
MapMessage JAVA format, which is slightly different from the original
output because map values are quoted.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
@@ -18,8 +18,7 @@ package org.apache.logging.log4j.core.pattern;
 
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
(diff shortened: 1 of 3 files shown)


cakofony:
Are we okay with adding quotes to map values here?

travisspencer:
We have customers that parse these log messages from the DB. Would be nice to not change the format.

cakofony:
That makes sense, thanks for the quick reply! Another potential solution would be to implement `MapFormat.JAVA_UNQUOTED` which exactly matches the current behavior, but we only have to define and maintain it in one place. Would you be interested in implementing that?

cakofony:
Closing this pr, @travisspencer is taking care of the fix :-)

travisspencer:
Sure! I'm here to help :)

travisspencer:
ecd05cb225423418f1db91526ed444d8f1ced55f is what you had in mind? If so, and if there's nothing more to change, I'll squash all the commits in that branch and wait for green on the CI.

cakofony:
Looks great, thank you!

PR #250, opened 2019-01-10
LOG4J2-2530 Generalize check for MapMessage

travisspencer:
This fixes LOG4J2-2530 by generalizing the check for `MapMessage` to catch not only `StringMapMessage` objects but also `StructuredMapMessgage` objects, fixing an accidental breaking change introduced in 76aff58919dc5030367ea24c7c8ef42dca3b75cd.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
@@ -18,8 +18,10 @@ package org.apache.logging.log4j.core.pattern;
 
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
-import org.apache.logging.log4j.message.StringMapMessage;
-import org.apache.logging.log4j.util.IndexedReadOnlyStringMap;
+import org.apache.logging.log4j.message.MapMessage;
+import org.apache.logging.log4j.message.MapMessage.MapFormat;
+
+import java.util.Objects;
 
 /**
  * Able to handle the contents of the LogEvent's MapMessage and either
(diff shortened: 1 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
```diff
@@ -60,9 +60,9 @@ public static MapPatternConverter newInstance(final String[] options) {
      */
     @Override
     public void format(final LogEvent event, final StringBuilder toAppendTo) {
-        StringMapMessage msg;
-        if (event.getMessage() instanceof StringMapMessage) {
-            msg = (StringMapMessage) event.getMessage();
+        MapMessage msg;
+        if (event.getMessage() instanceof MapMessage) {
+            msg = (MapMessage) event.getMessage();
```
cakofony: Unfortunately this doesn't quite work because we cannot rely on casting map values to strings on line 82

log4j-samples/log4j-samples-loggerProperties/src/main/java/org/apache/logging/log4j/lookup/MapMessageLookup.java
```diff
@@ -52,7 +53,7 @@
     @Override
     public String lookup(final LogEvent event, final String key) {
         final Message msg = event.getMessage();
-        if (msg instanceof StringMapMessage) {
+        if (msg instanceof MapMessage) {
```
cakofony: Appending `Object` values to the stringbuilder on 69 is potentially dangerous, `MapMessage.get(String)` works around this by taking advantage of `ParameterFormatter.deepToString`.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/MapPatternConverter.java
travisspencer: just remove the cast?

log4j-samples/log4j-samples-loggerProperties/src/main/java/org/apache/logging/log4j/lookup/MapMessageLookup.java
travisspencer: added

cakofony:
Would you mind adding tests to make sure we don't reintroduce this regression? The implementation looks great :-)

travisspencer:
Will do that and the paperwork tomorrow or over the weekend.

travisspencer:
- Tests added
- Paperwork sent in
- CI server is green
- Commits are squashed

This PR can be merged and the issue can be closed, as far as I'm concerned.

If there's something that needs to be done further on my part, please let me know.

PR #252, opened 2019-01-14
Fixed message length with TLS and Syslog // fixes LOG4J2-2532

Shortfinga:
LOG4J2-2532

Length must not be part of syslog message.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/TlsSyslogFrame.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/TlsSyslogFrame.java
@@ -41,7 +41,7 @@ public class TlsSyslogFrame {
 
     @Override
     public String toString() {
-        return Integer.toString(byteLength) + Chars.SPACE + message;
+        return message;
     }
 
     @Override


Shortfinga:
Can you point me to the chapter in RFC5424 where this "feature" is mentioned? Also can you tell me how I can disable this "feature"? I searched for this behavior quite some time and did not find anything.
Thanks

rgoers:
It isn't documented correctly. The only doc for it is at http://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/layout/Rfc5424Layout.html#createLayout(org.apache.logging.log4j.core.net.Facility,%20java.lang.String,%20int,%20boolean,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20boolean,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20java.lang.String,%20boolean,%20org.apache.logging.log4j.core.layout.LoggerFields[],%20org.apache.logging.log4j.core.config.Configuration) and I don't believe even that is correct. As I recall, either syslog-ng or rsyslog accepts a message length prefix to allow you to have multi-line messages. I don't think this option really has anything to do with TLS but I am not sure. This was added in https://issues.apache.org/jira/browse/LOG4J2-338, so it has nothing to do with RFC-5424.

rgoers:
Closing this PR since the TLS support for RFC 5424 Syslog records is covered in https://tools.ietf.org/html/rfc5425 which specifies that the 2 byte length prefix is required.

PR #253, opened 2019-01-25
Log4 j2 913 - Add support for HTTP/S configuration monitoring. Support Spring Cloud Config

rgoers:
For your initial review. I still need to update the web site and look into spring cloud push notifications, so don't merge this yet.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/CompositeTriggeringPolicy.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/CompositeTriggeringPolicy.java
@@ -50,6 +50,7 @@ public final class CompositeTriggeringPolicy extends AbstractTriggeringPolicy {
     @Override
     public void initialize(final RollingFileManager manager) {
         for (final TriggeringPolicy triggeringPolicy : triggeringPolicies) {
+            LOGGER.debug("Initializing triggering policy ", triggeringPolicy.toString());
             triggeringPolicy.initialize(manager);
         }
     }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
@@ -21,6 +21,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.net.URI;
 import java.net.URL;
+import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -28,6 +29,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+
+import javax.net.ssl.HttpsURLConnection;
 
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
@@ -38,6 +41,9 @@ import org.apache.logging.log4j.core.config.plugins.util.PluginManager;
 import org.apache.logging.log4j.core.config.plugins.util.PluginType;
 import org.apache.logging.log4j.core.lookup.Interpolator;
 import org.apache.logging.log4j.core.lookup.StrSubstitutor;
+import org.apache.logging.log4j.core.net.ssl.LaxHostnameVerifier;
+import org.apache.logging.log4j.core.net.ssl.SslConfiguration;
+import org.apache.logging.log4j.core.net.ssl.SslConfigurationFactory;
 import org.apache.logging.log4j.core.util.FileUtils;
 import org.apache.logging.log4j.core.util.Loader;
 import org.apache.logging.log4j.core.util.NetUtils;
@@ -127,6 +133,8 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
     protected final StrSubstitutor substitutor = new StrSubstitutor(new Interpolator());
 
     private static final Lock LOCK = new ReentrantLock();
+
+    private static final String HTTPS = "https";
 
     /**
      * Returns the ConfigurationFactory.
@@ -294,7 +302,22 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
     protected ConfigurationSource getInputFromString(final String config, final ClassLoader loader) {
         try {
             final URL url = new URL(config);
-            return new ConfigurationSource(url.openStream(), FileUtils.fileFromUri(url.toURI()));
+            URLConnection urlConnection = url.openConnection();
+            if (url.getProtocol().equals(HTTPS)) {
+                SslConfiguration sslConfiguration = SslConfigurationFactory.getSslConfiguration();
+                if (sslConfiguration != null) {
+                    ((HttpsURLConnection) urlConnection).setSSLSocketFactory(sslConfiguration.getSslSocketFactory());
+                    if (!sslConfiguration.isVerifyHostName()) {
+                        ((HttpsURLConnection) urlConnection).setHostnameVerifier(LaxHostnameVerifier.INSTANCE);
+                    }
+                }
+            }
+            File file = FileUtils.fileFromUri(url.toURI());
+            if (file != null) {
+                return new ConfigurationSource(urlConnection.getInputStream(), FileUtils.fileFromUri(url.toURI()));
+            } else {
+                return new ConfigurationSource(urlConnection.getInputStream(), url, urlConnection.getLastModified());
+            }
         } catch (final Exception ex) {
             final ConfigurationSource source = ConfigurationSource.fromResource(config, loader);
             if (source == null) {
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
@@ -28,11 +28,18 @@ import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLConnection;
 import java.util.Objects;
 
+import javax.net.ssl.HttpsURLConnection;
+
 import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.net.ssl.LaxHostnameVerifier;
+import org.apache.logging.log4j.core.net.ssl.SslConfiguration;
+import org.apache.logging.log4j.core.net.ssl.SslConfigurationFactory;
 import org.apache.logging.log4j.core.util.FileUtils;
 import org.apache.logging.log4j.core.util.Loader;
+import org.apache.logging.log4j.core.util.Source;
 import org.apache.logging.log4j.util.LoaderUtil;
 
 /**
@@ -43,13 +50,18 @@ public class ConfigurationSource {
     /**
      * ConfigurationSource to use with Configurations that do not require a "real" configuration source.
      */
-    public static final ConfigurationSource NULL_SOURCE = new ConfigurationSource(new byte[0]);
+    public static final ConfigurationSource NULL_SOURCE = new ConfigurationSource(new byte[0], null, 0);
+    private static final String HTTPS = "https";
 
     private final File file;
     private final URL url;
     private final String location;
     private final InputStream stream;
-    private final byte[] data;
+    private volatile byte[] data;
+    private volatile Source source = null;
+    private final long lastModified;
+    // Set when the configuration has been updated so reset can use it for the next lastModified timestamp.
+    private volatile long modifiedMillis;
 
     /**
      * Constructs a new {@code ConfigurationSource} with the specified input stream that originated from the specified
@@ -64,6 +76,13 @@ public class ConfigurationSource {
         this.location = file.getAbsolutePath();
         this.url = null;
         this.data = null;
+        long modified = 0;
+        try {
+            modified = file.lastModified();
+        } catch (Exception ex) {
+            // There is a problem with the file. It will be handled somewhere else.
+        }
+        this.lastModified = modified;
     }
 
     /**
@@ -79,6 +98,24 @@ public class ConfigurationSource {
         this.location = url.toString();
         this.file = null;
         this.data = null;
+        this.lastModified = 0;
+    }
+
+    /**
+     * Constructs a new {@code ConfigurationSource} with the specified input stream that originated from the specified
+     * url.
+     *
+     * @param stream the input stream
+     * @param url the URL where the input stream originated
+     * @param lastModified when the source was last modified.
+     */
+    public ConfigurationSource(final InputStream stream, final URL url, long lastModified) {
+        this.stream = Objects.requireNonNull(stream, "stream is null");
+        this.url = Objects.requireNonNull(url, "URL is null");
+        this.location = url.toString();
+        this.file = null;
+        this.data = null;
+        this.lastModified = lastModified;
     }
 
     /**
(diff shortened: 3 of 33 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -294,7 +302,22 @@ static String extractClassLoaderUriPath(final URI uri) {
...
         try {
             final URL url = new URL(config);
-            return new ConfigurationSource(url.openStream(), FileUtils.fileFromUri(url.toURI()));
+            URLConnection urlConnection = url.openConnection();
+            if (url.getProtocol().equals(HTTPS)) {
+                SslConfiguration sslConfiguration = SslConfigurationFactory.getSslConfiguration();
+                if (sslConfiguration != null) {
+                    ((HttpsURLConnection) urlConnection).setSSLSocketFactory(sslConfiguration.getSslSocketFactory());
+                    if (!sslConfiguration.isVerifyHostName()) {
+                        ((HttpsURLConnection) urlConnection).setHostnameVerifier(LaxHostnameVerifier.INSTANCE);
```
jvz: I mean, at least you're not setting it globally, but why is this here?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -79,6 +98,24 @@ public ConfigurationSource(final InputStream stream, final URL url) {
...
+
+    /**
+     * Constructs a new {@code ConfigurationSource} with the specified input stream that originated from the specified
+     * url.
+     *
+     * @param stream the input stream
+     * @param url the URL where the input stream originated
+     * @param lastModified when the source was last modified.
+     */
+    public ConfigurationSource(final InputStream stream, final URL url, long lastModified) {
```
jvz: So I'll note here that for 3.x, we should certainly be redoing this API. It exposes way too much internal state.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/HttpWatcher.java
```diff
@@ -0,0 +1,186 @@
...
+    private static final String XML = "application/xml";
+    private static final String PROPERTIES = "text/x-java-properties";
+    private static final String TEXT = "text/plain";
+    private static final int NOT_MODIFIED = 304;
+    private static final int OK = 200;
+    private static final int BUF_SIZE = 1024;
+    private static final String HTTP = "http";
+    private static final String HTTPS = "https";
+
+    public HttpWatcher(final Configuration configuration, final Reconfigurable reconfigurable,
```
jvz: Should this be private?

log4j-core/src/main/java/org/apache/logging/log4j/core/net/ssl/SslConfigurationFactory.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Creates an SSL configuration from Log4j properties.
+ */
+public class SslConfigurationFactory {
+
+    private static final Logger LOGGER = StatusLogger.getLogger();
+    private static SslConfiguration sslConfiguration = null;
+
+    private static final String trustStorelocation = "log4j2.trustStore.location";
+    private static final String trustStorePassword = "log4j2.trustStore.password";
```
jvz: While this will resolve to `log4j2.trustStorePassword` at runtime, you should at least stay consistent in your naming scheme within this single class (see `log4j2.keyStorePassword`).

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Watcher.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.logging.log4j.core.config.Reconfigurable;
+
+/**
+ * Watches for changes in a Source and performs an action when it is modified.
+ *
+ * @see WatchManager
+ */
+public interface Watcher {
+
+    public static final String CATEGORY = "Watcher";
```
jvz:
```suggestion
    String CATEGORY = "Watcher";
```

The three keywords are all implied by a "field" in an interface.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/WatcherFactory.java
```diff
@@ -0,0 +1,89 @@
...
+            return null;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T extends Watcher> T instantiate(String name, final Class<T> clazz, final Configuration configuration,
+        final Reconfigurable reconfigurable, final List<ConfigurationListener> listeners, long lastModifiedMillis) {
+        Objects.requireNonNull(clazz, "No class provided");
+        try {
+            Constructor constructor = clazz.getConstructor(Configuration.class, Reconfigurable.class, List.class, long.class);
```
jvz:
```suggestion
            Constructor<T> constructor = clazz.getConstructor(Configuration.class, Reconfigurable.class, List.class, long.class);
```

That way you don't need a generic unchecked cast below.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/WrappedFileWatcher.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (c) 2019 Nextiva, Inc. to Present.
```
jvz: Needs updating.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/WatcherFactory.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * Copyright (c) 2019 Nextiva, Inc. to Present.
```
jvz: Needs updating.

log4j-core/src/test/java/org/apache/logging/log4j/core/util/WatchHttpTest.java
```diff
@@ -0,0 +1,175 @@
...
+    private static final String XML = "application/xml";
+
+    private static final boolean IS_WINDOWS = PropertiesUtil.getProperties().isOsWindows();
+
+    @BeforeClass
+    public static void beforeClass() {
+        try {
+            formatter = FastDateFormat.getInstance("EEE, dd MMM yyyy HH:mm:ss", TimeZone.getTimeZone("UTC"));
+        } catch (Exception ex) {
+            System.err.println("Unable to create date format.");
```
jvz: You could use `LowLevelLogUtil` here if you want.

log4j-core/src/test/java/org/apache/logging/log4j/core/util/WatchHttpTest.java
```diff
@@ -0,0 +1,175 @@
...
+        final ConfigurationScheduler scheduler = new ConfigurationScheduler();
+        scheduler.incrementScheduledItems();
+        final WatchManager watchManager = new WatchManager(scheduler);
+        watchManager.setIntervalSeconds(1);
+        scheduler.start();
+        watchManager.start();
+        try {
+            watchManager.watch(new Source(url.toURI(), previous.getTimeInMillis()), new HttpWatcher(configuration, null,
+                listeners, previous.getTimeInMillis()));
+            Thread.sleep(1500);
```
jvz: Is there any way you could instead expose a `CoundDownLatch` for testing purposes? Then you could `await()` it instead of allowing any problems under CI load.

log4j-spring-cloud-config/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2CloudConfigLoggingSystem.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Override
+    protected String[] getStandardConfigLocations() {
+        String[] locations = super.getStandardConfigLocations();
+        PropertiesUtil props = new PropertiesUtil(new Properties());
+        String location = props.getStringProperty(ConfigurationFactory.CONFIGURATION_FILE_PROPERTY);
+        if (location != null) {
+            List<String> list = Arrays.asList(super.getStandardConfigLocations());
+            list.add(location);
+            locations = list.toArray(new String[list.size()]);
```
jvz: From the latest benchmarks I remember reading about, it's generally faster to just pass 0 here instead of the real size since `toArray()` has to check `list.size()` first anyways.

log4j-spring-cloud-config/log4j-spring-cloud-config-server/src/main/java/org/apache/logging/log4j/spring/cloud/config/server/controller/Log4jResourceController.java
```diff
@@ -0,0 +1,200 @@
...
+ * @author Dave Syer
+ * @author Daniel Lavoie
+ *
+ */
+@RestController
+@RequestMapping(method = RequestMethod.GET, path = "${spring.cloud.config.server.prefix:}/resource")
+public class Log4jResourceController {
+
+    @Autowired
+	private ResourceRepository resourceRepository;
```
jvz: Seems like this file has a mix of tabs and spaces.

log4j-spring-cloud-config/log4j-spring-cloud-config-server/src/main/java/org/apache/logging/log4j/spring/cloud/config/server/controller/Log4jResourceController.java
```diff
@@ -0,0 +1,200 @@
...
+        name = resolveName(name);
+        label = resolveLabel(label);
+        Resource resource = this.resourceRepository.findOne(name, profile, label, path);
+        if (checkNotModified(request, resource)) {
+            // Content was not modified. Just return.
+            return null;
+        }
+        // ensure InputStream will be closed to prevent file locks on Windows
+        try (InputStream is = resource.getInputStream()) {
+            String text = StreamUtils.copyToString(is, Charset.forName("UTF-8"));
```
jvz: Can use `StandardCharsets.UTF_8` here.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
rgoers: HostNameVerifier? I copied this code from the HttpAppender but since we haven't configured yet I had to get creative in how the SSL settings would be obtained.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
rgoers: Good luck with that. All this information has to be passed from one configuration to another to allow reconfiguration to occur.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/HttpWatcher.java
rgoers: No. It is called by unit tests.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/ssl/SslConfigurationFactory.java
rgoers: Fixed

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Watcher.java
rgoers: Fixed

log4j-core/src/test/java/org/apache/logging/log4j/core/util/WatchHttpTest.java
rgoers: It is just a unit test. I am not sure what using LowLevelLogUtil would gain me here.

log4j-core/src/test/java/org/apache/logging/log4j/core/util/WatchHttpTest.java
rgoers: Turns out I didn't need to sleep at all anyway...

log4j-spring-cloud-config/log4j-spring-cloud-config-server/src/main/java/org/apache/logging/log4j/spring/cloud/config/server/controller/Log4jResourceController.java
rgoers: This file was acquired from Spring. They use tabs. I'll reformat it.

log4j-spring-cloud-config/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2CloudConfigLoggingSystem.java
rgoers: I had always thought that if you pass in an array that toArray would just use it.

log4j-spring-cloud-config/log4j-spring-cloud-config-server/src/main/java/org/apache/logging/log4j/spring/cloud/config/server/controller/Log4jResourceController.java
rgoers: Fixed

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
rgoers: It is handled in SslConfigurationFactory.

log4j-spring-cloud-config/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2CloudConfigLoggingSystem.java
jvz: The key here is that even if you pass it an array, toArray has to check the length of the passed array against the length of the array being copied. toArray will resize the array if needed, so there's less if-checks executed in the array of length 0 execution path. Basically, toArray() is programmed to construct the new array regardless, so save yourself the bytecode.

rgoers: You must be thinking of a different method. The javadoc for this says "Returns a fixed-size list backed by the specified array. (Changes to the returned list "write through" to the array.)". Looking at the implementation that is exactly what it does. Passing an array of size 0 would require me to manually populate the list from the array, which is the whole point of this method.

jvz: Oh I think you're right. Found more info [here](https://shipilev.net/blog/2016/arrays-wisdom-ancients/) and it doesn't seem to be an issue in this scenario (or in general anymore; this old issue seems to have been fixed). The key difference here is that calling `size()` on an array list is `O(1)`, so it's not as big an issue. If you were calling, say, `Collection::toArray` on a collection that doesn't cache its size (random glance at some standard collection classes indicates they normally do cache that), or worse, if the collection changed between checking the size and calling toArray, then that causes the performance latency.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/CompositeTriggeringPolicy.java
```diff
@@ -50,6 +50,7 @@ private CompositeTriggeringPolicy(final TriggeringPolicy... triggeringPolicies)
     @Override
     public void initialize(final RollingFileManager manager) {
         for (final TriggeringPolicy triggeringPolicy : triggeringPolicies) {
+            LOGGER.debug("Initializing triggering policy ", triggeringPolicy.toString());
```
cakofony: nit: I don't think toString isn't necessary here, the logger will get the string value if debug is enabled

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
jvz: I've considered in the past porting over or implementing a subset of the [`Resource`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/io/Resource.html) interface from Spring. Do you think such an API would be more useful than having to inspect the ConfigurationSource union-like class?

PR #256, opened 2019-02-11
(doc) Clarify Default Rollover Strategy

PPazderski:
I got surprised by the rollover rename strategy because I haven't read the manual carefully enough.
This pull request may help to improve the manual at this point.
It will:
1. reorder the archived log file names to list the newest file first. This should clarify the difference between fileIndex=min and fileIndex=max in addition to the description column.
2. swap description of fileIndex=min and fileInex=max to list the default one first.


rgoers:
Looks OK to me.

PR #257, opened 2019-02-20
LOG4J2-2552 Allow access to the backlog state of the disruptor to improve program…

ptogher:
…atic reconfiguration of async appenders.

This change is to help support programmatically switching async appenders. In cases where it is important that logged messages go to the appender that was configured at the time of writing the log event, the flushing the appender buffer to disk is not sufficient, as the log events may still be in the disruptor. This change provides access to test if the disruptor has items in its backlog. This becomes useful in an application that is controlling its logging config and logging events to ensure that no new logging events are generated between removing one appender and adding another, so that it can also ensure the change does not happen until all pending log events are processed.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerContext.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerContext.java
@@ -106,6 +106,13 @@ public class AsyncLoggerContext extends LoggerContext {
         super.stop(timeout, timeUnit);
         return true;
     }
+    
+    /**
+     * Returns {@code true} if the underlying disruptor still has unprocessed events.
+     */
+    public boolean hasBacklog() {
+        return loggerDisruptor.hasBacklog();
+    }
 
     /**
      * Creates and returns a new {@code RingBufferAdmin} that instruments the ringbuffer of the {@code AsyncLogger}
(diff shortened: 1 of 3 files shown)


ptogher:
I am not sure if the build errors are due to my changes. I get the same errors if i do a clean master build. However, last Travis master build was OK. I cannot kick off a new build to verify. Does the PR build process use different build settings? I can see  PR https://github.com/apache/logging-log4j2/pull/256 has the same issue, and they only changed a document.

ptogher:
The problem appears to be the revapi.json, or the reference version. The deprecated methods have been removed on master, but I think the revapi plugin is comparing against the last release, and so it thinks they have just been deleted. Not sure how the last master build manged to work though. Was the check skipped as the methods were actually supposed to be deleted?

ptogher:
Have raised a Jira, if that helps... LOG4J2-2552

remkop:
Hi @ptogher, sorry I haven't had a chance to look at this earlier.

The information you are looking for is already available in current versions of Log4j, although in slightly less convenient form, if you do:

```
RingBufferAdmin admin = asyncLoggerContext.createRingBufferAdmin();
boolean hasBacklog = admin.getRemainingCapacity() != admin.getBufferSize();
```

Please be aware that if you have an application that is logging from multiple threads, and one of these threads asks the question "do we have a logging backlog", the answer may be stale immediately. This is simply the nature of multi-threading. The answer is only reliable if the application logs from a single thread, or if you can somehow stop all threads from producing log events.

ptogher:
Ooh, many thanks, That's really useful. I missed this when i was looking for a solution. Completely understand about the threading. We have very specific thread controls in place to manage this.

PR #260, opened 2019-03-01
[LOG4J2-2561] JEP223 version detection fix for JDK 9 and up

ulrichenslin:
Fix version detection for JDK 9 and up

### DIFF
--- a/log4j-api/src/test/java/org/apache/logging/log4j/util/ConstantsTest.java
+++ b/log4j-api/src/test/java/org/apache/logging/log4j/util/ConstantsTest.java
@@ -1,0 +1,20 @@
+package org.apache.logging.log4j.util;
+
+import org.junit.Test;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Properties;
+
+import static org.junit.Assert.*;
+
+public class ConstantsTest {
+
+    @Test
+    public void testJdkVersionDetection() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        assertEquals(1, Constants.getMajorVersion("1.1.2"));
+        assertEquals(8, Constants.getMajorVersion("1.8.2"));
+        assertEquals(9, Constants.getMajorVersion("9.1.1"));
+        assertEquals(11, Constants.getMajorVersion("11.1.1"));
+    }
+}
--- a/log4j-api/src/test/java/org/apache/logging/log4j/util/ConstantsTest.java
+++ b/log4j-api/src/test/java/org/apache/logging/log4j/util/ConstantsTest.java
@@ -2,16 +2,12 @@ package org.apache.logging.log4j.util;
 
 import org.junit.Test;
 
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.Properties;
-
 import static org.junit.Assert.*;
 
 public class ConstantsTest {
 
     @Test
-    public void testJdkVersionDetection() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+    public void testJdkVersionDetection() {
         assertEquals(1, Constants.getMajorVersion("1.1.2"));
         assertEquals(8, Constants.getMajorVersion("1.8.2"));
         assertEquals(9, Constants.getMajorVersion("9.1.1"));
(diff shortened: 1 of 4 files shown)


jvz:
Can you add a jira issue for this as well? https://issues.apache.org/jira/browse/LOG4J2

ulrichenslin:
Jira LOG4J2-2561 added for this PR.

ulrichenslin:
Added test for the detection testing.

@remkop Thanks for the advise on reseting the java.version.  I missed that in my initial test.

remkop:
update: looking at this on my phone, I missed that the PR also includes a test. Looks good to me.

remkop:
Oh, you just added that? Ok, thanks!

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractStringLayout.java
```diff
@@ -134,7 +134,13 @@ private static boolean isPreJava8() {
         final String version = System.getProperty("java.version");
```
cakofony: Can we make this method delegate to a package-private `static boolean isPreJava8(String javaVersion)` which we can invoke directly from the test? The reflective method is likely to cause failures if we refactor this code in the future, and setting the `java.version` system property may have other consequences which I'd prefer to avoid.

ulrichenslin: I agree, but wanted to make the code change as small as possible, since it is a very minor change.  

If we want to be more adventures, we should move the org.apache.logging.log4j.util.Constants#getMajorVersion to core, and just have everything call that.  Then the version detection is centralised and can we can test it as a public method.

How to you advise we move forward on this.

rgoers: Personally, I dislike checking for versions. I much prefer checking for features. But seeing that this was already implemented I wouldn't expect you to change that.

I am a bit confused by your question though. It looks like Constants has a getMajorVersion method that has the same problem. I would recommend you just fix that and then have AbstractStringLayout reference Constants.JAVA_MAJOR_VERSION

ulrichenslin: Please see updated PR.  This is a cleaner implementation.

log4j-core/src/test/java/org/apache/logging/log4j/core/layout/AbstractStringLayoutTest.java
```diff
@@ -15,7 +15,10 @@
  * limitations under the license.
  */
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.nio.charset.Charset;
+import java.util.Properties;
```
cakofony: Could you please fix the unused imports in this file?

log4j-api/src/test/java/org/apache/logging/log4j/util/ConstantsTest.java
```diff
@@ -0,0 +1,20 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Properties;
+
+import static org.junit.Assert.*;
+
+public class ConstantsTest {
+
+    @Test
+    public void testJdkVersionDetection() throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
```
cakofony: nit: No need to declare these exception types now that we're calling `getMajorVersion` directly.

ulrichenslin: The code has been cleaned up.

ulrichenslin:
The tests on the CI is failing, but the failures are not related to the changes made in this PR.  

How do we move forward?

ulrichenslin:
@jvz Thanks for approving the PR.

To get this ported to 2.11, do I just create another PR with the same change on a 'release-2.x' branch, with the same Jira number?

jvz:
That's one simple way to do it. I don't recall if I've merged any backported commits myself, though many of the other developers here will merge a commit to both branches concurrently.

rgoers:
I'd appreciate a new PR. I have tried various options like cherry picking but never had much luck. I usually end up doing it by hand.

ulrichenslin:
@rgoers @jvz 
New for release 2 PR on :  https://github.com/apache/logging-log4j2/pull/262

PR #263, opened 2019-03-21
Fix jackson layout performance regression introduced with LOG4J2-2341

cakofony:
This makes two changes to the previous fix:
1. Special casing applies only to LogEvents which implement Message,
   where previously it applied to any event which was not a
   Log4jLogEvent
2. The special case code no longer uses Log4jLogEvent.createMemento,
   but instead wraps the existing event in a simple wrapper LogEvent
   which delegates directly to the original event.

### DIFF
--- a/log4j-layout-jackson/src/main/java/org/apache/logging/log4j/jackson/AbstractJacksonLayout.java
+++ b/log4j-layout-jackson/src/main/java/org/apache/logging/log4j/jackson/AbstractJacksonLayout.java
@@ -22,15 +22,21 @@ import java.nio.charset.Charset;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
+import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.ThreadContext;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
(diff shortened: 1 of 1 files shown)


log4j-layout-jackson/src/main/java/org/apache/logging/log4j/jackson/AbstractJacksonLayout.java
```diff
@@ -216,11 +222,126 @@ public Object getLogEvent() {
     protected static final String DEFAULT_EOL = "\r\n";
     protected static final String COMPACT_EOL = Strings.EMPTY;
     private static LogEvent convertMutableToLog4jEvent(final LogEvent event) {
-        // TODO Jackson-based layouts have certain filters set up for Log4jLogEvent.
-        // TODO Need to set up the same filters for MutableLogEvent but don't know how...
-        // This is a workaround.
```
cakofony: These comments do not appear to be correct. Rhe code is confusing because we register a Jackson property filter using id `org.apache.logging.log4j.core.impl.Log4jLogEvent`, however the code is not at all specific to Log4jLogEvent.

cakofony:
Any idea why we have a second unreferenced`AbstractJacksonLayout` in the `org.apache.logging.log4j.jackson.layout` package? release-2.x only provides AbstractJacksonLayout in `org.apache.logging.log4j.jackson.layout`, so perhaps something got lost while splitting out `log4j-layout-jackson`?

cakofony:
Closing this out, @rgoers is including a fix with LOG4J2-913

PR #267, opened 2019-04-30
LOG4J2-2598: GzCompressAction supports configurable compression levels

carterkozak:
Updated the implementation to use the BUF_SIZE constant for the
GZIPOutputStream output buffer as well.


remkop:
Nice! Thanks for carefully preserving backwards compatibility.

carterkozak:
Thanks for looking over this @remkop :-)

PR #269, opened 2019-05-20
LOG4J2-2609: Implement Synchronous AsyncQueueFullPolicy

carterkozak:
https://issues.apache.org/jira/browse/LOG4J2-2609

* Implement "Synchronous" AsyncQueueFullPolicy
* AsyncQueueFullPolicy short values are case insensitive

Benchmarks for [LOG4J2-2606](https://issues.apache.org/jira/browse/LOG4J2-2606):
```
Benchmark                                                      (queueFullPolicy)   Mode  Cnt        Score         Error  Units
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads            ENQUEUE  thrpt    3   369559.791 ± 1307230.236  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads        SYNCHRONOUS  thrpt    3  1059713.811 ± 1910806.587  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread                 ENQUEUE  thrpt    3  1451843.607 ±  355606.050  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread             SYNCHRONOUS  thrpt    3  1393328.851 ±  589759.385  ops/s
```

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncQueueFullPolicyFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncQueueFullPolicyFactory.java
@@ -47,6 +47,7 @@ public class AsyncQueueFullPolicyFactory {
     static final String PROPERTY_VALUE_DEFAULT_ASYNC_EVENT_ROUTER = "Default";
     static final String PROPERTY_VALUE_DISCARDING_ASYNC_EVENT_ROUTER = "Discard";
     static final String PROPERTY_NAME_DISCARDING_THRESHOLD_LEVEL = "log4j2.DiscardThreshold";
+    static final String PROPERTY_VALUE_SYNCHRONOUS_ASYNC_EVENT_ROUTER = "Synchronous";
 
     private static final Logger LOGGER = StatusLogger.getLogger();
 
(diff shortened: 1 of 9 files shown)


garydgregory:
It key IMO to also have documentation that explains the new feature and how it compares to other polices.

carterkozak:
Ack, I will add docs to this PR later today

carterkozak:
I may not end up committing the synchronous policy because it's not terribly helpful outside of testing. If we (carefully) synchronize around the enqueue operation we can get more throughput, avoid poorly ordered events, and avoid creating thread-local buffers outside of the background thread:

```
Benchmark                                                      (queueFullPolicy)   Mode  Cnt        Score        Error  Units
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads            ENQUEUE  thrpt    3  1302070.388 ± 713643.057  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads        SYNCHRONOUS  thrpt    3  1151360.276 ±  99047.131  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread                 ENQUEUE  thrpt    3  1330328.259 ± 749542.377  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread             SYNCHRONOUS  thrpt    3  1381320.895 ± 507036.686  ops/s
```

remkop:
I agree that the SYNCHRONOUS policy won't be very useful in reality. (A few years ago when this was the default policy we regularly got jumbled logs when our app was logging too much; my team members hated it: it made the log files useless...)

When you say "synchronize around the enqueue operation", that sounds wrong to me. The point of the Disruptor is to avoid synchronization in the app and leave that to the WaitStrategy. Shouldn't we just use a different WaitStrategy instead? The Disruptor offers a range of options:

* [BlockingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/BlockingWaitStrategy.java) - essentially this gives synchronization, if I understand correctly...
* [BusySpinWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/BusySpinWaitStrategy.java) - the other extreme: sacrifice a core to get the absolute minimal latency
* [LiteBlockingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/LiteBlockingWaitStrategy.java)
* [LiteTimeoutBlockingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/LiteTimeoutBlockingWaitStrategy.java)
* [PhasedBackoffWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/PhasedBackoffWaitStrategy.java)
* [SleepingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/SleepingWaitStrategy.java)
* [TimeoutBlockingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/TimeoutBlockingWaitStrategy.java)
* [YieldingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/YieldingWaitStrategy.java)

garydgregory:
Why not make it configurable so users and pick and configure each one?

remkop:
@garydgregory It already is configurable: https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java#L64

remkop:
The question is: what is the default to use in Log4j that gives the best balance between performance and safety.

garydgregory:
I see, but does it make sense to add the missing ones?

remkop:
Probably yes, but the [documentation|https://logging.apache.org/log4j/2.x/manual/async.html#SysPropsAllAsync] would need to be updated with details on the trade-offs for the newly added wait strategies. Let's wait to hear what the Disruptor maintainers come back with.

carterkozak:
@remkop I've tested with several `WaitStrategy` implementations and haven't observed a great deal of difference between them in this benchmark for either CPU utilization or throughput.

I think part of the issue is that disruptor provides order across threads, so each time an event from the queue is handled, all waiting threads are notified, and must all do work each time space becomes available in the queue. Synchronizing on enqueue limits the number of threads which must work every time an event is offered, allowing other blocking threads to idle instead.

I don't have a great deal of experience with this library outside of log4j, so it's very possible I'm running in the wrong direction. Hopefully the disruptor maintainers can provide helpful feedback :-)

carterkozak:
Closing this PR as I've broken it into a few pieces:
1. Initial benchmarks have already merged: https://github.com/apache/logging-log4j2/commit/15aadc24773bdf2b2b537899594d513d3eb48e2c
2. Case insensitive `AsyncQueueFullPolicy` short names: https://github.com/apache/logging-log4j2/pull/272
3. Proposal to fix LOG4J2-2606: https://github.com/apache/logging-log4j2/pull/273 We can move discussion to that PR.

PR #272, opened 2019-05-21
LOG4J2-2611: AsyncQueueFullPolicy configuration short values are case insensitive

carterkozak:


remkop:
I reviewed this PR to make configuration case-insensitive and improve the javadoc by adding a warning for the SYNCHRONOUS event route.

Looks good to merge.

(As an aside, I've started to dislike the "SYNCHRONOUS" name for that event route. It sounds too innocuous. :-) It should really be called something like "BYPASS_QUEUE" or something... Anyway, a bit late to change that perhaps.)

carterkozak:
Thanks Remko. I like the sound of `BYPASS_QUEUE` and agree that `SYNCHRONOUS` makes `ENQUEUE` sound fast or asynchronous by comparison. We could deprecate and add the new type, though that's probably more trouble than it's worth...

remkop:
Maybe this name change is something to do in log4j 3.0.

garydgregory:
You can add the new name and deprecate the old one, no?

carterkozak:
That is correct.

PR #273, opened 2019-05-21
LOG4J2-2606: Substantially improve async logging performance under heavy load

carterkozak:
This change applies synchronization to AsyncLoggerDisruptor blocking enqueue
operations when the asynchronous logging queue is completely full.
This new behavior may be disabled using the boolean property
`AsyncLogger.SynchronizeEnqueueWhenQueueFull` (default true).

Alternatives tested:
* All available lmax disruptor `WaitStrategy` implementations.
  None of the available implementations provided a substantial
  difference in throughput or CPU utilization. In all cases
  my processor was stuck at or near 100% across all six cores.
* Based on feedback from https://github.com/LMAX-Exchange/disruptor/issues/266
  I attempted avoiding the blocking enqueue entirely in favor of
  `LockSupport.parkNanos` with values ranging from 1,000 through
  1,000,000. 1,000,000 provided the most throughput among parkNanos
  values tested, but is too long for most scenarios.
* Semaphore instead of synchronized. Testing with permits
  equivalent to the available processor count yielded higher
  CPU utilization and lower throughput than a semaphore
  with a single permit. Given that most work occurs on
  the (single) background thread, there's not much reason
  to allow multiple enqueues simultaneously.

Updated benchmarks adding "ENQUEUE_UNSYNCHRONIZED" following
the old unsynchronized path, where "ENQUEUE" shows an improvement.
Also added benchmarks for mixed asynchronous logging.

```
Benchmark                                                      (asyncLoggerType)       (queueFullPolicy)   Mode  Cnt        Score         Error  Units
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads      ASYNC_CONTEXT                 ENQUEUE  thrpt    3  1196852.082 ±  656921.903  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads      ASYNC_CONTEXT  ENQUEUE_UNSYNCHRONIZED  thrpt    3   299829.995 ±  672293.830  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads      ASYNC_CONTEXT             SYNCHRONOUS  thrpt    3  1119699.755 ±  985194.962  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads       ASYNC_CONFIG                 ENQUEUE  thrpt    3  1255978.339 ±  208596.091  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads       ASYNC_CONFIG  ENQUEUE_UNSYNCHRONIZED  thrpt    3   252966.661 ±   31303.171  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads       ASYNC_CONFIG             SYNCHRONOUS  thrpt    3  1050306.069 ± 1564741.862  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread           ASYNC_CONTEXT                 ENQUEUE  thrpt    3  1256681.352 ± 1858635.191  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread           ASYNC_CONTEXT  ENQUEUE_UNSYNCHRONIZED  thrpt    3  1357446.702 ±  825753.509  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread           ASYNC_CONTEXT             SYNCHRONOUS  thrpt    3  1240065.456 ±  832767.348  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread            ASYNC_CONFIG                 ENQUEUE  thrpt    3  1376691.658 ±  614955.853  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread            ASYNC_CONFIG  ENQUEUE_UNSYNCHRONIZED  thrpt    3  1403978.404 ±  776352.491  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread            ASYNC_CONFIG             SYNCHRONOUS  thrpt    3  1347219.410 ± 2608889.236  ops/s
```

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigDelegate.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigDelegate.java
@@ -46,6 +46,11 @@ public interface AsyncLoggerConfigDelegate {
      */
     EventRoute getEventRoute(final Level level);
 
+    /**
+     * Enqueues the {@link LogEvent} on the mixed configuration ringbuffer.
+     * This method must only be used after {@link #tryEnqueue(LogEvent, AsyncLoggerConfig)} returns <code>false</code>
+     * indicating that the ringbuffer is full, otherwise it may incur unnecessary synchronization.
+     */
     void enqueueEvent(LogEvent event, AsyncLoggerConfig asyncLoggerConfig);
 
     boolean tryEnqueue(LogEvent event, AsyncLoggerConfig asyncLoggerConfig);
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
@@ -20,7 +20,9 @@ package org.apache.logging.log4j.core.async;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 
+import com.lmax.disruptor.EventTranslatorVararg;
 import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Marker;
 import org.apache.logging.log4j.core.AbstractLifeCycle;
 import org.apache.logging.log4j.core.jmx.RingBufferAdmin;
 import org.apache.logging.log4j.core.util.Log4jThreadFactory;
@@ -32,6 +34,7 @@ import com.lmax.disruptor.TimeoutException;
 import com.lmax.disruptor.WaitStrategy;
 import com.lmax.disruptor.dsl.Disruptor;
 import com.lmax.disruptor.dsl.ProducerType;
+import org.apache.logging.log4j.message.Message;
 
 /**
  * Helper class for async loggers: AsyncLoggerDisruptor handles the mechanics of working with the LMAX Disruptor, and
@@ -42,6 +45,8 @@ import com.lmax.disruptor.dsl.ProducerType;
 class AsyncLoggerDisruptor extends AbstractLifeCycle {
     private static final int SLEEP_MILLIS_BETWEEN_DRAIN_ATTEMPTS = 50;
     private static final int MAX_DRAIN_ATTEMPTS_BEFORE_SHUTDOWN = 200;
+
+    private final Object queueFullEnqueueLock = new Object();
 
     private volatile Disruptor<RingBufferLogEvent> disruptor;
     private String contextName;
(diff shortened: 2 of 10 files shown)


carterkozak:
@remkop I can't seem to tag you as a reviewer. I'd appreciate your feedback at your convenience :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -46,6 +46,11 @@
     private static final int RINGBUFFER_DEFAULT_SIZE = 256 * 1024;
     private static final int RINGBUFFER_NO_GC_DEFAULT_SIZE = 4 * 1024;
 
+    // LOG4J2-2606: Disruptor spins enqueuing events across multiple threads when the queue is full.
+    // CPU utilization is significantly reduced by restricting access to the enqueue operation.
+    static final boolean SYNCHRONIZE_ENQUEUE_WHEN_QUEUE_IS_FULL = PropertiesUtil.getProperties()
+            .getBooleanProperty("AsyncLogger.SynchronizeEnqueueWhenQueueFull", true);
```
carterkozak: TODO: Update documentation with this value.

remkop:
I’ll take a look.

remkop:
It still seems strange to me that the application (Log4j in this case) needs to do synchronization when using the Disruptor.

I'm also surprised that all the WaitStrategies were reported to show the same behaviour when the ringbuffer is full and under load. The proposed change (synchronizing on a mutex object) seems to do the same as the [BlockingWaitStrategy](https://github.com/LMAX-Exchange/disruptor/blob/master/src/main/java/com/lmax/disruptor/BlockingWaitStrategy.java) provided by the Disruptor. So I would expect at least the BlockingWaitStrategy to show some improvement...

Can we experiment a bit before moving ahead? 
@carterkozak You reported this happened in your production environment. Can I ask how you've dealt with that? Is your organization waiting for a quick Log4j fix, or do you have some other interim solution?

carterkozak:
> Can we experiment a bit before moving ahead?
@carterkozak You reported this happened in your production environment. Can I ask how you've dealt with that? Is your organization waiting for a quick Log4j fix, or do you have some other interim solution?

Absolutely, I'd much rather get this done correctly than quickly. For the time being we've implemented a custom `AsyncQueueFullPolicy` which uses `EventRoute.SYNCHRONOUS` instead of `ENQUEUE`. Event order isn't terribly problematic in our case because because our events use a custom json format, and flow through an automated pipeline where they're indexed and presented in order based on timestamps.

> I'm also surprised that all the WaitStrategies were reported to show the same behaviour when the ringbuffer is full and under load. The proposed change (synchronizing on a mutex object) seems to do the same as the BlockingWaitStrategy provided by the Disruptor. So I would expect at least the BlockingWaitStrategy to show some improvement...

The BlockingWaitStrategy in disruptor master does appear much closer to what I've written than the version in 3.4.2, however if I backport that version of the file to this branch and run benchmarks (values in queuFullPolicy=ENQUEUE_UNSYNCHRONIZED), the java process CPU utilization exceeds 1000% (across 6 cores + hyperthreading) and throughput is lower than the timeout strategy. (disclaimer: benchmarks run on a laptop, cpu throttling and other processes produce noise)

```
Benchmark                                                           (queueFullPolicy)   Mode  Cnt        Score         Error  Units
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads                 ENQUEUE  thrpt    3  1269003.278 ±  858705.609  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads  ENQUEUE_UNSYNCHRONIZED  thrpt    3   296842.480 ± 1060494.701  ops/s
ConcurrentAsyncLoggerToFileBenchmark.concurrentLoggingThreads             SYNCHRONOUS  thrpt    3  1095597.447 ±  767282.544  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread                      ENQUEUE  thrpt    3  1231449.118 ± 1874760.323  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread       ENQUEUE_UNSYNCHRONIZED  thrpt    3  1329153.582 ± 2409041.282  ops/s
ConcurrentAsyncLoggerToFileBenchmark.singleLoggingThread                  SYNCHRONOUS  thrpt    3  1336944.414 ±  636546.820  ops/s
```

carterkozak:
The WaitStrategy appears to be designed to awaken all producer threads each time an event from the queue is consumed. With more blocked threads, there will be more wakeups and CPU churn.

The more I think about this, the more I like the idea of a semaphore with configurable permits (default = 1 based on benchmarks) over my current mutex approach. That will allow us to optimize throughput in cases that I haven't fully considered while reducing the risk of N spinning threads. Otherwise there's not much reason to allow WaitStrategy to be configured at all.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
```diff
@@ -202,32 +209,95 @@ private boolean hasLog4jBeenShutDown(final Disruptor<RingBufferLogEvent> aDisrup
...
-            LOGGER.warn("[{}] Ignoring log event after log4j was shut down: {} [{}] {}", contextName,
-                    translator.level, translator.loggerName, translator.message.getFormattedMessage()
-                            + (translator.thrown == null ? "" : Throwables.toStringList(translator.thrown)));
+            logWarningOnNpeFromDisruptorPublish(translator);
             return false;
         }
     }
 
-    void enqueueLogMessageInfo(final RingBufferLogEventTranslator translator) {
+    void enqueueLogMessageQueueFullBlocking(final RingBufferLogEventTranslator translator) {
```
remkop: I would suggest to remove "Blocking" from the method name; whether it actually blocks depends on configuration.
I do like that you added "queue full" to the method name, that is informative.
What about `enqueueLogMessageWhenQueueFull`?

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
```diff
@@ -202,32 +209,95 @@ private boolean hasLog4jBeenShutDown(final Disruptor<RingBufferLogEvent> aDisrup
...
+            // Note: we deliberately access the volatile disruptor field afresh here.
+            // Avoiding this and using an older reference could result in adding a log event to the disruptor after it
+            // was shut down, which could cause the publishEvent method to hang and never return.
             return disruptor.getRingBuffer().tryPublishEvent(translator);
         } catch (final NullPointerException npe) {
             // LOG4J2-639: catch NPE if disruptor field was set to null in stop()
-            LOGGER.warn("[{}] Ignoring log event after log4j was shut down: {} [{}] {}", contextName,
-                    translator.level, translator.loggerName, translator.message.getFormattedMessage()
-                            + (translator.thrown == null ? "" : Throwables.toStringList(translator.thrown)));
+            logWarningOnNpeFromDisruptorPublish(translator);
```
remkop: I like the exception handling in a separate method. Looks much cleaner!

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
```diff
@@ -202,32 +209,95 @@ private boolean hasLog4jBeenShutDown(final Disruptor<RingBufferLogEvent> aDisrup
...
+                        marker, // 4
+                        msg, // 5
+                        thrown); // 6
+            }
         } catch (final NullPointerException npe) {
             // LOG4J2-639: catch NPE if disruptor field was set to null in stop()
             LOGGER.warn("[{}] Ignoring log event after log4j was shut down: {} [{}] {}", contextName,
-                    translator.level, translator.loggerName, translator.message.getFormattedMessage()
-                            + (translator.thrown == null ? "" : Throwables.toStringList(translator.thrown)));
+                    level, fqcn, msg.getFormattedMessage()
```
remkop: Maybe factor out this exception handling to a separate method also?

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigDelegate.java
```diff
@@ -46,6 +46,11 @@
      */
     EventRoute getEventRoute(final Level level);
 
+    /**
+     * Enqueues the {@link LogEvent} on the mixed configuration ringbuffer.
+     * This method must only be used after {@link #tryEnqueue(LogEvent, AsyncLoggerConfig)} returns <code>false</code>
+     * indicating that the ringbuffer is full, otherwise it my incur unnecessary synchronization.
```
remkop: Nit: "my incur" -> "may incur"

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -153,4 +155,43 @@ public Long call() {
...
+    static Semaphore createAsyncConfigQueueFullEnqueuSemaphore() {
+        return createAsyncQueueFullEnqueuSemaphore("AsyncLoggerConfig.AsyncQueueFullEnqueueSemaphorePermits");
+    }
+
+    /**
+     * LOG4J2-2606: Disruptor spins enqueuing events across multiple threads when the queue is full.
+     * CPU utilization is significantly reduced by restricting access to the enqueue operation.
+     * Non-positive values disable the semaphore.
+     */
+    private static Semaphore createAsyncQueueFullEnqueuSemaphore(String property) {
```
remkop: Nit: ...EnqueuSemaphore -> ...EnqueueSemaphore

Also, I'm not sure that it is accurate to say "Disruptor spins", that makes it sound like this is by design. What about something like this:
> LOG4J2-2606: users encountered excessive CPU utilization with Disruptor v3.4.2 when the application
> was logging more than the underlying appender could keep up with and the ringbuffer became full,
> especially when the number of application threads vastly outnumbered the number of cores.
> CPU utilization is significantly reduced by...

remkop:
I did an initial review, but I haven't looked at what the trade-offs are for the use of Semaphor versus synchronized block or other alternatives. I need some more time for that. :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
carterkozak: Good idea, thanks for the suggestion

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
carterkozak: Should allow the fast path to inline more readily as well :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
carterkozak: > Nit: ...EnqueuSemaphore -> ...EnqueueSemaphore

Oops, that's embarrassing! Fixed and updated the comment based on your suggestion.

carterkozak:
Thank you for the review, no rush on the remaining pieces. I don't plan to merge any of this until you're comfortable with it.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -153,4 +155,45 @@ public Long call() {
...
+        return createAsyncQueueFullEnqueueSemaphore("AsyncLogger.AsyncQueueFullEnqueueSemaphorePermits");
+    }
+
+    /**
+     * Creates a new {@link Semaphore} or null based on
+     * <pre>AsyncLoggerConfig.AsyncQueueFullEnqueueSemaphorePermits</pre>.
+     */
+    static Semaphore createAsyncConfigQueueFullEnqueueSemaphore() {
+        return createAsyncQueueFullEnqueueSemaphore("AsyncLoggerConfig.AsyncQueueFullEnqueueSemaphorePermits");
+    }
```
carterkozak: TODO: Add these properties to our documentation. I'm holding off until we're happy with the design.

carterkozak:
Friendly ping on this. Not a rush, just want to make sure we don't forget about it for 2.12.0 :-)

remkop:
Thanks for the reminder!

I thought a synchronized block would express the intentions better because it’s simpler. 
The PR introduces a system property `AsyncLogger.AsyncQueueFullEnqueueSemaphorePermits` that suggests that users can configure multiple permits but I’m not clear on when users should or shouldn’t do this, and what the trade-offs are for configuring multiple permits. If there’s an actual use case for this then it’s fine. If it’s just future-proofing then I’d prefer not to make it configurable.

carterkozak:
Thanks @remkop! I don't have a use case where the semaphore would be better than a synchronized block, it was an attempt to allow consumers to take advantage of alternative disruptor WaitStrategy implementations while bounding the maximum number of allowed disruptor waiting threads. I will replace the semaphore with a simple synchronized block.

PR #274, opened 2019-06-02
[LOG4J2-2600] Add declarative Groovy configuration DSL

jvz:
See https://issues.apache.org/jira/browse/LOG4J2-2600

This is currently a draft PR as it doesn't even seem to work properly at the moment. However, the gist of what's going to be supportable is here. I may be able to re-use the Component class from the configuration builder API, but that API is a bit too strongly typed for this use case.


rgoers:
Anything going on here or should it be closed?

jvz:
This PR is stalled for now. I'll be revisiting this later, and it can be resurrected then.

rgoers:
Closing per @jvz so I don't keep looking at it.

PR #275, opened 2019-06-11
Log4 j2 2621

rgoers:
Moves plugin support to its own module. Generates a Java file for recording plugins instead of Log4jPlugins.dat. Maintains support for Plugins generated in Log4j 2 2.x.


remkop:
I took a look. Many files changed but vast majority is import changes.

Files of interest:
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/PluginBuilder.java (118)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginAttributeVisitor.java (27)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginBuilderAttributeVisitor.java (26)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginConfigurationVisitor.java (10)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginElementVisitor.java (18)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginNodeVisitor.java (13)
M log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginValueVisitor.java (17)
M log4j-core/src/main/java/org/apache/logging/log4j/core/osgi/BundleContextSelector.java (77)
M log4j-core/src/main/java/org/apache/logging/log4j/core/selector/BasicContextSelector.java (11)
M log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ClassLoaderContextSelector.java (47)
M log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ContextSelector.java (31)
M log4j-core/src/main/java/org/apache/logging/log4j/core/selector/JndiContextSelector.java (36)

Overall I could not spot any issues but to be honest I did not look at the context selector stuff in much detail.

I could not find in the diff the place where the Java file for recording plugins is emitted.

rgoers:
See https://github.com/apache/logging-log4j2/blob/LOG4J2-2621/log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/processor/PluginProcessor.java method writeClassFile at line 161.

rgoers:
It just occurred to me that I haven't yet updated the documentation. I will have to do that before merging.

remkop:
OK thanks. I was looking at this with a view to in the future add another visitor that generates an additional file for the plugins (see https://issues.apache.org/jira/browse/LOG4J2-2604). The current code does not cater for additional visitors, but that is fine, we can cross that bridge when we get there.

rgoers:
The modifications necessary for LOG4J2-2604 don't conflict at all with these changes. It would do something similar to what writeServiceFile does, although writing the actual JSON would be more complicated.

rgoers:
OK. I thought about it. Services implemented with ServiceLoader are only loaded once. That would mean only a single instance of a Layout or Filter could be created. That obviously doesn't work. So in the end we would end up with something more or less like what is implemented here.

remkop:
About the shading issue, how do these changes improve things?

rgoers:
Presumably shading should provide support for ServiceLoader since it is standard java. We already use it to load Log4J core. The only difficulty will be for people who want to create plugins for use in OSGi as they will require an Activator to register their plugins as a standard OSGi service.

remkop:
(Taking another look at [PluginProcessor](https://github.com/apache/logging-log4j2/blob/LOG4J2-2621/log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/processor/PluginProcessor.java)...)

I see now: the generated `${pkg}.plugins.Log4jPlugins.java` is created in a different location for each package. So the source file for plugins in log4j itself, and the source file for custom plugins will be generated in different locations and therefore can be shaded (they won't overwrite each other).

log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ContextSelector.java
```diff
@@ -26,6 +27,36 @@
  */
 public interface ContextSelector {
 
+    long DEFAULT_STOP_TIMEOUT = 50;
```
jvz: You can use `java.time.Duration` at this point.

PR #276, opened 2019-06-12
Update SLF4J to 1.7.26

martin-g:
1.7.26 has been released at Feb 16th 2019.

### DIFF
--- a/log4j-slf4j-impl/pom.xml
+++ b/log4j-slf4j-impl/pom.xml
@@ -31,7 +31,7 @@
     <log4jParentDir>${basedir}/..</log4jParentDir>
     <docLabel>SLF4J Documentation</docLabel>
     <projectDir>/slf4j-impl</projectDir>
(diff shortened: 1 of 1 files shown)


rgoers:
This had to be reverted as SLF4J 1.7.26 breaks binary compatibility.

martin-g:
@rgoers Can you give me some link(s) to discussion about the issue with .26 ? I cannot find anything in SLF4J's JIRA or GitHub

rgoers:
I merged the PR and quickly remembered that we had tried to do it before and couldn't because the build failed. Ceki removed the EventData class from slf4j-ext. Log4j provides support for that class so the upgrade causes a compilation error. Since there is 1.7.26 provides nothing new Log4j needs to support there is really no reason for us to upgrade as doing so would require all Log4j users to have to upgrade their SLF4J version.

martin-g:
Thank you for the explanation!

PR #278, opened 2019-06-16
LOG4J2-2631: RoutingAppender PurgePolicy implementations don't remove …

carterkozak:
Note that this makes a behavior change to RoutingAppender.getAppenders
where Routes based on appender references are no longer included.
PurgePolicy implementations should be entirely unaware of the
existance of reference based routes because those appenders may
be used elsewhere in the configuration.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/routing/RoutingAppender.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/routing/RoutingAppender.java
@@ -28,7 +28,6 @@ import javax.script.Bindings;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.Core;
 import org.apache.logging.log4j.core.Filter;
-import org.apache.logging.log4j.core.LifeCycle;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.appender.AbstractAppender;
 import org.apache.logging.log4j.core.appender.rewrite.RewritePolicy;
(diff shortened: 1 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/routing/RoutingAppender.java
```diff
@@ -224,13 +224,16 @@ public void append(LogEvent event) {
             control.callAppender(event);
         }
 
-        if (purgePolicy != null) {
+        if (purgePolicy != null
+                // LOG4J2-2631: PurgePolicy implementations do not need to be aware of appenders that
+                // were not created by this RoutingAppender.
+                && !referencedAppenders.containsKey(key)) {
```
carterkozak: This check is a bit odd, I considered another approach where the concurrent map uses a value with additional metadata describing whether or not the appender is based on a reference. Unfortunately that approach would still require two concurrent maps in order to avoid concurrency issues, so I opted against it in favor of the containsKey check here.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/routing/RoutingAppender.java
```diff
@@ -279,8 +290,12 @@ private Appender createAppender(final Route route, final LogEvent event) {
...
     }
 
+    /**
+     * Returns an unmodifiable view of the appenders created by this {@link RoutingAppender}.
+     * Note that this map does not contain appenders that are routed by reference.
+     */
     public Map<String, AppenderControl> getAppenders() {
-        return Collections.unmodifiableMap(appenders);
+        return createdAppendersUnmodifiableView;
     }
```
carterkozak: Concerns with this behavior change?

PR #279, opened 2019-06-16
LOG4J2-2629: Avoid losing log events when the PurgePolicy races a log event

carterkozak:
Stacked on top of #278


rgoers:
You will also have to apply this to release-2.x. As far as I am concerned you don't need another PR - just commit it.

carterkozak:
Done, thanks for the quick review @rgoers!

PR #282, opened 2019-06-20
LOG4J2-2634: Break up large AsyncLogger methods for inlining

carterkozak:


carterkozak:
@rgoers could you provide guidance on whether I should add a changelog and/or jira ticket for this sort of thing? Thanks!

rgoers:
Yes, you should create a Jira and a changelog.

PR #283, opened 2019-06-21
LOG4J-2635: Fix java.version check in ThreadNameCachingStrategy

fgunbin:


fgunbin:
@garydgregory No problem, though I viewed duplication in test file as a kind of test that strings are right.
Add to org.apache.logging.log4j.core.util.Constants, right?

remkop:
@fgunbin if these are only locally used I would prefer not to define them in `Constants` but just define them locally.

carterkozak:
I think we can use the enum `name()` from `ThreadNameCachingStrategy` instead of defining new constants.

remkop:
Makes sense.

fgunbin:
Pushed the update, please check.  
There's Travis failure somewhere in Log4j Sample Configuration Service build, and I cannot see the reason, could you help please?

PR #284, opened 2019-06-22
LOG4J2-2639 - Allow logging calls to be constructed using a builder pattern

rgoers:
This PR adds methods to the Logger interface to construct and log an event using builder methods. It also allows the location information to either be provided or constructed at lower cost. It takes advantage of Java 8 default methods to maintain backward compatibility.


log4j-api/src/main/java/org/apache/logging/log4j/internal/LogBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.apache.logging.log4j.Marker;
+import org.apache.logging.log4j.message.Message;
+
+/**
+ * Class Description goes here.
+ */
+public interface LogBuilder {
+
+    public static final LogBuilder INSTANCE = new LogBuilder() {};
```
carterkozak: I think this would be easier to follow if this was named something along the lines of `NOP`, `LogBuilder.INSTANCE` sounds a lot like the DefaultLogBuilder.

Should these be default methods? While it's nice to avoid causing breaks when we add new methods, it's unfortunate that we will be potentially throwing data away.

log4j-api/src/main/java/org/apache/logging/log4j/internal/LogBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+    default LogBuilder withMessage(String msg) {
+        return this;
+    }
+
+    default LogBuilder withMessage(Object msg) {
+        return this;
+    }
+
+
+    default LogBuilder withParameters(Object... params) {
```
carterkozak: How does this interact with `withMessage(Message)`? Do we expect to overload this method with variable numbers of arguments similarly to existing logger.<level> methods?

Would it be possible to use a single method which adds a parameter?

```java
log.info()
  .message("{}, {}!)
  .parameter("Hello")
  .parameter("World")
  .log();
```

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
```diff
@@ -2734,4 +2748,108 @@ public void warn(final String message, final Object p0, final Object p1, final O
...
+
+    /**
+     * Constuct a trace log event.
+     * @return a LogBuilder.
+     * @since 3.0
+     */
+    @Override
+    public  LogBuilder atTrace() {
+        if (isTraceEnabled()) {
+            return logBuilder.get().setLevel(Level.TRACE);
```
carterkozak: I think we need something similar to `ReusableParameterizedMessage.reserved` for this to work, otherwise we will clobber threadlocal state with something like this:

```java
private void func() {
  log.atInfo().withMessage("Value {}").withParameters(getValue()).log();
}

private Object getValue() {
    log.atInfo().withMessage("Getting Value").log();
    return value;
}
```

log4j-api/src/main/java/org/apache/logging/log4j/internal/LogBuilder.java
rgoers: Yes, NOOP_LOGBUILDER would probably be better.

Actually, the chance of throwing data away is minimal. AbstractLogger implements the methods that virtually all Loggers use.

log4j-api/src/main/java/org/apache/logging/log4j/internal/LogBuilder.java
rgoers: The parameters are passed to the message factory along with the format string when constructing a message. If a Message is supplied the parameters would be ignored.

You can specify multiple parameters on the parameters method and you can invoke the parameters method as many times as you want. I need to add another parameters method that accepts a supplier.

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
rgoers: Ouch. I hadn't thought of that. That means the threadLocal will need to maintain a list of LogBuilders instead of just one.

rgoers:
Calling log twice on the same LogBuilder would result in 2 identical log messages, other than the timestamp. 

We could easily add the current thread id to DefaultLogBuilder to verify it is only used on a single thread but is it worth the overhead of checking that on every method call?

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
carterkozak: Could we keep a single builder with a `reserved` flag? If the current value is reserved, we can remove it from the threadlocal and create a new instance for the nested logger. I think that works well for ReusableParameterizedMessage, and limits the per-thread memory we can retain. It think nested log statements are uncommon enough that we don't need to make them allocation free, though they may be more common using a builder API than the standard style APIs.

carterkozak:
> Calling log twice on the same LogBuilder would result in 2 identical log messages, other than the timestamp

I think parameter values are mutated in gc free mode after the first event is logged, I would expect the second invocation (assuming it is executed after a few microseconds) would have all null parameters.

> We could easily add the current thread id to DefaultLogBuilder to verify it is only used on a single thread but is it worth the overhead of checking that on every method call?

Maybe only the `log()` invocation?

rgoers:
@carterkozak I have updated the code based on your comments above.

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
```diff
@@ -2755,77 +2755,114 @@ public void warn(final String message, final Object p0, final Object p1, final O
      * @since 3.0
      */
     @Override
-    public  LogBuilder atTrace() {
+    public  LogBuilder trace() {
         if (isTraceEnabled()) {
-            return logBuilder.get().setLevel(Level.TRACE);
+            DefaultLogBuilder builder = logBuilder.get();
+            if (builder.isInUse()) {
+                return new DefaultLogBuilder(this);
```
remkop: Wouldn't we need to call `setLevel(Level.TRACE)` on `new DefaultLogBuilder(this)` before returning it?

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
```diff
@@ -2755,77 +2755,114 @@ public void warn(final String message, final Object p0, final Object p1, final O
...
+    public LogBuilder debug() {
         if (isDebugEnabled()) {
+            DefaultLogBuilder builder = logBuilder.get();
+            if (builder.isInUse()) {
+                return new DefaultLogBuilder(this);
+            }
             return logBuilder.get().setLevel(Level.DEBUG);
         } else {
-            return LogBuilder.INSTANCE;
+            return LogBuilder.NOOP;
```
remkop: The above 7 lines are repeated for each log level. Would it be possible to extract this to a helper method?

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
rgoers: Yes, unless a second constructor is added that accepts the level.

log4j-api/src/main/java/org/apache/logging/log4j/spi/AbstractLogger.java
rgoers: Isn’t that just replacing one line with another?

rgoers: There is no need for a new method. I have modified them to all call atLevel().

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+        this.level = level;
+        this.marker = null;
+        this.throwable = null;
+        this.location = null;
+        this.object = null;
+        this.msg = null;
+        this.textMessage = null;
+        this.supplier = null;
+        this.parameters = null;
+        return this;
```
remkop: I believe the `setLevel` method should be moved to `Logger` (and renamed to `atLevel`), and removed from the `LogBuilder` class.

That will allow all of the remaining methods to skip processing if the log level is not applicable. That is one of the value propositions of [fluent logging APIs like this](https://google.github.io/flogger/benefits).

We lose that benefit if we allow users to change the log level later in the call chain: for example, this usage is possible:

```java
// assume log level is WARN
logger.trace().withMessage("This {} should {} be {} ignored {}")
    .withParameters("statement", "immediately", "completely", "for better performance")
    .setLevel(Level.TRACE) // oops, we cannot ignore the above after all
    .log();
```

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
```diff
@@ -0,0 +1,179 @@
...
+         }
+        return this;
+    }
+
+    public boolean isInUse() {
+        return inUse;
+    }
+
+    public void log() {
+        if (!inUse) {
```
remkop: We can learn from [flogger](https://google.github.io/flogger/anatomy.html)'s approach.

Can we avoid the call to `log()` at the end? All that this does is signal that the log statement is done, like a call to `build()` at the end of a builder chain.

How about if we make the `log` method do more work? Something like this:

```java
logger.atTrace().log("Hi, {}", () -> fetchWorld());
logger.atTrace().withMarker(marker).withThrowable(t).log("Found error while {}", currentTask);
```

I propose we remove the `withMessage` and `withParameters` methods and replace them with overloaded `log` methods that take these parameters:

* Message
* Object
* Supplier<Message>
* String, Object...
* String, Supplier<Object>...
* (potentially other...)


Another benefit of this approach is that it would prevent users from doing the wrong thing, like for example:

```java
// the API should not allow this:
Message msg = msg();
logger.atTrace().withMessage(msg)
    .withMessage("It is now possible to use the API incorrectly")
    .withMessage(() -> differentMsg())
    .withParameters("a", "b", "c")
    .withParameters(() -> getParams())
    .withMessage("hi!")
    .log(); // now, which of the above should actually be logged?
```

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
rgoers: setLevel is dangerous and is not intended to be called by end users as it provides no value to them. As such it is not part of the LogBuilder interface and only exists in DefaultLogBuilder. As your example above shows, it would be nonsensical to create the LogBuilder by calling the trace method and then call setLevel() to set the level to something else. The main purpose of setLevel is to clear the LogBuilder for reuse.

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
rgoers: I should note that at Gary's request I removed "at" from all the logger methods. so the first would be
`logger.trace().log("Hi, {}", () -> fetchWordl());`

After reading the above a few times I think I am OK with this.

remkop: About mixing Suppliers with other objects, this is still possible if (a) we cater for it in our implementation and (b) client code is willing to cast lambdas to `org.apache.logging.log4j.util.Supplier`. For example, this works fine in Java 8:

```java
static void log(String pattern, Object... something) {
    System.out.println(pattern);
    for (Object element : something) {
        if (element instanceof Supplier) {
            System.out.println(((Supplier) element).get());
        } else {
            System.out.println("Plain string: " + element);
        }
    }
}
    log("Hello {} {}", 123, (Supplier)() -> "I'm a little lambda");
}
```

remkop: About the logger method naming: I'm generally in favor of short names and I liked Gary's suggestion. However, looking at it some more, in this case there may be something to be said for prefixing the new "fluent" methods with `at`;

* it makes the fluent methods stick out from the "normal" `info(String)`, `trace(Message)`, etc Logger  methods. I think it is nice if they look different, because they are quite different.
* users coding in an IDE will have the advantage of typing "at<Ctl+space>" and get completions for the fluent methods.
* it is nice if the `atLevel(Level)` method is consistently named with the `atInfo`, `atDebug`, etc other fluent methods.

@garydgregory (and others), what do you think?

carterkozak:
I'm worried about memory leaks in calls that never invoke `log()`

```java
logger.info().message("Hello, {}!", "World").marker(getMarker() /* throws! */).log();
```

In this case, we would apply data to a message factory, building a message object, but never release threadlocal parameter references.

We could do something along the lines of `info(Consumer<LogBuilder>)` and remove `build()` from the LogBuilder, however the fluent usage would never be garbage free. We would get a guarantee that we could release resources on both success and failure, but the API would be more difficult to use.

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
```diff
@@ -0,0 +1,187 @@
...
+    private Object[] parameters;
+    private volatile boolean inUse;
+    private long threadId;
+
+    public DefaultLogBuilder(Logger logger, Level level) {
+        this.logger = logger;
+        this.level = level;
+        this.inUse = false;
+        this.threadId = Thread.currentThread().getId();
+        this.inUse = true;
```
carterkozak: `inUse` is set twice in this constructor

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
```diff
@@ -0,0 +1,187 @@
...
+    private final Logger logger;
+    private Level level;
+    private Marker marker;
+    private Throwable throwable;
+    private StackTraceElement location;
+    private Object object;
+    private Message msg;
+    private String textMessage;
+    private Supplier<Message> supplier;
+    private Object[] parameters;
```
carterkozak: The current implementation retains references parameter values between fluent logger uses, this can cause objects that are expected to be short lived to make their way into tenured space.

carterkozak:
This doesn't support filters which enable higher verbosity logging based on the presence of a marker. That sort of filtering will be difficult using this API, but it may be possible if we avoid the no-op logger based on the presence of filters. I'm not sure what that will do to performance though.

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
rgoers: OK. I will fix that.

rgoers:
There should be no risk of a memory leak if log() is not called. The LogBuilders are owned by a thread local. If log() is not called then the next call to an "at" method would clear them for re-use. Even if one was dynamically created it would be garbage collected as soon as the it went out of scope.

log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
rgoers: These have all been removed with the changes Remko requested. I don't believe I have committed them yet.

carterkozak:
> There should be no risk of a memory leak if log() is not called. The LogBuilders are owned by a thread local. If log() is not called then the next call to an "at" method would clear them for re-use. Even if one was dynamically created it would be garbage collected as soon as the it went out of scope.

Some applications record errors infrequently, I don't think it's sufficient to hold the previous parameters until the next time a logger is used with an enabled level. I've fixed a couple permutations of this issue in our existing message/event implementations based on third party bug reports.

I don't think it's a problem if we hold a reference to an object array of null values, but if we hold references to dynamically provided parameter objects I think it can cause issues.

rgoers:
As I said, the message and parameter are no longer stored in the builder, so I really don't think there is a problem.

remkop:
I will take another detailed look, but this is shaping up very nicely now I think.

log4j-api/src/main/java/org/apache/logging/log4j/LogBuilder.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    default void log(String message, Object... params) {
+    }
+
+    default void log(String message, Supplier<?>... params) {
+    }
+
+    default void log(Message message) {
+    }
```
carterkozak: I really like this approach, thanks @rgoers!

PR #287, opened 2019-06-27
[LOG4J2-2688] adding support for session attributes in log4j-web

rmannibucau:
Follow up of https://issues.apache.org/jira/browse/LOG4J2-2523 comments, can need a new ticket though.

### DIFF
--- a/log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
+++ b/log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
@@ -19,6 +19,7 @@ package org.apache.logging.log4j.web;
 // this comment if no longer relevant
 
 import java.security.Principal;
+import java.util.Objects;
 import java.util.stream.Stream;
 
 import javax.servlet.ServletContext;
@@ -98,8 +99,7 @@ public class WebLookup extends AbstractLookup {
                 final HttpSession session = HttpServletRequest.class.cast(req).getSession(false);
                 if (session != null) {
                     final String name = key.substring(SESSION_ATTR_PREFIX.length());
-                    final Object attribute = session.getAttribute(name);
-                    return attribute == null ? null : String.valueOf(attribute);
+                    return Objects.toString(session.getAttribute(name), null);
                 }
             }
             return null;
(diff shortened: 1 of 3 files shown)


log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
```diff
@@ -91,6 +92,19 @@ public String lookup(final LogEvent event, final String key) {
...
         }
 
+        if (key.startsWith(SESSION_ATTR_PREFIX)) {
+            final ServletRequest req = getRequest();
+            if (HttpServletRequest.class.isInstance(req)) {
+                final HttpSession session = HttpServletRequest.class.cast(req).getSession(false);
+                if (session != null) {
+                    final String name = key.substring(SESSION_ATTR_PREFIX.length());
+                    final Object attribute = session.getAttribute(name);
+                    return attribute == null ? null : String.valueOf(attribute);
```
garydgregory: Hide the ternary: `return Objects.toString(session.getAttribute(name), null);`

rmannibucau:
@rgoers good catch, added.

log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
```diff
@@ -91,6 +93,18 @@ public String lookup(final LogEvent event, final String key) {
                     HttpServletRequest.class.cast(req).getParameter(name) : null;
         }
 
+        if (key.startsWith(SESSION_ATTR_PREFIX)) {
+            final ServletRequest req = getRequest();
+            if (HttpServletRequest.class.isInstance(req)) {
```
jvz: Why the reflection? You could change this to be:

```
if (req instanceof HttpServletRequest) {
  final HttpSession session = ((HttpServletRequest) req).getSession(false);
  // ...
}
```

rmannibucau:
Sure, created LOG4J2-2688

log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
rmannibucau: Guess it is mainly a coding habit and the preference over a debuggable method (java.lang.Class#cast). Not strong objection if you feel it must be change

PR #289, opened 2019-06-30
LOG4J2-2644 - Improve performance of getting location info

rgoers:
Improves the performance of getting location info.


log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -72,15 +78,40 @@ private StackLocator() {
     }
 
     public StackTraceElement calcLocation(final String fqcnOfLogger) {
-        return stackWalker.walk(
-                s -> s.dropWhile(f -> !f.getClassName().equals(fqcnOfLogger)) // drop the top frames until we reach the logger
-                        .dropWhile(f -> f.getClassName().equals(fqcnOfLogger)) // drop the logger frames
-                        .findFirst())
-                .get()
-                .toStackTraceElement();
+        return walker.walk(s -> s.filter(LOCATOR.get().setFqcn(fqcnOfLogger)).findFirst()).get().toStackTraceElement();
```
carterkozak: We may want `CallerLocator` to implement `Function<Stream<StackWalker.StackFrame>, Optional<StackWalker.StackFrame>>` to avoid the lambda allocation for `s -> s.filter(LOCATOR.get().setFqcn(fqcnOfLogger)).findFirst()`
```suggestion
        return walker.walk(LOCATOR.get().setFqcn(fqcnOfLogger)).get().toStackTraceElement();
```

using something like this:

```java
static final class CallerLocator implements Predicate<StackWalker.StackFrame>,
        Function<Stream<StackWalker.StackFrame>, Optional<StackWalker.StackFrame>> {

    private String fqcn;

    private boolean foundFqcn = false;
    private boolean foundLogger = false;

    CallerLocator setFqcn(String fqcn) {
        this.fqcn = fqcn;
        this.foundFqcn = false;
        this.foundLogger = false;
        return this;
    }

    @Override
    public boolean test(StackWalker.StackFrame t) {
        final String className = t.getClassName();
        if (!foundLogger) {
            if (!foundFqcn) {
                foundFqcn = className.equals(fqcn);
                return false;
            } else if (!className.equals(fqcn)) {
                foundLogger = true;
                return true;
            }
        }
        return false;
    }

    @Override
    public Optional<StackWalker.StackFrame> apply(Stream<StackWalker.StackFrame> stackFrameStream) {
        return stackFrameStream.filter(this).findFirst();
    }
}
```

I'm not sure how much that helps if at all, I haven't looked into how much allocation StackWalker does internally.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -31,6 +32,11 @@
 
     private final static StackLocator INSTANCE = new StackLocator();
 
+    private final static ThreadLocal<CallerLocator> LOCATOR = new ThreadLocal<CallerLocator>() {
+        public CallerLocator initialValue() {
+            return new CallerLocator();
+        }
+    };
```
carterkozak: Since this is java 9, we have the option to take advantage of the java 8`ThreadLocal.withInitial(CallerLocator::new)`

carterkozak:
Any chance you have benchmarks which show the performance before and after this change?

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -72,15 +78,40 @@ private StackLocator() {
...
+                if (!foundFqcn) {
+                    foundFqcn = className.equals(fqcn);
+                    return false;
+                } else if (!className.equals(fqcn)) {
+                    foundLogger = true;
+                    return true;
+                }
+            }
+            return false;
+        }
```
carterkozak: It took me a few reads through to understand how this works, would you mind adding a comment to make it easier to understand?

Maybe something along these lines (if I understand correctly)
```java
/**
 * States:
 * 1. foundFqcn=false, foundLogger=false Nested in logging code, we must first
 *     work our way up to frames from the loggers fully qualified class name.
 * 2. foundFqcn=true, foundLogger=false Once we've found the logger, the next
 *     class we encounter which does not match the loggers fully qualified class
 *     name is the caller.
 * 3. foundFqcn=true, foundLogger=true Terminal state, if execution continues after
 *     the logger is found, the caller is doing something incorrectly.
 */
public boolean test(StackWalker.StackFrame t) {
    final String className = t.getClassName();
    if (!foundLogger) {
        if (!foundFqcn) {
            foundFqcn = className.equals(fqcn);
            return false;
        } else if (!className.equals(fqcn)) {
            foundLogger = true;
            return true;
        }
    }
    // Unexpected usage, CallerLocator.test has been invoked after the logger has been found.
    return false;
}
```

rgoers:
For the benchmarks see the related Jira issue.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -31,6 +33,7 @@
 
     private final static StackLocator INSTANCE = new StackLocator();
 
+    private final static ThreadLocal<FqcnCallerLocator> LOCATOR = ThreadLocal.withInitial(FqcnCallerLocator::new);
```
carterkozak: Should we check if threadlocals are enabled before using this since it's not a standard type like StringBuilder?

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -72,15 +75,38 @@ private StackLocator() {
...
+
+        public FqcnCallerLocator setFqcn(String fqcn) {
+            this.fqcn = fqcn;
+            return this;
+        }
+
+        @Override
+        public StackWalker.StackFrame apply(Stream<StackWalker.StackFrame> stackFrameStream) {
+            boolean foundFqcn = false;
+            Object[] frames = stackFrameStream.toArray();
```
carterkozak: Is this going to be more expensive for logging that occurs with a large stack?

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
rgoers: What do you mean - "check if threadlocals are enabled"? A ThreadLocal is a standard Java class...

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
rgoers: I don't know but I suspect not. The test I am running has a stack size of 35 entries. My understanding is the StackFrames are optimized for when you access less than 8., but it might just be grabbing them in batches of 8. I haven't checked that. I do know that the stream logic is adding considerable overhead. I suspect this would be even faster if there were methods that could do this inside of StackWalker without having to create the Stream.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
carterkozak: See `org.apache.logging.log4j.core.util.Constants.ENABLE_THREADLOCALS` and `org.apache.logging.log4j.util.Constants.ENABLE_THREADLOCALS`.

I think we disable most threadlocals in web-apps to avoid N tomcat webapps leaving thread-local state on each webserver thread. If memory serves, a ThreadLocal of a type from the standard library can be shared between all threads, even if they have loaded different log4j jars, but I'm having trouble finding the comment where I read that.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
carterkozak: I see, I didn't realize it was chunked internally, that's great! It's unfortunate that streams don't JIT very well compared to standard loops/iterators, the array sounds like the way to go if it performs better than the fluent stream approach.

rgoers: Ultimately I will have to create a separate project at GitHub with JMH benchmarks to show how the variations behave and then open tickets against the JDK to see what can be done if it can be shown that using Streams isn't the most performant way to access stack traces.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
remkop: Hi Carter, yes. Not sure which comment you were looking for, but in general ThreadLocals that hold a reference to an application class (e.g. a Log4j class) may cause memory leaks in application servers that keep their thread pools up when a web app is stopped and started. If the ThreadLocal only contains classes that are in the JDK library, such memory leaks will not occur.

log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
rgoers: It seems we already have JMH benchmarks. I ran them. The results are in the Jira issue.

PR #290, opened 2019-07-04
[LOG4J2-2650] Fix ArrayIndexOutOfBoundsException when the mac variable is an ipv6

mattiabertorello:


rgoers:
I don't understand this fix. NetUtils.getMacAddress is supposed to return a MAC address or emulate one. If it is returning a MAC address with the wrong number of bytes that is what needs to be fixed, not UuidUtil.

mattiabertorello:
Yes you are right, I was not sure what part of the code fix. 
Because I didn't think that the `NetUtils.getMacAddress` method should return also an emulate MAC. In fact this line `mac = localHost.getAddress();`  was pretty obscure to me why was there.
In any case I replace it with these lines 

```

// Emulate a mac address with an IP v4 or v6
final byte[] address = localHost.getAddress();
// Take only 6 bytes if the address is an IPv6 otherwise will pad with two zero bytes
mac = Arrays.copyOf(address, 6);
```

garydgregory:
Where are we on this?

rgoers:
Thanks for the nudge. I meant to re-review this and life got in the way.

PR #291, opened 2019-07-10
[LOG4J2-2604/2649] Change MethodHandle to Constructor to adapt graalvm

Mhhli:
See  https://issues.apache.org/jira/browse/LOG4J2-2649
        https://issues.apache.org/jira/browse/LOG4J2-2604
Log4j2 core does not support static compilation because Graalvm does not support MethodHandle method. 
I change the useing in log4j2 to constructor.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ContextDataFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ContextDataFactory.java
@@ -16,9 +16,7 @@
  */
 package org.apache.logging.log4j.core.impl;
 
-import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
-import java.lang.invoke.MethodType;
+import java.lang.reflect.Constructor;
 import java.util.Map;
 import java.util.Map.Entry;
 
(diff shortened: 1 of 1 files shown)


Mhhli:
The pom.xml of log4j-api-java9 hased change jdk9 to jdk11, but the .travis.yml has not changed for jdk11

remkop:
I don't disagree with these changes, but please do 2 things:
* respect the spacing of the original file: space between `if` and `(`, space between `try` and `{`, etc.
* Please add a comment to `ContextDataFactory.DEFAULT_CONSTRUCTOR` as to _why_ this should be a `Constructor` and not a `MethodHandle`. The comment may contain JIRA or GitHub issue numbers or links.

Mhhli:
> I don't disagree with these changes, but please do 2 things:
> 
> * respect the spacing of the original file: space between `if` and `(`, space between `try` and `{`, etc.
> * Please add a comment to `ContextDataFactory.DEFAULT_CONSTRUCTOR` as to _why_ this should be a `Constructor` and not a `MethodHandle`. The comment may contain JIRA or GitHub issue numbers or links.

I'm sorry about the space. Now I have changed them.

In [LOG4J2-2649](https://issues.apache.org/jira/browse/LOG4J2-2649), the reporter said some reason about using graalvm to static compile.  In graalvm substratevm [LIMITATIONS](https://github.com/oracle/graal/blob/master/substratevm/LIMITATIONS.md), graalvm is not support `MethodHandle` now, so I think the MethodHandle may need to return `Constructor`.

remkop:
Thanks for changing the whitespace. About the reason for the changes, I understand the background. 

What I meant was: please add that comment with the reason for using `Constructor` to the source code (maybe as a line comment on `ContextDataFactory.DEFAULT_CONSTRUCTOR`) so that future maintainers will not change it back to `MethodHandle`.

Thanks for your quick changes.

Mhhli:
> Thanks for changing the whitespace. About the reason for the changes, I understand the background.
> 
> What I meant was: please add that comment with the reason for using `Constructor` to the source code (maybe as a line comment on `ContextDataFactory.DEFAULT_CONSTRUCTOR`) so that future maintainers will not change it back to `MethodHandle`.
> 
> Thanks for your quick changes.

Thanks. I add the comment on `ContextDataFactory.DEFAULT_CONSTRUCTOR`.

jvz:
Ugh, and the API for MethodHandle is so much nicer than the reflection API for these sorts of things. Does the JVM optimize calls to `Method.invoke()` to work the same as `MethodHandle.invoke()`? If that's the case, then we could avoid using that API. Otherwise, I was looking at using `MethodHandle` far more in an update to the plugin system for 3.0, and if there's good reason to avoid that, I'd like to know now rather than later.

PR #292, opened 2019-07-18
LOG4J2-2656: Add support of ignoreKafkaConnectionError flag

akrikheli:
See https://issues.apache.org/jira/browse/LOG4J2-2656

### DIFF
--- a/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaAppender.java
+++ b/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaAppender.java
@@ -59,6 +59,9 @@ public final class KafkaAppender extends AbstractAppender {
         @PluginAttribute(value = "syncSend", defaultBoolean = true)
         private boolean syncSend;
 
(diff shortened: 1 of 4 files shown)


rgoers:
I am continuing to think about this as I am not sure if this is the right way to fix the problem.

carterkozak:
It feels a bit odd for the Kafka appender to work differently from other appenders when creation fails. I wonder if it would be better if we provided an appender wrapper (similar to failover appender) which can create appenders lazily and log the failure without stopping the application? Such an appender could apply to any existing appender, and wouldn't require special boolean flags in each implementation.

rgoers:
@carterkozak The AppenderControl already wraps Appenders to catch the exceptions they throw. There is already a generic way to cause it to stop suppressing the exceptions that are thrown while it is executing. But off the top of my head I don't recall if there is a standard way to handle exceptions while the Appender is created. Most of the ones I have seen just the Appender to be null, but other stuff still may succeed.

carterkozak:
A workaround might be to use a RoutingAppender with a single default route of the KafkaAppender defined in place. That should lazily attempt to create the delgate KafkaAppender, logging failures and continuing.

PR #298, opened 2019-07-22
Synchronize before looping over synchronized collection

emopers:
In StatusConfiguration.java, `errorMessages = Collections.synchronizedCollection(new LinkedList<String>())` is iterated over without synchronization. According to [Java API](https://docs.oracle.com/javase/6/docs/api/java/util/Collections.html#synchronizedCollection%28java.util.Collection%29), it is recommended that user manually synchronizes each iteration over the synchronized collection to reduce non-determinism.
This pull request improved the condition by synchronizing a loop which loops over the synchronized collection.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/status/StatusConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/status/StatusConfiguration.java
@@ -217,9 +217,12 @@ public class StatusConfiguration {
     }
 
     private void migrateSavedLogMessages() {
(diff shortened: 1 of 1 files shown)


rgoers:
Have you actually encountered a problem? If so, please file a Jira issue with the stack trace. In reviewing the code all that is happening is that any status messages that were accumulated during the time the Configuration object was created up to the time the StatusLogger is available are being logged. When this method gets control nothing should be generating new status messages so there shouldn't be a problem.

rgoers:
An alternative change was made

PR #299, opened 2019-07-31
[LOG4j2-1946] Fix problem with purgeAscending if an old file was deleted

IgorPerelyotov:
I ran into the same problem as described in the ticket https://issues.apache.org/jira/browse/LOG4J2-1946
There are an unit test reproducing the problem and the fix.


rgoers:
I tested these locally on release-2.x and verified the problem and the fix.

PR #300, opened 2019-08-10
call #StatusLogger.getLogger() only when it's needed.

qxo:
when Log4jWebSupport.IS_LOG4J_AUTO_INITIALIZATION_DISABLED=true should not invoke #StatusLogger.getLogger() in Log4jServletContainerInitializer
Otherwise there will be initialization issue: custom config issue

### DIFF
--- a/log4j-web/src/main/java/org/apache/logging/log4j/web/Log4jServletContainerInitializer.java
+++ b/log4j-web/src/main/java/org/apache/logging/log4j/web/Log4jServletContainerInitializer.java
@@ -34,14 +34,14 @@ import org.apache.logging.log4j.status.StatusLogger;
  */
 public class Log4jServletContainerInitializer implements ServletContainerInitializer {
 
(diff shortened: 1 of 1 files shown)


qxo:
> This change looks fine to me. Are there any specific startup scenarios you came across that warrants this?
> 
> Edit: I see what you mean. Can you also file a Jira ticket for this? https://issues.apache.org/jira/projects/LOG4J2/issues
> 
> Also, if you rebase from master, you should get a working Travis config.

1. rebase from master was done:)
2. the startup scenarios was :
   we use custom ServletContextListener for init the app, include the logging system
   so we need ` Log4jWebSupport.IS_LOG4J_AUTO_INITIALIZATION_DISABLED=true `
  but the `static final LOGGER=StatusLogger.getLogger() ` on Log4jServletContainerInitializer will break custom init for log4j2

3. the apache Jira is so slow from China... sorry for this

PR #301, opened 2019-08-19
LOG4J-2676: Backport use of javax.annotation.processing.Messager to release-2.x.

greggdonovan:
Add a test for Bazel support for the PluginProcessor, which does not work on 2.12.1 due to System.out calls, but works with this PR.


rgoers:
I understand the desire to get rid of System.out calls but why is bazel being added?

rgoers:
I committed the change without the bazel stuff.

PR #306, opened 2019-08-29
[LOG4J2-2678] Add LogEvent timestamp to Kafka ProducerRecord

fedexist:
### What I've added:

* `timestamp` field for KafkaAppender configuration

### What I've changed

* `KafkaManager#send` signature from

`public void send(final byte[] msg)`

to

`public void send(final byte[] msg, final Long eventTimestamp) `


garydgregory:
Should this really be optional? Why not just send the timestamp all the time? That would be simpler IMO.

fedexist:
> 
> 
> Should this really be optional? Why not just send the timestamp all the time? That would be simpler IMO.

I agree, that would be easier. I thought this could be optional in order to preserve the current behaviour, but in reality, if you use Kafka (like me), you will always want to have an event time timestamp embedded with your record.

garydgregory:
Unless this will break existing apps, I say we make it optional.

rgoers:
@garydgregory How could making it optional possibly break existing apps? Making it optional would be done to preserve the existing behavior and not break existing apps and is the only reason one would want to make it optional.  
Personally, I think it should be optional but the default should be to include it.

garydgregory:
I was asking the OP how it could break, since that person might have domain-specific knowledge I am not aware of. The reason for not making this an options is that it is simpler. IOW, when would I _not_ want the timestamp?

fedexist:
I honestly wouldn't know whether there's any _actual_ use case where the producer timestamp (i.e. the current behaviour) is preferred to the event/record timestamp. Usually, if the record timestamp is not reliable, the timestamp is embedded within the message (i.e. a field in a JSON object) for it to be extracted by the processing framework being used.
As I said, in my (relatively limited - 2 years, I admit) experience, the focus while using Kafka is the record timestamp and, moreover, the fact that, with log4j, we're in a true logging scenario makes it more plausible that someone using the KafkaAppender is actually interested to the LogEvent timestamp. 

Still, making this an option is playing it safe, I guess.

jvz:
If you rebase from master, you should get a working Travis config to test this.

As for the use case, making the event timestamp match the log event timestamp makes perfect sense. Using a sent time as it currently defaults to is basically a behavioral bug as it makes the messages' own timestamps differ from the log event.

jvz:
Seems like you broke the history somehow. Can you try using `git pull --rebase origin/master`?

fedexist:
> 
> 
> Seems like you broke the history somehow. Can you try using `git pull --rebase origin/master`?
> 
> Note that CI passed, so the updated config works fine.

rebase is the bane of my existence :)

log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaAppender.java
```diff
@@ -83,7 +83,7 @@ public boolean isSyncSend() {
             return syncSend;
         }
 
-        public boolean isSendTimestamp() { return sendTimestamp; }
+        public boolean isSendEventTimestamp() { return sendEventTimestamp; }
```
garydgregory: This needs to be split into >1 line to match the coding style.

fedexist: Fixed.

rgoers:
I applied this to master and when I ran the test I get
```
[INFO] Running org.apache.logging.log4j.kafka.appender.KafkaAppenderTest
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.442 s <<< FAILURE! - in org.apache.logging.log4j.kafka.appender.KafkaAppenderTest
[ERROR] testAppenderNoEventTimestamp(org.apache.logging.log4j.kafka.appender.KafkaAppenderTest)  Time elapsed: 0.025 s  <<< FAILURE!
java.lang.AssertionError: Appender named KafkaAppenderNoEventTimestamp was null.
	at org.apache.logging.log4j.kafka.appender.KafkaAppenderTest.testAppenderNoEventTimestamp(KafkaAppenderTest.java:155)

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   KafkaAppenderTest.testAppenderNoEventTimestamp:155 Appender named KafkaAppenderNoEventTimestamp was null.
[INFO] 
[ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0

```
Also, if you would like this as part of a Log4j 2.x release please provide an equivalent patch to the release-2.x branch.

fedexist:
> I applied this to master and when I ran the test I get
> 
> ```
> [INFO] Running org.apache.logging.log4j.kafka.appender.KafkaAppenderTest
> [ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.442 s <<< FAILURE! - in org.apache.logging.log4j.kafka.appender.KafkaAppenderTest
> [ERROR] testAppenderNoEventTimestamp(org.apache.logging.log4j.kafka.appender.KafkaAppenderTest)  Time elapsed: 0.025 s  <<< FAILURE!
> java.lang.AssertionError: Appender named KafkaAppenderNoEventTimestamp was null.
> 	at org.apache.logging.log4j.kafka.appender.KafkaAppenderTest.testAppenderNoEventTimestamp(KafkaAppenderTest.java:155)
> 
> [INFO] 
> [INFO] Results:
> [INFO] 
> [ERROR] Failures: 
> [ERROR]   KafkaAppenderTest.testAppenderNoEventTimestamp:155 Appender named KafkaAppenderNoEventTimestamp was null.
> [INFO] 
> [ERROR] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0
> ```
> 
> Also, if you would like this as part of a Log4j 2.x release please provide an equivalent patch to the release-2.x branch.

Hi @rgoers, I had this issue too, but then I found out that this error is caused by the old `KafkaAppenderTest.xml` still getting loaded, instead of the updated one (which is containing the new appender config). This file is within the test resources of the `log4j-core` module. If you run `mvn install` for log4j-core, then it will install the updated `log4j-core` module and tests for log4j-kafka should pass.

I guess it would be reasonable to move `KafkaAppenderTest.xml` and `KafkaAppenderCloseTimeoutTest.xml` to the `log4j-kafka` test resources, anyway.

PR #307, opened 2019-08-30
[LOG4J2-2652] JSON output wrong when using additonal fields

IgorPerelyotov:
Fix and tests for the issue https://issues.apache.org/jira/projects/LOG4J2/issues/LOG4J2-2652 

The cause of the issue is: 
1. Jackson is configured with custom serializer for Message added through setSerializer https://github.com/apache/logging-log4j2/blob/33009e1163227a20b526ac04b9af79b90d439bf5/log4j-layout-jackson/src/main/java/org/apache/logging/log4j/jackson/SimpleModuleInitializer.java#L24 
2. Annotations for LogEvent is mixed through mix-in mechanism 
3. MutableLogEvent implements both interfaces LogEvent and Message.

When Jackson looking for an appropriate serializer for MutableLogEvent, it selects MessageSerializer and not UnwrappingBeanSerializer because MutableLogEvent implements Message.

The bug report notes that the problem is not reproduced in 2,10 version. There was converting of MutableLogEvent to Log4jEvent and Log4jEvent implements single LogEvent interface,
https://github.com/apache/logging-log4j2/blob/6275d8aac4e392eacd81d942a1b97856c92b698f/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractJacksonLayout.java#L306

As the fix I added JsonSerialize annotation with parameter "as" on the getter method in order to tell Jackson that the field must be serialized as LogEvent. Jackson will not consider MessageSerializer if actual type of this LogEvent is MutableLogEvent.

### DIFF
--- a/log4j-layout-jackson-json/src/main/java/org/apache/logging/log4j/jackson/json/layout/JsonLayout.java
+++ b/log4j-layout-jackson-json/src/main/java/org/apache/logging/log4j/jackson/json/layout/JsonLayout.java
@@ -19,6 +19,7 @@ package org.apache.logging.log4j.jackson.json.layout;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonRootName;
 import com.fasterxml.jackson.annotation.JsonUnwrapped;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import org.apache.logging.log4j.core.Layout;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.Configuration;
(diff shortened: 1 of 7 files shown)


jvz:
If you rebase from master, you should get a working Travis config to test this.

Pebin:
any progress on this? I could eventually help with testing if needed :)

drx777:
I would also appreciate if this is fixed soon, it's completely breaking our logging with kibana. I'd be happy to help if I can.

rgoers:
This fix PR has been applied to master but the code in master is significantly different than on the release-2.x branch so it cannot be applied to the 2.x release series.

cyrixsimon:
Are there any plans to make a release 2.13.1 in the next time? I need this fix really soon in an official release, because the bug is breaking the logs.

rgoers:
My hope is to be able to do a release this weekend, but I still have a few more issues that need to be resolved.

PR #309, opened 2019-09-14
broken link on https://logging.apache.org/log4j/2.x/manual/thread-con…

s50600822:
https://logging.apache.org/log4j/2.x/manual/thread-context.html's ```Including the ThreadContext when writing logs``` has a broken link on PatternLayout.
Screen Shot 2019-09-15 at 12 40 02 am
The class isn't in API, linking to the layout page seems most reasonable.


rgoers:
Thanks for letting us know. I corrected the web site, master and release-2.x branches. I didn't commit your patch since you changed it to an absolute link.

PR #311, opened 2019-10-06
LOG4J2-2701: update jackson 2.9.9->2.9.10, minor jackson usage cleanup

cowtowncoder:
Update Jackson patch version (as per Jira issue): also minor simplification for Jackson usage.
I had a look at other usage and everything makes sense; the only thing I was wondering was whether some transient `ObjectMapper` instances could use statically created ones, in case methods are called more than once.
 But not being familiar enough with codebase decided not to try to change those yet.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/json/JsonConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/json/JsonConfiguration.java
@@ -16,8 +16,6 @@
  */
 package org.apache.logging.log4j.core.config.json;
 
-import java.io.ByteArrayInputStream;
-import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


jvz:
There's an open discussion somewhere around wrapping `ObjectMapper` into a plugin of some sort so it can be configured or extended by the config.

cowtowncoder:
That would probably make sense and isolate details better.

jvz:
Found it: https://issues.apache.org/jira/browse/LOG4J2-2670

rgoers:
If there is a 2.10 version then master should upgrade to that. release-2.x should upgrade to 2.9.10.

rgoers:
I updated master to 2.10.1 and release-2.x to 2.9.10.

cowtowncoder:
Makes sense.

PR #316, opened 2019-11-27
LOG4J2-2725 - Added try/finally around event.execute() for RingBufferLogEventHandler

anikosss:
Added try/finally around event.execute() for RingBufferLogEventHandler to clear memory correctly in case of exception/error


carterkozak:
Thanks for your contribution @anikosss!

PR #317, opened 2019-11-27
LOG4J2-2725 - Added try/finally around event.execute() for RingBufferLogEventHandler 

anikosss:
Added try/finally around event.execute() for RingBufferLogEventHandler to clear memory correctly in case of exception/error


carterkozak:
Great find! Is there any chance we can test this? [EventParameterMemoryLeakTest](https://github.com/apache/logging-log4j2/blob/master/log4j-core/src/test/java/org/apache/logging/log4j/core/EventParameterMemoryLeakTest.java) might provide some inspiration.

carterkozak:
I have a test which reproduces, pushing shortly.

PR #320, opened 2019-12-10
Layouts doc possible fail

elfogre:
Not compact json means pretty style (previous it said not pretty)


rgoers:
After re-reading this I realized the previous text,  compact (a.k.a not "pretty"),  is correct but confusing.  compact does mean the output will not be pretty printed.  The fact that it is preceded by "not" makes it look like "not compact" is the same as "not pretty".   

I will correct the phrasing to be better.

PR #322, opened 2019-12-20
LOG4J2-2677: Rollover and Deletion handle file deletions gracefully

carterkozak:
It's common for multiple appenders to write (and roll) log files
into a single directory. When these appenders are also responsible
for cleaning up after themselves, they must also search that
directory for files to delete. These searches are executed in
parallel by multiple appenders and have a chance to delete files
the other is checking before the file can be checked against
a regex or glob.
Missing files can safely be assumed to have been deleted.


carterkozak:
Still needs a changelog and tests, not ready to merge.

carterkozak:
merged

PR #323, opened 2019-12-20
LOG4J2-2741 - Fix race condition in computation of prevFireTime in Cr…

ahr-huber:
…onTriggeringPolicy

There was a race condition in the computation of prevFireTime - which is used as time
for the file pattern - could be wrong which lead to the possibility of overwriting files,
because a file name could be generated twice. The cause was CronExpression.getPrevFireTime()
needs a Date in the interval (prevFireTime, currentFireTime], but the Date provided was
created at or after the currentFireTime.
Fixed by using the exact currentFireTime as provided by the scheduler.


ahr-huber:
Closing. Turned out this was already fixed on master. Ticket was LOG4J2-2575. The fix from master was cherry picked.

PR #324, opened 2019-12-23
LOG4J2-2415 - Lock contention in classloader because of missing org.slf4j.ext.EventData

turbanoff:
Make checking class `org.slf4j.ext.EventData` presence only once in static initialization of `org.apache.logging.slf4j.Log4jLogger`.
Also changed order of conditions inside `log` method. JVMs are better at inlining static final fields. See https://shipilev.net/jvm/anatomy-quarks/15-just-in-time-constants/

### DIFF
--- a/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java
+++ b/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java
@@ -42,16 +42,16 @@ public class Log4jLogger implements LocationAwareLogger, Serializable {
 
     private static final long serialVersionUID = 7869000638091304316L;
     private static final Marker EVENT_MARKER = MarkerFactory.getMarker("EVENT");
+    private static final EventDataConverter CONVERTER = createConverter();
+
     private final boolean eventLogger;
     private transient ExtendedLogger logger;
     private final String name;
-    private transient EventDataConverter converter;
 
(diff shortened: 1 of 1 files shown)


log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java
```diff
@@ -411,13 +411,17 @@ private void writeObject(final ObjectOutputStream aOutputStream) throws IOExcept
...
-            return null;
+        if (eventDataLoaded == null) {
+            try {
+                LoaderUtil.loadClass("org.slf4j.ext.EventData");
+                eventDataLoaded = true;
+            } catch (final ClassNotFoundException cnfe) {
+                eventDataLoaded = false;
+            }
         }
+        return eventDataLoaded ? new EventDataConverter() : null;
```
carterkozak: Instead of modifying this method, what if we update the `converter` field to be static and call this method on class initialization? That way instead of requiring work (and allocation!) every time a logger is created, it's done once when the application loads.

turbanoff: done

carterkozak: Great, thanks!

PR #325, opened 2019-12-24
LOG4J2-2745: LOG4J2-2744: Upgrade slf4j-1.8 to beta4 (from alpha2)

carterkozak:
This deletes the EventDataConverter entirely from log4j-slf4j18-impl
entirely because EventData is not supported in 1.8.


carterkozak:
merged

PR #326, opened 2019-12-30
LOG4J2-2750: Extended throwable class loading can be disabled

carterkozak:
Implemented a new `log4j2.enable.extended.throwable.class.loading`
flag which may be set to `false` in order to safely avoid performance
pitfalls of class loading.


carterkozak:
TODOS:
- Documentation for the new flag
- Changelog

rgoers:
@carterkozak I added support for requiresLocation() to LoggerConfig and various the various components a few releases ago. However, requiresLocation() should really be used as advice. If it is set to false it is still possible there is a custom component that didn't register its need for location information. With that in mind is there a way to leverage that somehow?

jacky1193610322:
is this ok？

PR #327, opened 2019-12-30
LOG4J2-2752: MutableLogEvent and RingBufferLogEvent lazily create buffers

carterkozak:
When reusable events are used with non-reusable message implementations
(ParameterizedMessageFactory) there's no need to create these buffers.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEvent.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEvent.java
@@ -58,12 +58,7 @@ public class RingBufferLogEvent implements LogEvent, ReusableMessage, CharSequen
 
         @Override
         public RingBufferLogEvent newInstance() {
-            final RingBufferLogEvent result = new RingBufferLogEvent();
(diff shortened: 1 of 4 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEvent.java
```diff
@@ -416,12 +409,18 @@ public void clear() {
...
-                parameters[i] = null;
+            if (parameters != null) {
+                Arrays.fill(parameters, null);
             }
+        } else {
+            // A user may have manually logged a ReusableMessage implementation, when thread locals are
+            // disabled we remove the reference in order to avoid permanently holding references to these
+            // buffers.
+            messageText = null;
+            parameters = null;
```
carterkozak: Previously this allowed for unexpected growth, though bounded by the ring buffer size.

carterkozak:
Opened for comments for the night, I'll plan to merge tomorrow unless there are concerns. Overall I think this reduces complexity, and reduces memory use in my scenario.

carterkozak:
merged

PR #328, opened 2019-12-31
LOG4J2-2754: LoaderUtil.getClassLoaders may discover additional loaders

carterkozak:
The utility no longer erroneously returns a result with a null element in some
environments.
Also updated loadClass to avoid internally throwing and catching a null
pointer exception when getThreadContextClassLoader returns null.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
@@ -21,11 +21,9 @@ import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Enumeration;
 import java.util.LinkedHashSet;
-import java.util.List;
 import java.util.Objects;
 
 /**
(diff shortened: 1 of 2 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
```diff
@@ -106,30 +104,31 @@ public ClassLoader run() {
...
+        final Collection<ClassLoader> classLoaders = new LinkedHashSet<>();
         final ClassLoader tcl = getThreadContextClassLoader();
-        classLoaders.add(tcl);
-        // Some implementations may use null to represent the bootstrap class loader.
-        final ClassLoader current = LoaderUtil.class.getClassLoader();
-        if (current != null && current != tcl) {
-            classLoaders.add(current);
-            final ClassLoader parent = current.getParent();
-            while (parent != null && !classLoaders.contains(parent)) {
-                classLoaders.add(parent);
```
carterkozak: This loop failed to modify the condition

log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
```diff
@@ -106,30 +104,31 @@ public ClassLoader run() {
     }
 
     public static ClassLoader[] getClassLoaders() {
-        final List<ClassLoader> classLoaders = new ArrayList<>();
+        final Collection<ClassLoader> classLoaders = new LinkedHashSet<>();
```
carterkozak: LinkedHashSet is more expensive in environments with few values, but performs better in the worst case without sacrificing order.

carterkozak:
@rgoers I'm not as familiar with the consumers of `LoaderUtil.getClassLoaders()` as you are, I'd appreciate your feedback before this merges, at your convenience -- I'm not in a rush to merge this, I happened to notice the loop while looking at something else.

carterkozak:
Thanks Ralph, tests all pass, I've merged this commit. It's not a clean cherry-pick to master because fbde9cd0ef was not applied to release-2.x.

PR #330, opened 2020-01-10
Fixed log message in RollingFileAppender

ChristophKaser:
An error message in RollingFileAppender uses a placeholder for the name but does not specify the name argument in the logging call. This PR adds the missing argument.


carterkozak:
Good catch,Thanks! I'll get this merged when I have a moment.

rgoers:
Create LOG4J2-2776 for this.

PR #331, opened 2020-01-13
Fixed check for RollingFileManager,directWrite

ChristophKaser:
Bugfix: Use the interface DirectFileRolloverStrategy instead of the concrete implementation DirectWriteRolloverStrategy when deciding whether to use the pattern to generate a file name.

This fixes a NPE I get with a custom DirectWriteRolloverStrategy and no file name

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingFileManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingFileManager.java
@@ -104,7 +104,7 @@ public class RollingFileManager extends FileManager {
         this.patternProcessor = new PatternProcessor(pattern);
         this.patternProcessor.setPrevFileTime(initialTime);
         this.fileName = fileName;
-        this.directWrite = rolloverStrategy instanceof DirectWriteRolloverStrategy;
+        this.directWrite = rolloverStrategy instanceof DirectFileRolloverStrategy;
         this.fileExtension = FileExtension.lookupForFile(pattern);
     }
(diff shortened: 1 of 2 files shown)


carterkozak:
Thanks @ChristophKaser, would you mind opening a jira issue to accompany this change?

ChristophKaser:
Done: https://issues.apache.org/jira/browse/LOG4J2-2758

carterkozak:
It looks like we have five occurrences of `instanceof DirectWriteRolloverStrategy` across `RollingFileManager` and `RollingRandomAccessFileManager`, should they all be updated? Could we add a test to prevent regressions in the future?

ChristophKaser:
I only count three: 

- The one I replaced
- [Two](https://github.com/apache/logging-log4j2/blob/231f669df3f16c03e35e92460787f73f265ee3ca/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingFileManager.java#L158) [checks](https://github.com/apache/logging-log4j2/blob/231f669df3f16c03e35e92460787f73f265ee3ca/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java#L94) that the fileName is null when DirectWriteRolloverStrategy is used. 
I am unsure about those checks - in theory it may be possible to build a RolloverStrategy that implements DirectFileRolloverStrategy but also uses the fileName. 
But I don't think the interface is meant to be used that way, so I guess it would make sense to replace those as well. What do you think?

I added a simple regression test to this patch.

PR #332, opened 2020-01-15
LOG4J2-2760: Fixes RollingFileAppender with DirectWriteRolloverStrategy and HtmlLayout

ChristophKaser:
This PR consist of a regression test showcasing the problem(s) and a fix.

See https://issues.apache.org/jira/browse/LOG4J2-2760


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
```diff
@@ -164,6 +164,7 @@ public boolean hasOutputStream() {
     protected OutputStream getOutputStream() throws IOException {
         if (outputStream == null) {
             outputStream = createOutputStream();
+            setOutputStream(outputStream); // Needed so the header will be written
```
rgoers: It seems like this should be
```
    if (outputSteam == null) {
        setOutputStream(createOutputStream);
```

PR #335, opened 2020-01-28
Import of LogstashLayout as JsonTemplateLayout

vy:
This is the very first draft of my work on contributing [log4j2-logstash-layout](/vy/log4j2-logstash-layout) into Log4j core. Please note that this is still a work in progress. I want to have your feedback and support on the following issues which I will share below.

## Goals

Below is the list of goals I plan to deliver in order:

- [X] Update manual
- [X] Replace `Uris` with `LoaderUtils`
- [X] Add directive for "nanos of epoch millis", "millis of epoch seconds", etc.
- [X] Move `pattern`, `timeZone`, `locale` parameters into `timestamp` key level.
- [ ] Add benchmark to `log4j-perf`
- [X] Rename `log4j-layout-jackson-json-template` to `log4j-layout-json-template`
- [X] Replace TLAs with a user-provided 3-way caching strategy

I will update this list upon progress.

## Module and package name

I chose `JsonTemplateLayout` as the new module name, which clearly communicates the intent. I am not totally convinced with the package name `o.a.l.log4j.jackson.json.template.layout` though. I have my doubts about whether we can remove Jackson dependency in the future or not. This ambition does look neither feasible, nor practical in the short-term. That said, from an API perspective, why shall an internal detail like `jackson` get exposed in the package name? I am considering to place it under `o.a.l.log4j.layout.json.template`. Any ideas?

## Documentation

`LogstashLayout` has quite some configuration knobs which renders it inconvenient to explain it in `/src/site/asciidoc/manual/layouts.adoc`. I have skimmed through the source code with the hope of finding a similar documentation pattern, but failed to do so. Would somebody mind shedding some light onto the convention I should follow here, please?

## `MapMessage#getFormattedMessage()` JSON encoding bug

As reported in [LOG4J2-2703](https://issues.apache.org/jira/browse/LOG4J2-2703), `MapMessage#getFormattedMessage()` is incorrectly formatting `Object`s for JSON. As a workaround in `LogstashLayout`, I had a `LogstashLayout.Builder#mapMessageFormatterIgnored` property, which I have dragged into `JsonTemplateLayout.Builder` too. The property instructs the layout to ignore `MapMessage#getFormattedMessage()` and do its own magic to properly encode the `MapMessage` into JSON. Shall I keep this ad-hoc fix or rather resolve LOG4J2-2703 first?

## Performance tests

I have some test utility classes that I want to leverage in performance tests, but test classpath of `log4j-jackson-json-template` module is not accessible by `log4j-perf` module. What is the work around here?

## What to do with `JsonLayout`?

I do not intend to implement `JsonLayout` by leveraging `JsonTemplateLayout`, at least, not now. I would rather just leave it be and get deprecated eventually. Put another way, is this PR the right place to address this issue?

## Missing features compared to `JsonLayout`

You might have noticed that `JsonTemplateLayout` misses certain features (e.g., `compact`, `complete`, `header`, `footer`, etc.) compared to `JsonLayout`. This is a deliberate decision of mine. If you have any objections, please chime in.

## Thread locals

Even though the existing TLA usage of `LogstashLayout` is dragged in, I am going to replace it with an object pool. I have spotted `ThreadLocalVsPoolBenchmark` and will start from there.


log4j-layout-jackson-json-template/src/main/resources/JsonLayout.json
```diff
@@ -0,0 +1,27 @@
+{
+  "instant": {
+    "epochSecond": "${json:timestamp:epoch:divisor=1e9,integral}",
+    "nanoOfSecond": 0
```
vy: I don't have a directive for `nanoOfSecond`, yet. Would you rather want to have it in the first release?

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/JsonTemplateLayout.java
```diff
@@ -0,0 +1,678 @@
...
+        return context;
+    }
+
+    private void reloadSerializationContext(
+            final Exception cause,
+            final JsonTemplateLayoutSerializationContext oldContext) {
+        try {
+            oldContext.close();
+        } catch (final Exception error) {
+            System.err.println("context close attempt suppressing the error:");
```
vy: Have a better idea here rather than reaching out for `System.err`?

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/JsonTemplateLayout.java
```diff
@@ -0,0 +1,678 @@
...
+            cause.printStackTrace(System.err);
```
vy: Have a better idea here rather than reaching out for `System.err`?

carterkozak: The StatusLogger is likely what you want here, it's a meta-logger for reporting issues with the logging implementation.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Throwables.java
```diff
@@ -40,12 +40,26 @@ private Throwables() {
      * @return the deepest throwable or the given throwable
      */
     public static Throwable getRootCause(final Throwable throwable) {
+
```
jvz: Seems like there's a broken test related to this change?

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/JsonTemplateLayout.java
```diff
@@ -0,0 +1,678 @@
...
+                throw new IllegalArgumentException(
+                        "maxStringLength requires a positive integer");
+            }
+            Objects.requireNonNull(objectMapperFactoryMethod, "objectMapperFactoryMethod");
+        }
+
+    }
+
+    // We need this ugly model and its builder just to be able to allow
+    // key-value pairs in a dedicated element.
```
jvz: At first I thought this was incorrect, but it turns out that everything else that uses KeyValuePair does so in the same flattened way.

For small plugin classes like this (including named wrappers like this), you can just use a static plugin factory method instead of the builder class. The builder is available for the plugins that use several options where maintaining huge static methods is bad style (i.e., like half of them in core).

jvz:
> I chose `JsonTemplateLayout` as the new module name, which clearly communicates the intent. I am not totally convinced with the package name `o.a.l.log4j.jackson.json.template.layout` though. I have my doubts about whether we can remove Jackson dependency in the future or not. This ambition does look neither feasible, nor practical in the short-term. That said, from an API perspective, why shall an internal detail like `jackson` get exposed in the package name? I am considering to place it under `o.a.l.log4j.layout.json.template`. Any ideas?

No need to put jackson in early like that I think. The package was moved around for modularization in 3.x which makes me think we can still rename things.

> `LogstashLayout` has quite some configuration knobs which renders it inconvenient to explain it in `/src/site/asciidoc/manual/layouts.adoc`. I have skimmed through the source code with the hope of finding a similar documentation pattern, but failed to do so. Would somebody mind shedding some light onto the convention I should follow here, please?

You can use other formats if that makes it easier, though we may have to get creative.

> I have some test utility classes that I want to leverage in performance tests, but test classpath of `log4j-jackson-json-template` module is not accessible by `log4j-perf` module. What is the work around here?

You can add the tests jar as a dependency to log4j-perf.

> I do not intend to implement `JsonLayout` by leveraging `JsonTemplateLayout`, at least, not now. I would rather just leave it be and get deprecated eventually. Put another way, is this PR the right place to address this issue?

Can likely be done as a follow up.

log4j-layout-jackson-json-template/src/main/resources/JsonLayout.json
vy: @rgoers, any ideas?

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/JsonTemplateLayout.java
vy: @carterkozak, thanks! Fixed.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Throwables.java
vy: There was a code typo. Fixed it and added a test to check against causal chains with loops.

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/JsonTemplateLayout.java
vy: Yes, I have thought of sticking to `@PluginFactory` for `EventTemplateAdditionalFields`, but eventually settled with the `Builder`. I am inclined to keep the *building* of the `@Plugin`s consistent within the same file/module.

I will mark the conversation as resolved. Please don't hesitate to open it up again, if you have objections.

vy:
> > I chose `JsonTemplateLayout` as the new module name, which clearly communicates the intent. I am not totally convinced with the package name `o.a.l.log4j.jackson.json.template.layout` though. I have my doubts about whether we can remove Jackson dependency in the future or not. This ambition does look neither feasible, nor practical in the short-term. That said, from an API perspective, why shall an internal detail like `jackson` get exposed in the package name? I am considering to place it under `o.a.l.log4j.layout.json.template`. Any ideas?
> 
> No need to put jackson in early like that I think. The package was moved around for modularization in 3.x which makes me think we can still rename things.

:+1: I will rename the package to `o.a.l.log4j.layout.json.template` once the dust settles down. I don't want to introduce such a big change right now and corrupt the conversation history.

> > `LogstashLayout` has quite some configuration knobs which renders it inconvenient to explain it in `/src/site/asciidoc/manual/layouts.adoc`. I have skimmed through the source code with the hope of finding a similar documentation pattern, but failed to do so. Would somebody mind shedding some light onto the convention I should follow here, please?
> 
> You can use other formats if that makes it easier, though we may have to get creative.

I am okay with AsciiDoc, my point was, shall I create a dedicated page (e.g., `json-template.adoc`) just for `JsonTemplateLayout`, given the rest of the `Layout`s are briefly summarized in `layouts.adoc`. Anyway... I will go toward this direction -- better to ask for excuse, rather than waiting permission. :wink: 

> > I have some test utility classes that I want to leverage in performance tests, but test classpath of `log4j-jackson-json-template` module is not accessible by `log4j-perf` module. What is the work around here?
> 
> You can add the tests jar as a dependency to log4j-perf.

Hrm... I didn't know that was possible. I will investigate how I can do that. If you happen to feel like helping, hints are welcome.

> > I do not intend to implement `JsonLayout` by leveraging `JsonTemplateLayout`, at least, not now. I would rather just leave it be and get deprecated eventually. Put another way, is this PR the right place to address this issue?
> 
> Can likely be done as a follow up.

:+1:

vy:
@remkop, please note the revamp in ed1dda4. The results on my machine are as follows:

```
$ uname -a
Linux tahta 4.15.0-70-generic #79-Ubuntu SMP Tue Nov 12 10:36:11 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ grep "model name" /proc/cpuinfo | head -n 1
model name      : Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz

$ cat /proc/cpuinfo | grep processor | wc -l
8

$ java -version
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.2+9)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.2+9, mixed mode)

$ java -jar log4j-perf/target/benchmarks.jar ".*ThreadLocalVsPool.*" -f 1 -wi 10 -i 20 -tu ns -bm sample -t 4

                            sample  27754662      492.377 ± 6.634  ns/op
NoOp·p0.00                  sample                263.000          ns/op
NoOp·p0.50                  sample                353.000          ns/op
NoOp·p0.90                  sample                610.000          ns/op
NoOp·p0.95                  sample                746.000          ns/op
NoOp·p0.99                  sample               1516.000          ns/op
NoOp·p0.999                 sample              10800.000          ns/op
NoOp·p0.9999                sample              26240.000          ns/op
NoOp·p1.00                  sample            9797632.000          ns/op
                            sample  20835106      352.366 ± 0.354  ns/op
ThreadLocal·p0.00           sample                251.000          ns/op
ThreadLocal·p0.50           sample                324.000          ns/op
ThreadLocal·p0.90           sample                331.000          ns/op
ThreadLocal·p0.95           sample                573.000          ns/op
ThreadLocal·p0.99           sample                785.000          ns/op
ThreadLocal·p0.999          sample               2384.000          ns/op
ThreadLocal·p0.9999         sample              15232.000          ns/op
ThreadLocal·p1.00           sample            1386496.000          ns/op
                            sample  20989233      599.494 ± 0.432  ns/op
RightProvisionedJc·p0.00    sample                265.000          ns/op
RightProvisionedJc·p0.50    sample                523.000          ns/op
RightProvisionedJc·p0.90    sample                841.000          ns/op
RightProvisionedJc·p0.95    sample                942.000          ns/op
RightProvisionedJc·p0.99    sample               1194.000          ns/op
RightProvisionedJc·p0.999   sample               5032.000          ns/op
RightProvisionedJc·p0.9999  sample              18976.000          ns/op
RightProvisionedJc·p1.00    sample            1038336.000          ns/op
                            sample  20839035      624.378 ± 3.719  ns/op
UnderProvisionedJc·p0.00    sample                267.000          ns/op
UnderProvisionedJc·p0.50    sample                547.000          ns/op
UnderProvisionedJc·p0.90    sample                776.000          ns/op
UnderProvisionedJc·p0.95    sample                931.000          ns/op
UnderProvisionedJc·p0.99    sample               1466.000          ns/op
UnderProvisionedJc·p0.999   sample               5928.000          ns/op
UnderProvisionedJc·p0.9999  sample              19008.000          ns/op
UnderProvisionedJc·p1.00    sample            9011200.000          ns/op
```

Either I am missing something big or G1 in JDK 11 is creating miracles. I would appreciate some insight here.

remkop:
I’ll take a look but it will take me some time.

jvz:
To add a test jar as a dependency, add `<type>test-jar</type>` to the `<dependency>` definition.

vy:
> To add a test jar as a dependency, add `<type>test-jar</type>` to the `<dependency>` definition. See for example [log4j-core-its/pom.xml](https://github.com/apache/logging-log4j2/blob/master/log4j-core-its/pom.xml).

Yep, indeed there are plenty of examples, found it, thanks.

remkop:
Showing only the average in ns/operation:

```
NoOp                        sample  27754662      492.377 ± 6.634  ns/op
ThreadLocal                 sample  20835106      352.366 ± 0.354  ns/op
RightProvisionedJc          sample  20989233      599.494 ± 0.432  ns/op
UnderProvisionedJc          sample  20839035      624.378 ± 3.719  ns/op
```

**What do these numbers mean?**

It is the number of nanoseconds it takes on average to serialize a log event into a StringBuilder using the `%d %5p [%t] %c{1} %X{transactionId} - %m%n` PatternFormat. (Note that the log event has no context data, so the `%X{transactionId}` formatter is skipped.)

**What can we conclude from these numbers?**

The `NoOp` benchmark should be renamed `Allocate`, because it creates a `new StringBuilder(1024)` every invocation. Allocation does not seem too expensive here. Perhaps you have a lot of memory and the benchmark finishes before GC kicks in?

The `ThreadLocal` benchmark shows it is a bit faster to reuse an existing StringBuilder instance cached in a ThreadLocal.

The `Jc` benchmarks use the `org.jctools.queues.MpmcArrayQueue` data structure which seems a bit slower than a simple `ThreadLocal`.

@vy Does that answer your question?

remkop:
If you wonder why the `MpmcArrayQueue` benchmark numbers are higher than simple allocation, I am not sure either.

My first guess would be lock contention. With 4 threads that could be significant. I don't believe that `MpmcArrayQueue` is lock-free (but it has been a while, I could be wrong).

vy:
> The NoOp benchmark should be renamed Allocate

Check.

> I don't believe that `MpmcArrayQueue` is lock-free

The definition of what is lock-free is quite [a debate](https://groups.google.com/forum/#!topic/lock-free/MnhH9AlLfOc) on its own. But according to [README](/JCTools/JCTools#jctools), `MPMC` is lock-less (what ever that means) and bounded.

I am just baffled by the fact how come `AllocatePool` can yield such a miraculous performance that is almost on par with `ThreadLocalPool`. I have shared this question with @richardstartin and he kindly gave the following feedback:

> I think that the G1 write barrier is the cost in the MPMC case (I haven't run the benchmark though) and I would be interested to know the marginal cost of G1 in this case. The most interesting metric is MPMC/G1 vs MPMC/other GC

I will experiment with smaller heap sizes and bigger `StringBuilder` capacities.

richardstartin:
To be clear, that’s just a hunch. I haven’t run the benchmarks let alone profiled them.

vy:
And the following question still stands open: Given TLA results constitute the baseline, what is the pool implied maximum performance degradation we are okay with? With a rough look, MPMC performs 2.5x worse compared to TLAs on 4 cores. I will dig further... Comments are welcome.

remkop:
Well, what is the trade off? What benefits do we gain in return of this worse performance?

remkop:
> I would be interested to know the marginal cost of G1 in this case. The most interesting metric is MPMC/G1 vs MPMC/other GC

I don’t think we’re measuring the cost of GC anywhere, are we?

richardstartin:
> I don’t think we’re measuring the cost of GC anywhere, are we?

At this scale, there are so many confounding factors which can explain a difference, and you won't get an explanation without profiling the benchmark. One quite big factor at this scale is the choice of garbage collector, and _not_ because of the time it spends collecting garbage.

I just guessed (in what I thought was a private conversation) that the cost might be explained by the little bit of machine code the JIT compiler inserts into reference assignments to track inter-region references (the write barrier). In my experience, this can really show up in microbenchmarks running G1, especially with some concurrency. My advice was simply to run the benchmark with a different garbage collector and see how much it affects the numbers.  

Since I was tagged here, my advice would be to think twice about optimising this, and think about the future. If someone imports this library, they may well never upgrade it, but stack allocation which might be implemented in the JIT compiler would:

* remove the garbage (benefit vs heap allocation)
* avoid any contention (benefit vs pooling)
* avoid an increase in the number of GC roots (benefit vs thread local)

Applications would benefit automatically when they upgrade the JVM, but not if you optimise this now.

vy:
Triggered by @richardstartin's observation, why don't we allow all 3 co-exist in the code base? After all, the call site doesn't care about the caching mechanics. We can let the user decide via a configuration knob ala `cache=<none|thread-local|queueClassName,capacity>`.

remkop:
Not sure I agree... What is the trade-off? What do users gain when they select something other than ThreadLocal?

vy:
> Not sure I agree... What is the trade-off? What do users gain when they select something other than ThreadLocal?

If you'd recall, we had this discussion in the mailing list in [December](http://mail-archives.apache.org/mod_mbox/logging-dev/201912.mbox/%3CCAP7pH7sHb-nP8iGJybruCkno1FzajrO8PSUBPCOAkpZP2dNYvA%40mail.gmail.com%3E) and [January](http://mail-archives.apache.org/mod_mbox/logging-dev/202001.mbox/%3CCACmp6koucj6N2%3DFW_xM4sX2ZCLdYkHZSs9535NYjfY5wtWiLGQ%40mail.gmail.com%3E). There the raised concern was TLAs, in particular, in the domain of web applications where there is a proliferation of threads and `AbstractStringLayout`, used by many `String`-emitting layouts, is not taking `log4j2.enableThreadlocals` flag into account, etc.

As described in [this ticket](/vy/log4j2-logstash-layout/issues/50#issuecomment-574671508), `JsonTemplateLayout` has a memory footprint which can be summarized with the following rough formula:

```
threadCount * (
    (stackTraceEnabled ? 1 : 0) * maxByteCount +    // for stacktraces
    maxByteCount +    // for JSON
    epsilon    // for the rest (<100 bytes)
)
```

While this is a perfectly affordable footprint where the thread count matches the available CPU cores, it is not something desirable in a (web) application container. My suggestion is to leave this decision to user (granted we incorporate sane defaults) and support each strategy. Is this a slick Netty/gRPC app? Stick to TLAs. Is this an app running on Tomcat? Stick to pooling. Is this an embedded app with relatively low load? Stick to allocation.

remkop:
@vy Understood. That makes sense to me.

log4j-layout-jackson-json-template/src/main/java/org/apache/logging/log4j/jackson/json/template/layout/util/Uris.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    private static String readClassPathUri(
+            final URI uri,
+            final Charset charset)
+            throws IOException {
+        final String spec = uri.toString();
+        final String path = spec.substring("classpath:".length());
+        final URL resource = Uris.class.getClassLoader().getResource(path);
```
vy: @jvz, @carterkozak, @remkop, I guess this is wrong, right? I should rather be using `log4j.core.util.Loader`, right? Speaking of which, shall I move all this content to `Loader` instead?

jvz: Use whatever the appropriate ClassLoader is for the context. The Loader/Util classes are practically hacks at this point due to most Java applications being either flat classpaths or web applications with a well-known ClassLoader model. The more fun environments like Java module paths and OSGi are why these util classes exist, and we sometimes get that wrong.

In the revamped system I've been working on, I've been doing everything I can to avoid the need to ever load a class by name like this, though your best bet currently would be to obtain the ClassLoader of a relevant parent class or known class in the jar to ensure you've selected the correct one. For loading files here, you'll want the ClassLoader of a Class from the jar where the resource is located as usual. If you're trying to load arbitrary files like `log4j2.xml`, that can likely be loaded from the TCCL as that should be set to the main application's loader which is where you typically expect to start from. If you're loading extensions from other jars dynamically, that's where it gets tricky and might require separate implementations for different module systems.

jvz: So, to answer your original question which I somehow missed, correct, this looks incorrect. If you want to load arbitrary classpath resources, you can try using `LoaderUtil.findResources` which appears to have been designed for this use case, but I'm not sure if that function will work properly in all environments (better than nothing I guess).

log4j-layout-jackson-json-template/src/main/resources/JsonLayout.json
rgoers: Sorry for not responding sooner.  I am fine with skipping features so long as we are not trying to be completely compatible with the existing JSONLayout.

vy:
@jvz, I could not get the test-jar thing working. I have pushed my progress: 7084934. I will be so glad if you can spare some time to help me with figuring out the missing piece.

jvz:
It's possible that there's a corresponding configuration to enable for the test jar to make it available as a dependency like in log4j-core:test-jar.

vy:
@jvz, I have the cloned `maven-jar-plugin` configuration in `/log4j-layout-json-template/pom.xml` and I can see both `log4j-layout-json-template-3.0.0-SNAPSHOT.jar` and `log4j-layout-json-template-3.0.0-SNAPSHOT-tests.jar` in `/log4j-layout-json-template/target` directory. But still Maven falls over its face:

```
[ERROR] Failed to execute goal on project log4j-perf: Could not resolve dependencies for project org.apache.logging.log4j:log4j-perf:jar:3.0.0-SNAPSHOT: Could not find artifact org.apache.logging.log4j:log4j-layout-json-template:jar:tests:3.0.0-SNAPSHOT
```

I will really appreciate some more help, please.

jvz:
I'm not really sure what else there is to configure. I can't get it working at the moment myself.

vy:
@remkop, in case you are interested in, dae3be0 introduces the abstraction for different caching strategies we have discussed: dummy, thread-local, and queue.

vy:
For the records, I am done. The only missing bit is getting test-jar working for log4j-perf module.

rgoers:
OK. I hope I can look at that this weekend. I am also planning on trying to get a release out this weekend so I would expect to merge the PR right after that.

Ralph

log4j-api/src/main/java/org/apache/logging/log4j/util/Strings.java
```diff
@@ -55,14 +55,23 @@ public static String dquote(final String str) {
...
-        return s == null || s.trim().isEmpty();
+        if (s == null || s.isEmpty()) {
+            return true;
+        }
+        for (int i = 0; i < s.length(); i++) {
+            char c = s.charAt(i);
+            if (!Character.isWhitespace(c)) {
+                return false;
+            }
+        }
```
rgoers, line 73: Why is this necessary? Does trim() now work correctly?

log4j-core/src/test/java/org/apache/logging/log4j/core/GcFreeLoggingTestUtil.java
```diff
@@ -16,35 +16,33 @@
...
 
-import com.google.monitoring.runtime.instrumentation.AllocationRecorder;
-import com.google.monitoring.runtime.instrumentation.Sampler;
+import java.io.File;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Pattern;
```
rgoers, line 34: The order of imports no longer matches the style guide - http://logging.apache.org/log4j/2.x/javastyle.html. I should note though that we have lots of checkstyle errors in the existing code (they should be addressed).

src/site/markdown/manual/cloud.md
```diff
@@ -118,76 +118,69 @@ being configured with a list of hosts and ports so high availability is not an i
...
-      gelf {
-        host => "localhost"
-        use_tcp => true
-        use_udp => false
-        port => 12222
-        type => "gelf"
-      }
-    }
+We will configure Logstash to listen on TCP port 12345 for payloads of type JSON
+and then forward these to (either console and/or) an Elasticsearch server.
```
rgoers, line 164: In my testing gelf has to be configured so that it uses null to delimit log events, otherwise events that contain exceptions get split into two log events. Did you do something magical to make the tcp input with JSON suddenly behave properly?

src/site/markdown/manual/cloud.md
```diff
@@ -118,76 +118,69 @@ being configured with a list of hosts and ports so high availability is not an i
...
+Template Layout](layouts.html#JSONTemplateLayout) allows full schema
+customization and bundles ELK-specific layouts by default, which makes it a
+great fit for the bill.
+
+    <Socket name="Logstash"
+            host="${sys:logstash.host}"
+            port="12345"
+            protocol="tcp"
+            bufferedIo="true">
+        <JsonTemplateLayout eventTemplateUri="classpath:EcsLayout.json">
```
rgoers, line 139: Where does this layout say to use Null as a delimiter (or some other character besides newline)?

vy:
> The logging in the cloud docs need to be discussed or reverted.

How shall we proceed? Do you want to discuss in the mailing list or here?

rgoers:
mailing list. This is already too long.

log4j-api/src/main/java/org/apache/logging/log4j/util/Strings.java
```diff
@@ -286,4 +295,23 @@ public static String join(final Iterator<?> iterator, final char separator) {
...
     }
 
+    /**
+     * Creates a new string repeating given {@code str} {@code count} times.
+     * @param str input string
+     * @param count the repetition count
+     * @return the new string
+     * @throws IllegalArgumentException if either {@code str} is null or {@code count} is negative
+     */
+    public static String repeat(final String str, final int count) {
```
carterkozak, line 305: I think this is only used in test code, should we put it into a test util instead of log4j-api main api? Otherwise it could use some optimziation around edge cases, for example if count is zero we return empty string, one we return the input `str`.

log4j-api/src/main/java/org/apache/logging/log4j/util/Strings.java
carterkozak: Trim allocates new strings, I like this optimization assuming the behavior is the same. It might be nice to separate this optimization into another change that's not related to the new json layout.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Throwables.java
```diff
@@ -40,12 +40,26 @@ private Throwables() {
...
+        Throwable parent = throwable;
         Throwable cause;
-        Throwable root = throwable;
-        while ((cause = root.getCause()) != null) {
-            root = cause;
+        while ((cause = parent.getCause()) != null) {
+            parent = cause;
+            if (parent == slowPointer) {
+                throw new IllegalArgumentException("loop in causal chain");
```
carterkozak, line 54: This isn't something we expect to happen often, but I wonder if we should return the input `throwable` and debug log to the status logger in this case?

log4j-layout-json-template/src/main/java/org/apache/logging/log4j/layout/json/template/JsonTemplateLayout.java
```diff
@@ -0,0 +1,562 @@
...
+        } finally {
+            contextRecycler.release(context);
+        }
+    }
+
+    @Override
+    public void encode(final LogEvent event, final ByteBufferDestination destination) {
+
+        // Acquire a context.
+        final Context context = acquireContext();
```
carterkozak, line 224: no changes requested, but it might be helpful to implement Closeable (without throwing an IOException) to use structured try-with-resources blocks around context use.

log4j-layout-json-template/src/main/java/org/apache/logging/log4j/layout/json/template/resolver/EventResolverFactories.java
```diff
@@ -0,0 +1,64 @@
...
+
+enum EventResolverFactories {;
+
+    private static final Map<String, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>>> RESOLVER_FACTORY_BY_NAME =
+            createResolverFactoryByName();
+
+    private static Map<String, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>>> createResolverFactoryByName() {
+
+        // Collect resolver factories.
+        final List<EventResolverFactory<? extends EventResolver>> resolverFactories = Arrays.asList(
```
rgoers, line 35: This seems somewhat brittle to me but this can be improved as the need arises.

ramazanpolat:
Hi, do you have a documentation for `JsonTemplateLayout`?

vy:
@ramazanpolat, you can access [the latest version of the manual here](https://github.com/apache/logging-log4j2/blob/master/src/site/asciidoc/manual/json-template-layout.adoc.vm). Due to `.vm` (Velocity template) suffix, GitHub is not able to render the AsciiDoc properly. You can copy the content and either paste it into [asciidocLive](https://asciidoclive.com) or use your own IDE (IDEA, Eclipse, etc.) for the rendered view.

The changes are landed both on `master` and `release-2.x` branches. With the upcoming Log4j 2.x release (which I am trying to steer the rest of the Log4j maintainers toward as of this writing) the layout will be available in `log4j-layout-template-json` artifact ID and the website will be updated accordingly.

ramazanpolat:
Thank you @vy
I think you made a great addition to log4j2 with this work.

ramazanpolat:
@vy One quick question: Will `JsonTemplateLayout` have the ability to jsonify the message object passed to Logger?

e.g: Currently with using JsonLayout, this code:
```java
  logger.info("This is an info message");
```
produces something like this:
```json
{
  "timeMillis" : 1598613343782,
  "thread" : "main",
  "level" : "INFO",
  "loggerName" : "com.example.Test",
  "message" : "This is an info message",
  "endOfBatch" : false,
  "loggerFqcn" : "org.apache.logging.log4j.spi.AbstractLogger",
  "contextMap" : { },
  "threadId" : 1,
  "threadPriority" : 5
}
```
But if we pass an object to the logger like this (considering the object constructor is `MyDummyEventBean(String msg, String user, String ip)`:
```java
  logger.error(new MyDummyEventBean("Something bad happened", "john", "10.1.2.3");
```
JsonLayout gets appends `.toString()` result of the `MyDummyEventBean` and produces this:
```json
{
  "timeMillis" : 1598613343782,
  "thread" : "main",
  "level" : "INFO",
  "loggerName" : "com.example.Test",
  "message" : "{\"message\":\"Something bad happened\",\"userId\":\"john\",\"ip\":\"10.1.2.3\"}"
  "endOfBatch" : false,
  "loggerFqcn" : "org.apache.logging.log4j.spi.AbstractLogger",
  "contextMap" : { },
  "threadId" : 1,
  "threadPriority" : 5
}
```

vy:
@ramazanpolat, yes, `JsonTemplateLayout` can do that. You just need to set `stringified` to `false` (the default) as follows:

```json
{
  "$resolver": "message",
  "stringified": false
}
```

Further, `LogstashLayout` can also do that. In `LogstashLayout`, you just need to use `${json:message:json}` directive rather than `${json:message}`.

HTH.

ramazanpolat:
@vy Sorry but `LogstashLayout` spits my object's as toString() output:

```json
{
  "line" : 123,
  "class" : "tr.gov.sgk.logclient.DAO",
  "message" : "MyLogBean2{message='LogBean Message', userId='LogBean userId', ip='LogBean IP'}",
  "level" : "ERROR",
  "method" : "errorWithExceptionLogstash1",
  "logger" : "logstashLogger"
}
```
Maybe I'm doing something wrong. My `LogstashJsonEventLayoutV1.son` file:
```json
{
  "line": "${json:source:lineNumber}",
  "class": "${json:source:className}",
  "message": "${json:message:json}",
  "level": "${json:level}",
  "method": "${json:source:methodName}",
  "logger": "${json:logger:name}"
}
```
My `LogstashLayout` in `log4j2.xml`:
```xml
<LogstashLayout eventTemplateUri="classpath:LogstashJsonEventLayoutV1.json"
                prettyPrintEnabled="true"
                locationInfoEnabled="true"
                stackTraceEnabled="true" />
```
The class I want to be printed as Json(**maybe this class needs to implement an interface?**):
```java
public class MyLogBean2 {
    private String message;
    private String userId;
    private String ip;

    public MyLogBean2() {    }

    public MyLogBean2(String message, String userId, String ip) {
        this.message = message; this.userId = userId; this.ip = ip;
    }

    public String getMessage() { return message;}

    public void setMessage(String message) {this.message = message;}

    public String getUserId() {return userId;}

    public void setUserId(String userId) {this.userId = userId;}

    public String getIp() {return ip;}

    public void setIp(String ip) {this.ip = ip;}

    @Override
    public String toString() {
        return "MyLogBean2{" +
                "message='" + message + '\'' +
                ", userId='" + userId + '\'' +
                ", ip='" + ip + '\'' +
                '}';
    }
   
}
```

vy:
1. Unless there is a reason to do so, I would avoid custom classes to represent the logging structure. In your case a simple `logger.info(new MapMessage(Map.of("message", message, "userId", userId, "ip", ip)))` fits the bill.
2. You should have either used `logger.info(new ObjectMessage(customLogBean))` or implement `Message` interface in the `CustomLogBean` class.

diepet:
Hi, I am aware of this development after open this Jira issue: [LOG4J2-2936](https://issues.apache.org/jira/browse/LOG4J2-2936)

Please, could you tell me:

1) When will be the `JsonTemplateLayout` released?

2) Is there someway to resolve the message parameters? As explained in the ticket, I would like to isolate the message parameters in a different JSON section.

3) If one of the additional fields configured under `<EventTemplateAdditionalField>` resolves an empty value, the relative JSON field will be rendered with an empty string as value or not rendered at all?

Thanks in advance and congrats for your work!

vy:
Let me try to answer your questions @diepet:

> 1. When will be the `JsonTemplateLayout` released?

Next Log4j 2.x release (2.14.0?) will contain the `JsonTemplateLayout`. @rgoers is busy with rolling the release out.

> 2. Is there someway to resolve the message parameters? As explained in the Jira issue linked above, I would like to isolate and to render the message parameters in a different JSON section.

`JsonTemplateLayout` doesn't provide a built-in resolver for message parameters. (AFAIK, neither `PatternLayout` does.) That said, I question this approach. (I will comment more about this in the linked Jira issue.) 

> 3. If one of the additional fields configured under `<EventTemplateAdditionalField>` resolves an empty value, the relative JSON field will be rendered with an empty string as value or not rendered at all?

Yes, it will be rendered as is. If it is `null`, `""`, or `"\t\n"`, etc. whatever, it will be rendered intact. `JsonTemplateLayout` doesn't employ any blank value elimination.

diepet:
Thanks @vy for the feedbacks!

Ok, so let's move the discussions for the point 2 in the Jira issue.

Could be possible, in the future, to plan a new `<EventTemplateAdditionalField>` boolean attribute for disabling the rendering of these kind of fields in case of empty/null values?

I am looking forward to download the new log4j2 release!

vy:
_[Sorry for the late reply, I am pretty much swamped with my day job nowadays.]_

@diepet, I have deliberately left out _blank value elimination_ in `JsonTemplateLayout` (while I was the one who has introduced it in `LogstashLayout`) mostly due to the following reasons:

- It incurs an extra runtime cost.
- Most of the time you don't care. You persist logs in a storage system, e.g., ELK stack, and there null value elimination is provided out of the box by the storage engine in the most efficient way.

If you have a use case which necessitates such a feature, I would like to hear more about it.

diepet:
Ehy @vy, better a late reply than a null reply!

Thanks, your points  are clear and understandable.

I am looking forward to using the message parameters resolver discussed in [Jira](https://issues.apache.org/jira/browse/LOG4J2-2936?page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel&focusedCommentId=17210684)!

vy:
Thanks for the feedback @diepet, keep 'em coming!

PR #337, opened 2020-01-31
Log4j 2LOG4J2-2770 Fix NPE in JMX reconfiguration

bkuker:
I think this needs to store the data so that resetInputStream() works.
The jmx function LoggerContextAdmin::getConfigText throws a NPE if you have previously called LoggerContextAdmin::setConfigText


rgoers:
Can you create a Jira issue for this. It would also be great to include a unit test that fails without this fix.

bkuker:
Thanks, yes I'll do that Monday!

PR #343, opened 2020-02-23
Conditionally perform status logging calculations in PluginRegistry

MariusVolkhart:
Debug status logging is typically disabled, but formatting calculations were performed regardless.

Moving the formatting logic inside of a conditional reduces the startup time of applications. The reduction is mostly attributable to the fact that the class initialization of DecimalFormat is non-trivial for the JVM, as it loads a variety of formatting information.

### DIFF
--- a/log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/util/PluginRegistry.java
+++ b/log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/util/PluginRegistry.java
@@ -39,6 +39,8 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import static org.apache.logging.log4j.util.Unbox.box;
+
 /**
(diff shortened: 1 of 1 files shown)


log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/util/PluginRegistry.java
```diff
@@ -320,10 +326,12 @@ public void loadPlugins(ClassLoader classLoader, Map<String, List<PluginType<?>>
         }
 
         final long endTime = System.nanoTime();
```
carterkozak: Can we move these nanoTime invocations into the conditional as well?

rgoers: This is a good idea but it makes more sense to me to have the whole message generated as a lambda. Working on modifying it to do that now.

rgoers:
I have applied a version of this that uses lambdas. Thanks for the good work!

PR #351, opened 2020-03-20
LOG4J2-1360 - Provide a Log4j implementation of System.Logger

kevinleturc:
You can find below a POC of JEP 264 implementation with log4j. I may have missed some implementation details, let me know if there's something I can do.

The related ticket: [LOG4J2-1360](https://issues.apache.org/jira/browse/LOG4J2-1360)

### DIFF
--- a/log4j-jpl/pom.xml
+++ b/log4j-jpl/pom.xml
@@ -156,6 +156,80 @@
       </plugin>
     </plugins>
   </build>
+  <reporting>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-changes-plugin</artifactId>
+        <version>${changes.plugin.version}</version>
+        <reportSets>
+          <reportSet>
+            <reports>
+              <report>changes-report</report>
+            </reports>
+          </reportSet>
+        </reportSets>
+        <configuration>
+          <issueLinkTemplate>%URL%/show_bug.cgi?id=%ISSUE%</issueLinkTemplate>
+          <useJql>true</useJql>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>${checkstyle.plugin.version}</version>
+        <configuration>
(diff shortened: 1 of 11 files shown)


log4j-jpl/pom.xml
```diff
@@ -0,0 +1,255 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+      <scope>runtime</scope>
```
rgoers: Why is this a runtime dependency? Could this implementation use just the Log4j API to bind to an alternate implementation? If so then log4j-core would just be a test dependency.

kevinleturc: You were right, we just need log4j-core for testing. Scope was changed to `test`.

kevinleturc:
I added the site section with the right maven plugins to make the build succeed.

In this addition, I faced several issues with tools such as surefire, javadoc, spotbugs or pmd due to toolchain usage. I created a `java11-module` maven profile, activated if jdk is higher than 11, to not loose to many things with the regular build under jdk 8.
- adding module-info.java makes surefire 2.21 failing.
we need surefire 2.13 to make forked VM works with toolchain, but with this version we don't test the modular system -> run the build with surefire 2.13 and add under the java 11 profile a surefire 2.21 configuration with forked VM disabled in order to test modular system
- javadoc doesn't succeed to generate site for classes whose source/target is 9 -> compile everything with a Java 11 LTS
- sportbugs and pmd run on compiled classes and doesn't support toolchain plugin -> I moved their reports under the java 11 profile

rgoers:
I am having a difficult time getting the site build to work with this. When I use Java 11 it fails in the JMX GUI module because jconsole isn't provided with Java 11. When I build with Java 8 the javadoc plugin fails in this new module with:

`Exit code: 1 - error: module not found: org.apache.logging.log4j.jpl`

I believe this is because it is using the wrong version of javadoc even though it says it is using the Java 11 version. When I run mvn javadoc:javadoc with Java 11 I don't get this error. I am not prepared to commit this until I can get the site build to work.

rgoers:
This was merged.

PR #353, opened 2020-03-23
Fix initialization when security is enabled

dumitrubogdanmihai:
File:exists may cause an AccessControlException in case of an strict SecurityManager.
This would fail the initialization.


rgoers:
Fixed by PR #355

PR #355, opened 2020-03-25
LOG4J2-2761: Fix FileUtils#fileFromUri to works correctly with Securi…

uschindler:
…tyManager and also fix URI parsing to behave sane with standards


uschindler:
This did not happen locally:

```
[ERROR] Failures: 
[ERROR]   FileOutputTest.testConfig:46 Status output file does not exist
Expected: file exists is <true>
     but: file exists was <false>
[ERROR]   TimeFilterTest.testTime:193 expected same:<NEUTRAL> was not:<DENY>
````

uschindler:
I figured out where the issue is: If the file does not exist, `fileFromUri(URI)` returns null. The previous code worked. The problem was a bit of fall-through logic. Should be fixed now. Running tests and will push afterwards.

PR #357, opened 2020-04-20
Support Location Info in GELF Layout

CrazyBills:
GELF Layout supports message pattern but couldn't print location info

So I created this PR to fix this issue, it will check if the message pattern contains location info.


rgoers:
Code was merged.

PR #358, opened 2020-04-27
LOG4J2-2832: Updated strategy name to existing strategy

asbachb:
Instead of using a hard coded name the class name of the interface was used which
a) Increases visibility where the interface is referenced
b) Prevents that the message needs to be adjusted when the interface name changes


asbachb:
@rgoers is there anything todo which prevents this from merging?

PR #359, opened 2020-05-15
LOG4J2-2344: Fix exception message in PropertiesConfigurationBuilder.createFilter

dengliming:


vy:
Thanks so much @dengliming!

PR #360, opened 2020-05-20
LOG4J2-2847 - Extend Log4j-config.xsd

sebthom:
This commit modifies the Log4j-config.xsd in the following ways:
a) the type elements are re-order in a way that related types are physically close to each other, that helps with extending the schema in the future
b) the Configuration type is extended by the missing attributes
c) validation for boolean/integer types is added
d) uniqueness is enforced when declaring custom levels, properties and loggers
e) new configuration elements are added (filters, appenders)
f) the default values and description of attributes are added

### DIFF
--- a/log4j-core/src/main/resources/Log4j-config.xsd
+++ b/log4j-core/src/main/resources/Log4j-config.xsd
@@ -16,149 +16,1295 @@
  limitations under the License.
 
 -->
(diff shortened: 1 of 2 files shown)


sebthom:
@rgoers I added a test case that validates the config xml files found under /src/test/resources against the XSD.

PR #361, opened 2020-05-29
LOG4J2-2858: More flexible configuration of WaitStrategy of Disruptor

stepan2271:
LOG4J2-2858: added additional properties for setting sleepTimeNs and retries parameters of SleepingWaitStrategy


log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -60,36 +60,44 @@
     private DisruptorUtil() {
     }
 
-    static long getTimeout(final String propertyName, final long defaultTimeout) {
-        return PropertiesUtil.getProperties().getLongProperty(propertyName, defaultTimeout);
+    static WaitStrategy createWaitStrategy(final String propertyName) {
+        final String strategyStr = PropertiesUtil.getProperties().getStringProperty(propertyName, DisruptorWaitStrategy.TIMEOUT.toString());
+        LOGGER.trace("property {}={}", propertyName, strategyStr);
+        final DisruptorWaitStrategy strategy = DisruptorWaitStrategy.valueOf(Strings.toRootUpperCase(strategyStr));
```
remkop, line 66: I'm not in favor of using an enum and doing a `valueOf` lookup this way. This lookup means that an invalid value (like "INVALID") will result in an `IllegalArgumentException` being thrown. This is not desirable.

In the previous code, an invalid value would result in the `default` case being selected in the `switch` statement, and the logic would correctly fall back to the default `TimeoutBlockingWaitStrategy`.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -60,36 +60,44 @@
...
+        return strategyKey.startsWith("AsyncLogger.")
+                ? "AsyncLogger." + additionalKey
+                : "AsyncLoggerConfig." + additionalKey;
     }
 
-    static WaitStrategy createWaitStrategy(final String propertyName, final long timeoutMillis) {
-        final String strategy = PropertiesUtil.getProperties().getStringProperty(propertyName, "TIMEOUT");
-        LOGGER.trace("property {}={}", propertyName, strategy);
-        final String strategyUp = strategy.toUpperCase(Locale.ROOT); // TODO Refactor into Strings.toRootUpperCase(String)
-        switch (strategyUp) { // TODO Define a DisruptorWaitStrategy enum?
```
remkop, line 79: This TODO comment ("Define an enum") is misleading. Let's replace this with a comment explaining that we deliberately don't use an enum so that invalid values will result in the default timeoutstrategy being chosen without exceptions being thrown.

(As an aside, I believe the value of defining a private enum is limited anyway. What matters to me is that the valid values are well-documented and that this documentation is easy to find.)

stepan2271:
> I put some comments on the mechanics of the proposed changes, but I actually have a bigger question:
> 
> what problem are we trying to solve?
> Can you explain the use case where it would help to be able to fine-tune sleep time and number of retries?

What I want to achieve is to log messages with low resource consumption and at the same time garbage free (since I have low latency code in the same JVM with no gc pauses allowed). Currently, I don`t see an option that satisfies these requeirements. If I use BlockingWaitStrategy or TimeoutBlockingWaitStrategy then I have garbage. If I use SleepingWaitStrategy with default parameters, YieldingWaitStrategy or BusyspinWaitStrategy then logging consumes resources that I don`t want to allocate there.

If you have any other suggestions on how to handle my use case, I would be happy to use your approach.

remkop:
Understood, thanks for the clarification!

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
stepan2271: Ok. Changed it back to String

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
stepan2271: Ok. TODO removed. Proper comment on deliberate usage is added.

Regrding documentation, I haven`t found any documentation on property related to `timeoutMillis` parameter of `TimeoutBlockingWaitStrategy`. I think the best place is to add documentation on all these three properties (`timeoutMillis`, `sleepTimeNs`, `retries`) to `logging-log4j2/src/site/asciidoc/manual/async.adoc`. Do you think it is worth to do it? And what is the best place then?

remkop:
Thanks for making those changes.

The system properties used in Log4j2 are documented in this section in the Configuration page of the manual: https://github.com/apache/logging-log4j2/blob/master/src/site/asciidoc/manual/configuration.adoc#system-properties

Please document the additional system properties for both the `AsyncLogger` prefix and  the `AsyncLoggerConfig` prefix, to follow the convention of the existing docs.

stepan2271:
> Thanks for making those changes.
> 
> The system properties used in Log4j2 are documented in this section in the Configuration page of the manual: https://github.com/apache/logging-log4j2/blob/master/src/site/asciidoc/manual/configuration.adoc#system-properties
> 
> Please document the additional system properties for both the `AsyncLogger` prefix and the `AsyncLoggerConfig` prefix, to follow the convention of the existing docs.

Thanks for reviewing! Documentation is added to correspondent sections

stepan2271:
Last two builds failed in spite of the fact that my last commits were to documentation only:(

remkop:
Thank you for the contribution, @stepan2271 !

stepan2271:
Thanks, @remkop! Should I do smth to fix build? Seems like the build fails, but it is unrelated to my changes. 

```
[INFO] Results:

[INFO] 

[ERROR] Failures: 

[ERROR]   TimeFilterTest.testTime:199 Time 1588326633252 is not within range: start=1590915600000, end=1590919200000, timezone=America/Los_Angeles expected same:<NEUTRAL> was not:<DENY>

[INFO] 

[ERROR] Tests run: 1837, Failures: 1, Errors: 0, Skipped: 12

[INFO] 
```

stepan2271:
Build is fixed after merge with master. @remkop could you please merge, since I don`t have write access

remkop:
@carterkozak Could you also review this PR?

carterkozak:
Sure thing, I’ll take a look tomorrow :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -60,36 +60,46 @@
...
+                return new YieldingWaitStrategy();
+            case "BLOCK":
+                return new BlockingWaitStrategy();
+            case "BUSYSPIN":
+                return new BusySpinWaitStrategy();
+            case "TIMEOUT":
+                return new TimeoutBlockingWaitStrategy(timeoutMillis, TimeUnit.MILLISECONDS);
+            default:
+                return new TimeoutBlockingWaitStrategy(timeoutMillis, TimeUnit.MILLISECONDS);
+        }
```
carterkozak, line 88: At some point we could allow custom WaitStrategy implementations to be provided, likely most helpful for demonstrating changes like this without waiting a full release cycle. This change doesn't make it any more difficult to implement later on, and in the majority of cases we use simple configurations without custom code. :+1:

carterkozak:
Curious if anyone has suggestions for backporting documentation to the 2.x branch? It's difficult manually transforming between adoc and rst.

PR #363, opened 2020-05-29
LOG4J2-2859 - Fix typo that mergeFactory should be mergeStrategy

quaff:
according to https://github.com/apache/logging-log4j2/blob/d304e034dfd3e3e5ca115b6b6e6f22d72d606850/log4j-core/src/main/java/org/apache/logging/log4j/core/config/composite/CompositeConfiguration.java#L50

### DIFF
--- a/log4j-api/src/test/java/org/apache/logging/log4j/util/LegacyPropertiesCompatibilityTest.java
+++ b/log4j-api/src/test/java/org/apache/logging/log4j/util/LegacyPropertiesCompatibilityTest.java
@@ -39,7 +39,7 @@ public class LegacyPropertiesCompatibilityTest {
     public static Object[][] data() {
         return new Object[][]{
             {"log4j2.configurationFile", "log4j.configurationFile"},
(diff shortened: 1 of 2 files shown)


vy:
@jvz, you appear to be the one who has committed these unit tests in the first place. How come they were not failing given the wrong property was used? (Sorry for the short-sighted conclusion. I have just skimmed through the sources of which I am not much familiar with. I will try to spare time to dig this further, if needed.)

rgoers:
The test is in the API module and is only testing the property name resolution. The CompositeConfiguration is in Core so these properties can't be used for anything in this test.

Matt shows up on this because he went through all of Log4j scrounging up the properties and "normalized" and documented them. He probably made a simple error with this. The fact that the documentation is wrong is worse then the test being wrong. But this fix looks good to me.

jvz:
Yeah, looks like a typo, but the test still works because it was testing the name mangling.

rgoers:
Patch was manually applied. For some reason merging the branch caused test failures in log4j-taglib for me.

PR #366, opened 2020-06-27
Allow clients to set the source manually.

shawjef3:
The purpose of this is to make it possible to improve performance. Getting the StackTraceElement for a message is somewhat expensive. Creating it manually is tedious and no one should do it, but the Scala plugin can provide the source location at compile time. Possibly other tools could do the same thing. This enables that to be possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,23 @@ JIRA issue in the Pull Request.
 
 ## Building From Source
 
-Log4j requires Apache Maven 3.x. To build from source and install to your local Maven repository, execute the following:
+Log4j requires Apache Maven 3.x. Java 9 needs to be installed and present in your `~/.m2/toolchains.xml`. See the following for an example.
+ 
+ ```$xml
+<toolchains>
+	<toolchain>
+    <type>jdk</type>
+    <provides>
+      <version>9</version>
+    </provides>
+    <configuration>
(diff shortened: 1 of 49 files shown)


rgoers:
This change is modifying the signatures of virtually all the Messages, which will break some existing users. What's more, clients who want to set the source manually already can. Just use  one of Logger's "at" methods to get a LogBuilder and then use withLocation(stackTraceElement).

garydgregory:
The breaking could be less using default methods FWIW...

rgoers:
That is true, but it seems odd to add the StackTraceElement to things like traceEntry & traceExit and then ALSO add them to Messages. The location information isn't logically part of a Message.

garydgregory:
Yeah, I do not have a use case for that, I just wanted to mention it for
completeness sake as well as guidance for contributing: don't break binary
compatibility of the public API, also small PRs are better than larger ones.

Gary

rgoers:
@shawjef3 Based on our comments, what do you want to do with this PR?

shawjef3:
@rgoers  thanks for the pointer to `Logger#atLevel`. It looks like that will work for most of it, if not all. I'll have to copy implementations of parts of AbstractLogger, like `throwing`, since I won't be able to use the public method, but it'll work.

PR #367, opened 2020-06-28
avoid initialize 'volatile' fields with default values

turbanoff:
It's known that writing to a  `volatile` field has performance impact.
It's true even if code initializes field with _default_ value.
This PR cleans such redundant initialization and improves startup performance.
 See similar improvement in JDK codebase: https://bugs.openjdk.java.net/browse/JDK-8145680.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocatorUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocatorUtil.java
@@ -26,7 +26,7 @@ import java.util.Stack;
  */
 public final class StackLocatorUtil {
     private static StackLocator stackLocator = null;
-    private static volatile boolean errorLogged = false;
+    private static volatile boolean errorLogged;
 
     static {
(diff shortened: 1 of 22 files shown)


vy:
Hey @turbanoff! Thanks for the heads up and PR, appreciated! I will need some time to wrap my mind around the change set and approve it -- unless some other committer steps forward earlier.

rgoers:
I've looked at the linked documentation and am fine with this in principle. I have not looked to see if leaving any of the fields uninitialized would cause problems, but I find that doubtful as I would expect compile errors if that were the case (although I have not tried to compile it).

carterkozak:
This isn't something I had considered but it makes sense, thanks for the contribution and for teaching me something new! I'll have to look into writing an error-prone plugin to detect and refactor this pattern when I have time :-)

carterkozak:
I've created a ticket for this contribution: https://issues.apache.org/jira/browse/LOG4J2-2898
Merging momentarily.

PR #379, opened 2020-07-01
LOG4J2-2764 Implement CI pipeline via GitHub Actions

vy:
Note that this change set updates the Maven Wrapper scripts too.

An example successful run is accessible [here](/vy/log4j2/actions/runs/153931860).


rgoers:
Assuming you have tested this I don't see any issues.

vy:
@jvz, do you still want to review this or shall I go ahead and merge it?

PR #381, opened 2020-07-02
LOG4J2-2882: Add support for JUL filters

ebourg:
This implements basic support for JUL filters as described in [LOG4J2-2882](https://issues.apache.org/jira/browse/LOG4J2-2882). The log calls are delegated to the super class only when a filter is configured, I hope this preserves the performance in the general case. I left out the tracing methods because I wasn't sure delegating to the super class would generate a message with the right semantic.

### DIFF
--- a/log4j-jul/src/main/java/org/apache/logging/log4j/jul/ApiLogger.java
+++ b/log4j-jul/src/main/java/org/apache/logging/log4j/jul/ApiLogger.java
@@ -119,22 +119,38 @@ public class ApiLogger extends Logger {
 
     @Override
     public void log(final Level level, final String msg) {
-        logger.log(LevelTranslator.toLevel(level), msg);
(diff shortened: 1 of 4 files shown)


rgoers:
Now that I look at this I find what was originally implemented rather odd. The setLevel() and setParent() methods throw exceptions and yet the setFilter method is still allowed. It seems to me if setFilter is supported then setLevel should be too (i.e. it would call setLevel on the underlying Log4j Logger). But if setLevel isn't supported then I don't understand why setFilter would be. In any case, I wouldn't have setLevel throw an exception. At most I would just ignore the call.

ebourg:
I agree that `setLevel()` shouldn't throw an exception. It could maybe just log a one time warning, but that's not worth breaking the application just because log4j was plugged in.

PR #387, opened 2020-07-15
LOG4J2-2867: RingBufferLogEventTranslator uses a static ContextDataInjector

carterkozak:
Previously a new instance was created for each translator, which
occurred on every thread. Some applications do not reuse threads
efficiently, and shouldn't be over-penalized.


carterkozak:
committed to release-2.x and master

PR #388, opened 2020-07-15
LOG4J2-2837: Disruptor and JUL no longer recursively start the AsyncLoggerDisruptor

carterkozak:
Previously logging initialization would result in an AsyncLoggerContext
initializing an AsyncLoggerDisruptor, which would access a logger from
JUL, starting the process again. This change avoids recursive calls with
a state check.


remkop:
Thanks for taking care of this, Carter!

carterkozak:
I appreciate the reviews, thanks @remkop :-)
I still need to finish full test runs locally for these changes, then I'll merge along with #361, likely tomorrow.

remkop:
If you can merge  #361, that would be great! Thank you!

carterkozak:
committed to release-2.x and master

PR #394, opened 2020-08-05
LOG4J2-2892 - Add support for newline delimited messages in GelfLayout

jakluk:
A pull request for issue [https://issues.apache.org/jira/browse/LOG4J2-2892](https://issues.apache.org/jira/browse/LOG4J2-2892).

This PR adds ability to delimit Graylog Extended Log Format (GELF) messages with a newline. It can be useful when you want to append messages in the Graylog format into a file, or when you have to feed the messages into a consumer which supports only newlines as a delimiter. The documentation of GELF only specifies usage of a NULL delimiter and doesn't discourage from using some other. It also requires that newlines inside the message body must be escaped, so multi-line stack traces shouldn't be an issue.

New optional boolean parameter "includeNewLineDelimiter" is introduced to GelfLayout, nothing changes for current users if they decide not to use it.


jakluk:
@rgoers Thank you for the quick review. There was no test whether the null byte delimiter is present so I've added that one as well.

PR #396, opened 2020-08-10
Load pre-loaded lookups via SPI, rather than hard-code in Interpolator.

tbwork:
issue link: [Provide a way to add customized StrLookUps in Interpolator initialization.](https://issues.apache.org/jira/projects/LOG4J2/issues/LOG4J2-2904?filter=reportedbyme)


tbwork:
What's wrong with the CI? The change should have nothing to do with the error. :)

jvz:
No worries about the Travis check. The GitHub Actions checks are the important ones. We need to disable our old Travis checks.

rgoers:
The only concern I have with this is the overhead of ServiceLoader during startup. I recently had to make ServiceLoader run asynchronously during startup while loading Custom ThreadContextDataProviders. This will likely cause the same problem. We should make that more generic so that components such as them can register their service for asynchronous initialization. But that would be done in a different PR.

rgoers:
@jvz I dislike ServiceLookup as it is describing its implementation method, not its purpose.

tbwork:
@rgoers Agree! This PR is intended to do that as you proposed. Hope this could be supported ASAP :) . Have a nice day.

tbwork:
@rgoers Do you mean all the pre-loaded Lookups should be loaded via ServiceLoader? I am glad to contribute to this part if you do not mind. :)

rgoers:
Yes, that is exactly what I mean. Feel free to modify this PR to do that.

tbwork:
@rgoers ok.

tbwork:
@jvz @rgoers It seems the Log4jPlugins class is moved from org.apache.logging.log4j.core.plugins. However in Activator class, Log4jPlugins is still imported as "org.apache.logging.log4j.core.plugins.Log4jPlugins".

tbwork:
@rgoers @jvz Hi, are you busy? Could you get round to take a review and merge :).

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/JndiLookup.java
```diff
@@ -40,6 +39,24 @@
     /** JNDI resource path prefix used in a J2EE container */
     static final String CONTAINER_JNDI_RESOURCE_PATH_PREFIX = "java:comp/env/";
 
+    public JndiLookup(){
+        // JNDI check
+        try {
+            // [LOG4J2-703] We might be on Android
+            Loader.newCheckedInstanceOf("org.apache.logging.log4j.core.lookup.JndiLookup", StrLookup.class);
```
jvz, line 46: This is trying to load itself here?

log4j-docker/src/main/java/org/apache/logging/log4j/docker/DockerLookup.java
```diff
@@ -46,6 +46,15 @@
     private final Container container;
 
     public DockerLookup() {
+
+        try {
+            // docker check.
+            Loader.newCheckedInstanceOf("org.apache.logging.log4j.docker.DockerLookup", StrLookup.class);
```
jvz, line 52: These checks are attempting to load the class they're in. They won't load properly when `SystemLoader.load()` tries to load the class itself due to the static references to optional classpath code. This check you've added is only useful for code outside this exact file. The same goes for the other classes you've added these to.

log4j-kubernetes/src/main/java/org/apache/logging/log4j/kubernetes/KubernetesLookup.java
```diff
@@ -63,16 +63,29 @@
...
+        return BuiltInLookupCategory.KUBERNETES;
     }
 
     KubernetesLookup(Pod pod, Namespace namespace, URL masterUrl) {
         this.pod = pod;
         this.namespace = namespace;
         this.masterUrl = masterUrl;
+
+        try {
+            Loader.newCheckedInstanceOf("org.apache.logging.log4j.kubernetes.KubernetesLookup", StrLookup.class);
```
jvz, line 80: Same here.

log4j-spring-cloud-config/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/SpringLookup.java
```diff
@@ -40,9 +40,21 @@
     private static final Pattern DEFAULT_PATTERN = Pattern.compile(DEFAULT + PATTERN);
 
     public SpringLookup() {
+        try {
+            //Spring check.
+            Loader.newCheckedInstanceOf("org.apache.logging.log4j.spring.cloud.config.client.SpringLookup", StrLookup.class);
```
jvz, line 45: Same here.

log4j-web/src/main/java/org/apache/logging/log4j/web/WebLookup.java
```diff
@@ -43,6 +48,30 @@
...
     private static final String INIT_PARAM_PREFIX = "initParam.";
 
+    private static final Logger LOGGER = StatusLogger.getLogger();
+
+
+    public  WebLookup(){
+        super();
+        if (Constants.IS_WEB_APP) {
+            try {
+                Loader.newCheckedInstanceOf("org.apache.logging.log4j.web.WebLookup", StrLookup.class);
```
jvz, line 58: Same here.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/JmxRuntimeInputArgumentsLookup.java
```diff
@@ -20,33 +20,51 @@
...
+
+    public static final JmxRuntimeInputArgumentsLookup JMX_SINGLETON;
+
+
     public JmxRuntimeInputArgumentsLookup(final Map<String, String> map) {
         super(map);
+        // JMX check
+        try {
+            // We might be on Android
+            Loader.newCheckedInstanceOf("org.apache.logging.log4j.core.lookup.JmxRuntimeInputArgumentsLookup",
```
jvz, line 61: Same here.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/Interpolator.java
```diff
@@ -39,157 +32,95 @@
...
+
+    /**
+     * Pure lookups are used to resolve variables without prefix.
+     * @since 3.0.0
+     */
+    private final Set<AbstractPureLookup> pureLookups = new TreeSet<AbstractPureLookup>(
+            /*
+             *   Comparator for ordering pure StrLookup instances by priority.
+             */
+            new Comparator<AbstractPureLookup>(){
```
jvz, line 50: This should use a lambda.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/Interpolator.java
```diff
@@ -16,20 +16,13 @@
  */
 package org.apache.logging.log4j.core.lookup;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
jvz, line 19: No star imports please!

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -52,10 +52,7 @@
 import org.apache.logging.log4j.core.config.plugins.util.PluginBuilder;
 import org.apache.logging.log4j.core.filter.AbstractFilterable;
 import org.apache.logging.log4j.core.layout.PatternLayout;
-import org.apache.logging.log4j.core.lookup.Interpolator;
-import org.apache.logging.log4j.core.lookup.MapLookup;
-import org.apache.logging.log4j.core.lookup.StrLookup;
-import org.apache.logging.log4j.core.lookup.StrSubstitutor;
+import org.apache.logging.log4j.core.lookup.*;
```
jvz, line 55: No star imports please!

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/AbstractPureLookup.java
```diff
@@ -0,0 +1,33 @@
+package org.apache.logging.log4j.core.lookup;
+
+
+/**
+ *
+ * Use to resolve those variables without prefix.
+ *
+ * @author tommy.tesla
```
jvz, line 8: Please don't use the `@author` tag in javadoc.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/BuiltInLookupCategory.java
```diff
@@ -0,0 +1,116 @@
+package org.apache.logging.log4j.core.lookup;
+
+/**
+ * Built-in category names' home.
+ *
+ * @author tommy.testla
+ * @since 3.0.0
+ */
+public interface BuiltInLookupCategory {
```
jvz, line 9: Two things about this:

1. We do not use interfaces for holding string constants; we use a `public final class` with `public static final` constants and a private constructor. Using interfaces this way is an anti-pattern.
2. Due to how you've refactored these to be dynamically looked up via services, these constants are only referenced in their respective classes now. There's no reason to factor out the constants at this point.

rgoers:
These changes are not what I expected. All the Lookups should remain plugins. You should have only had to modify Interpolator to load the required lookups via ServiceLoader and then added the Lookups you wanted to load to the file used by ServiceLoader.  The Lookups would be loaded again as normal as Plugins during configuration.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/JndiLookup.java
tbwork: Hah, my fault. Your r right. How about just removing them? It is not necessary in terms of SPI mode.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/Interpolator.java
tbwork: Rogar that.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/Interpolator.java
tbwork: Rogar that.

tbwork:
@rgoers Hi, rgoers. It seems my first pull request was what you want? Here is my plan:
1. Keep all the old lookup loaded as plugins.
2. Provide a SPI mode for custom lookup.

Is it right? Or what do you suggest in detail?

tbwork:
@rgoers I propably misunderstood your opinion:

 **_In essence, StrLookup would become the service interface and every plugin to be pre-loaded would be listed in the services file instead of hard coded in Interpolator._**

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/BuiltInLookupCategory.java
tbwork: Ok, got it.

tbwork:
@rgoers Hi, rgoers. How to understand your comments:

 **_The Lookups would be loaded again as normal as Plugins during configuration._**

If all pre-loaded lookups are already loaded, why those lookups still need a reload as plugins?

I suggest to separete these two types of lookups: lookups loaded by SPI  and lookups loaded by plugins. Please see the new codes and free feel to give some advices :)

rgoers:
Well, the Interpolator constructor that has the hard-coded list of lookups is initializing the lookups that could be using while constructing the configuration. The other lookups would be used at runtime. There is no reason that I can think of that all the Lookups couldn't be initialized once via ServiceLoader except that it would break compatibility with users who have created their own Lookups as plugins. We could possibly load  the serviceloader Lookups once and then add the ones defined as plugins later.

ppkarwasz:

PR #397, opened 2020-08-11
[LOG4J2-2906] - Fix Log4j2 cloud logging system config loading

sjrd218:
Wrap the list coming from the getStandardConfigLocations in a list that can be added to.


rgoers:
Patch has been applied but it was done manually since the affected code moved to another branch.

PR #401, opened 2020-08-16
run CI build for pull requests

sullis:


jvz:
Is this PR only being built then because of the config update here?

BTW, our workflow is broken at the moment anyways. Hopefully we can figure it out soon, but something seems to have changed in the setup java action which isn't using the same environment variables anymore.

jvz:
In the meantime, I've enabled Jenkins builds for PRs from collaborators at least.

sullis:
> Is this PR only being built then because of the config update here?
> 

Yes.

jvz:
I noticed one potential issue here which I'm not sure how to configure: PRs that are created from branches on origin (e.g., how Dependabot is configured) end up being built twice due to being pushed to a branch and a PR.

PR #402, opened 2020-08-16
Added two articles recently published

fmarchioni:
I've added two fresh articles based on the latest releases of WildFly application Server and Spring Boot 2.

### DIFF
--- a/src/site/asciidoc/articles.adoc
+++ b/src/site/asciidoc/articles.adoc
@@ -43,6 +43,10 @@ guide for up-to-date and detailed information on how to configure and use Log4j 
 
 == English
 
+* http://www.mastertheboss.com/jboss-server/jboss-log/how-to-use-log4j2-in-your-wildfly-applications[How to configure Log4j2 in your WildFly applications]
+(August 16, 2020)
+* http://www.masterspringboot.com/configuration/logging/configuring-log4j2-in-spring-boot-applications[Configuring Log4j2 in a Spring Boot 2 application]
+(August 16, 2020)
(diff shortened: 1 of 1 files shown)


rgoers:
A couple of comments on the Srping Boot article:
1. In the first two paragraphs you say "new" in each. The first commit for Log4j 2 was 10 years ago and the first GA release was 6 years ago. I'd hardly call that new.
2. You use the phrase "substantially based on Logback". That is quite incorrect and would violate Logback's licensing were it true.  Logback originated from Log4j 1. Log4j 2 took ideas from Logback and Log4j 1 but was almost entirely new code. 
3. While using spring-boot-starter-log4j2 is an option I recommend importing log4j-bom instead. That way you control the version of Log4j 2 you are using.
4. Log4j provides direct Spring Boot support. In 2.13.3 it is included in the log4j-spring-cloud-config module. The code that only requires Spring Boot is moving to a new log4j-spring-boot module in 2.14.0.
5. It is worth looking into the support the log4j-spring-cloud-config module provides as it supports placing your log4j2.xml file in Spring Cloud config where it can be shared by all the instances of your application dynamically updating them all when it changes.

fmarchioni:
Thanks for taking the time to review the article.
I've fixed point 1) and 2)
Also for point 3) I've added a section to show how to use log4j-bom both both for Spring Boot and WildFly article.
About 4) and 5) that's a great suggestion. Being more specific to spring cloud, I'd like to discuss it in a new article which focuses on spring cloud and log4j.

rgoers:
Thanks for making the changes.

PR #425, opened 2020-09-21
Update KafkaManager.java

wuqian0808:
Avoid kafka thread leak during loggercontext reconfigure multiple times.

### DIFF
--- a/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaManager.java
+++ b/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaManager.java
@@ -154,10 +154,9 @@ public class KafkaManager extends AbstractManager {
 
     public void startup() {
         // Avoid kafka thread leak
-        if (producer != null){
-            producer.close();
+        if (producer == null){
+            producer = producerFactory.newKafkaProducer(config);
         }
-        producer = producerFactory.newKafkaProducer(config);
     }
 
     public String getTopic() {
--- a/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaManager.java
+++ b/log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaManager.java
@@ -153,6 +153,10 @@ public class KafkaManager extends AbstractManager {
     }
 
     public void startup() {
+        // Avoid kafka thread leak
+        if (producer != null){
+            producer.close();
+        }
         producer = producerFactory.newKafkaProducer(config);
     }
 


wuqian0808:
Configure kafkaappender in log4j2.xml, then call below steps serveral times. The kafka thread will increase.         LoggerContext.getContext(false).reconfigure();
        LoggerContext.getContext(false).reconfigure();
        LoggerContext.getContext(false).reconfigure();

gengyuanzhe:
In my opition, `org.apache.logging.log4j.kafka.appender.KafkaManager#releaseSub` should close the kafka producer when `reconfigure()`. The `close()` should not be invoked in`startup ()`, it is confused.

wuqian0808:
> In my opition, `org.apache.logging.log4j.kafka.appender.KafkaManager#releaseSub` should close the kafka producer when `reconfigure()`. The `close()` should not be invoked in`startup()`, it is confused.

wuqian0808:
> In my opition, `org.apache.logging.log4j.kafka.appender.KafkaManager#releaseSub` should close the kafka producer when `reconfigure()`. The `close()` should not be invoked in`startup()`, it is confused.

gengyuanzhe:
I tried your case, and you are right, the KafkaAppender do have the problem of thead leak. But **close when startup** is not the correct way  to solve this problem. Try to find why the **count in abstractmanager** increase each time when reconfigure(), fix it and make sure the kafkaManager close when prev.stop(). You'd better open an [issue](https://issues.apache.org/jira/browse/LOG4J2) to track this problem. And if you can provide unit test that failed without your PR, and success with your PR, it will help a lot to make this PR  accepted ASAP.

wuqian0808:
> I tried your case, and you are right, the KafkaAppender do have the problem of thead leak. But **close when startup** is not the correct way to solve this problem. Try to find why the **count in abstractmanager** increase each time when reconfigure(), fix it and make sure the kafkaManager close when prev.stop(). You'd better open an [issue](https://issues.apache.org/jira/browse/LOG4J2) to track this problem. And if you can provide unit test that failed without your PR, and success with your PR, it will help a lot to make this PR accepted ASAP.

Sure. I will find a better way to fix this issue

wuqian0808:
> I tried your case, and you are right, the KafkaAppender do have the problem of thead leak. But **close when startup** is not the correct way to solve this problem. Try to find why the **count in abstractmanager** increase each time when reconfigure(), fix it and make sure the kafkaManager close when prev.stop(). You'd better open an [issue](https://issues.apache.org/jira/browse/LOG4J2) to track this problem. And if you can provide unit test that failed without your PR, and success with your PR, it will help a lot to make this PR accepted ASAP.

PR #426, opened 2020-09-25
Avoid kafka thread leak - LOG4J2-2916

wuqian0808:


wuqian0808:

wuqian0808:
open another request with testcase

PR #427, opened 2020-09-28
Add logic to detect kafka broker is valid or not

wuqian0808:
Add logic to detect kafka broker is valid or not during app startup.
User should be noticed with kafka broker available or not.
Mark kafka appenders as not disabled if kafka broker not available. Then Logevents won't be stored in this appender.
Otherwise, the kafka producer will retry unlimited with lots of log. If user didn't add blocking parameter to asyncappender, all appenders in this logger will be affected.


wuqian0808:
open with antoher pull request

PR #428, opened 2020-09-28
LOG4J2-2916 Avoid redundant Kafka producer instantiation causing thread leaks.

wuqian0808:


wuqian0808:
Testcase added. Is there anything else I can do?

wuqian0808:
@jvz  kindly help to advise

jvz:
@wuqian0808 could you fix the filename of the test file? Without the .java extension, I'm pretty sure the compiler or Maven will complain.

wuqian0808:
> @wuqian0808 could you fix the filename of the test file? Without the .java extension, I'm pretty sure the compiler or Maven will complain.

File name changed

jvz (approved):
Test failures appear to be unrelated. Looks good to me!

wuqian0808:
@rgoers Could u help to advise? Or is there anything else should I do?

wuqian0808hundsun:
@vy  Could u help to advise if there is anything else should I do before merging?

log4j-kafka/src/test/java/org/apache/logging/log4j/kafka/appender/KafkaThreadTest.java
```diff
@@ -0,0 +1,33 @@
...
+        int kafkaThreadCount = 0;
+        if (maps != null && maps.size() >0) {
+            for (Thread thread : maps.keySet()) {
+                if (thread.getName().startsWith("kafka-producer")){
+                    kafkaThreadCount++;
+                }
+            }
+        }
+        assertTrue(kafkaThreadCount <= 5);
+    }
```
vy, line 31: Magical numbers in the tests, e.g., 5 here, have strong potential to bite us in the back later on. Would you mind revising this test as follows, please?

```java
@Test
void test_kafka_producer_threads_are_not_leaking_after_context_restart() {

    // Determine the initial number of threads.
    final LoggerContext context = ctx.getLoggerContext();
    final int initialThreadCount = kafkaProducerThreadCount();

    // Perform context restarts.
    final int contextRestartCount = 3;
    for (int i = 0; i < contextRestartCount; i++) {
        context.getContext(false).reconfigure();
    }

    // Verify the final thread count.
    final int lastThreadCount = kafkaProducerThreadCount();
    assertThat(lastThreadCount).equalsTo(initialThreadCount);

}

private static int kafkaProducerThreadCount() {
    // Return the number of threads prefixed with "kafka-producer".
}
```
Note that,
- This is sort of a pseudo-code I jotted in the comment box, hence you need to fill it in yourself.
- Try to use JUnit 5, if possible. (You can see examples in the source code.)

vy (changes requested):
Looks really nice! Would you also mind updating `changes.xml` too, please?

vy:
Nevermind, applied your changes manually. Will backport it to 2.x too. Thanks so much!

PR #430, opened 2020-10-05
Fix NPE in MDCContextMap

SiegfriedTobias1:
PR for https://issues.apache.org/jira/browse/LOG4J2-2939
Accomodate for the fact that MDC.getCopyOfContextMap() may return null.

### DIFF
--- a/log4j-to-slf4j/src/main/java/org/apache/logging/slf4j/MDCContextMap.java
+++ b/log4j-to-slf4j/src/main/java/org/apache/logging/slf4j/MDCContextMap.java
@@ -71,7 +71,8 @@ public class MDCContextMap implements CleanableThreadContextMap {
 
     @Override
     public boolean containsKey(final String key) {
-        return MDC.getCopyOfContextMap().containsKey(key);
+        Map<String, String> map = MDC.getCopyOfContextMap();
+        return map != null && map.containsKey(key);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


carterkozak (commented):
Thanks, the change looks good. Would you mind adding a test that fails without the fix to prevent regressions?

carterkozak:
Thanks! I'll handle the merge and cherry-pick to relevant release branches.

SiegfriedTobias1:
Thanks for the review. I added a test, as requested.

log4j-to-slf4j/src/test/java/org/apache/logging/slf4j/LoggerTest.java
```diff
@@ -165,5 +166,12 @@ public void mdc() {
         assertThat(list.strList, hasSize(2));
         assertTrue("Incorrect year", list.strList.get(0).startsWith("2010"));
     }
+
+    @Test
+    public void mdcNullBacked() {
+        assertNull("Setup wrong", MDC.getCopyOfContextMap());
+        assertTrue(ThreadContext.isEmpty());
+        assertFalse(ThreadContext.containsKey("something"));
+    }
```
garydgregory, line 175: I would separate these into two tests, one for each API.

garydgregory: An edge case worth testing IMO is getting a null key.

SiegfriedTobias1: Thank you for the feedback! I split the tests and added some tests concerning null keys.

PR #431, opened 2020-10-06
Fix nested project spotbugs configuration

carterkozak:


carterkozak:
Running here to trigger CI before committing to the release branch.

PR #434, opened 2020-10-30
LOG4J2-2954 Retain strong reference to shutdown callbacks

henryptung:
Note: Made two attempts to fix this, but the latter seems a little better. Changes:

- Change the ShutdownCallbackRegistry API back to "register it, we'll call it" - no special GC semantics to worry about, and no need to fix random other uses of the API.
- Move the leak-avoidance Reference semantics out of ShutdownCallbackRegistry and into LoggerContext itself. Since the SoftReference is now to the LoggerContext (rather than an inline lambda), the callback will stay "alive" as long as the LoggerContext itself does.

This seems like a much more sustainable approach to the leak problem - API stays intuitive, and those who need GC guards can implement it themselves as needed, following this pattern and choosing which key objects to reference.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/DefaultShutdownCallbackRegistry.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/DefaultShutdownCallbackRegistry.java
@@ -45,7 +45,9 @@ public class DefaultShutdownCallbackRegistry implements ShutdownCallbackRegistry
 
     private final AtomicReference<State> state = new AtomicReference<>(State.INITIALIZED);
     private final ThreadFactory threadFactory;
-    private final Collection<Cancellable> hooks = new CopyOnWriteArrayList<>();
+
+    // use references to prevent memory leaks
+    private final Collection<Reference<Cancellable>> hooks = new CopyOnWriteArrayList<>();
     private Reference<Thread> shutdownHookRef;
 
     /**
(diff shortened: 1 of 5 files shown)


carterkozak:
While I agree that the change to registering a runnable without requiring the caller holds a reference to the acknowledgement is easier to understand, I'm a little worried that this change could impact third party components in unexpected ways (cause memory leaks), so we may want to fix the regression without changing that behavior of `ShutdownCallbackRegistry`.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/DefaultShutdownCallbackRegistry.java
```diff
@@ -87,34 +86,33 @@ public void run() {
...
     private static class RegisteredCancellable implements Cancellable {
-        // use a reference to prevent memory leaks
-        private final Reference<Runnable> hook;
+        private Runnable callback;
         private Collection<Cancellable> registered;
 
         RegisteredCancellable(final Runnable callback, final Collection<Cancellable> registered) {
+            this.callback = callback;
             this.registered = registered;
-            hook = new SoftReference<>(callback);
```
carterkozak, line 96: I'm not sure why this used a `SoftReference` rather than `WeakReference` since the former is notorious for GC problems, and I don't think we want to hold the callback reference until the heap is otherwise full.

src/changes/changes.xml
```diff
@@ -274,6 +274,9 @@
       <action issue="LOG4J2-2939" dev="ckozak" type="fix" due-to="Constantin Hirsch">
         Fix NPE in MDCContextMap on 'contains' and 'isEmpty' invocations.
       </action>
+      <action issue="LOG4J2-2954" dev="htung" type="fix">
```
carterkozak, line 277: Sorry, this part is a little counterintuitive:
```suggestion
      <action issue="LOG4J2-2954" dev="dev who approves and merges the change" type="fix" due-to="Henry Tung">
```

log4j-core/src/main/java/org/apache/logging/log4j/core/util/DefaultShutdownCallbackRegistry.java
```diff
@@ -87,42 +92,41 @@ public void run() {
...
         }
 
         @Override
         public void cancel() {
-            hook.clear();
-            registered.remove(this);
+            callback = null;
+            registered.removeIf(ref -> ref.get() == this);
```
carterkozak: Might as well include a null check as well to remove reference objects whose target has been garbage collected while we're at it.

jvz:
From what I remember, the hard part here was making sure any default shutdown callbacks registered in web apps didn't cause classloader leaks. The shutdown hook is a fun place to debug things since not necessarily all the classes in log4j are loaded at that point (or may be unloaded already if not strongly referenced somewhere), so it's mostly just finding the correct boundary point to prevent memory leaks at both the classpath level as well as stronger classloaders like in servlets.

carterkozak:
I think the difference between the first and second commit on this PR is whether or not the JeroMqManager class (and related context classes) should be allowed to be unloaded if they're only referenced by the shutdown hook. I'm' not sure whether or not that's important.

carterkozak:
I think the approach in the first commit is less risky, so I'm going to merge that after I verify tests pass locally to ensure this makes the 2.14.0 release.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/DefaultShutdownCallbackRegistry.java
carterkozak: I'm going to update this to include the null check to remove collected values, and add a null check on registered so that subsequent `cancel()` invocations do not NPE.

rgoers:
@carterkozak Are you doing something with this?

PR #438, opened 2020-11-15
LOG4J2-2963: Relax access level for code reuse

annakukliansky:


garydgregory:
I'm personally not a fan of opening up the guts of these classes for public consumption, it just handcuffs us to provide binary compatibility. 

I'd think it is better for users to provide their own clean implementations of messages. Reuse by inheritance is brittle.

vy:
I agree with @garydgregory here. @annakukliansky, would you share your use case, please? If you elaborate on what you are trying to achieve, maybe we can point to a better alternative.

vy:
Closing due to inactivity and insufficient rationale.

PR #440, opened 2020-11-17
LOG4J2-2964 Merge packages from several Configurations in Composite Configuration

valfirst:


vy:
@valfirst, thanks so much for the PR!

PR #441, opened 2020-11-17
LOG4J2-2962 Enrich "map" resolver by unifying its backend with "mdc" resolver.

vy:


log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/ReadOnlyStringMapResolver.java
```diff
@@ -0,0 +1,386 @@
...
+                final ReadOnlyStringMap map = mapAccessor.apply(logEvent);
+                return map != null && map.containsKey(key);
+            }
+
+            @Override
+            public void resolve(final LogEvent logEvent, final JsonWriter jsonWriter) {
+                final ReadOnlyStringMap map = mapAccessor.apply(logEvent);
+                final Object value = map == null ? null : map.getValue(key);
+                if (stringified) {
+                    final String valueString = String.valueOf(value);
```
carterkozak, line 204: I realize this code has just moved and isn't new in this commit, but we might want to use something like `ParameterizedMessage.deepToString(value)` for better handling of standard types, and exception handling in the case toString throws.

rgoers (approved):
This all looks good to me.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/ReadOnlyStringMapResolver.java
vy: Good catch @carterkozak! I did not know about `ParameterizedMessage.deepToString(value)`. I will resolve this conversation to get this PR merged and address your remark in a separate ticket, [LOG4J2-2966](https://issues.apache.org/jira/browse/LOG4J2-2966).

PR #443, opened 2020-11-29
LOG4J2-2967 Fix JTL parameter resolution for messages with too few parameters.

carterkozak:
When messages don't contain enough parameters, index-based lookups
result in a JSON null token rather than failing to produce a log event.

This change also modifies behavior of non-index-based resolution
when messages have no parameters. Previously the value would be null,
now we produce an empty list. Note that this is a change for only
ParameterizedMessage, not for messages which implement
ParameterVisitable.


carterkozak:
@vy I'd appreciate your review on this at your convenience :-)

vy:
@carterkozak Great catch with a complete fix! Thanks so much for sparing time!

PR #445, opened 2020-12-06
slf4j 1.7.30

sullis:
slf4j 1.7.30 was released in December 2019

https://github.com/qos-ch/slf4j/releases


rgoers:
Updating the parent pom version is insufficient. Log4j-slf4j-impl provides support for the EventData object that was removed in SLF4J 1.7.25. Log4j-slf4j-impl also specifies 1.7.25 as its version, so changing the parent pom probably won't accomplish what you want. Upgrading the version is probably fine in master, but that requires the EventData support be removed. We do not want to upgrade the release-2.x branch since that will break backward compatibility. However, user's can certainly upgrade to a more recent version of SLF4J (although it appears SLF4J is apparently no longer being maintained) so long as they don't require the EventData support.

sullis:
obsolete.

PR #450, opened 2020-12-24
Document how to disable message pattern lookups

quaff:
It's worth to add a section in manual, I've struggled this problem.

### DIFF
--- a/src/site/asciidoc/manual/configuration.adoc
+++ b/src/site/asciidoc/manual/configuration.adoc
@@ -1146,6 +1146,13 @@ included, such as ``${main:\--file:-app.properties}`. This would use the
 `MainMapLookup` for a key named `--file`. If the key is not found then
 <code>app.properties</code> would be used as the default value.
 
+[#DisablingMessagePatternLookups]
+== Disables Message Pattern Lookups
+A message is processed (by default) by lookups, for example if you defined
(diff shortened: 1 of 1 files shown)


src/site/asciidoc/manual/configuration.adoc
```diff
@@ -1146,6 +1146,13 @@ included, such as ``${main:\--file:-app.properties}`. This would use the
 `MainMapLookup` for a key named `--file`. If the key is not found then
 <code>app.properties</code> would be used as the default value.
 
+[#DisablesMessagePatternLookups]
```
garydgregory, line 1149: DisablesMessagePatternLookups -> DisablingMessagePatternLookups

src/site/asciidoc/manual/configuration.adoc
```diff
@@ -1146,6 +1146,13 @@ included, such as ``${main:\--file:-app.properties}`. This would use the
...
+== Disables Message Pattern Lookups
+Message is processed (by default) by lookups, for example if you defined
```
garydgregory, line 1151: Message is processed (by default)
->
A message is processed

garydgregory: Actually it might be simpler for a native English speaker to rewrite the whole thing instead of editing here which is awkward.

quaff: @garydgregory You could merge this PR then polish it.

PR #452, opened 2020-12-28
LOG4J2-2972 Respawn AsyncAppender thread on failures.

vy:


carterkozak:
I think the only way this can happen is if someone calls [Thread.stop()](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#stop--), so we must consider the intent when that occurs. Is it accidental, where the asyncappender interaction is collateral damage, or is the system attempting to stop all threads? In the former case we may want to avoid breaking the logger, but in the latter we would not want to restart the background thread.

If we're restarting the background thread, why not simply catch ThreadDeath to avoid thread churn and potential bugs that result from the indirection?

Based on the original request when we began catching Error, I don't think the caller was stopping background threads (or using asynchronous logging at all) but rather calling Thread.stop on an application thread that happened to log. Using synchronous logging it would swallow the ThreadDeath error silently and allow the application thread to continue. I think we can allow ThreadDeath to be propagated, and even kill background threads -- if that's what the user wants and the security manager allows it, on their own heads be it!

What do you think?

Also note that as written this PR only applies to the AsyncAppender, not the disruptor based fully async AsyncLoggerContextSelector or mixed sync/async logging.

vy:
> I think the only way this can happen is if someone calls [Thread.stop()](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#stop--), so we must consider the intent when that occurs. Is it accidental, where the asyncappender interaction is collateral damage, or is the system attempting to stop all threads? In the former case we may want to avoid breaking the logger, but in the latter we would not want to restart the background thread.

The use case is reported [here](/apache/logging-log4j2/commit/56436ad2176eac000d2821690e4373f097b76670#r44892412).

> If we're restarting the background thread, why not simply catch ThreadDeath to avoid thread churn and potential bugs that result from the indirection?
> 
> Based on the original request when we began catching Error, I don't think the caller was stopping background threads (or using asynchronous logging at all) but rather calling Thread.stop on an application thread that happened to log. Using synchronous logging it would swallow the ThreadDeath error silently and allow the application thread to continue. I think we can allow ThreadDeath to be propagated, and even kill background threads -- if that's what the user wants and the security manager allows it, on their own heads be it!

The major improvement this change set brings is that there are no smart whitelist of failures that are subject to handling. Anything above an `Exception`, we will not catch, which is the best practice while dealing with `Throwable`s. At the beginning, we were just catching `Exception`s. Then we figured there are certain `Error`s (e.g., `ExceptionInInitializerError`) that are harmless. Consequently we extended the catch block to `Throwable`s. Then somebody reported we shouldn't catch `ThreadDeath`. Evidently, we can't oversee all the potential corner cases while dealing with exceptions. Hence, in this patch, I just catch `Exception`s and respawn the thread on unexpected failures. I think, this is a way simpler and future-proof approach.

> Also note that as written this PR only applies to the AsyncAppender, not the disruptor based fully async AsyncLoggerContextSelector or mixed sync/async logging.

If we agree on the strategy employed here, I can extend it to wherever applicable. (The changes need to be duplicated on `master` anyway.)

carterkozak:
> The major improvement this change set brings is that there are no smart whitelist of failures that are subject to handling.

The downside is when an error is raised, we pay the full cost of thread creation and context switching rather than catching and continuing.

Can we compare this solution with a potential alternative:
The `AppenderControl` is reverted to catch `Exception`, not `Throwable`/`Error`, however `AsyncAppender.AsyncThread.run()` is updated to catch `Throwable` in both `callAppenders` and the `errorAppender.callAppender` try/catch? I believe this would match how the disruptor-based `AsyncContextSelector` implementation works already.

vy:
> The downside is when an error is raised, we pay the full cost of thread creation and context switching rather than catching and continuing.

Correct. Though note that this is not worsening the situation compared to 2.13.0 and earlier. Up until 2.13.0, we were catching `RuntimeException` and `Exception`s. In 2.14.0, we have replaced these two catch blocks with a single `Throwable` catch. Then the hell broke loose.

> Can we compare this solution with a potential alternative:
> The `AppenderControl` is reverted to catch `Exception`, not `Throwable`/`Error`, however `AsyncAppender.AsyncThread.run()` is updated to catch `Throwable` in both `callAppenders` and the `errorAppender.callAppender` try/catch?

How are we gonna handle `ThreadDeath` in this case?

In summary, I am really reluctant to add any kind of `catch (Throwable ...)` clauses.

> I believe this would match how the disruptor-based `AsyncContextSelector` implementation works already.

Any code pointers?

carterkozak:
> Then the hell broke loose.

I don't think it's that bad! We fixed a problem for some users, and caused an issue for others. I'd consider the behavior in 2.14 an improvement over previous releases, but I agree that we should allow ThreadDeath to propagate in one way or another.

> How are we gonna handle ThreadDeath in this case?

In my proposal above, the AsyncAppender would catch it and warn/ignore as it catches all throwables, but the AppenderControl would allow ThreadDeath to propagate, bringing back 2.13.x behavior. The difference is that the AppenderControl can throw errors again, but background threads must be robust against them. This PR makes the AsyncAppender robust against Error by transitioning to another thread, but I'm not sure why we need to change threads instead of allowing the original thread to continue.

> In summary, I am really reluctant to add any kind of `catch (Throwable ...)` clauses.

I think the reason this strikes me as odd is that this implementation is already using a `catch (Throwable ...)`, but without writing it directly. We're using an uncaught exception handler to catch `Throwable` and handle it by logging to the statuslogger, and start another thread. If we explicitly write the try/catch instead, we can produce identical behavior without creating new thread

> Any code pointers?

The `catch Throwable` occurs outside log4j2, in the disruptor implementation:

https://github.com/LMAX-Exchange/disruptor/blob/c1ea0d4a6d6144d338a9ff8cb0d46f2a7038b968/src/main/java/com/lmax/disruptor/BatchEventProcessor.java#L185-L190

I hope I'm not coming off too strong here, I really appreciate the work you've put into fixing this bug!

vy:
> I hope I'm not coming off too strong here

Not at all! Somebody is thinking along with me free of charge! What else I can ask for? I am really grateful that you take time for this. :bow:

> In my proposal above, the AsyncAppender would catch it and warn/ignore as it catches all throwables, but the AppenderControl would allow ThreadDeath to propagate, bringing back 2.13.x behavior. The difference is that the AppenderControl can throw errors again, but background threads must be robust against them.

But once `AsyncAppender` catches the `ThreadDeath`, it is basically dead. Who is gonna restart that thread?

> This PR makes the AsyncAppender robust against Error by transitioning to another thread, but I'm not sure why we need to change threads instead of allowing the original thread to continue.

Because the original thread is dead as indicated by `ThreadDeath`?

> > In summary, I am really reluctant to add any kind of `catch (Throwable ...)` clauses.
> 
> I think the reason this strikes me as odd is that this implementation is already using a `catch (Throwable ...)`, but without writing it directly. We're using an uncaught exception handler to catch `Throwable` and handle it by logging to the statuslogger, and start another thread. If we explicitly write the try/catch instead, we can produce identical behavior without creating new thread

Indeed the code is still possessing a `catch (Throwable ...)` behavior, but we delegate handling of non-`Exception` throwables to the JDK, rather than putting a plaster to every new hole we discover. 

> The `catch Throwable` occurs outside log4j2, in the disruptor implementation:
> 
> https://github.com/LMAX-Exchange/disruptor/blob/c1ea0d4a6d6144d338a9ff8cb0d46f2a7038b968/src/main/java/com/lmax/disruptor/BatchEventProcessor.java#L185-L190

Really ~weird~ interesting... I still think `catch (Throwable ...)` is a time bomb waiting to be exploited by a bug report. You are handling `OutOfMemoryError`s, `ThreadDeath`s, etc. I will try poking somebody from Disruptor to comment on this.

carterkozak:
> But once `AsyncAppender` catches the `ThreadDeath`, it is basically dead. Who is gonna restart that thread?

In this case  `ThreadDeath` is a `Throwable` like any other, after we catch it, it's handled and gone. Calling `Thread.stop()` results in that exception being thrown from another thread. The background thread is still both alive and in a normal state (e.g. `RUNNABLE`) so it can continue to process the queue.

carterkozak:
I think the `OutOfMemoryError` handling is critical. Consider an application which gets a spike of requests and fills heap (outside of the logging components) however an asynchronous logging event happens to trigger the first `OutOfMemoryError`. In this case we wouldn't want to stop processing the disruptor queue and block the application because that would make things even worse, we prefer to gracefully swallow the failure and provide as much detail as possible (see [AbstractAsyncExceptionHandler.java](https://github.com/apache/logging-log4j2/blob/913e87d7747aa8d90b7bd3236c4dca2e3307336a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AbstractAsyncExceptionHandler.java)).

In an ideal world everyone would use `-XX:+CrashOnOutOfMemoryError` to avoid thinking about OMEs, but that has drawbacks in some environments as you might imagine :-)

vy:
@carterkozak, after giving it some thought, I have decided to go with your proposal. In a nutshell, what the most recent change set incorporates is as follows:

- `AppenderControl` catch clauses is demoted from `Throwable` to `Exception`.
- `AsyncAppender.AsyncThread` is moved to `AsyncAppenderEventForwarder` with plenty of refactoring for simplification.
- On shutdown, `AsyncAppenderEventForwarder` forwards all the remaining `LogEvent`s, whereas `AsyncAppender.AsyncThread` was ignoring non-`Log4LogEvent`s.
- `AsyncAppenderExceptionHandlingTest` is improved for `ThreadDeath`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,147 @@
...
+            event.setEndOfBatch(queue.isEmpty());
+            forwardOne(event);
+            eventCount++;
+        }
+        LOGGER.trace(
+                "{} has processed the last {} remaining event(s).",
+                getName(), eventCount);
+    }
+
+    void forwardOne(final LogEvent event) {
```
garydgregory, line 95: I would rename to "forward" since you give the method one argument always and it's not like we have to distinguish with a forwardTwo method ;-) or am I missing something?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // Forward the event to all registered appenders.
+        boolean succeeded = false;
+        // noinspection ForLoopReplaceableByForEach (avoid iterator instantion)
+        for (int appenderIndex = 0; appenderIndex < appenders.size(); appenderIndex++) {
+            final AppenderControl control = appenders.get(appenderIndex);
```
garydgregory, line 101: Less clutter with a for each loop IMO.

garydgregory (approved):
A bit of a challenge to follow the bouncing ball ;-) but I would like to confirm/ask that with the new behavior that flushes events on shutdown, that there is no chance of an app even appearing to hang/delay/wait on shutdown... IOW, when I shutdown my app, I want it to go away _now_ (within a reasonable but _very_ short time).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: +1 I don't think forward is correct here or in the class name.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: I believe this approach avoids iterative allocation, with the constraint that it requires a RandomAccess list implementation.

carterkozak: I suppose we could convert the list to an array on creation and get the best of both worlds :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: Oops, I misread your comment. I think it's odd that this object is both a thread, and called a forwarder. I think that made sense before, but the background thread is a receiver, and this provides functionality to submit events. Naming things is one of the two hardest problems in CS, and I don't have better alternatives in mind.

To be clear I'm not suggesting it must be renamed, just thinking aloud.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,147 @@
...
+            if (!stopped) {
+                stopped = true;
+                LOGGER.trace("{} is signaled to stop.", getName());
+            }
+        }
+
+        // There is a slight chance that the thread is not started yet, wait for
+        // it to run. Otherwise, interrupt+join might block.
+        // noinspection StatementWithEmptyBody
+        while (Thread.State.NEW.equals(getState()));
```
carterkozak, line 135: Is this still necessary?

carterkozak:
One point of clarification, I don't think the `Log4jLogEvent` change results in any different behavior, each event (with the exception of `SHUTDOWN_LOG_EVENT`) is transformed to a `Log4jLogEvent` in the `AsyncAppender.append` method [here](https://github.com/apache/logging-log4j2/pull/452/files#diff-6b30d73942796a35c3bcf93d46127d03cb6959fb69fb8eb84c4c6fca13bb68e6R155)

I still need to review the shutdown mechanism for thread safety, although at a glance the queue empty checks that exist on release-2.x are entirely threadsafe either...

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Yes. If `stop()` is called right after `start()`, there is a slight chance that `stop()` will proceed even though the thread is not started yet. Consequently interrupt+join blocks. Without this line, if you would try manually running `AsyncAppenderExceptionHandlingTest`, it fails randomly, though rarely.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: @garydgregory, okay, I will rename it to `forward()`.

@carterkozak, what about *dispatcher*?

vy: For the records, with `forwardOne()`, I was trying to stick to the Spring Data conventions: `findOne()`, `findAll()`, etc.

vy:
> A bit of a challenge to follow the bouncing ball ;-) but I would like to confirm/ask that with the new behavior that flushes events on shutdown, that there is no chance of an app even appearing to hang/delay/wait on shutdown... IOW, when I shutdown my app, I want it to go away _now_ (within a reasonable but _very_ short time).

Yes, AFAIC, when `stop()` is called, thread is interrupted to transition from `forwardAll()` to `forwardRemaining()`, where the latter is non-blocking. This is also analogous to the earlier behavior.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: I wonder if we could add a thread state check around join instead? If the thread is still initializing, we've already set `stopped` to `true`, so it should exit almost immediately if I understand correctly.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,147 @@
...
+                // we can do about it.
+            }
+        }
+
+    }
+
+    void stop(final long timeoutMillis) throws InterruptedException {
+
+        // Mark the completion, if necessary.
+        synchronized (this) {
```
carterkozak, line 125: I think we can replace this synchronization with an atomic getAndSet (since there are only two possible values) or compareAndSet using AtomicBoolean or an AtomicIntegerFieldUpdater (flat object layout, avoids pointer indirection).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: You reasoning makes perfect sense. I wish I would be able to answer your question. The only thing I know is, without this line, `stop()` blocks sometimes. Maybe something for you investigate during your thread-safety check?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
garydgregory: I am not a fan of following Spring's naming, we need to pick names that make sense FOR LOG4J ;-)

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Doh! :facepalm: A relic from the refactoring. Fixed by replacing with an `AtomicBoolean`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Renamed to `forward()`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,147 @@
...
+
+        // There is a slight chance that the thread is not started yet, wait for
+        // it to run. Otherwise, interrupt+join might block.
+        // noinspection StatementWithEmptyBody
+        while (Thread.State.NEW.equals(getState()));
+
+        // Interrupt the thread. (Note that there is neither a check on
+        // "stopped", nor a synchronization here. It is okay to interrupt
+        // concurrently and/or multiple times.)
+        interrupt();
```
carterkozak, line 139: So the previous implementation would submit a sentinel event through the queue to wake the background thread instead of relying on thread interruption.
I think we should avoid interrupting the thread if at all possible due to the subtleties of java interruption. Interruption will actually close sockets if any blocking operations are in progress! This means a socket appender may surprisingly fail to deliver final events. I recall some oddities with file i/o as well.

Perhaps we could try to enqueue a sentinel event and fall back to interruption if there's no space in the queue?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,146 @@
...
+        // Forward the event to all registered appenders.
+        boolean succeeded = false;
+        // noinspection ForLoopReplaceableByForEach (avoid iterator instantion)
+        for (int appenderIndex = 0; appenderIndex < appenders.size(); appenderIndex++) {
+            final AppenderControl control = appenders.get(appenderIndex);
+            try {
+                control.callAppender(event);
+                succeeded = true;
+            } catch (final Throwable ignored) {
+                // If no appender is successful, the error appender will get it.
```
carterkozak, line 107: Unsure what the behavior was before, but I think it's worth a statuslogger debug line here :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        // Fallback to the error appender if none has succeeded so far.
+        if (!succeeded && errorAppender != null) {
+            try {
+                errorAppender.callAppender(event);
+            } catch (final Throwable ignored) {
+                // If the error appender also fails, there is nothing further
+                // we can do about it.
```
carterkozak, line 117: might as well record the throwable here as well

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Makes sense. Will do.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
```diff
@@ -0,0 +1,146 @@
...
+            event.setEndOfBatch(queue.isEmpty());
+            forward(event);
+            eventCount++;
+        }
+        LOGGER.trace(
+                "{} has processed the last {} remaining event(s).",
+                getName(), eventCount);
+    }
+
+    void forward(final LogEvent event) {
```
carterkozak, line 96: naming opinions/bikeshedding, feel free to ignore: Since this is called externally when the queue is full in some circumstances, perhaps the behavior would be more obvious if it was named `callAppenders` or even `callAppendersOnCurrentThread`

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: `dispatcher` or `handler`. No strong opinion, I'm terrible at naming.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Done.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Done.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Done.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: I will kindly pass this.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: @carterkozak, I guess we are sort of done, right? This `while` block discussion is the only one left there. I don't have a strong opinion here. How shall we proceed?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Renamed the class to `AsyncAppenderEventDispatcher`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
carterkozak: I'll take another read through this change today -- happy to defer changes on this part to a separate PR.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
```diff
@@ -0,0 +1,172 @@
...
+                control.callAppender(event);
+                succeeded = true;
+            } catch (final Throwable error) {
+                // If no appender is successful, the error appender will get it.
+                // It is okay to simply log it here.
+                if (LOGGER.isTraceEnabled()) {
+                    final String message = String.format(
+                            "%s has failed to call appender %s",
+                            getName(), control.getAppenderName());
+                    LOGGER.trace(message, error);
```
carterkozak, line 119: nit: I think we can use interpolation curly braces instead of String.format. Same on line 132.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
```diff
@@ -0,0 +1,172 @@
...
+        }
+        LOGGER.trace("{} has stopped.", getName());
+    }
+
+    private void dispatchRemaining() {
+        int eventCount = 0;
+        while (true) {
+            // Note the non-blocking Queue#poll() method!
+            final LogEvent event = queue.poll();
+            if (event == null || event == STOP_EVENT) {
```
carterkozak, line 90: The previous implementation would allow events that managed to be submitted after the sentinel, but before inputs resulted in a throw to be processed. I think we want a `continue` on `event == STOP_EVENT` rather than `break`.

It's not clear that this will make a difference in practice. Continuing to poll after STOP_EVENT doesn't guarantee we record all such events, only if they're already visible in the queue.

carterkozak (approved):
Thanks, @vy! Change looks good to me :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
vy: Here I try to match against the `trace(..., Throwable)` method. How can I employ *interpolation curly braces* to achieve that?

carterkozak: I believe as long as there are fewer curly braces than args, and the last arg is a Throwable, it works. (worth a quick validation running locally though).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
vy: Shouldn't we simply discard anything that is sent after the `stop()` signal?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
vy: Didn't know such a thing existed at all! :open_mouth: Fixed.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventForwarder.java
vy: Let's get all done here. I really want to get this merged and poke Ralph for 2.14.1.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AsyncAppenderEventDispatcher.java
carterkozak: I defer to your judgement whether or not to make a change here, this was meant to elaborate upon the comment and complexities of the scenario. We can make changes separately from this PR even if we decide the behavior should be different :-)

vy: I think your argument makes sense. Replace `break` with `continue`.

PR #454, opened 2020-12-29
LOG4J2-2985 Add eventTemplateRootObjectKey parameter to JsonTemplateLayout.

vy:


carterkozak:
I started reviewing this and realized I don't know as much as I should about json-template-layout as a whole!

This feature has a lot in common with the additional fields component in that it provides shorthand to avoid rewriting existing templates.

I wonder if it would be helpful to have a `templateUri` resolver that could solve this in an admittedly more verbose way:
```json
{
  "event": {
    "$resolver": "templateUri",
    "uri": "classpath:LogstashJsonEventLayoutV1.json"
  }
}
```

I'm not suggesting against the current approach, just recording my thoughts. What do you think?

I took a crack at implementing such a thing as an opportunity to explore JTL [here](https://github.com/apache/logging-log4j2/compare/release-2.x...carterkozak:ckozak/nested_templates?expand=1). I don't mind throwing away my code as the motivation was primarily exploratory. The JTL code is a joy to work with :-)

vy:
Hey @carterkozak, thanks so much for taking time to review the changes and going even further to implement an alternative approach. I am not inclined the go with a new template resolver to resolve the actual template. This is sort of a *meta* directive which I am hesitant to introduce, unless there is a use case that I fail to see.

@carterkozak, may I assume you are okay with my changes? If so, I am gonna merge them to both `release-2.x` and `master`.

carterkozak:
Yep, I think the changes are good, thanks! My commit was primarily an exercise to familiarize myself with the code :-)

One oddity the meta revolver surfaced was that the context contains data directly describing the root object (root attribute, additional properties) that may be better decomposed into the template model rather than the context. No changes requested, just something that might be worth considering.

vy:
> Yep, I think the changes are good, thanks! My commit was primarily an exercise to familiarize myself with the code :-)

Thanks, will merge it!

> One oddity the meta revolver surfaced was that the context contains data directly describing the root object (root attribute, additional properties) that may be better decomposed into the template model rather than the context. No changes requested, just something that might be worth considering.

Very good insight! I indeed smelled the same while working on the issue. Just to make sure we are on the same page, what do you exactly mean by the *template model*? Do you mean converting, say, *event template additional fields* to a template resolver?

carterkozak:
> Just to make sure we are on the same page, what do you exactly mean by the template model? Do you mean converting, say, event template additional fields to a template resolver?

I think that's close (or the same, but I'm not articulating my idea well!). In my mind a template resolver is able to produce a single component (json number, string, object, or array) which doesn't quite align with additional properties, because they apply to a separate component. Adding additional fields could work by passing the data through to `JsonReader.read(template)` to supplement the template, though you would have a better idea than I do where the correct place for this would be!

I think expanding the root property into a TemplateResolver would be feasible since it doesn't modify an existing object.

I wonder if the object-specific values could be passed to the template parsing code without being part of the more general context object?

vy:
Okay, indeed we were on the same page. I share the same concerns. I will try to refactor these bits – given they are internal, the changes should not cause any compatibility issues.

I had stated a couple of times in the past, I want to make JTL resolvers pluggable. So that users can introduce their own custom resolvers. Though before doing that, we need to iron out these API issues, that is, what to include in the context models and what not.

PR #456, opened 2020-12-30
ci: add --no-transfer-progress Maven flag

sullis:


garydgregory:
Good idea. I've done this for all/most of the Apache Commons builds except for the ones using old versions of Maven.

vy:
Applied and cherry-picked to `release-2.x` too.

PR #461, opened 2021-01-19
Fix formatting in log4j-spring-cloud-config-sample-application README

szpak:
Random trivial change as I was reading that file. It renders much greatly.


vy:
Thanks @szpak! I will see if we can backport this to `release-2.x` as well.

vy:
Fits to `release-2.x` like a glove. :+1:

PR #466, opened 2021-02-17
LOG4J2-3020 LOG4J2-3021 Improve Logger javadoc for throwing() and catching().

vy:


log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
```diff
@@ -80,59 +80,61 @@
...
      */
-    void catching(Level level, Throwable t);
+    void catching(Level level, Throwable throwable);
 
     /**
-     * Logs an exception or error that has been caught. Normally, one may wish to provide additional information with an
-     * exception while logging it; in these cases, one would not use this method. In other cases where simply logging
-     * the fact that an exception was swallowed somewhere (e.g., at the top of the stack trace in a {@code main()}
-     * method), this method is ideal for it.
+     * Logs an exception or error that has been caught at the {@link Level#ERROR ERROR} level.
```
garydgregory, line 88: Simpler to say "Logs a Throwable..." here IMO since the argument IS a Throwable and I can define a subclass that is neither an Exception or Error.

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
```diff
@@ -292,12 +294,12 @@
 
     /**
      * Logs a message at the {@link Level#DEBUG DEBUG} level including the stack trace of the {@link Throwable}
-     * <code>t</code> passed as parameter.
+     * <code>throwable</code> passed as parameter.
      *
      * @param message the message to log.
-     * @param t the exception to log, including its stack trace.
+     * @param throwable the exception to log, including its stack trace.
      */
```
garydgregory, line 301: Same comment as above... might as well be consistent  ;-)

garydgregory (changes requested):
I like all the changes :-) see my 2 comments.

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
vy: Corrected.

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
vy: Corrected.

Marcono1234:
Thanks! Though as pointed out in LOG4J2-3021, the documentation for `Logger.throwing(...)` would now be incorrect for the `java.util.logging` implementation which logs at `TRACE`.

Would it make sense to change these methods (in Log4j 3.x?) to log at `DEBUG` level because as mentioned in the Jira report descriptions, logging at `ERROR` will likely cause redundant and irritating information, especially when used as suggested by the documentation.

vy:
> Would it make sense to change these methods (in Log4j 3.x?) to log at `DEBUG` level? As mentioned in the Jira report descriptions logging at `ERROR` will likely cause redundant and irritating information, especially when used as suggested by the documentation.

@Marcono1234, sounds practical, at first sight. To further consult to the rest of the developers, would you mind moving this discussion to [the dev mailing list](https://logging.apache.org/log4j/2.x/mail-lists.html), please?

Marcono1234:
> To further consult to the rest of the developers, would you mind moving this discussion to [the dev mailing list](https://logging.apache.org/log4j/2.x/mail-lists.html), please?

I am not very deeply invested into Log4j2 development, and am not a fan of mailing lists (though your Pony Mail one looks pretty great compared to a lot other mailing lists!), therefore I won't write to the mailing list for now. However, feel free (but don't feel forced) to mention it yourself if you think that what I described is an issue. Sorry for the late response.

PR #468, opened 2021-02-19
Apply automated Assertj refactors to core modules

carterkozak:
I've applied some refactors I wrote a while ago: https://github.com/palantir/assertj-automation
Separated into two commits, first migrates from other asserts to assertj, and a second which optimizes assertj use.


log4j-1.2-api/src/test/java/org/apache/log4j/CategoryTest.java
```diff
@@ -78,23 +78,23 @@ public void testForcedLog() {
...
         final List<LogEvent> list = appender.getEvents();
         int events = list.size();
-        assertTrue("Number of events should be 1, was " + events, events == 1);
+        assertThat(events).describedAs("Number of events should be 1, was " + events).isEqualTo(1);
         LogEvent event = list.get(0);
         Message msg = event.getMessage();
-        assertNotNull("No message", msg);
-        assertTrue("Incorrect Message type", msg instanceof ObjectMessage);
+        assertThat(msg).describedAs("No message").isNotNull();
+        assertThat(msg instanceof ObjectMessage).describedAs("Incorrect Message type").isTrue();
```
carterkozak, line 85: I thought I'd written a transformation for these, perhaps not!
Should end up with:
```suggestion
        assertThat(msg).describedAs("Incorrect Message type").isInstanceof(ObjectMessage.class);
```

log4j-1.2-api/src/test/java/org/apache/log4j/CallerInformationTest.java
```diff
@@ -57,9 +57,9 @@ public void testMethodLogger() throws Exception {
         logger.warn("brains~~~");
         logger.info("Itchy. Tasty.");
         final List<String> messages = app.getMessages();
-        assertEquals("Incorrect number of messages.", 5, messages.size());
+        assertThat(messages).describedAs("Incorrect number of messages.").hasSize(5);
```
rgoers, line 60: Not that I object to these but I am not sure why the new version is "better".  In this case they both effectively say "Throw an exception with this message if the message list doesn't contain 5 items and to me both are equally clear.

log4j-1.2-api/src/test/java/org/apache/log4j/CategoryTest.java
```diff
@@ -78,23 +78,23 @@ public void testForcedLog() {
...
         category.log(Priority.INFO, "Hello, World");
         events = list.size();
-        assertTrue("Number of events should be 1, was " + events, events == 1);
+        assertThat(events).describedAs("Number of events should be 1, was " + events).isEqualTo(1);
         event = list.get(0);
         msg = event.getMessage();
-        assertNotNull("No message", msg);
-        assertTrue("Incorrect Message type", msg instanceof ObjectMessage);
+        assertThat(msg).describedAs("No message").isNotNull();
+        assertThat(msg instanceof ObjectMessage).describedAs("Incorrect Message type").isTrue();
```
rgoers, line 95: I do notice a pattern. Every one of the replacements is longer than the original. Of course, with IDE code completion that doesn't necessarily imply more typing.

log4j-1.2-api/src/test/java/org/apache/log4j/CallerInformationTest.java
carterkozak: In the case the test fails fails, il the new code is able to give us more data to understand what’s broken. The failure message includes the list contents in addition to expected and actual lengths.

log4j-1.2-api/src/test/java/org/apache/log4j/CategoryTest.java
carterkozak: That’s true. In many cases we can reduce the number of assertions, for example then null check can be removed in favor of an assertThat(message).isInstanceOf(ObjectMessage.class) which provides a human readable failure in the null case as well.

the other downside is without an auto-formatter these lines aren’t broken properly. I’ve tried setting up a formatted but haven’t made it very far yet.

carterkozak: I should have probably used “as” instead of “describedAs”, they do the exact same thing but the former is nine characters shorter.

PR #469, opened 2021-02-24
Fix windows build on master

perry2of5:
This pull fixes the build on Windows.


log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -205,6 +205,32 @@ public void reset() {
...
+        			hasSystemStreamWriter = true;
+        		} else {
+        			listenersToClose.add(listener);
+        			
+        		}
+        	}
+        }
+        if (!hasSystemStreamWriter) {
+        	registerListener(new StatusConsoleListener(getLevel()));
+        }
```
vy, line 229: I am still reading through the sources... Though I am confused by this step: Why do we have to have a system stream listener?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -205,6 +205,32 @@ public void reset() {
             clear();
         }
     }
+    
+    /**
+     * Stop all the listeners except those writing to standard out or error. If
+     * no loggers write to standard out or error, then add one and stop the 
+     * others.
+     */
+    public void prepareToStop() {
```
vy, line 214: Why not simply extending from `AutoCloseable` and overriding its `close()`? I guess, there have been a discussion about this in the mailing list, though it is good to communicate the context here as well.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -405,6 +405,7 @@ public boolean stop(final long timeout, final TimeUnit timeUnit) {
             }
         }
         LOGGER.debug("Stopped LoggerContext[name={}, {}] with status {}", getName(), this, true);
+    	StatusLogger.getLogger().prepareToStop();
```
vy, line 408: Again, I think I am missing some context here. Would you mind elaborating a bit on the necessity of this line here, please?

log4j-kafka/src/test/java/org/apache/logging/log4j/kafka/builder/ConfigurationBuilderTest.java
```diff
@@ -106,14 +112,27 @@ private void addTestFixtures(final String name, final ConfigurationBuilder<Built
                 INDENT + "</Loggers>" + EOL +
             "</Configuration>" + EOL;
 
-    // TODO make test run properly on Windows
     @Test
     public void testXmlConstructing() throws Exception {
         //assumeTrue(System.lineSeparator().length() == 1); // Only run test on platforms with single character line endings (such as Linux), not on Windows
```
vy, line 117: I guess we can remove this as well.

src/changes/changes.xml
```diff
@@ -31,6 +31,19 @@
          - "remove" - Removed
     -->
     <release version="3.0.0" date="2021-MM-DD" description="GA Release 3.0.0">
+      <action issue="LOG4J2-3024" dev="mattsicker" type="fix" due-to="Tim Perry">
+        Cleanly shutdown status logger when log4j is stopping to ensure file handles
+        are promptly released to avoid resource leakage.
+      </action>
```
vy, line 37: Would you mind including only your changes, please? (Is Matt, @jvz, really responsible for this change?)

jvz: The `dev` attribute value should be filled with whoever merges the patch.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
perry2of5: The commentary in the mailing list indicates the status logger shouldn't be closed, we need to redirect the status logger to standard out iff there is no remaining listener that writes to a file.  If the log4j code is running in the container the prepareToStop functionality should stop the listener for the context that is shutting down. This code is flawed because it stops the listeners for all contexts. I'll push some changes that stop listeners on a context-by-context basis.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
perry2of5: Something needs to tell the StatusLogger there is a stop event. Otherwise it never releases the file handles for the listeners writing to a file. 

In the code on master on windows the status logger log file is never closed so the file remains locked which causes the test to fail. Now the LoggerContext is telling the StatusLogger that a shutdown event is happening. In the code I'll push later, I've added a paremeter to say which context is shutting down so the StatusLogger can remove the listeners for that context.

log4j-kafka/src/test/java/org/apache/logging/log4j/kafka/builder/ConfigurationBuilderTest.java
perry2of5: Good point.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
perry2of5: In the dev email list, on Feb 23, "What is going on with Master?" thread:

I said:

My thought is we choose one of two options:

Option A:
1) check if any StatusLogger is writing to standard out or standard error. If not, add one.
2) stop any loggers that don't write to standard out or standard error.

Option B:
1) stop any loggers that don't write to standard out or standard error.

Option A could cause the log messages to be split across two destinations, but they all get sent somewhere. Option B could lose shutdown messages when writing to a file, but by that point it may not matter.


Ralph replied:

I would suggest that if it is writing to something other than System.out that it be redirected back there and then the OutputStream be closed. However, I’ve not looked at the code recently so I am not sure what it takes to do that.

src/changes/changes.xml
perry2of5: What is the correct thing for me to put in changes.xml?

<action issue="LOG4J2-3024" dev="UPDATE-ME" type="fix" due-to="Tim Perry">

perry2of5:
I'm going to cancel this. The fix is unwieldly and not well thought out.

PR #470, opened 2021-02-25
[LOG4J2-3028] OutputStreamManager#flushBuffer always clears the buffer.

jkozlowski:


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java
```diff
@@ -278,10 +278,13 @@ protected synchronized void flushDestination() {
      */
     protected synchronized void flushBuffer(final ByteBuffer buf) {
         ((Buffer) buf).flip();
```
carterkozak, line 280: So previously in the exceptional case, we would fail out before calling clear, leaving ourselves with buffer position zero, limit N (potentially a small fraction of the capacity, based on potentially a single event).
Subsequent attempts to flush the buffer can only further reduce the available buffer space until our limit and position converge. At that point our `buf.remaining() > 0` condition can never succeed if I'm reasoning through this correctly, and we cannot write any data.
Great catch!

carterkozak (approved):

vy:
@carterkozak, I am joining to the party late, sorry. What is the status of this PR? Can we merge it?

vy:
@jkozlowski, thanks so much for the contribution! Would you mind updating the `changes.xml` too, please? That is,

```xml
<action issue="LOG4J2-3028" dev="carterkozak" type="fix" due-to="Jakub Kozlowski">
  OutputStreamManager does not clear buffer on exception.
</action>
```

jkozlowski:
@vy should be done now

jkozlowski:
Hmm, there were merge conflicts, so I merged develop and now builds aren't green; I wonder if it's an actual issue or a flake?

carterkozak:
It looks like a build flake, I'll take care of the merge in the next hour

carterkozak:
@vy Yep, I've taken care of merging this :-)

PR #471, opened 2021-03-01
LOG4J2-2948 Replace HashSet with IdentityHashMap in ParameterFormatter to detect cycles.

vy:


vy:
@rgoers, would you mind reviewing the changes, please? Note that I don't create a new `HashSet` for each internal call, which has used to be the case. According to tests, this should be fine.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -538,24 +536,28 @@ private static void appendArray(final Object o, final StringBuilder str, Set<Str
                     } else {
                         str.append(", ");
                     }
-                    recursiveDeepToString(current, str, new HashSet<>(dejaVu));
+                    recursiveDeepToString(current, str, effectiveDejaVu);
```
carterkozak, line 539: By removing the `new HashSet` as we traverse the collection, I think we end up producing potentially confusing data. I think we may want to keep the collection creation despite the additional allocation cost.

We should add tests for this case ;-)

Consider `[Foo{A}, Foo{A}]`. It duplicates data, however it's not recursive. We only want to catch cycles here, not duplication, e.g. `A=[Foo{delegate=A}]` if that makes sense. What do you think?

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -448,7 +450,7 @@ static String deepToString(final Object o) {
      * @param str    the StringBuilder that o will be appended to
      * @param dejaVu a list of container identities that were already used.
```
Marcono1234, line 451:
```suggestion
     * @param dejaVu a set of container objects directly or transitively containing {@code o}
```

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -448,7 +450,7 @@ static String deepToString(final Object o) {
...
      */
-    static void recursiveDeepToString(final Object o, final StringBuilder str, final Set<String> dejaVu) {
+    static void recursiveDeepToString(final Object o, final StringBuilder str, final Set<Object> dejaVu) {
```
Marcono1234, line 453: Would it make sense to make this method private and instead add a package-private overload `recursiveDeepToString(Object, StringBuilder)`?
Most of the calls to this method have `null` as dejaVu, and especially the calls from `MapMessage` should not have to care about how recursion tracking is implemented.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
vy: Good catch! Putting the cloning back with a unit test.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
vy: :+1: Will do.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
vy: Makes sense, implemented.

vy:
@carterkozak, would you mind skimming through the changes one last time, please? If you approve, I am gonna merge this into `release-2.x` and will get busy with porting it to `master`.

log4j-api/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
```diff
@@ -180,13 +182,29 @@ public void testDeepToString() throws Exception {
...
+        final Object item = Collections.singletonList(0);
         list.add(1);
+        list.add(item);
+        list.add(2);
+        list.add(item);
+        list.add(3);
+        final String actual = ParameterFormatter.deepToString(list);
+        final String expected = "[1, [0], 2, [0], 3]";
+        assertEquals(expected, actual);
+    }
```
carterkozak, line 196: Thanks!

carterkozak (approved):
Looks good to me. Thanks, Volkan!

PR #472, opened 2021-03-03
Allow custom resolvers to be registered for JsonTemplateLayout

radut:
This PR adds ability to register custom resolvers for JsonTemplateLayout

### DIFF
--- a/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/EventResolverFactories.java
+++ b/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/EventResolverFactories.java
@@ -19,19 +19,19 @@ package org.apache.logging.log4j.layout.template.json.resolver;
 import org.apache.logging.log4j.core.LogEvent;
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 final class EventResolverFactories {
 
-    private EventResolverFactories() {}
+    private EventResolverFactories() {
+    }
 
-    private static final Map<String, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>>> RESOLVER_FACTORY_BY_NAME =
-            createResolverFactoryByName();
+    private static final Map<String, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>>> RESOLVER_FACTORY_BY_NAME = new ConcurrentHashMap<>();
 
-    private static Map<String, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>>> createResolverFactoryByName() {
+    static {
 
         // Collect resolver factories.
         final List<EventResolverFactory<? extends EventResolver>> resolverFactories = Arrays.asList(
@@ -57,7 +57,7 @@ final class EventResolverFactories {
         for (final EventResolverFactory<? extends EventResolver> resolverFactory : resolverFactories) {
             resolverFactoryByName.put(resolverFactory.getName(), resolverFactory);
         }
-        return Collections.unmodifiableMap(resolverFactoryByName);
+        RESOLVER_FACTORY_BY_NAME.putAll(resolverFactoryByName);
 
     }
 
@@ -65,4 +65,14 @@ final class EventResolverFactories {
         return RESOLVER_FACTORY_BY_NAME;
     }
 
+    public static void registerResolver(String name, TemplateResolverFactory<LogEvent, EventResolverContext, ? extends TemplateResolver<LogEvent>> templateResolverFactory) {
+        if (RESOLVER_FACTORY_BY_NAME.containsKey(name)) {
+            throw new IllegalArgumentException("Resolver with same name : " + name + " already registered");
+        }
+        RESOLVER_FACTORY_BY_NAME.put(name, templateResolverFactory);
+    }
+
+    public static void unregisterResolver(String name) {
+        RESOLVER_FACTORY_BY_NAME.remove(name);
+    }
 }


vy:
Hey @radut! This feature is something definitely on my radar, see [LOG4J2-3004](https://issues.apache.org/jira/browse/LOG4J2-3004). The reason we haven't made resolver registry public is indeed to disallow tampering this static list, until we migrate these to Log4j 2 plugins similar to the one in `PatternLayout`.

Compared to plugins, there are a couple of disadvantages of mutating a static list of resolvers.
1. They will affect all `JsonTemplateLayout` instantiations, which might not be desirable. Put another way, you cannot change them per `JsonTemplateLayout` instance.
2. It is not Log4j idiomatic – plugins are the way for extending an existing functionality.

Hence, I will sadly need to reject this change. Though feel free to submit a new PR (or re-open and update this one) adding plugin support to `JsonTemplateLayout`.

Curious about one thing: **What new resolvers were you planing to introduce?** If they address a common need, it is easier to land that into a release compared to a fully-fledged plugin support.

radut:
`JsonTemplateLayout` is a good idea, you could define specific constants like `"@version": 1`(number) or lets say `"source" : "log4j"`(string). What I am looking for is rewrite the mdc resolver to be able to write numbers as json numbers, instead of string literals, usefull for indexing and searching in ElasticSearch with function like between x and y, without having another transformer on the way. 

But most importantly I want to be able to pass specific `Markers` with a `JSON/map` attached to logging event, which is very usefull for that specific logging event. Currently it is possible to write only the name of the marker in my case something like `LS_MARKER` (https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/layout/src/main/java/org/apache/logging/log4j/core/LogStashMarker.java) , but instead I want to override the resolver and pull the `JSON` info and write it (eventually flatten) into the same json log event.

Something similar I have done in the past here : https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/demo/src/main/java/com/radut/log4j/demo/LoggerMarkerTest.java#L52  with the serializerd : https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/layout/src/main/java/org/apache/logging/log4j/core/jackson/MarkerJsonSerializer.java a fork from https://github.com/majikthys/log4j2-logstash-jsonevent-layout.

basically there is a `Map<String, Object>` which is serialized as JSON within a field or flattened.

vy:
> `JsonTemplateLayout` is a good idea, you could define specific constants like `"@version": 1`(number) or lets say `"source" : "log4j"`(string). What I am looking for is rewrite the mdc resolver to be able to write numbers as json numbers, instead of string literals, usefull for indexing and searching in ElasticSearch with function like between x and y, without having another transformer on the way.

MDC resolver is capable of rendering JSON objects, you don't necessarily need to _stringify_ them. You can toggle `stringified` flag of the `mdc` resolver. How doesn't this address your case?

> But most importantly I want to be able to pass specific `Markers` with a `JSON/map` attached to logging event, which is very usefull for that specific logging event. Currently it is possible to write only the name of the marker in my case something like `LS_MARKER` (https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/layout/src/main/java/org/apache/logging/log4j/core/LogStashMarker.java) , but instead I want to override the resolver and pull the `JSON` info and write it (eventually flatten) into the same json log event.
> 
> Something similar I have done in the past here : https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/demo/src/main/java/com/radut/log4j/demo/LoggerMarkerTest.java#L52 with the serializer : https://github.com/radut/log4j2-logstash-layout/blob/67f3ea74bf061ee27a4c5405c59745a64a663a91/layout/src/main/java/org/apache/logging/log4j/core/jackson/MarkerJsonSerializer.java a fork from https://github.com/majikthys/log4j2-logstash-jsonevent-layout.
> 
> basically there is a `Map<String, Object>` which is serialized as JSON within a field or flattened.

Yeah... That is really a custom use case. Though good to know; appreciated that you have shared it, thank you. I will try to prioritize LOG4J2-3004 in my backlog.

radut:
stringified doesn't seem to be working(or at least, not as I expect it to):
log4j2.xml
```
<Console name="console" target="SYSTEM_OUT" follow="true">
 <JsonTemplateLayout eventTemplateUri="classpath:LogstashJsonEventLayout.json" stackTraceEnabled="true" />
<Console>
```



```
ThreadContext.push("tcVal2", 2);
```
doesn't get outputed at all
image



```
ThreadContext.put("tcVal2", "2"); // allows only string as value
```
the mdc outputs "tcVal2":"2" as string even if `stringified` is `true` or `false`


image

image

vy:
So, I guess, both of us are right.

1. Yes, `JsonTemplateLayout` can serialize any object type, not only strings. `JTL` accesses to the thread context data using `LogEvent#getContextData()` method of return type `ReadOnlyStringMap`. There it uses the following method of the `ROSM`: `<V, S> void forEach(final TriConsumer<String, ? super V, S> action, final S state)`. See that here the `value` can be any `Object`. Check out MDC tests in `JsonTemplateLayoutTest` to see how there we test serialization for different types in addition to strings.
2. Yes, you cannot _easily_ put anything into MDC using `ThreadContext` class provided methods.

Note my _"easily"_ remark. I think you can do the following to put non-String values into MDC:

```java
ObjectThreadContextMap contextMap = (ObjectThreadContextMap) ThreadContext.getThreadContextMap();
contextMap.putValue("string", "foo");
contextMap.putValue("integer", 1);
contextMap.putValue("map", Collections.singletonMap("foo", "bar"));
contextMap.putValue("array", new Object[]{"foo", 0xDEADBEEF});
```

Though this will only work if `Constants.ENABLE_THREADLOCALS` is `true`:

```
boolean IS_WEB_APP = PropertiesUtil
        .getProperties()
        .getBooleanProperty("log4j2.is.webapp", isClassAvailable("javax.servlet.Servlet"));

boolean ENABLE_THREADLOCALS = !IS_WEB_APP &&
        PropertiesUtil
                .getProperties()
                .getBooleanProperty("log4j2.enable.threadlocals", true);
```

**Would you mind checking if this works for you?** In the meantime, I will raise this shortcoming of `ThreadContext` in the Log4j `dev` mailing list.

vy:
I think this shortcoming has already been raised in the past, see [LOG4J2-1648](https://issues.apache.org/jira/browse/LOG4J2-1648). Apparently such a feature would have other implications. I will try to dig further into this.

radut:
@vy this `ObjectThreadContextMap` worked. Altough is there any way to still have this in a webApp ?
I didn't tried yet in a spring boot application, but there the `javax.servlet.Servlet` will be available.

vy:
Yes, just set both of the `log4j2.is.webapp` and `log4j2.enable.threadlocals` system properties to `true`.

PR #473, opened 2021-03-09
ensure EncodingPatternConverter#handlesThrowable is implemented

rmannibucau:
Giess the title is enough to understand what happens ;).
Long story short the handlesThrowable() logic to skip forced (by default) append of the exception does not work as soon as encoding is used which is almost always the case when outputing JSON/XML using PatternFormatter (since Json appender required 3rd party libs which are often undesired).
This PR implements this method by checking the nested formatters.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/EncodingPatternConverter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/EncodingPatternConverter.java
@@ -17,6 +17,8 @@
 package org.apache.logging.log4j.core.pattern;
 
 import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


rmannibucau:
up?

vy:
Thanks for the contribution @rmannibucau! We had some discussion in the mailing list regarding the failing tests of this PR, though apparently we forgot about the ticket itself. :sweat_smile:

I will fix `changes.xml` for you – try to update it next time, please. I will also try to back-port this to `release-2.x`.

vy:
One last remark, a JIRA ticket is always handy. It helps us to link changes with tickets. I will create one for you.

vy:
> since Json appender required 3rd party libs which are often undesired

`JsonTemplateLayout` doesn't require any 3rd party libs, FYI.

PR #475, opened 2021-03-16
LOG4J2-2940: Context selectors are aware of ClassLoader dependency

carterkozak:
This allows LoggerContext lookups to avoid searching for the calling
class to discover a classloader when the ContextSelector implementation
is declared to be agnostic to the class loader.

Custom LoggerContextFactory and ContextSelector implementations
should be updated to override the new `isClassLoaderDependent` method.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/BasicAsyncLoggerContextSelector.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/BasicAsyncLoggerContextSelector.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.log4j.core.async;
+
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.impl.ContextAnchor;
+import org.apache.logging.log4j.core.selector.ContextSelector;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Returns either this Thread's context or the default {@link AsyncLoggerContext}.
+ * Single-application instances should prefer this implementation over the {@link AsyncLoggerContextSelector}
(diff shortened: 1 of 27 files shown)


log4j-slf4j18-impl/src/main/java/org/apache/logging/slf4j/Log4jLoggerFactory.java
```diff
@@ -47,8 +47,12 @@ protected Logger newLogger(final String name, final LoggerContext context) {
...
     @Override
     protected LoggerContext getContext() {
-        final Class<?> anchor = StackLocatorUtil.getCallerClass(FQCN, PACKAGE);
-        return anchor == null ? LogManager.getContext() : getContext(StackLocatorUtil.getCallerClass(anchor));
+        final Class<?> anchor = LogManager.getFactory().isClassLoaderDependent()
+                ? StackLocatorUtil.getCallerClass(FQCN, PACKAGE)
+                : null;
+        return anchor == null
+                ? LogManager.getContext()
+                : getContext(StackLocatorUtil.getCallerClass(anchor));
```
carterkozak, line 55: In the case where the context selector is classloader dependent, I imagine we could merge the `anchor = StackLocatorUtil.getCallerClass(FQCN, PACKAGE)` and `StackLocatorUtil.getCallerClass(anchor)` calls into a single sweep, potentially cutting the work in half.

rgoers: You would have to make a new method that gets the caller of the caller.

carterkozak: That's what I had in mind. We could provide an overload which takes an additional integer parameter of additional frames to skip.
I'd prefer to implement that optimization separately from this change, and also implement another ContextSelector equivalent to `BasicContextSelector` for fully asynchronous logging so folks don't have to write their own to avoid classloader lookups.

carterkozak: I can iterate on this PR with each component in an isolated commit, that may be easiest to review. Let me know if you think the current approach adding `boolean isClassLoaderDependent()` is reasonable.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/BasicAsyncLoggerContextSelector.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Returns either this Thread's context or the default {@link AsyncLoggerContext}.
+ * Single-application instances should prefer this implementation over the {@link AsyncLoggerContextSelector}
+ * due the the reduced overhead avoiding classloader lookups.
+ */
+public class BasicAsyncLoggerContextSelector implements ContextSelector {
```
carterkozak, line 34: I still need to add a commit to mention this class in the documentation.

log4j-slf4j18-impl/src/main/java/org/apache/logging/slf4j/Log4jLoggerFactory.java
rgoers: Yes, the optimization should be done separately. I'm not exactly sure what you mean by iterating on this PR. I would suggest merging this one and then do the other work as follow-up items later. So long as what you merge works we should be fine.

carterkozak:
Merged support for `isClassLoaderDependent` into release-2.x and master.

PR #476, opened 2021-03-22
LOG4J2-3004 Add plugin support to JsonTemplateLayout.

vy:
This change set aims to replace statically bound JSON template resolver factories with plugins. Along this pursuit, following side kicks are introduced:

* Added support for inline stack trace element templates in `exception` and `exceptionRootCause` resolvers.
* Added `PluginUtil` class.
* Updated `TypeConverterRegistry` to handle competing converters for the same type.

Judging from the `json-template-layout.html` generated via `./mvnw site -DskipTests=true -Dmaven.doap.skip=true`, there is an AsciiDoc puzzler waiting to be addressed: usage of `xref:event-template-resolver-exception[]
  and xref:event-template-resolver-exceptionRootCause[]` renders garbled text and link.


vy:
Note that this issue addresses the request in #472.

erohana:
is this feature ready for use ?
and from what version ?

vy:
@erohana, as pointed in [the JIRA ticket](https://issues.apache.org/jira/browse/LOG4J2-3004), this PR will be shipped with `2.15.0`, which is yet to be released. This said, you can use the `2.15.0-SNAPSHOT` versions. Or... create some noise in the mailing list and elaborate on why do you need this release, etc.

What is the particular use case you have in mind?

erohana:
i need to create a custom exception resolver to filter the exception stacktrace, we use spring framework for our application, and spring framework stacktraces are huge so I want to filter it out to make it more readable

vy:
@erohana, would [`exception[RootCause].stringified.truncation.pointMatcher{Strings,Regexes}`](https://github.com/apache/logging-log4j2/blob/release-2.x/src/site/asciidoc/manual/json-template-layout.adoc.vm#L612), that (sadly, again) will be shipped with `2.15.0`, address your issue? If not, why?

erohana:
i tried to use this feature, but it doesn't; filter the stacktrace, it rather filter out anything after the match.
for example if I have the following stacktace :
A
B
B
B
C
C

I want to filter out all B to have the following result:
A
C
C

the truncation feature will give the following result if I add B pointMatcher:
A

in this case C will be filtered out which is not the result i want.

i hope u understand it

erohana:
checkl this example https://logging.paluch.biz/stack-trace-filter.html
this is exactly what i want

vy:
@erohana, mind creating a Log4j JIRA ticket? Please share the example and the link(s) too.

radut:
Do you have any example how we can write custom resolvers for this plugin ?

vy:
@radut, we have [a whole section](/apache/logging-log4j2/blob/release-2.x/src/site/asciidoc/manual/json-template-layout.adoc.vm#L1529) dedicated to this, yet 2.15.0 is not released, hence these are not visible in the website. 2.15.0 is targeted for the second half of August.

radut:
I was hoping that I could test in master branch until the release of 2.15.0, maybe small adjustments could be made until release... I would like to override `MarkerResolver` or write a custom one, register it, gain access to `LogEvent` and `JsonWriter`

vy:
Is working with `2.15.0-SNAPSHOT` (which is available in Maven Central) an option for you?

radut:
cannot find the file here : https://repo1.maven.org/maven2/org/apache/logging/log4j/log4j-core/2.15.0-SNAPSHOT/log4j-core-2.15.0-SNAPSHOT.jar

vy:
@radut The following repository should work: https://repository.apache.org/content/repositories/snapshots/

PR #477, opened 2021-03-23
MapMessage put methods should not mandate String values

henryw374:
the underlying Map is typed <String,Object> so the put methods on MapMessage can also be.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/MapMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/MapMessage.java
@@ -197,7 +197,7 @@ public class MapMessage<M extends MapMessage<M, V>, V> implements MultiFormatStr
      * @param candidateKey The name of the data item.
      * @param value The value of the data item.
      */
-    public void put(final String candidateKey, final String value) {
+    public void put(final String candidateKey, final Object value) {
         if (value == null) {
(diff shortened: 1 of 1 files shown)


vy:
@henryw374, I have mentioned about your PR in the dev mailing list, there is [a discussion going on](https://mail-archives.apache.org/mod_mbox/logging-dev/202103.mbox/%3CCAP7pH7sG%2B82UUZK91w2%2BQO%2BaJceZm5zFaCO4UwwKGtdiw_y5Cg%40mail.gmail.com%3E). Long story short, it is not as easy as it seems. I'd suggest you to join the discussion, if you want to get this resolved soon.

henryw374:
Thanks. I'll look in the forum

henryw374:
I've sent a subscribe message to the list now. 

To add a bit more info, there is a constructor of MapMessage which takes a single map arg :

    
```
public MapMessage(final Map<String, V> map) {
        this.data = new SortedArrayStringMap(map);
    }
```

which I am using already with things like maps of maps etc - which get printed to nested json. So the put* methods are inconsistent with that.

rgoers:
The discussion on the mailing list died off without any clear resolution. I'm not sure what to do with this.

vy:
IIRC, as indicated in the mailing list discussion, the necessary set of changes is not as simple as this PR implies. I am taking the liberty to close this ticket. Feel free to re-open it if you have any objections.

PR #478, opened 2021-03-23
LOG4J2-3054: BasicContextSelector takes the default context into account

carterkozak:
https://issues.apache.org/jira/browse/LOG4J2-3054


carterkozak:
Incoming test case, will conflict with the other branch I'm working on, however I'd prefer to merge this first and avoid duplicating bugs in the new BasicAsyncLoggerContextSelector.

carterkozak:
merged

PR #479, opened 2021-03-24
LOG4J2-2940 StackWalker optimization and avoidance

carterkozak:
This adds a new `BasicAsyncLoggerContextSelector` to allow fully asynchronous logging without the overhead of classloader lookups, as well as an improvement to walk the stack at most once rather than twice when accessing slf4j loggers.


carterkozak:
merged

PR #480, opened 2021-03-28
Master java11

rgoers:
Moves master to Java11. 

1. Removes all -java9 modules.
2. Required version upgrades to lots of stuff.
3. Log4j-api and Log4j-plugins are true JPMS modules. Not much else is yet. Getting a test-jar created that is usable by down-stream components required doing some heaving hacking since the classes in the test jars being passed to downstream modules cannot be in the same package as the component they originated from when the modulepath is being used. Any classes that belong in a test jar are now placed in a java-test directory that is compiled and packaged before the "normal" test classes are compiled. This work has not been done for log4j-core.
2. PreciseClock is not supported. When enabled SystemClock causes the zero GC tests to fail on everything due to Instants being created for the timestamp being logged.
3. tests in log4j-osgi are disabled. They fail and I have no idea why.


log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -16,75 +16,23 @@
...
- * behavior for the rest of Java 7. However, the method is deprecated in Java 8, and current Java 9 development has not
- * addressed this API. Therefore, the functionality of this class cannot be relied upon for all future versions of Java.
- * It does, however, work just fine in Sun JDK 1.6, OpenJDK 1.6, Oracle/OpenJDK 1.7, and Oracle/OpenJDK 1.8. Other Java
- * environments may fall back to using {@link Throwable#getStackTrace()} which is significantly slower due to
- * examination of every virtual frame of execution.
- * </p>
+ * <em>Consider this class private.</em> Determines the caller's class.
  */
-public final class StackLocator {
+public class StackLocator {
```
carterkozak, line 29: Might as well keep around the `final` modifier.
```suggestion
public final class StackLocator {
```

log4j-core/src/main/java/org/apache/logging/log4j/core/time/internal/SystemClock.java
```diff
@@ -16,13 +16,19 @@
...
+
 import org.apache.logging.log4j.core.time.Clock;
+import org.apache.logging.log4j.core.time.MutableInstant;
+import org.apache.logging.log4j.core.time.PreciseClock;
 
 /**
  * Implementation of the {@code Clock} interface that returns the system time.
+ * @since 2.11.0
  */
+// Precise clock is not implemented because the instant() method in the init method is not garbage free.
```
carterkozak, line 29: Hmm, this would result in a regression in some places, we use microsecond precision everywhere at this point at work since events can be scrambled within millisecond buckets by our indexer. I suspect we'd be better of implementing PreciseClock and updating our tests, but I have a very specific use case. Either way we should swing back to add another clock implementation to support whichever implementation SystemClock does not.

carterkozak (commented):
I've read through a few files, and plan to continue tomorrow. What I've seen looks great, thanks for all the work!

log4j-core/src/main/java/org/apache/logging/log4j/core/time/internal/SystemClock.java
rgoers: I agree. If it is possible to detect that we are not in "garbage free mode" then we could go ahead with the init method. I admittedly took a shortcut has I spent nearly a week getting this done and I didn't want to solve every problem in Log4j.

log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocator.java
rgoers: Interesting. I didn't actually modify the class. Just moved the version from the Java9 branch in. I wonder why this and the @PerformanceSensitive annotation weren't ever added to that.

perry2of5:
Building on windows, Zulu Java 11, Maven 3.6.3, I see this warning in log4j-api, log4j-plugins, and others:
[WARNING] * Required filename-based automodules detected: [org.osgi.core-6.0.0.jar]. Please don't publish this project to a public artifact repository!

rgoers:
@perry2of5 Thanks. That is likely related to why I couldn't get the log4j-osgi to run its tests. I'll have to figure out what that warning means and fix it.

rschmitt:
Just to confirm: does this mean that the next version of log4j2 will drop support for JDK8?

remkop:
My understanding is that we will use the Java 11 compiler to generate Java-8 bytecode, so Log4j2 can be used with Java 8 at runtime.

garydgregory:
> My understanding is that we will use the Java 11 compiler to generate Java-8 bytecode, so Log4j2 can be used with Java 8 at runtime.

I do not think so, 2.x will remain Java 8, but 3.0 in master will be Java 11 only.

You should never use Java 11 to compile to Java 8 unless you are compiling against the JRE 8 jars (which I am not sure makes sense) because Java has changed some API signatures in the NIO/Buffer area.

remkop:
My bad, thanks Gary!

rgoers:
Yup, Gary is correct. The 2.x releases will remain on Java 8. The 3.0 releases will require Java 11.

Ralph

rschmitt:
Where can I read more about the plans for 3.0?

rgoers:
You would have to wade through the developers mailing list. We've talked about creating confluence pages but never done it.

PR #481, opened 2021-04-02
LOG4J2-3060 Fix integer overflow in DefaultErrorHandler

Noobgam:
When configured incorrectly, the ErrorHandler::error method will be called very often. As an example of such behavior, you can try stopping the appender manually.

This will result in `ErrorHandler::error("Attempted to append to non-started appender {...}")` which may or may not be called often. In our scenario, this happened ~4k-20k times per second. Although everything seems normal at first, exceptionCount keeps getting incremented until it overflows after 2^31 increments.

When it overflows, every such method call will enter the branch leading to very high CPU consumption in useless code and a significant spike in `System.out` throughput
image

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
@@ -45,18 +45,13 @@ public class DefaultErrorHandler implements ErrorHandler {
         this.appender = appender;
     }
 
-
     /**
      * Handle an error with a message.
      * @param msg The message.
      */
     @Override
     public void error(final String msg) {
-        final long current = System.nanoTime();
-        if (current - lastException > EXCEPTION_INTERVAL || exceptionCount++ < MAX_EXCEPTIONS) {
-            LOGGER.error(msg);
-        }
-        lastException = current;
+        handleErrorMessage(msg, null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


carterkozak:
Thank you for reporting, great find! Would you mind filing jira issue for tracking [here](https://issues.apache.org/jira/browse/LOG4J2)?

Noobgam:
Sure https://issues.apache.org/jira/browse/LOG4J2-3060

rgoers (approved):
This looks good to me. It needs to be back-ported to releaes-2.x. I believe we should be able to cherry pick the commit.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
```diff
@@ -84,14 +75,24 @@ public void error(final String msg, final Throwable t) {
      */
     @Override
     public void error(final String msg, final LogEvent event, final Throwable t) {
+        handleErrorMessage(msg, t);
+        if (!appender.ignoreExceptions() && t != null && !(t instanceof AppenderLoggingException)) {
+            throw new AppenderLoggingException(msg, t);
+        }
+    }
+
+    private void handleErrorMessage(
```
rgoers, line 84: All the parameters really should be on the method declaration since it isn't going to exceed the max line length.

rgoers (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
Noobgam: Fixed style and squashed with a reasonable commit name

zhurs:
@rgoers
I met this problem in my production.
What can we do to merge this fix?

PR #484, opened 2021-04-08
LOG4J2-3056 Refactor MD5 usage for sharing sensitive information

vy:


rgoers (approved):
This looks good to me.

jvz (approved):

carterkozak (approved):
Thanks!

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
```diff
@@ -16,44 +16,55 @@
...
+     *
+     * @return string composed of 32 hexadecimal digits of the calculated hash
+     */
+    public static String md5(final String input) {
+        Objects.requireNonNull(input, "input");
         try {
             final MessageDigest digest = MessageDigest.getInstance("MD5");
-            digest.update(string.getBytes());
+            digest.update(input.getBytes());
             final byte[] bytes = digest.digest();
```
Marcono1234, line 55: Could also simplify this to use [`digest(byte[])`](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html#digest-byte:A-):
```suggestion
            final byte[] bytes = digest.digest(input.getBytes());
```

However, as written in LOG4J2-3056, `String.getBytes()` uses the default charset which is very likely not the correct one.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
```diff
@@ -16,44 +16,55 @@
...
                 if (hex.length() == 1) {
                     md5.append('0');
                 }
                 md5.append(hex);
             }
             return md5.toString();
-        } catch (final Exception ex) {
-            return string;
+        } catch (final NoSuchAlgorithmException error) {
+            throw new RuntimeException(error);
```
Marcono1234, line 66: Might be good to add a comment explaining why it is acceptable to throw an unspecific `RuntimeException` here, that is, the [documentation](https://docs.oracle.com/javase/8/docs/api/java/security/MessageDigest.html) requires that every Java platform must support MD5.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
```diff
@@ -16,44 +16,55 @@
...
+     * @return string composed of 32 hexadecimal digits of the calculated hash
+     */
+    public static String md5(final String input) {
+        Objects.requireNonNull(input, "input");
         try {
             final MessageDigest digest = MessageDigest.getInstance("MD5");
-            digest.update(string.getBytes());
+            digest.update(input.getBytes());
             final byte[] bytes = digest.digest();
             final StringBuilder md5 = new StringBuilder();
```
Marcono1234, line 56: Could make this slightly more efficient by sizing the builder accordingly:
```suggestion
            final StringBuilder md5 = new StringBuilder(bytes.length * 2);
```

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
```diff
@@ -16,44 +16,55 @@
...
 
-    private static final int MASK = 0xff;
-
-    private NameUtil() {
-    }
+    private NameUtil() {}
 
     public static String getSubName(final String name) {
-        if (Strings.isEmpty(name)) {
+        if (Strings.isBlank(name)) {
```
Marcono1234, line 33: In theory this is a behavior change unrelated to this issue, might be good to omit it from this pull request?
(Though it is unlikely that this has a big impact)

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
```diff
@@ -16,44 +16,55 @@
...
 
-    public static String md5(final String string) {
+    /**
+     * Calculates the <a href="https://en.wikipedia.org/wiki/MD5">MD5</a> hash
+     * of the given input string.
+     * <p>
+     * <b>MD5 has severe vulnerabilities and should not be used for sharing any
+     * sensitive information.</b> This function should only be used to create
+     * unique identifiers, e.g., configuration element names.
+     *
```
Marcono1234, line 47: If you already add such good and extensive documentation, it be good to add `@param input` as well.

Marcono1234 (commented):
Thanks for the fast response from all of you on the issue!
I have left some review comments, hopefully they are useful.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
vy: Thanks, I've applied your suggestion. Though there is not much we can do for the missing charset, since at the present call sites they don't have access to that and allowing that access breaks the method footprints.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
vy: Applied.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
vy: Applied.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
vy: I can buy that, fixed.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/NameUtil.java
vy: Applied.

vy:
@Marcono1234, I am really thankful for all your reviews, not just this particular one. Would you mind sharing your real name, please, so that I can use it rather than `Marcono1234` in `changes.xml`?

I see that you are reading through the sources, if I may say, up to its minute details. I am curious about your motivation. Would you liked to be involved in more? Do you have any other (architectural?) feedback too? If so, please don't hesitate to share them in the `dev` mailing list.

Marcono1234:
> Would you mind sharing your real name, please, so that I can use it rather than `Marcono1234` in `changes.xml`?

Thanks for crediting me! I am fine with `changes.xml` crediting me as Marcono1234, so unless that is an issue for you or the Apache Software Foundation, I would like to leave it as is.

> I see that you are reading through the sources, if I may say, up to its minute details. I am curious about your motivation.

Some issues I raised were based on me using Log4j 2. However, this one I noticed through a [CodeQL](https://codeql.github.com/docs/) query. I have recently been experimenting with CodeQL and wrote a query for finding improper handling of `NoSuchAlgorithmException` (query can be found [here](https://github.com/Marcono1234/codeql-java-queries/blob/390603d95c2aae5fafe081982a6cd8690470d2be/codeql-custom-queries-java/queries/security/NoSuchAlgorithmException-not-properly-handled.ql)), which spotted this method.

> Do you have any other (architectural?) feedback too?

I have created some Jira issues in the past, but I don't have any extensive feedback to share.

> If so, please don't hesitate to share them in the `dev` mailing list.

I am not a big fan of mailing lists, even though your "Pony Mail!" has a pretty neat web-interface. However, maybe I will change my mind regarding joining the mailing list in the future.

PR #486, opened 2021-04-29
LOG4J2-2965: Fix JUL/AsyncLoggerContext deadlock

carterkozak:


src/changes/changes.xml
```diff
@@ -75,6 +75,9 @@
       <action issue="LOG4J2-2940" dev="ckozak" type="fix">
         Slf4j implementations walk the stack at most once rather than twice to determine the caller's class loader.
       </action>
+      <action issue="LOG4J2-2965" dev="ckozak" type="fix">
+        Fixed a deadlock between the AsyncLoggerContextSelector and java.util.logging.LogManager by updating Disruptor to 3.4.4.
```
carterkozak, line 79: Unclear if this should be an `update` or `fix`.

carterkozak:
committed

PR #489, opened 2021-05-04
LOG4J2-3087 Fix race in JsonTemplateLayout where a timestamp could end up unquoted

klaren:
[LOG4J2-3087](https://issues.apache.org/jira/browse/LOG4J2-3087)

There is a "race" when constructing a `JsonTemplateLayout` that can produce invalid JSON.

In the constructor of `FormatResolverContext` the `formattedTimestampBuilder` is populated with a formated, but unquoted, timestamp based on the current time. Later in `FormatResolver.resolve()` we check if the log event matches the current time and send the cached value to `jsonWriter.writeRawString()`, which can be unquoted if the layout and log event was created in the same millisecond.

### DIFF
--- a/log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutTimestampTest.java
+++ b/log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutTimestampTest.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.log4j.layout.template.json;
+
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.DefaultConfiguration;
+import org.junit.jupiter.api.RepeatedTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JsonTemplateLayoutTimestampTest
+{
+    private static final Configuration CONFIGURATION = new DefaultConfiguration();
+
+    /**
+     * Tests LOG4J2-3087 race when creating layout on the same instant as the log event would result in an unquoted date in the JSON
+     */
+    @RepeatedTest( 20 )
+    void test_timestamp_pattern_race() {
+        final JsonTemplateLayout layout = JsonTemplateLayout
+                .newBuilder()
+                .setConfiguration(CONFIGURATION)
+                .setEventTemplate("{\"t\":{\"$resolver\":\"timestamp\",\"pattern\":{\"format\":\"yyyy-MM-dd\"}}}")
+                .build();
+
+        final LogEvent logEvent = LogEventFixture.createLiteLogEvents(1).get(0);
+        final String json = layout.toSerializable(logEvent);
+        assertThat(json).matches("\\{\"t\":\"\\d{4}-\\d{2}-\\d{2}\"}\\R*");
+    }
+}
--- a/log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolverTest.java
+++ b/log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolverTest.java
@@ -14,32 +14,38 @@
  * See the license for the specific language governing permissions and
  * limitations under the license.
  */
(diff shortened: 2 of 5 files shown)


src/changes/changes.xml
```diff
@@ -66,6 +66,9 @@
         Allow a PatternSelector to be specified on GelfLayout.
       </action>
       <!-- FIXES -->
+      <action issue="LOG4J2-3087" dev="klaren" type="fix">
```
vy, line 69:
```suggestion
      <action issue="LOG4J2-3087" dev="vy" type="fix" due-to="Anton Klarén">
```

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutTimestampTest.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.logging.log4j.layout.template.json;
+
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.DefaultConfiguration;
+import org.junit.jupiter.api.RepeatedTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JsonTemplateLayoutTimestampTest
```
vy, line 26: I have two requests here:

1. Move this class to `...log4j.layout.template.json.resolver.TimestampResolverTest` – `JsonTemplateLayoutTest` is pretty bloated and I want to create `*ResolverTest` classes for each individual resolver.
2. Force some concurrent access here by using, e.g., an `ExecutorService`. (`JsonTemplateLayoutConcurrentEncodeTest` might come handy for inspiration.)

vy (changes requested):
Great finding and spot on fix @klaren! Thanks so much! I have some minor change requests, if you don't mind.

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutTimestampTest.java
klaren: I moved it and reworked it to target `TimestampResolver` directly.
Regarding a concurrent test; all public methods are marked as synchronized so I don't think that will be necessary. And this issue is present for a single thread as well.

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolverTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+class TimestampResolverTest
+{
+    private static final TemplateResolverConfig TEMPLATE_RESOLVER_CONFIG = new TemplateResolverConfig(
+            Collections.singletonMap("pattern", Collections.singletonMap("format", "yyyy-MM-dd")));
+
+    /**
+     * Tests LOG4J2-3087 race when creating layout on the same instant as the log event would result in an unquoted date in the JSON
+     */
+    @RepeatedTest( 20 )
```
klaren, line 36: Alternatively, we could expose `TimestampResolver.FormatResolverContext#calendar` and set the same time on it as the event. This will be more predictable and would not require us to "brute-force" it.

vy: I would definitely favor a deterministic approach like that. This said, judging from the sources of `TimestampResolver`, code needs to be polluted a bit. Nevertheless, it is doable:

```diff
Index: log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java
===================================================================
diff --git a/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java b/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java
--- a/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java	(revision a130d8ade602f36a1416eb9ef962935ea07dc4ea)
+++ b/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java	(date 1620285824290)
@@ -195,6 +195,8 @@
 
     private final EventResolver internalResolver;
 
+    static Calendar CALENDAR_TEST_PROBE = null;
+
     TimestampResolver(final TemplateResolverConfig config) {
         this.internalResolver = createResolver(config);
     }
@@ -229,7 +231,7 @@
                 final FastDateFormat timestampFormat) {
             this.timestampFormat = timestampFormat;
             this.formattedTimestampBuilder = new StringBuilder();
-            this.calendar = Calendar.getInstance(timeZone, locale);
+            this.calendar = CALENDAR_TEST_PROBE = Calendar.getInstance(timeZone, locale);
             timestampFormat.format(calendar, formattedTimestampBuilder);
         }
 

```
I would appreciate it if you can go forward with this proposal of yours.

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutTimestampTest.java
vy: Thank you!

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolverTest.java
klaren: I took a different approach instead without using a static field since that will break when running any concurrent tests.

vy: :man_facepalming: Yours is way better!

PR #491, opened 2021-05-04
LOG4J2-3088 LOG4J2-3089 Fix windows build master

perry2of5:
This pull fixes the windows build. I have tested on windows 10 using maven 3.6.3 and Zulu's OpenJDK 11 build.

It adds an annotation @DisableOnWindows to prevent the FileOutputTest test from running on Windows in log4j-core. This test cannot pass because the listener to the StatusLogger is never shut down which means the file handle to the log file is never released. On Windows, this means the log file can't be deleted which fails the test.

It fixes the XML comparison in ConfigurationBuilderTest in log4j-kafka to compare the XML as XML documents instead of as Strings which should prevent the OS dependent quirks from causing the test to fail.

It also fixes an intermittent nuisance in JsonTemplateLayoutNullEventDelimiterTest which fails when the port is unavailable on windows. I changed the test to catch the exception and print an error rather than fail the build. Ideally we'd find an available port dynamically, write out the XML config to a temp file with the targeted port, run the test, and then clean up the temporary config XML file. I took the easy way out here.


log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutNullEventDelimiterTest.java
```diff
@@ -71,13 +71,24 @@ public void test() throws Exception {
...
-
-        } catch (BindException bindException) {
-            System.err.println("Required port " + PORT + " not available, skipping " + getClass().getSimpleName());
+        } finally {
+            tempConfig.delete();
         }
 
     }
 
+    public void makeTempConfig(final int serverPort, final File tempConfigFile) throws IOException {
```
vy, line 80: This approach looks way better, thanks @perry2of5! Would you use `final` modifier on variables, wherever possible, please?

perry2of5: Yes.

log4j-core/src/test/java/org/apache/logging/log4j/junit/DisableOnWindows.java
```diff
@@ -0,0 +1,31 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.RetentionPolicy;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(DisableOnWindowsCondition.class)
+public @interface DisableOnWindows {
```
perry2of5, line 30: Should we make this more general? DisableOnOS? Then we'd have @DisableOnOS("windows") rather than @DisableOnWindows. Also, is this the correct place in the project structure to put this helper class? Or should it get moved in with other classes that help with testing? If so, where would that be?

perry2of5: It looks like I should  have just used @DisabledOnOs(WINDOWS). I'll drop these new classes and switch it to use org.junit.jupiter.api.condition.DisabledOnOs

perry2of5: I switched the JUnit @DisabledOnOs(WINDOWS), built, tested, and pushed to github.

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutNullEventDelimiterTest.java
```diff
@@ -23,38 +23,38 @@
...
-                "nullEventDelimitedJsonTemplateLayoutLogging.xml");
-    }
-
-    // Note that this port is hardcoded in the configuration file too!
-    private static final int PORT = 50514;
-
     @Test
     public void test() throws Exception {
-
+        File tempConfig = File.createTempFile("nullEventDelimitedJsonTemplateLayoutLogging", ".xml");
```
vy:
```suggestion
        final File tempConfig = File.createTempFile("nullEventDelimitedJsonTemplateLayoutLogging", ".xml");
```

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutNullEventDelimiterTest.java
```diff
@@ -71,13 +71,24 @@ public void test() throws Exception {
...
+            tempConfig.delete();
         }
 
     }
 
+    public void makeTempConfig(final int serverPort, final File tempConfigFile) throws IOException {
+        try (FileOutputStream fos = new FileOutputStream(tempConfigFile);
+                OutputStreamWriter osw = new OutputStreamWriter(fos, StandardCharsets.UTF_8);
+                BufferedWriter bw = new BufferedWriter(osw);) {
+            for (String line : Files.readAllLines(
```
vy:
```suggestion
        try (final FileOutputStream fos = new FileOutputStream(tempConfigFile);
                final OutputStreamWriter osw = new OutputStreamWriter(fos, StandardCharsets.UTF_8);
                final BufferedWriter bw = new BufferedWriter(osw);) {
            for (final String line : Files.readAllLines(
```

log4j-layout-template-json/src/test/java/org/apache/logging/log4j/layout/template/json/JsonTemplateLayoutNullEventDelimiterTest.java
perry2of5: This is a good change to make. I thought I already made it as commit d70da9e7214d7baa3643363b849e5ec124b7b748. I see that commit in my branch, did it not show up here? I'm still figuring out git so any advice on how I should have made this work.

PR #496, opened 2021-05-24
Allow JsonLayout output JSON `null` for ContextMap

porshkevich:
For now JsonLayout write string "null" to result output if ContextMap contains values equals Java Null.
But JSON have own `null` concept. I change ContextDataSerializer for write normal JSON `null` in case ContextMap contains Null.


vy:
Thanks for the contribution @porshkevich, much appreciated. Please try to target `release-2.x` branch next time. (I will merge your changes and cherry-pick them onto `release-2.x`in a minute.)

I am also curious about your motivation for using `JsonLayout` rather than its successor `JsonTemplateLayout`. Have you considered switching to `JsonTemplateLayout`?

porshkevich:
Thanks for fast merge. I will take into account your wishes for the future.

`JsonTemplateLayout` It's like a novelty. I missed his appearance a bit. When projects go through a dependency update cycle, i will try to use new layout. Thank you for that news.

PR #505, opened 2021-06-04
Fix incorrect LogManager accessor used by LOG4J2-2940

carterkozak:
The `getContext()` accessor with no args has been used as a fallback
for Log4jLoggerFactory slf4j implementations for a while, but it's
much more likely to be used now that LOG4J2-2940 is resolved.
Without the `false` argument, the first slf4j LoggerFactory.getLogger
call will not cause log4j to initialize itself.


carterkozak:
merged

PR #515, opened 2021-06-14
Update dependabot to track github action and dockerfile dependencies

chenrui333:


rgoers:
Can you please explain why we want to do this? To be honest, I don't find much value in dependabot in the first place. All it really tells me is what the most current version of a dependency is. That doesn't mean Log4j 2 should use it.

chenrui333:
> Can you please explain why we want to do this? To be honest, I don't find much value in dependabot in the first place. All it really tells me is what the most current version of a dependency is. That doesn't mean Log4j 2 should use it.

Dependabot actually can do two things:
- the regular dependency updates, like you said, it is really just a dependency notification thing (but it does have some sort of view in it, commit diff history, release notes, and compatibility score), I generally find it pretty convenient managing the dependencies, but it could be some maintenance overhead.
- security patches, it would raise the security patch PRs

By adding github action and dockerfile tracking, it would enable both.

Let me know what you think.

rgoers:
Thanks but that didn't really answer my question. What does enabling it for actions and dockerfiles do? What is the benefit?

chenrui333:
> Thanks but that didn't really answer my question. What does enabling it for actions and dockerfiles do? What is the benefit?

Like I said, the benefit would be more like notifying about the dependency changes.

rgoers:
dependency changes of what? I understand that the line specifying Maven lets the plugin know to look at the pom.xml to look for dependencies. I have no idea what it means to specify action or dockerfiles. After all, actions are just some instructions of steps to perform. I don't recall versions of anything being specified in an action. I don't believe our dockerfiles do either. So what are these actually checking for new dependencies?

jvz:
We have a Docker file for building that we don't really use for anything. I think we do a pretty good job of updating dependencies normally, but a Dependabot config for Maven plugins would be useful.

rgoers:
@jvz We already have that. This PR adds configuration for docker and actions. Before deciding whether to accept the PR I need to know exactly what that configuration causes to happen.

garydgregory:
We use Dependabot in Apache Commons, super useful IMO, you get a PR and
GitHub build to go along with a dependency change which gives you more
confidence that a dependecy is OK to update.

Gary

rgoers:
@garydgregory We use dependabot here as well, although it isn't nearly as useful as it is in Commons, especially since it doesn't update changes.xml. The issue here isn't dependabot but what the heck does it mean to add it to actions and dockerfiles.

vy:
I am in favor of this change due to:
1. `find . -name Dockerfile` acknowledges that we have `Dockerfile` usages in the code base. It is a good practice to make sure that the employed base images are up-to-date and doesn't use an outdated OS with security vulnerabilities. (I don't think `dependabot` is smart enough to help us with `maven-docker-plugin`-based usages, e.g., in `log4j-layout-template-json`, yet.)
2. GitHub Actions are similar to Log4j, just a library, though targeting the GitHub CI pipeline. (For instance, we use `scacap/action-surefire-report@v1` to publish test results and `actions/upload-artifact@v2` to upload test reports. See `/.github/workflows/main.yml` for details.) The more we wait for an upgrade, the painful it will be. And we all know what kind of a burden it is for library maintainers to keep an old version alive.

rgoers:
Ok. At least now I have an idea of what these are supposed to do. However, looking at the Dockerfile I see no versions of anything. If it is just checking the "uses" declarations on actions that might be helpful to know when new versions are released.

PR #516, opened 2021-06-14
workflow: add action to handle stale PRs

chenrui333:
It would be good to add some stale action to handle the stale PRs.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,16 @@
+name: 'Close stale issues and PRs'
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          stale-pr-message: 'This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.'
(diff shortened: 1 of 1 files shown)


rgoers:
I'm not sure I agree with this. It seems too simplistic. Sometimes PRs are good but we are slow to apply them.  Prior to every release I review all the PRs and close those I believe will never progress.

chenrui333:
yeah, you can add some labels of having stale bot ignoring the `in progress` PRs.

jvz:
I think the best stale not setting would be notifying PR authors when they have merge conflicts to fix the issue. I don't think we get enough PRs to worry about closing old ones; we tend to work in batches anyways.

vy:
Thanks so much for letting us know of this feature, much appreciated @chenrui333! Please don't get demotivated by the questions from the rest of us. This is a part of the deal when you are contributing to a project. Looking forward to your future PRs!

I agree with the comments of @rgoers and @jvz. Log4j crew handles PRs in a rather unorthodox way given the scarcity of development resources, hence I take the liberty of closing this PR without merging it.

garydgregory:
I agree with Ralph, staleness is not interesting here. The work I put in here very much sporadic so the age of a PR rarely matters,.what matters more are conflicts.

PR #520, opened 2021-06-17
LOG4J2-3107: SmtpManager.createManagerName ignores port

sman-81:
This patch is short and affects only one class.

### DIFF
--- a/src/changes/changes.xml
+++ b/src/changes/changes.xml
@@ -31,6 +31,9 @@
          - "remove" - Removed
     -->
     <release version="3.0.0" date="2021-MM-DD" description="GA Release 3.0.0">
+      <action issue="LOG4J2-3107" dev="spannjp" type="fix" due-to="Markus Spann">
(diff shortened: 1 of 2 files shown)


src/changes/changes.xml
```diff
@@ -31,6 +31,9 @@
          - "remove" - Removed
     -->
     <release version="3.0.0" date="2021-MM-DD" description="GA Release 3.0.0">
+      <action issue="LOG4J2-3107" dev="spannjp" type="fix" due-to="Markus Spann">
```
vy, line 34:
```suggestion
      <action issue="LOG4J2-3107" dev="vy" type="fix" due-to="Markus Spann">
```

vy:
@sman-81, thanks so much for the contribution! Could you also add a test to verify this new behavior?

vy:
It might also be better to target this change to `release-2.x` branch, rather than `master`. (I will take care of cherry-pick'ing.)

PR #521, opened 2021-06-17
Rephrase outdated comment in SystemClock

hpple:
Hi!

This comment looks outdated.


vy:
@hpple, thanks for reminding us about that detail. Actually the comment still applies, though the wording needs to be improved indeed: _"The precise clock is not enabled by default, since access to it is not garbage free."_ And also the comment better be situated just above the `log4j2.usePreciseClock` variable. Would you mind updating your PR to adapt this, please?

hpple:
@vy 

done

vy:
@hpple, thanks so much for the contribution and the prompt response! Merging it right away.

PR #527, opened 2021-06-22
Add TimeBasedTriggeringPolicy#modulate default value to the manual

NgSekLong:
(doc)

**TimeBasedTriggeringPolicy** modulate default is `false` is not in the doc which might caused confusion. 

Would like to suggest adding the default as false in the doc for clarity, cheers!


vy:
Thanks for the improvement @NgSekLong! Keep them coming! (I will cherry-pick this onto `release-2.x` too.)

PR #528, opened 2021-06-22
Bugfix for LOG4J2-3107: SmtpManager.createManagerName ignores port

sman-81:
My fix for LOG4J2-3107 changes two files and adds one. The change is now targeted at the release-2.x branch not master.
Hope you find it useful.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/net/SmtpManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/net/SmtpManager.java
@@ -106,7 +106,7 @@ public class SmtpManager extends AbstractManager {
             protocol = "smtp";
         }
 
-        final String name = createManagerName(to, cc, bcc, from, replyTo, subject, protocol, host, port, username, password, isDebug, filterName);
+        final String name = createManagerName(to, cc, bcc, from, replyTo, subject, protocol, host, port, username, isDebug, filterName);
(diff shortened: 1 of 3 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/net/SmtpManager.java
```diff
@@ -166,9 +172,7 @@ private static String createManagerName(
         sb.append(isDebug ? ":debug:" : "::");
         sb.append(filterName);
 
-        final String hash = NameUtil.md5(sb.toString());
-        return "SMTP:" + hash;
-
+        return "SMTP:" + sb.toString();
```
vy, line 175: Given `password` is in `sb`, this is not a good idea. Would you mind reverting this line, please?

sman-81: Hello @vy,
thanks for your feedback.
I've removed password from creation of smtp manager name, as it is irrelevant for creating a unique per-configuration name/key (cannot have same user name with different password on same host).
Greetings,
Markus

vy:
Thanks so much for the fix @sman-81!

jvz:
@vy your cherry pick broke master. Can you take a look?

vy:
[The associated GitHub Actions run](/apache/logging-log4j2/actions/runs/970729092) has succeeded. What is exactly broken?

jvz:
Ralph figured it out earlier today. It was in the wrong package in master.

PR #542, opened 2021-07-01
LOG4J2-2785: Added support for pattern Layout to abbreviate logger or class names except n rightmost words

sman-81:
Hi, I've added a specialized abbreviator named `DynamicWordAbbreviator` to address the specific requirement of this feature request. Javadoc added and JUnit test updated to cover the new feature.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
@@ -17,10 +17,10 @@
 package org.apache.logging.log4j.core.pattern;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import org.apache.logging.log4j.util.PerformanceSensitive;
-
 
 /**
  * NameAbbreviator generates abbreviated logger and class names.
@@ -55,6 +55,11 @@ public abstract class NameAbbreviator {
                 return DEFAULT;
             }
 
+            NameAbbreviator dwa = DynamicWordAbbreviator.create(trimmed);
+            if (dwa != null) {
+                return dwa;
+            }
+
             boolean isNegativeNumber;
             final String number;
 
@@ -82,7 +87,7 @@ public abstract class NameAbbreviator {
                         isNegativeNumber? MaxElementAbbreviator.Strategy.DROP : MaxElementAbbreviator.Strategy.RETAIN);
             }
 
-            final ArrayList<PatternAbbreviatorFragment> fragments = new ArrayList<>(5);
+            final List<PatternAbbreviatorFragment> fragments = new ArrayList<>(5);
             char ellipsis;
             int charCount;
             int pos = 0;
@@ -313,6 +318,12 @@ public abstract class NameAbbreviator {
             }
             return nextDot;
         }
+
+        @Override
+        public String toString() {
+            return String.format("%s[charCount=%s, ellipsis=%s]",
+                    getClass().getSimpleName(), charCount, Integer.toHexString(ellipsis));
+        }
     }
 
     /**
@@ -364,5 +375,11 @@ public abstract class NameAbbreviator {
                 }
             }
         }
+
+        @Override
+        public String toString() {
+            return String.format("%s[fragments=%s]", getClass().getSimpleName(), Arrays.toString(fragments));
+        }
     }
+
 }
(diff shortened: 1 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
```diff
@@ -365,4 +372,51 @@ public void abbreviate(final String original, final StringBuilder destination) {
...
+        }
+
+        @Override
+        public void abbreviate(String original, StringBuilder destination) {
+            if (original != null && destination != null) {
+                String[] words = original.split("\\.");
+                for (int i = 0; i < words.length - rightWordCount; i++) {
+                    words[i] = words[i].substring(0, 1);
+                }
+                destination.append(String.join(".", words));
```
carterkozak, line 416: This looks very expensive. String.split uses regex and allocates a new string for each segment. Substring allocates another string per abbreviated segment, and the final String.join creates a string representation of the data that's added to the StringBuilder. I think allocation can be entirely avoided here by tacking indexes, and writing directly to the Destination buffer.

It could be interesting to run the benchmarks comparing `1.1*` to the existing `1.`

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
```diff
@@ -365,4 +372,51 @@ public void abbreviate(final String original, final StringBuilder destination) {
...
+        private DynamicWordAbbreviator(int rightWordCount) {
+            this.rightWordCount = rightWordCount;
+        }
+
+        @Override
+        public void abbreviate(String original, StringBuilder destination) {
+            if (original != null && destination != null) {
+                String[] words = original.split("\\.");
+                for (int i = 0; i < words.length - rightWordCount; i++) {
+                    words[i] = words[i].substring(0, 1);
```
carterkozak, line 414: Probably worthwhile to add test coverage for an input string along the lines of `org..example.Name` with two periods in a row, I think this substring will throw an IndexOutOfBoundsException in that case.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
sman-81: Thanks for your feedback.
Afaik `String#split` does not create a `Pattern` object for certain one-character patterns such as here.
I've rewritten the method for improved efficiency.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NameAbbreviator.java
sman-81: Thanks, you're right. I've fixed this and improved test coverage.

carterkozak:
When I have some time to tinker, I think this feature could be added to the existing abbreviator. All existing patterns effectively use `rightWordCount=1`, however we could make that configurable and allow existing options to work with this.

Something along the lines of 
```java
int limit = input.length();
for (int i = 0; I < rightWordCount; I++) {
    limit = input.lastIndexOf('.', limit);
    if (limit < 0) {
        return input;
    }
}
// call into standard abbreviator.
```

sman-81:
@carterkozak Here's another small update to my pull request.
Can you merge it into the release branch please?

carterkozak:
@sman-81 we are all volunteers and have limited time to devote to review. Your most recent comment comes across as entitled and rude. Please work on that. In your position, I might have replied on the PR with "Is there anything I can do to get this into a mergable state?"

> you listen to their criticism
> you incorporate their feedback

I asked you twice to avoid `string.split` in favor of `indexOf`/`lastIndexOf`, but it's still used.

sman-81:
@carterkozak

I'm a volunteer just like you.
My pull request is from July 1st, thus seven months old by now!
The last comment from anyone other than me was August 12, 2021.
Such discourages people from contributing to the project.

> I asked you twice to avoid string.split in favor of indexOf/lastIndexOf, but it's still used.

And you are wrong. `String#split` does not create a Pattern object for certain one-character patterns such as here, as I have pointed out to you (see comment on August 5, 2021).

carterkozak:
The point is that `string.split` is inefficient, regardless of internal details because it's allocating several new strings unnecessarily.

sman-81:
Hi there @carterkozak now `String#split` is gone. I have also squashed commits into one. Please take a look.

rgoers:
This patch was manually extracted and merged since the backing repo has been deleted.

sman-81:
Hard to believe my PR actually got merged after a little over 1,5 years :)

PR #549, opened 2021-07-14
LOG4J2-2955: More precise file size expression parsing; fail-fast behavior for invalid expressions

sman-81:
…avior for invalid expressions


rgoers:
I started to apply this and got an error from revapi that compatibility had been broken. The previous version passed in a default value and used that if there was an error. This version throws an exception which will cause Appender initialization to fail. I think I prefer logging an error and using the default.

rgoers:
Part of the patch was applied.

PR #553, opened 2021-08-03
LOG4J2-3083 Fix slf4j calling class lookup using both accessors

carterkozak:
`LoggerFactory.getLogger(name/class)` as well as `LoggerFactory.getILoggerFactory().getLogger(name)`.

### DIFF
--- a/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
+++ b/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
@@ -19,6 +19,7 @@ package org.apache.logging.log4j.util;
 import java.util.List;
 import java.util.Stack;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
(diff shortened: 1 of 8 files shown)


log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -40,6 +41,16 @@ public static StackLocator getInstance() {
     private StackLocator() {
     }
 
+    public Class<?> getCallerClass(final Class<?> sentinelClass, final Predicate<Class<?>> callerPredicate) {
```
carterkozak, line 44: Missing null checks
```suggestion
    public Class<?> getCallerClass(final Class<?> sentinelClass, final Predicate<Class<?>> callerPredicate) {
    if (sentinelClass == null) {
            throw new IllegalArgumentException("sentinelClass cannot be null");
        }
        if (callerPredicate == null) {
            throw new IllegalArgumentException("callerPredicate cannot be null");
        }
```

PR #557, opened 2021-08-04
LOG4J2-3133: log4j-slf4j-impl provides new singleton accessors from slf4j-api 1.7.14

carterkozak:
See https://github.com/qos-ch/slf4j/commit/ea3cca72cd5a9329a06b788317a17e806ee8acd0
and https://jira.qos.ch/browse/SLF4J-347


carterkozak:
merged

PR #558, opened 2021-08-04
Add missing JsonLayout properties

sanyarnd:
Add 3 missing properties from https://logging.apache.org/log4j/2.x/manual/layouts.html


vy:
Thanks so much for the contribution @sanyarnd!

PR #565, opened 2021-08-22
LOG4J2-2808 LoggerContext should skip looking up localhost when hostName is configured

asapha:
https://issues.apache.org/jira/browse/LOG4J2-2808

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
@@ -611,12 +611,7 @@ public class LoggerContext extends AbstractLifeCycle
 
             try { // LOG4J2-719 network access may throw android.os.NetworkOnMainThreadException
                 // LOG4J2-2808 don't block unless necessary
-                map.computeIfAbsent("hostName", new Function<String, String>() {
-                    @Override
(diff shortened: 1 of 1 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -609,7 +610,13 @@ public Configuration setConfiguration(final Configuration config) {
             final ConcurrentMap<String, String> map = config.getComponent(Configuration.CONTEXT_PROPERTIES);
 
             try { // LOG4J2-719 network access may throw android.os.NetworkOnMainThreadException
-                map.putIfAbsent("hostName", NetUtils.getLocalHostname());
+                // LOG4J2-2808 don't block unless necessary
+                map.computeIfAbsent("hostName", new Function<String, String>() {
```
garydgregory, line 614: You don't have to be so verbose here, use () -> or ::

asapha: Fixed. 👍

carterkozak:
I've applied the fix as 177a85708535cad6a3e0af9468f323135cc91fbd, thanks for the contribution!

PR #567, opened 2021-08-23
LOG4J2-3142: Use log event time millis for timestamp in log event adapter

jjmms:
https://issues.apache.org/jira/browse/LOG4J2-3142

When using the `LogEventAdapter.getTimeStamp()` we should return the `LogEvent.getTimeMillis()` as this is the equivalent timestamp value from Log4j2 to Log4j1 for compatibility.

### DIFF
--- a/log4j-1.2-api/src/test/java/org/apache/log4j/config/RewriteAppenderTest.java
+++ b/log4j-1.2-api/src/test/java/org/apache/log4j/config/RewriteAppenderTest.java
@@ -57,6 +57,7 @@ public class RewriteAppenderTest {
         Logger logger = LogManager.getLogger("test");
         ThreadContext.put("key1", "This is a test");
         ThreadContext.put("hello", "world");
+        long logTime = System.currentTimeMillis();
         logger.debug("Say hello");
         LoggerContext context = (LoggerContext) org.apache.logging.log4j.LogManager.getContext(false);
         Configuration configuration = context.getConfiguration();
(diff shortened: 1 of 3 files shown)


carterkozak:
Looks reasonable to me. Any chance you can provide a minimal test case? Something as simple as an additional assert in `RewriteAppenderTest` should do the trick, we can verify the value is greater than or equal to the system time when the test starts.

jjmms:
I have written a test but unable to run it due to errors in Maven build and test per CI errors:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project log4j-core: Compilation failure: Compilation failure:
[ERROR] error: module not found: org.fusesource.jansi
[ERROR] error: module not found: org.apache.commons.compress
[ERROR] error: module not found: com.fasterxml.jackson.dataformat.yaml
[ERROR] error: module not found: com.fasterxml.jackson.dataformat.xml
[ERROR] error: module not found: com.fasterxml.jackson.databind
[ERROR] error: module not found: com.fasterxml.jackson.core
[ERROR] error: module not found: com.conversantmedia.disruptor
[ERROR] error: module not found: org.osgi.framework
[ERROR] error: module not found: org.jctools.core
[ERROR] error: module not found: com.lmax.disruptor
[ERROR] error: module not found: org.apache.logging.log4j.plugins
[ERROR] error: module not found: org.apache.logging.log4j
```

carterkozak:
Sorry, the `master` branch isn't in the best state, I generally target `release-2.x` initially and port commits to master. CI builds are running correctly on that branch.

jjmms:
Updated PR with a test case to ensure log event timestamps happen after initial log time.

log4j-1.2-api/src/test/java/org/apache/log4j/config/RewriteAppenderTest.java
```diff
@@ -73,5 +74,6 @@ public void testRewrite() throws Exception {
         assertNotNull("No properties in the event", events.get(0).getProperties());
         assertTrue("Key was not inserted", events.get(0).getProperties().containsKey("key2"));
         assertEquals("Key value is incorrect", "Log4j", events.get(0).getProperties().get("key2"));
+        assertTrue("Timestamp is before point of logging", events.get(0).getTimeStamp() >= logTime);
     }
```
garydgregory, line 78: I'm not sure if we've fixed this elsewhere but I've actually seen this kind of test fail when running builds just around the daylight saving time change. Probably not worth making the test more complex.

jjmms: The milliseconds since epoch which the timestamp represents are not influenced by timezones and daylight saving time.

carterkozak: That's correct, this should only be able to fail if the system clock is updated mid-test, which I wouldn't expect.

garydgregory:
This PR is still based on master, it needs to be for the `release-2.x` branch.

jjmms:
Updated branch and PR to target `release-2.x` branch.

carterkozak:
Thank you for the contribution, @jjmms. Once the tests complete I'll get this merged into 2.x and master.

carterkozak:
I've added a changelog entry and committed this to release-2.x and master branches: b1596ed52406dbb93865b99ce7c9563a04258d22

PR #569, opened 2021-08-25
Log4j2-3147 fix kubernetes spring lookup

maccamlc:


maccamlc:
Bug fix so that the isClassAvailable check correctly looks for presence of [org/apache/logging/log4j/spring/boot/SpringEnvironmentHolder.java](https://github.com/apache/logging-log4j2/blob/master/log4j-spring-boot/src/main/java/org/apache/logging/log4j/spring/boot/SpringEnvironmentHolder.java) 

It seems this was broken in 1.14.0

PR #571, opened 2021-08-28
doc: fix typos

yaojingguo:


vy:
@yaojingguo, thanks so much for taking time to submit a PR to improve the documentation! Much appreciated! Though sadly these are not typos. For instance, linked statements apply to both `eventTemplate` and `eventTemplateUri`, hence the `eventTemplate[Uri]`.

yaojingguo:
>  For instance, linked statements apply to both eventTemplate and eventTemplateUri, hence the eventTemplate[Uri].

@vy Could you please kindly show me how to use `[Uri]`? For example,  what value should I pass to `Uri ` in `log4j.layout.jsonTemplate.eventTemplate[Uri]`?

https://logging.apache.org/log4j/2.x/manual/json-template-layout.html has:

截屏2021-08-30 下午11 03 24

But `JsonTemplateLayoutDefaults` also has the following code:
```
    public static String getEventTemplateUri() {
        return PROPERTIES.getStringProperty(
                "log4j.layout.jsonTemplate.eventTemplateUri",
                "classpath:EcsLayout.json");
    }
```

vy:
The entire documentation assumes that square brackets are used for optional parameters. Hence, in the case of, e.g., `eventTemplate[Uri]`, it means the statement applies to both `eventTemplate` and `eventTemplateUri`, not `Uri`. This syntax is heavily used in every programming environment and to the best of my knowledge, goes back to [BNF notation](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form).

An example from JSON Template Layout manual:
image

An example from `find` manual, i.e., `man find`:
image

Regarding your remark about `JsonTemplateLayoutDefaults`, both forms have its place, hence it is consistent too:
image

I hope this makes it clear.

PR #572, opened 2021-08-28
LOG4J2-3153: PreciseClock doesn't impact DatePatternConverter performance

carterkozak:


carterkozak:
Applied to release-2.x + master

PR #573, opened 2021-08-29
Reduce PatternLayout branching

carterkozak:
Prefer fewer conditionals based on types known ahead of time.
Serializer extends Serializer2 with a default implementation
of the encode method -- all of our implementations of
the Serializer interface also implement Serializer2, so the
conditional path was never used. This may allow more code to
be inlined.


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/LevelPatternConverter.java
```diff
@@ -123,7 +120,7 @@ private static String left(final Level level, final int length) {
      */
     @Override
     public void format(final LogEvent event, final StringBuilder output) {
-        output.append(levelMap == null ? event.getLevel().toString() : levelMap.get(event.getLevel()));
+
```
schlosna, line 123: Is this supposed to be empty, or should this method be removed and handled by super?

schlosna:
I would definitely be interested to see benchmark comparisons for before & after. I imagine the results may depend on the distinct number and shape of formatter patterns used in an application, as that [may push C2 to optimize differently for monomorphic, bimorphic, or megamorphic calls](https://shipilev.net/jvm/anatomy-quarks/16-megamorphic-virtual-calls/) and [some of the potential pitfalls of class hierarchy optimizations](https://shipilev.net/blog/2015/black-magic-method-dispatch/#_conclusion) (though logging tends to be a hot path in many apps, so worth the experimentation).

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/LevelPatternConverter.java
carterkozak: This method should be removed, it's effectively abstract but moving the base class from final -> abstract upset revapi.

carterkozak:
Thanks for the links! I'm planning to benchmark the individual changes as well as end-to-end impact. I'd meant to open this as a draft, this is currently very much an exploratory exercise :-)

I've pulled out the most impactful pieces into separate commits:
https://github.com/apache/logging-log4j2/commit/5078db0e90e2862d2027457f0e95ccb8dbd221fb
https://github.com/apache/logging-log4j2/commit/eff5f185a24c348b25d9d95f1871861c15622284

carterkozak:
I imagine there's something clever (aka risky/unmaintainable/fun) we could do with layouts using bytecode generation to avoid some amount of overhead. Definitely want to explore all other options first and validate that we fully grok the causes, however in benchmarking I found a hand-written layout performed ~15-20% better than PatternLayout with identical output.

carterkozak:
Initial benchmarks on this branch look promising -- writing through a RandomAccessFile appender to `/dev/null` `-Dlog4j2.enable.direct.encoders=false -Dlog4j2.formatMsgNoLookups=true` I'm seeing a 10-13% performance increase (with nolookups enabled). Rerunning with lookups enabled (but not used in my message string) didn't appear to meaningfully impact results.

I still need to do more precise/targetted benchmarking.

rgoers:
I cloned your repo and checked out this branch. When I run the build all the RollingFileAppender tests fail and RollingAppenderDeleteScriptFri13thTest is permanently stuck in the while loop.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/PatternParser.java
```diff
@@ -669,20 +668,20 @@ private int finalizeConverter(final char c, final String pattern, final int star
                 msg.append("] starting at position ");
             }
 
-            msg.append(Integer.toString(i));
+            msg.append(i);
             msg.append(" in conversion pattern.");
```
garydgregory, line 672: I do not think we should hide boxing and unboxing if we are making all of these performance based changes. I would think we would want to see as much of the mechanics as possible.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.util.OptionConverter;
+import org.apache.logging.log4j.util.PerformanceSensitive;
+
+
+/**
+ * Formats a string literal without substitution.
+ *
+ * This is an effectively-sealed internal type.
+ */
```
garydgregory, line 29: And since tag?

carterkozak:
Sorry, Ralph. I've updated this branch with a fix. I'm going to PR something into the release branch separately to improve messaging/behavior in that case because it's a little spooky.

`.core.appender.rolling.PatternProcessor` parses patterns and executes (simplified):
```java
List<PatternConverter> parsed = ...;
ArrayPatternConverter[] converters = parsed.toArray(new ArrayPatternConverter[0]);
```

Calling `toArray(subtype)` succeeds based on runtime type info, however it requires all subtypes to implement ArrrayPatternConverter. I think we should either:
1. Fail loudly and eagerly with an error message that states the ArrayPatternConverter requirement and the type of the invalid converter
2. Allow non-array converters, but pass a null value.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
```diff
@@ -0,0 +1,119 @@
...
+        void format(StringBuilder output) {
+            // no-op
+        }
+    }
+
+    private static final class Space extends SimpleLiteralPatternConverter {
+        private static final Space INSTANCE = new Space();
+
+        @Override
+        void format(StringBuilder output) {
```
garydgregory, line 101: Use final where we can IMO.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/PatternParser.java
carterkozak: This actually uses the `append(int)` overload, avoiding boxing and string allocation entirely :-)

https://github.com/openjdk/jdk/blob/b4b121018d16e531f3a51ff75ae37fdc374d530b/src/java.base/share/classes/java/lang/AbstractStringBuilder.java#L811-L834

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
```diff
@@ -0,0 +1,119 @@
...
+    public final boolean isVariable() {
+        return false;
+    }
+
+    @Override
+    public final boolean handlesThrowable() {
+        return false;
+    }
+
+    private static final class Empty extends SimpleLiteralPatternConverter {
```
garydgregory, line 88: "Noop" would reflect more what is happening here it feels to me.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
```diff
@@ -0,0 +1,119 @@
...
+        private static final Empty INSTANCE = new Empty();
+
+        @Override
+        void format(StringBuilder output) {
+            // no-op
+        }
+    }
+
+    private static final class Space extends SimpleLiteralPatternConverter {
```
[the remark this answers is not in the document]
garydgregory: Thanks. We probably don't even need the subclass if we keep a constant for SPACE.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
carterkozak: Updated with final parameter modifiers

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/SimpleLiteralPatternConverter.java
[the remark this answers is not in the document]
carterkozak: That's fair, I think it's helpful to avoid field access overhead for the hottest types, but it's not clear if there are enough non-space single-character literals to be helpful.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/LogEventFactory.java
```diff
@@ -28,8 +28,21 @@
 /**
  *
  */
-public interface LogEventFactory {
+public interface LogEventFactory extends LocationAwareLogEventFactory {
```
carterkozak, line 31: @rgoers I'd appreciate your input on this part of the change. Now that we're using java 8 we can simplify call-sites to invoke the location-aware implementation. It's possible though unlikely that implementations could exist specifically to avoid capturing location information, but our logger configs support configuring that directly.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/AbstractStringLayout.java
```diff
@@ -88,9 +88,10 @@ public boolean requiresLocation() {
         return false;
     }
 
-    public interface Serializer {
+    public interface Serializer extends Serializer2 {
```
carterkozak, line 91: Similarly, this collapses types to avoid unnecessary instanceof checks

carterkozak:
>  imagine the results may depend on the distinct number and shape of formatter patterns used in an application, as that may push C2 to optimize differently for monomorphic, bimorphic, or megamorphic calls and some of the potential pitfalls of class hierarchy optimizations

@schlosna In this case the additional subclasses shouldn't impact overall performance because the PatternLayout operates with an array of LogEventPatternConverter instances based on the input pattern. There are dozens of types that may be used, we're exchanging some for others to reduce the variables introduced into the system.

Running with java 11:

This PR:

```
$ java -jar log4j-perf/target/benchmarks.jar ".*PatternLayoutBenchmark.serializableMCDEx" -f 1 -wi 4 -i 4 -w 4s -r 4s -t 1 -bm avgt
...
Benchmark                                 Mode  Cnt   Score    Error  Units
PatternLayoutBenchmark.serializableMCDEx  avgt    4  84.154 ± 10.530  ns/op
```

(second run after running both branches a first time, to verify the results are reproducible)
```
Benchmark                                 Mode  Cnt   Score    Error  Units
PatternLayoutBenchmark.serializableMCDEx  avgt    4  87.192 ± 16.008  ns/op
```

Compared to the 2.x branch:
```
java -jar log4j-perf/target/benchmarks.jar ".*PatternLayoutBenchmark.serializableMCDEx" -f 1 -wi 4 -i 4 -w 4s -r 4s -t 1 -bm avgt
...
Benchmark                                 Mode  Cnt    Score   Error  Units
PatternLayoutBenchmark.serializableMCDEx  avgt    4  152.972 ± 2.614  ns/op
```

Second run of release-2.x after the second run of the PR branch:
```
Benchmark                                 Mode  Cnt    Score   Error  Units
PatternLayoutBenchmark.serializableMCDEx  avgt    4  154.212 ± 8.509  ns/op
```

carterkozak:
Testing locally passes, CI builds are also passing. Planning to merge this shortly.

PR #574, opened 2021-08-30
 LOG4J2-2978: add support for Jakarta EE 9

mseele:


rgoers (approved):
Thanks for doing this!  We will need to include some documentation changes too for this.

rgoers (approved):
Thanks. This looks great! I will get this merged as soon as I get a few minutes to also do the cherry-pick to master.

rgoers:
This PR fails to build. It is referencing hamcrest-all. I believe release-2.x was modified to use the hamcrest dependency instead.

mseele:
> This PR fails to build. It is referencing hamcrest-all. I believe release-2.x was modified to use the hamcrest dependency instead.

@rgoers I've fixed that problem. Everything should be fine, again.

PR #576, opened 2021-09-01
JTL InstantFormatter invalidates cached FixedDateFormat as expected

carterkozak:
Previously more precise patterns would be cached unexpectedly
within a millisecond interval, all using the microsecond or
nanosecond value of the first event that millisecond.


vy:
Great catch! Thanks so much @carterkozak!

PR #582, opened 2021-09-17
LOG4J2-3162 Improve immediateFlush docs regarding its write guarantees.

agsha:
https://issues.apache.org/jira/browse/LOG4J2-3162

In the documentation of RollingFileAppender

> immediateFlush boolean When set to true - the default, each write will be followed by a flush. 
> This will guarantee the data is written to disk but could impact performance.
> Flushing after every write is only useful when using this appender with synchronous loggers. 
> Asynchronous loggers and appenders will automatically flush at the end of a batch of events, 
> even if immediateFlush is set to false. This also guarantees the data is written to disk but is more efficient.
> This is misleading. The appender does not really guarantee that the data is written to disk . 
> All it does is call the underlying OutputStream.flush()method, 
> which does not guaratee that data is written to disk. It only guaratees that the data is handed over to the operating system
> 

This is misleading. The appender does not really `guarantee that the data is written to disk` . All it does is [call](https://github.com/agsha/logging-log4j2/blob/1d56dac20d0827300a49a8cd9f0324185451757e/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/OutputStreamManager.java#L264) the underlying [OutputStream.flush()](https://docs.oracle.com/javase/7/docs/api/java/io/OutputStream.html#flush()) method, which does not guaratee that data is written to disk. It only guaratees that the data is handed over to the operating system

### DIFF
--- a/log4j-core/src/main/resources/Log4j-config.xsd
+++ b/log4j-core/src/main/resources/Log4j-config.xsd
@@ -910,8 +910,9 @@
                <annotation>
                   <documentation>
                      When set to true, each write will be followed by a flush.
-                     This will guarantee the data is written to disk but could impact
-                     performance.
+                     This will guarantee that the data is passed to the operating system for writing;
+                     it does not guarantee that the data is actually written to a physical device
+                     such as a disk drive.
                   </documentation>
                </annotation>
             </attribute>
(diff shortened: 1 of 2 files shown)


vy:
@agsha Thanks so much for the contribution! I will backport this to `release-2.x` too.

PR #584, opened 2021-09-21
LOG4J2-3164: Refactor StoringStatusListener to improve test logic

wx930910:
### Jira

- [LOG4J2-3164](https://issues.apache.org/jira/browse/LOG4J2-3164)

#### Replace test class [StoringStatusListener](https://github.com/apache/logging-log4j2/blob/20f9a97dbe5928c3b5077bcdd2a22ac92e941655/log4j-core/src/test/java/org/apache/logging/log4j/core/config/plugins/validation/validators/ValidatingPluginWithFailoverTest.java#L86) by mocking object and improve test design.
<hr>

##### Motivation

 * Decouple test class `StoringStatusListener` from production interface `StatusListener`.
 * Remove the redundant test child class `StoringStatusListener`.
 * Remove the redundant overridden methods `close()`.
 * Make testing logic more explict by directly verify the input argument through `ArgumentCapture`.

<hr>

##### Key changed/added classes in this PR
 * Created mocking object to replace test subclass `StoringStatusListener`, decoupled test from production code.
 * Use ArgumentCapture to replace the new field created in `StoringStatusListener`.

<hr>

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/config/plugins/validation/validators/ValidatingPluginWithFailoverTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/config/plugins/validation/validators/ValidatingPluginWithFailoverTest.java
@@ -31,14 +31,19 @@ import org.apache.logging.log4j.status.StatusListener;
 import org.apache.logging.log4j.status.StatusLogger;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-
-import java.util.ArrayList;
-import java.util.List;
+import org.mockito.ArgumentCaptor;
 
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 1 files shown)


wx930910:
If we only want to make sure that the `log(StatusData)` method execution status, we can verify it by using `Mockito.times()` and get rid of the ArgumentCapture. It can make the test condition even more clear.

vy:
Thanks so much for the contribution @wx930910! Mocking is definitely a better approach, since it allows (backward-compatible) evolution of `StatusListener` without breaking tests. Not to mention that the intent is clearer as well.

Regarding the implementation, I think I would prefer a `verify(listener, never())` finalized with a `verifyNoMoreInteractions()`. If you would be kind enough to update your PR, I will be glad to merge it.

wx930910:
@vy Thanks for your comments! Make sence to me, I changed the code corresponding to your advice :+1:

PR #587, opened 2021-10-09
Avoid kafka with new topic configuration get overried by the old one

wuqian0808:
Avoid kafka with new topic configuration get overried by the old one


vy:
This PR is superseded by #590.

PR #593, opened 2021-10-29
LOG4J2-3182 Protect KafkaManager against start failures upon config changes.

wuqian0808:
With MonitorInterval, if new config start failed, the previous watchmanager and scheduler will never shut down.

Besides, the async threads will increase.

This is for fixing this issue

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
@@ -623,7 +623,12 @@ public class LoggerContext extends AbstractLifeCycle
                 map.putIfAbsent("hostName", "unknown");
             }
             map.putIfAbsent("contextName", contextName);
-            config.start();
+            try {
+            	config.start();
+            } catch (Exception e) {
+            	config.stop();
+            	return configuration;
+            }
             this.configuration = config;
             updateLoggers();
             if (prev != null) {
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
@@ -2,7 +2,6 @@ package org.apache.logging.log4j.core.config;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.junit.LoggerContextSource;
 import org.junit.Assert;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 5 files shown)


garydgregory:
This PR lacks a test case.

wuqian0808:
> This PR lacks a test case.

Test case added. Pls help to review and kindly help to advise

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
```diff
@@ -0,0 +1,69 @@
+package org.apache.logging.log4j.core.config;
```
vy, line 1: License header is missing.

log4j-core/src/test/resources/log4j2-3182-error.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
```
vy, line 1: License header is missing.

log4j-core/src/test/resources/log4j2-3182.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
```
vy, line 1: License header is missing.

log4j-core/src/test/resources/log4j2-3182-error.xml
```diff
@@ -0,0 +1,33 @@
...
+    </properties>
+
+    <appenders>
+        <Console name="CONSOLE" target="system_out">
+            <PatternLayout pattern="${PATTERN}" />
+        </Console>
+
+        <Kafka name="KafkaLog" topic="acm">
+            <PatternLayout pattern="%msg"/>
+            <Property name="bootstrap.servers">10.20.10.111111:9092</Property>
```
vy, line 14: Mind documenting the reason for the deliberately incorrectly typed IP address, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -623,7 +623,12 @@ public Configuration setConfiguration(final Configuration config) {
                 map.putIfAbsent("hostName", "unknown");
             }
             map.putIfAbsent("contextName", contextName);
-            config.start();
+            try {
+            	config.start();
+            } catch (Exception e) {
+            	config.stop();
+            	return configuration;
+            }
```
vy, line 631: We are swallowing the exception here. Shouldn't we check if it matches first and otherwise raise it?

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+
+    @Test
+    public void onChangeTest() {
+        String errorFileName = destDir + targetErrorFileName;
+        String originFileName = destDir + targetFileName;
+        String tmpFile = destDir + tmpFileName;
+        wait(10);
+        updateConfigFileModTime(originFileName, errorFileName, tmpFile);
+        wait(30);
```
vy, line 31: Given that the current test suite takes more than 30 minutes, I am reluctant to add a test that waits for 10+30=40 seconds.

log4j-core/src/test/resources/log4j2-3182.xml
```diff
@@ -0,0 +1,33 @@
...
+    </properties>
+
+    <appenders>
+        <Console name="CONSOLE" target="system_out">
+            <PatternLayout pattern="${PATTERN}" />
+        </Console>
+
+        <Kafka name="KafkaLog" topic="acm">
+            <PatternLayout pattern="%msg"/>
+            <Property name="bootstrap.servers">10.20.10.11:9092</Property>
```
vy, line 14: I don't think this is an OS/network-independent solution. How do you know that the Kafka server is on `10.20.10.11`? Further, doesn't this test depend on a running Kafka server? If so, where do we start it?

vy (changes requested):

vy:
@wuqian0808, mind updating the `changes.xml` too, please?

garydgregory:
Hi all,

Did someone make sure the tests fail if the changes in main are not applied? I can't check ATM.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
wuqian0808: I will change this part and raise the exception always

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
wuqian0808: OK. I will reduce this wait time

log4j-core/src/test/resources/log4j2-3182.xml
wuqian0808: It doesn't depend on a running server. This is for mocking a thread leak during monitorinterval file on change suituation.  Even the kafka ip address is not available, it won't affect this testcase

wuqian0808:
Open new clean pre instead

vy:
@wuqian0808, please don't abandon a PR like this. It becomes really difficult to keep track of what has happened and why.

PR #594, opened 2021-10-29
Log4 j2 3168 overwritten rotated logs

bwoester:
This fixes an issue in the rolling file appender. The increment counter in the filePattern didn't work if the filenames contained regex characters. In our case it was the '+' character introduced by the timezone in a timestamp (app-20211029T123000+0200.%i.log), that caused filenames to be missed by the `getEligibleFiles` method.

The fix quotes the parts of the filenames that are not meant to be treated as regex.

I think this is the actual cause for JIRA ticket https://issues.apache.org/jira/browse/LOG4J2-3168 not the compression mentioned in the ticket.

The first commit in this PR has to be reverted before merging, I kept it for now in case there's more work to be done on the PR.

Please let me know what you think.


rgoers:
Because of the 2 bad files I had to manually create a patch and apply it, but this patch has been applied.

bwoester:
Tanks a lot! :)

PR #597, opened 2021-11-08
LOG4J2-3060 LOG4J2-3185 Fix thread-safety issues in DefaultErrorHandler.

vy:
This change set supersedes #481 and #596.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/DefaultErrorHandler.java
@@ -40,9 +40,9 @@ public class DefaultErrorHandler implements ErrorHandler {
 
     private static final long EXCEPTION_INTERVAL_NANOS = TimeUnit.MINUTES.toNanos(5);
 
-    private volatile int exceptionCount = 0;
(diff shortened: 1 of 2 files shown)


garydgregory:
Hi @vy 
I am curious as to why not simply make methods synchronized. Also, isn't it the case that you do not need to make variables volatile if they are updated inside a synchronized method or block?

vy:
@garydgregory, thanks so much for the review. `volatile` modifiers are not needed indeed, removed them. I have used the `synchronized` block around the smallest possible group to minimize the wait duration under contention.

mzbonnt:
> @garydgregory, thanks so much for the review. `volatile` modifiers are not needed indeed, removed them. I have used the `synchronized` block around the smallest possible group to minimize the wait duration under contention.

consider AtomicInteger or AtomicLong, which is more light weight than synchronized block

garydgregory:
> > @garydgregory, thanks so much for the review. `volatile` modifiers are not needed indeed, removed them. I have used the `synchronized` block around the smallest possible group to minimize the wait duration under contention.
> 
> consider AtomicInteger or AtomicLong, which is more light weight than synchronized block

@mzbonnt You should look at the differences more carefully, considering multiple threads interweaving within these methods.

PR #598, opened 2021-11-09
LOG4J2-3182 Protect KafkaManager against start failures upon config changes

wuqian0808:
LOG4J2-3182 Protect KafkaManager against start failures upon config changes.

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
@@ -43,9 +43,9 @@ public class LoggerContextChangeTest {
         String errorFileName = destDir + targetErrorFileName;
         String originFileName = destDir + targetFileName;
         String tmpFile = destDir + tmpFileName;
-        wait(3);
+        wait(1);
         updateConfigFileModTime(originFileName, errorFileName, tmpFile);
-        wait(3);
+        wait(5);
         Set<Thread> allThreads = Thread.getAllStackTraces().keySet();
         Integer kafkaThreads = 0;
         Integer consoleThreads = 0;
@@ -80,6 +80,7 @@ public class LoggerContextChangeTest {
         }
         File destFile = new File(destFileName);
         if (destFile.exists()) {
+            destFile.setLastModified(System.currentTimeMillis());
             destFile.renameTo(new File(originFileName));
         }
     }
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


wuqian0808:
@garydgregory  Raise new clean PR instead. Add all modifications

vy:
This PR supersedes #593.

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.junit.Assert;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.util.Set;
+
+public class LoggerContextChangeTest {
```
vy, line 29: When I run this test without the changes you have introduced in `AsyncAppender` and `LoggerContext`, it still passes. I don't think this test stresses any aspect of either `LoggerContext` or `AsyncAppender` that has changed via this PR. Am I missing something?

vy (changes requested):

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
wuqian0808: But I get java.lang.AssertionError if loggercontext didn't change....

rgoers (commented):
I'm a little confused by the title of this PR. It mentions protecting KafkaManager but none of the changes are to the Kafka Appender. The implication here is that this would affect every component.

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
vy: Checking out _just_ `LoggerContextChangeTest.java` from your branch and running it against `release-2.x` simply succeeds for me:

```bash
git remote add wuqian0808 https://github.com/wuqian0808/logging-log4j2.git
git fetch wuqian0808
git checkout wuqian0808/LOG4J2-3182 -- log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
./mvnw  -pl :log4j-core test -Dtest=LoggerContextChangeTest
```

wuqian0808: You should also use the confiugred log4j2-3182.xml and log4j2-3182-error.xml. Otherwise you won't get kind of this issue

wuqian0808:
> a little confused by the title of this PR. It mentions protecting Kafk

Actually if the log4j2 configuration starts failed and with monitorinterval set. The scheduled thread will always increase. I just changed this tittle as requested. But this thread leak logic will affect all.

log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
vy: @wuqian0808, test succeeds anyway without your changes:

```bash
git remote add wuqian0808 https://github.com/wuqian0808/logging-log4j2.git
git fetch wuqian0808
git checkout wuqian0808/LOG4J2-3182 -- \
log4j-core/src/test/resources/log4j2-3182.xml \
log4j-core/src/test/resources/log4j2-3182-error.xml \
log4j-core/src/test/java/org/apache/logging/log4j/core/config/LoggerContextChangeTest.java
./mvnw clean package -DskipTests
./mvnw -pl :log4j-core test -Dtest=LoggerContextChangeTest
```

@wuqian0808, mind telling me what am I missing?

garydgregory:
@vy makes a very important point here that I would like to emphasize: if your test passes without the changes to the main tree, then your test does not prove that you've fixed anything.

PR #602, opened 2021-11-23
LOG4J2-3190 Fix ScriptAppenderSelector docs using removed importPackage() in Nashorn.

chubbard:
Fix for LOG4J2-3190.  Documentation referred to importPackage() method in ScriptAppenderSelector section which is broken in Nashorn so the fix was to remove the importPackage(java.lang) and replace it with fully qualified references.


vy:
Thanks so much @chubbard! Ported this to `master` too.

PR #614, opened 2021-12-10
(doc) Fix environment variable for formatMsgNoLookups

fxshlein:
The documentation currently says `FORMAT_MESSAGES_PATTERN_DISABLE_LOOKUPS` is the correct environment variable to disable the message pattern lookups, however after testing, it seems that `LOG4J_FORMAT_MSG_NO_LOOKUPS` is actually correct. (It also fits in with the other environment variables)

Even though this option is deprecated in the newest version, this could make people wrongfully think `FORMAT_MESSAGES_PATTERN_DISABLE_LOOKUPS` works on older deployments where the log4j version is not able to be updated.

### DIFF
--- a/src/site/asciidoc/manual/configuration.adoc
+++ b/src/site/asciidoc/manual/configuration.adoc
@@ -2479,7 +2479,7 @@ LoggerContext is started. For debug purposes.
 
 |[[formatMsgNoLookups]]log4j2.formatMsgNoLookups +
 ([[log4j2.formatMsgNoLookups]]log4j2.formatMsgNoLookups)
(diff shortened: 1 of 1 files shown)


vy:
Thanks so much for the heads up @fxshlein! Please note that this correction is against `master`, which is not released yet. All Log4j 2 releases & websites are derived from `release-2.x` branch, where this typo doesn't occur. (Correct me if I am wrong.)

fxshlein:
I was going off this: https://logging.apache.org/log4j/2.x/manual/configuration.html

Although its completely removed there now. This morning it was still there 😉

PR #618, opened 2021-12-12
Check javaClassName before JNDI lookup

jyemin:
Check that the class name in the `javaClassName` attribute exactly matches the class name in the `javaSerializedData`
attribute.

Background:

The new documentation states the following:

> When using LDAP only references to the local host name or ip address are supported along with any hosts or ip addresses listed in the log4j2.allowedLdapHosts property.

This seems to be the primary line of defense against CVE-2021-44228 (Log4Shell).  However, the second line of defense goes further:

> Only the Java primitive classes are supported by default as well as any classes specified by the log4j2.allowedLdapClasses property. 

Unfortunately, this guarantee seems easily subverted. 

Suppose that an attacker in control of an LDAP server returns a set of attributes like this:
 
```
javaClassName: "java.lang.String"
javaSerializedData: <serialized data for some attackable class, not actually java.lang.String!>
```

In this scenario, Log4J's `JndiManager#lookup(java.lang.String)` will determine that the object is safe to deserialize based on the value of the `javaClassName` attribute (because `java.lang.String` is on the default allow list), and then proceed to call `javax.naming.Context#lookup(java.lang.String)`, which will deserialize the data in `javaSerializedData`. One might think that the JDK's LDAP implementation for JNDI would first check that the object in `javaSerializedData` actually matches `javaClassName` before proceeding with deserialization, but that's not the case.  It seems that it will happily deserialize and return whatever is in `javaSerializedData`, regardless of whether the actual class matches `javaClassName`.  This is problematic, and in my opinion a security issue with the JDK.

This patch is an attempt to mitigate the issue. It has a number of problems:

1. The attack scenario is untested.  I'm unsure how to recreate the scenario given the capabilities of `embedded-ldap-junit`.
2. It violates the strict definition of `javaClassName` in RFC 2713 (see inline comment).

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/net/SerializationHelper.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/net/SerializationHelper.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.log4j.core.net;
+
+import java.nio.BufferUnderflowException;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
+import static java.io.ObjectStreamConstants.STREAM_MAGIC;
+import static java.io.ObjectStreamConstants.STREAM_VERSION;
+import static java.io.ObjectStreamConstants.TC_ARRAY;
+import static java.io.ObjectStreamConstants.TC_CLASS;
+import static java.io.ObjectStreamConstants.TC_CLASSDESC;
+import static java.io.ObjectStreamConstants.TC_ENUM;
+import static java.io.ObjectStreamConstants.TC_LONGSTRING;
+import static java.io.ObjectStreamConstants.TC_OBJECT;
+import static java.io.ObjectStreamConstants.TC_PROXYCLASSDESC;
+import static java.io.ObjectStreamConstants.TC_REFERENCE;
+import static java.io.ObjectStreamConstants.TC_STRING;
+
+// See https://docs.oracle.com/javase/7/docs/platform/serialization/spec/protocol.html for
+// specification of the Object Serialization Stream Protocol
+final class SerializationHelper {
+    // Return the class name of serializedObject, or null if the class name can not be extracted.
+    static String extractClassName(final byte[] serializedObject) {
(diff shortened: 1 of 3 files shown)


garydgregory:
You'll need a test that fails without the main changes to prove this actually works, please see our existing tests for inspiration. You might need to use mocking to provide the bad data.

jyemin:
> You'll need a test that fails without the main changes to prove this actually works

I agree, but I don't see a way to mock this.  I'd appreciate any ideas, so feel free to add commits to this PR.

garydgregory:
Then perhaps consider refactoring the code in a way that makes it testable. Otherwise, this is just one commit away from a regression :-(

rgoers:
This shouldn't be too hard to test. We just need to store the appropriate attributes in the test LDAP to simulate an attack.  Let me check this out and see if I can help with it.

jyemin:
Figuring out how to test this is definitely worthwhile, but I think I'm going to have to reconsider the approach.  I had a closer look at `ObjectInputStream`, and realized that the `resolveClass` method is not called in all circumstances.  In particular, it's not called if the object is a `String`, which I imagine is a common use case.

I'm going to work on an alternative approach in which the serialized data is examined directly rather than indirectly through `ObjectInputStream`.  How does that sound?

jambestwick:
@jyemin I agree with you ，need to test it

jyemin:
I refactored and reimplemented using a hopefully more reliable design.  The bulk of the code has been extracted into a helper class and unit tested.  The conditional in JndiManager still remains to be tested.  Thanks for looking into that one, @rgoers.

vy:
I am against any kind of changes to make JNDI-related code look secure. This has proven to be futile. JNDI is dangerous! Any measure to make it look secure is nothing but an illusion! Don't use it! Those who think otherwise can simply roll out their own security measures.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/SerializationHelper.java
```diff
@@ -0,0 +1,77 @@
...
+import static java.io.ObjectStreamConstants.TC_OBJECT;
+import static java.io.ObjectStreamConstants.TC_PROXYCLASSDESC;
+import static java.io.ObjectStreamConstants.TC_REFERENCE;
+import static java.io.ObjectStreamConstants.TC_STRING;
+
+// See https://docs.oracle.com/javase/7/docs/platform/serialization/spec/protocol.html for
+// specification of the Object Serialization Stream Protocol
+final class SerializationHelper {
+    // Return the class name of serializedObject, or null if the class name can not be extracted.
+    static String extractClassName(final byte[] serializedObject) {
```
caio-picpay, line 39: Maybe this method will always return null because of:

           if (buffer.getShort() != STREAM_MAGIC) {
                return null;
            }
            if (buffer.getShort() != STREAM_VERSION) {
                return null;
            }

jyemin: The tests show otherwise.  This is just a sanity check that the byte array actually represents a serialized object, which is required to start with STREAM_MAGIC and STREAM_VERSION, in sequence.

jyemin:
@vy I don't disagree with you, and if the decision is to remove `log4j2.allowedLdapClasses` entirely, along with all support for JNDI lookups, I'd be all for it.  But it seems unwise to continue supporting and documenting `log4j2.allowedLdapClasses` without a change like this one, given that currently that setting seems to offer no actual protection from an attacker, and so may mislead users who might otherwise assume that it's useful.

rgoers:
We have agreed we will be removing ldap(s) and all the allowed lists, so this PR won't be necessary. But we do appreciate the work you did.

jambestwick (commented):
closed

PR #623, opened 2021-12-13
LOG4J2-3211 - Remove Messge Lookups

rgoers:


log4j-core/src/test/java/org/apache/logging/log4j/core/layout/PatternLayoutLookupDateTest.java
```diff
@@ -38,7 +38,7 @@ public void testDateLookupInMessage(final LoggerContext context, @Named("List")
         final String template = "${date:YYYY-MM-dd}";
         context.getLogger(PatternLayoutLookupDateTest.class.getName()).info(template);
         final String string = listAppender.getMessages().get(0);
-        assertFalse(string.contains(template), string);
+        assertTrue(string.contains(template), string);
```
carterkozak, line 41: This class and log4j-list-lookups.xml could be deleted.

log4j-core/src/test/java/org/apache/logging/log4j/core/pattern/MessagePatternConverterTest.java
```diff
@@ -121,7 +121,7 @@ public void testLookup() {
                 .setMessage(msg).build();
         final StringBuilder sb = new StringBuilder();
         converter.format(event, sb);
-        assertEquals("bar", sb.toString(), "Unexpected result");
+        assertEquals("${foo}", sb.toString(), "Unexpected result");
```
carterkozak, line 124: Given the feature is gone, we can probably remove this test as it’s covered elsewhere

carterkozak (approved):
Strongly in favor of this. Thank you for putting it together, Ralph!

jvz (approved):
I'm down with this idea. I don't believe lookups should be applied to layouts at the message level anyways.

src/site/xdoc/manual/layouts.xml.vm
```diff
@@ -1497,18 +1497,6 @@ WARN  [main]: Message 2</pre>
                   The call site can look like this:
                 </p>
                 <pre class="prettyprint linenums">logger.info("@|KeyStyle {}|@ = @|ValueStyle {}|@", entry.getKey(), entry.getValue());</pre>
-                <p>
-                  Use <code>{lookups}</code> to log messages like <code>logger.info("Try ${esc.d}{date:YYYY-MM-dd}")</code>
```
remkop, line 1501: Would it make sense to put a note here, something like this:

> <p>From Log4j 2.16.0, support for lookups in log messages has been removed for security reasons. Both the <code>{lookups}</code> and the <code>{nolookups}</code> options on the <code>%m</code>,  <code>%msg</code> and  <code>%message</code> pattern are now silently ignored. </p>

remkop (approved):

log4j-core/src/test/java/org/apache/logging/log4j/core/pattern/MessagePatternConverterTest.java
rgoers: I didn't want to remove any test for this release. Just change them to prove the change was made.

PR #625, opened 2021-12-14
Remove static dependency on JndiLookup.class

rschmitt:
This change ensures that it remains safe to remove `JndiLookup.class`.


vy:
Thanks @rschmitt!

PR #629, opened 2021-12-14
Add new "When to use this the Log4j 1.x bridge" section to migration doc

vorburger:
@rgoers could something like this at the top of https://logging.apache.org/log4j/2.x/manual/migration.html be helpful? I'm happy to rephrase it if it have review feedback how what I'm trying to express here could be made clearer, or close this if it's too verbose for your taste and you feel it's not needed here. 

PS: Thank You for everything you are doing for Log4j! FYI https://twitter.com/vorburger/status/1469807072535584772

### DIFF
--- a/src/site/asciidoc/manual/migration.adoc
+++ b/src/site/asciidoc/manual/migration.adoc
@@ -2,6 +2,14 @@
 Ralph Goers <rgoers@apache.org>
 
 [#Log4j1.2Bridge]
+== When to use this the Log4j 1.x bridge
+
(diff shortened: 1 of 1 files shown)


vorburger:
https://issues.apache.org/jira/browse/LOG4J2-3229 created for failing builds on ASF Jenkins (GitHub actions are still running as of writing this).

vorburger:
See https://github.com/netty/netty/pull/11925 for an example of applying what I'm suggesting here to one of (many, of course) libraries and frameworks.

PR #631, opened 2021-12-16
async logger name can be empty。for compatible old version

zhuyou1234:
<AsyncLogger name="" level="info"
                     includeLocation="false" additivity="false">
</AsyncLogger>

should be allowed


vy:
@zhuyou1234, thanks for your contribution. I think allowing empty names were already a bug in the first place. Hence, I am inclined to keep the code in its current state (that is, enforcing non-empty names) with the cost of a slight backward incompatibility.

PR #632, opened 2021-12-16
Update Required.java

zhuyou1234:
tip message is not accurate。required check null and empty


mron7:
"empty" should be spelled correctly in the updated message before you merge this PR. The current PR has "empty" spelled wrong.

PR #636, opened 2021-12-17
[DOC] fix Duration string

michael-buerkle:
According to
https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/appender/rolling/action/Duration.html#parseCharSequence
a Duration looks like "P2D" or "PT20S".

In serveral instances our configuration did not work as expected and
didn't delete files despite them beeing older then the period that we
thought was configured. After some digging into the configuration we
found that the above mention string fixed the configuration.


garydgregory:
FYI: The current documentation is in the branch release-2.x

michael-buerkle:
@garydgregory: Thank you for that hint.

PR #637, opened 2021-12-17
feat: lock feature by properties(PropertiesPropertySource order in properties)

qxo:
so we can change the load order if we need:
such as lock some properties for force disable some feature we don't need.

ie:

SystemPropertiesPropertySource is priority=100, if we want log4j2.component.properties lock some properties(such as log4j2.enableJndi)
that we do not want to be changed by system properties or ENV.
with this feature, log4j2.component.properties as below:
```
log4j2.enableJndiLookup=false
log4j2.propertiesLoadPriority=1000
```
Our app do not need jndi lookup feature, so we want force disable it(Can't enable by -Dlog4j2.enableJndiLookup=true or by ENV)

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
@@ -29,7 +29,7 @@ public class PropertiesPropertySource implements PropertySource {
 
     private static final String PREFIX = "log4j2.";
 
-    private static final String KEY_PRIORITY = "log4j2.properties_load_priority";
+    private static final String KEY_PRIORITY = "log4j2.propertySourcePriority";
 
     private final int priority;
 
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
@@ -29,15 +29,21 @@ public class PropertiesPropertySource implements PropertySource {
 
     private static final String PREFIX = "log4j2.";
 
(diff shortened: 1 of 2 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
```diff
@@ -29,15 +29,21 @@
 
     private static final String PREFIX = "log4j2.";
 
+    private static final String KEY_PRIORITY = "log4j2.properties_load_priority";
```
vy, line 32:
```suggestion
    private static final String KEY_PRIORITY = "log4j2.propertySourcePriority";
```

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesPropertySource.java
```diff
@@ -29,15 +29,21 @@
...
+
+    private final int priority;
+
     private final Properties properties;
 
     public PropertiesPropertySource(final Properties properties) {
         this.properties = properties;
+        final String order = (String)properties.get(KEY_PRIORITY);
```
vy, line 40: Log4j has a fuzzy matching strategy, I am not sure if we can simply look for `KEY_PRIORITY` at this stage – the fuzzy key alternatives might not have been populated.

vy (changes requested):
Thanks so much for your contribution @qxo. I liked the idea, though I would appreciate it if you can create a JIRA ticket and prefix the PR title with the JIRA ticket ID. Once the JIRA ticket is created, other maintainers can weigh in too.

There are also some other stuff that needs to be addressed:

- `changes.xml` update
- tests
- docs

But these can wait for later.

ppkarwasz:

PR #638, opened 2021-12-17
feat: substitutionInVariablesEnabled configable.

qxo:
recursive substitute suck, we can switch it off in the log4j2.component.properties:
```
log4j2.enableSubstitutionInVariables=false
```


carterkozak (changes requested):
Thank you for the proposal, I am handling this in a different way for 2.17.0.

qxo:
> Thank you for the proposal, I am handling this in a different way for 2.17.0.

ok:)

PR #640, opened 2021-12-18
LOG4J2-3256: Reduce ignored package scope of KafkaAppender

dongjinleekr:


remkop (approved):
Looks good to me.

garydgregory (commented):
@dongjinleekr 
Thank you for your PR.
This feels pretty brittle. What about putting this package list in a text file in the resource folder?
Note that this PR is for master but our current work is based on the branch release-2.x

dongjinleekr:
@garydgregory So, you mean...

1. Make a list of `kafka-clients` packages in the resources directory.
2. When KafkaAppender is initialized, load 1.
3. Determine whether the given logging message should be ignored with 2.

Do I understand correctly?

garydgregory:
> @garydgregory So, you mean...
> 
> 1. Make a list of `kafka-clients` packages in the resources directory.
> 2. When KafkaAppender is initialized, load 1.
> 3. Determine whether the given logging message should be ignored with 2.
> 
> Do I understand correctly?

@dongjinleekr 
Exactly, my thought is to allow for a simpler way to manage this list.

dongjinleekr:
@garydgregory Great, with more maintainability! :+1:

dongjinleekr:
Hi @garydgregory,

After some trial and error, I concluded that maintaining the ignored package list in an array would be better than a file-based approach. Here is why:

- The effort to use an array is similar to a resource file.
- The code becomes much concise and straightforward.
- The array-based approach is much faster than the file-based one in initialization.

I also updated the PR to follow this way. How do you think? Is there any specific reason for the file-based approach?

log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaAppender.java
```diff
@@ -136,6 +136,18 @@ public B setRetryCount(final String retryCount) {
         return new Builder<B>().asBuilder();
     }
 
+    private static final String[] KAFKA_CLIENT_PACKAGES = new String[] { "org.apache.kafka.common", "org.apache.kafka.clients" };
```
garydgregory, line 139: If you use a static init block, you can sort the array and then bsearch() it in the  lookup method.

garydgregory (changes requested):
An array works for me. See sort and bsearch comment, using the Arrays class.

log4j-kafka/src/main/java/org/apache/logging/log4j/kafka/appender/KafkaAppender.java
dongjinleekr: It seems like we are checking the package prefix, not the exact match. Is there any way to do prefix match with binary search? (Or, using regex would be better?)

garydgregory: You are correct, I forgot that this is just a prefix match, discard my last comment. I'll review in the morning unless someone else does first.

dongjinleekr: Great. Happy holidays!! :smile:

garydgregory:
@dongjinleekr 

I modified GitHub patch #640 and pushed patches to branches release-2.x and master.

You are credited in changes.xml.

Thank you for your help!

PR #642, opened 2021-12-19
Add Link to Release History

KristjanESPERANTO:
It's a minor detail, but I didn't find any reference to the release history in the repository.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -1,3 +1,5 @@
+Look also at the [Release History](https://logging.apache.org/log4j/2.x/changes-report.html).
+
 # Apache Log4j 2.11.0 Release Notes
 
 ## List of Binary Compatibility Breaking Changes in log4j-core


larrywest42:
That's a helpful change, especially as the most recent version mentioned in "RELEASE-NOTE.md" is `2.11.0`, from 2018.

vy:
@KristjanESPERANTO, thanks for the contribution! This update needs to go into `announcement.vm` in `release-2.x` branch.

vy:
Hrm... I will do it myself. :+1:

KristjanESPERANTO:
When will the changes be transferred from the `release-2.x` branch to the master branch?

The `RELEASE-NOTES.md` in the master branch is still outdated.

vy:
> When will the changes be transferred from the release-2.x branch to the master branch?

[Never, since every release will have its own content.](https://lists.apache.org/thread/zrn7f1hz38x9q9vo8wmfbdb46w3wqtnm)

KristjanESPERANTO:
Then you will never find a reference to the current release notes in the master branch. If that's the way you want it, I don't understand it, but that's okay too :slightly_smiling_face: 

Thank you very much for your patience and your work on log4j! 🎆

vy:
I see your point @KristjanESPERANTO, maybe we should set the default branch to `release-2.x`. I guess this addresses your concern, right?

KristjanESPERANTO:
Yes, that would address my point. The master branch is now the entry point for people who want to start dealing with the code. Since I can also see current changes this branch, it seems a bit inconsistent to find old release notes in there.

I have no idea about best practice, I'm just trying to give some feedback here :-)

PR #646, opened 2021-12-21
LOG4J2-3264: Fix MapLookup to lookup MapMessage before DefaultMap

quaff:


log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/MapLookupTest.java
```diff
@@ -107,7 +107,7 @@ public void testLookupDefaultMapIsCheckedBeforeMapMessage() {
                 .setMessage(message)
                 .build();
         final MapLookup lookup = new MapLookup(map);
-        assertEquals("ADefault", lookup.lookup(event, "A"));
+        assertEquals("AEvent", lookup.lookup(event, "A"));
```
garydgregory, line 110: Hi @quaff 
Thank you for your PR.
I think you should test both a key that maps to a default and another that does not.
You want to make sure the test covers all the paths in the lookup method. 
TY!

garydgregory (changes requested):
Please see my comment in the test.

log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/MapLookupTest.java
quaff: @garydgregory Fixed.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/MapLookup.java
```diff
@@ -122,14 +122,14 @@ public String lookup(final LogEvent event, final String key) {
...
+        if (isMapMessage) {
+            final String obj = ((MapMessage) event.getMessage()).get(key);
             if (obj != null) {
                 return obj;
             }
         }
-        if (isMapMessage) {
-            return ((MapMessage) event.getMessage()).get(key);
+        if (map != null && map.containsKey(key)) {
+            return map.get(key);
```
carterkozak, line 132: `containsKey` immediately followed by `get` is unnecessary. If the map isn't null we can avoid the second traversal and return the result of `map.get(key)`.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/MapLookup.java
```diff
@@ -122,14 +122,14 @@ public String lookup(final LogEvent event, final String key) {
         if (map == null && !isMapMessage) {
             return null;
         }
```
carterkozak, line 124: I don't think this block saves us anything because we do the same checks below -- I'd remove it.

carterkozak (approved):

carterkozak:
I think this PR suggests correct behavior, however, before it merges we need to split out a separate lookup for configuration properties which is not impacted by map-message components. Otherwise properties may be read from structured data map messages unexpectedly when the config author intended to read from the `<Properties>` block in a config. I've filed https://issues.apache.org/jira/browse/LOG4J2-3270 where I'll fix that part up :-)

carterkozak:
@quaff I've resolved that issue, would you mind rebasing to `release-2.x`? I think you'll need to update one of the tests in `InterpolatorTest` as well. Thanks!

quaff:
> @quaff I've resolved that issue, would you mind rebasing to `release-2.x`? I think you'll need to update one of the tests in `InterpolatorTest` as well. Thanks!

@carterkozak Updated.

carterkozak:
@garydgregory You have an active disapproval on this PR, is there anything you're waiting for?

garydgregory (approved):
LGTM

garydgregory:
> @garydgregory You have an active disapproval on this PR, is there anything you're waiting for?

Approved, thank you for the reminder.

PR #647, opened 2021-12-21
LOG4J2-3270 Provide separation between MapMessage and properties lookups

carterkozak:
`MapLookup` includes special handling for `MapMessages` in addition
to the configuration properties map. This change aims to prevent
unwanted interactions between these two concerns.

### DIFF
--- a/src/site/xdoc/manual/lookups.xml
+++ b/src/site/xdoc/manual/lookups.xml
@@ -493,7 +493,6 @@ public static void main(String args[]) {
             <ol>
               <li>Provide the base for Properties declared in the configuration file.</li>
               <li>Retrieve values from MapMessages in LogEvents.</li>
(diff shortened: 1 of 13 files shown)


src/site/xdoc/manual/lookups.xml
```diff
@@ -493,7 +493,6 @@ public static void main(String args[]) {
             <ol>
               <li>Provide the base for Properties declared in the configuration file.</li>
               <li>Retrieve values from MapMessages in LogEvents.</li>
-              <li>Retrieve values set with <a href="../log4j-core/apidocs/org/apache/logging/log4j/core/lookup/MapLookup.html#setMainArguments%28java.lang.String[]%29">MapLookup.setMainArguments(String[])</a></li>
```
carterkozak, line 496: This interaction was removed some time ago, but remained in the documentation.

PR #650, opened 2021-12-22
JNDI enablement properties are loaded at most once

carterkozak:
Rather than checking properties after each invocation, jndi
properties are all read into static final boolean fields when
the JndiManager class is initialized, this way properties cannot
be mutated and refreshed at runtime, and checks are cheaper.
This format matches other log4j configuration points as set
in `Constants.java`.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/net/JndiManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/net/JndiManager.java
@@ -42,6 +42,10 @@ public class JndiManager extends AbstractManager {
     private static final String PREFIX = "log4j2.enableJndi";
     private static final String JAVA_SCHEME = "java";
 
+    private static final boolean JNDI_CONTEXT_SELECTOR_ENABLED = isJndiEnabled("ContextSelector");
+    private static final boolean JNDI_JMS_ENABLED = isJndiEnabled("Jms");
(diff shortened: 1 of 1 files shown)


garydgregory:
How is that going to work for tests that set and clear properties? You
don't want to set a property "forever"...

Gaty

carterkozak:
> How is that going to work for tests that set and clear properties? You don't want to set a property "forever"...

Good question. I'm letting CI run our tests to see if there's anything that will need to be updated. Most of our tests rely on similar static state because binding a logging framework is itself static. Each test class (in the core project, anyhow) runs in an isolated fork, so per-test properties should be fine as long as we don't expect to change the value within a single test case. It shouldn't be any different from the tests which set properties like `log4j2.enable.threadlocals`/etc.

garydgregory:
> > How is that going to work for tests that set and clear properties? You don't want to set a property "forever"...
> 
> Good question. I'm letting CI run our tests to see if there's anything that will need to be updated. Most of our tests rely on similar static state because binding a logging framework is itself static. Each test class (in the core project, anyhow) runs in an isolated fork, so per-test properties should be fine as long as we don't expect to change the value within a single test case. It shouldn't be any different from the tests which set properties like `log4j2.enable.threadlocals`/etc.

Fingers crossed then ;-)

carterkozak:
Local build succeeded :-)

garydgregory:
> Local build succeeded :-)

Ka-pow! :-)

PR #653, opened 2021-12-24
Release 2.x log4j-to-jul JDK Logging Bridge (LOG4J2-3282)

vorburger:
https://issues.apache.org/jira/browse/LOG4J2-3282


log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @Override
+    public void logMessage(final String fqcn, final Level level, final Marker marker, final Message message, final Throwable t) {
+        java.util.logging.Level julLevel = convertLevel(level);
+        if (!logger.isLoggable(julLevel)) {
+            return;
+        }
+        LogRecord record = new LogRecord(julLevel, message.getFormattedMessage()); // TODO getFormat() or getFormattedMessage() ?
+        record.setLoggerName(getName()); // TODO getName() or fqcn parameter? What's that?
+        record.setParameters(message.getParameters());
```
carterkozak, line 62: Good questions!

### 1. getFormattedMessage/getMessage/getParameters

JUL `LogRecord` and the log4j-api `Message` are similar in that they both provide a raw format string, and a parameter array. However, the log4j-api encodes how the format and parameters are combined in the `Message` implementation (`getFormattedMessage` + `formatTo(StringBuilder)` handle formatting) where JUL (default formatters anyhow) use [java.text.MessageFormat.format](https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/src/java.logging/share/classes/java/util/logging/Formatter.java#L92-L150).

Given that, we don't want to use `getFormat()` and definitely don't want to set the parameters here, otherwise there's JUL may do substitution on the user-provided message...and we don't want to go there again ;-)

### 2. Logger name/fqcn

We definitely want to set the logger name to `getName` which is the name of the logger as requested from the LogManager.
The `fqcn` is the fully qualified name of the logger implementation, e.g. `org.apache.logging.log4j.tojul.JULLogger`. This is useful to find the call-site, where we can traverse just above that fqcn. JUL supports this sort of thing using `getSourceClassName ` and `getSourceMethodName`, so it may be a good idea to support the functionality here, otherwise the `LogRecord` will always report `JULLogger.logMessage` as the caller. I'd subclass LogRecord to lazily evaluate the call-site when it's requested given the compute cost of StackWalker/etc.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
```diff
@@ -0,0 +1,246 @@
...
+        record.setThrown(message.getThrowable());
```
carterkozak, line 63: I think we need to check the `Throwable t` parameter
```suggestion
        record.setThrown(t == null ? message.getThrowable() : t);
```

carterkozak (commented):
I haven't had a chance to fully review, but I've tried to answer the questions you've left.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.logging.log4j.spi.AbstractLogger;
+
+/**
+ * Implementation of {@link org.apache.logging.log4j.Logger} that's backed by a {@link Logger}.
+ *
+ * This implementation currently ignores {@link Marker}.
+ *
+ * @author <a href="http://www.vorburger.ch">Michael Vorburger.ch</a> for Google
+ */
+public class JULLogger extends AbstractLogger {
```
carterkozak, line 36: My preference is final-by-default for new types

pom.xml
```diff
@@ -964,6 +964,19 @@
...
+        <artifactId>guava-testlib</artifactId>
+        <version>31.0.1-jre</version>
+        <scope>test</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.truth</groupId>
+        <artifactId>truth</artifactId>
+        <version>1.1.3</version>
+        <scope>test</scope>
+      </dependency>
```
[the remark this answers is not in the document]
garydgregory: I agree, it gives us a touch more flexibility. OTOH, if a new class does not declare protected methods, does it matter?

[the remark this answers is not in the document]
vorburger: Ah, right. So I personally really like https://truth.dev and assumed it was a "more modern" alternative to https://assertj.github.io/doc/... but it seems like AssertJ may have "caught up" since I last had a closer look at it (years ago) and also has that "fluent style" that I like in Truth (only now, or always had?). There is an internal process at work that is a tiny bit easier for me if I can use Truth instead of AssertJ, so I have to ask 😸 if this feedback is a "suggestion" or a "requirement", for merge? I totally do understand if you make it a hard requirement, for consistency across Log4j, and am happy to switch and adapt, but I thought I would ask just to double check once before doing so. 

> I agree, it gives us a touch more flexibility.

Which one, AssertJ or Truth?

> OTOH, if a new class does not declare protected methods, does it matter?

Sorry I did not understand this (my choice of assertion framework here wasn't related to testing declared protected methods).

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: `java.util.logging.LogRecord.setThrown(Throwable)` actually only does `this.thrown = thrown;` anyway, so this doesn't actually matter, but given that this has jumped out at you, let me commit your suggestion anyway, as it's clearer to read (to you).

vorburger: Oh sorry I misinterpreted - this is about using message's getThrowable() iff t is null! Please ignore my comment above. Either way, this is now applied. Thank You!

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: Sure, done. May I ask why, just for learning? Does this make a (real, noticeable, measurable) difference for performance? Or is it not for performance but "style" - you don't want classes to be "open" and extensible by users? Thanks!

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: 1. So KEEP `new LogRecord(julLevel, message.getFormattedMessage())` as-is, and REMOVE the `// TODO getFormat() or getFormattedMessage() ?` AND also remove the `record.setParameters(message.getParameters());` - am I getting this right? I've done this now, could you just double check I got this right and ACK 1. is fine.

2. I see. Let me get back to that point. (As I'm adding more test coverage I'm finding that other -more basic- things need a bit more work, first.)

pom.xml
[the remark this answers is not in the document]
carterkozak: I think Gary meant to reply to https://github.com/apache/logging-log4j2/pull/653#discussion_r775083886

I haven't used the Truth library outside of perhaps a couple bug-fixes in guava/error-prone. I will be able to review much more quickly and easily if you use frameworks I'm familiar with. Truth appears to bring in a few large transitive dependencies as well. Unless we're consolidating all the assertj/junit asserts into truth, we're probably better off avoiding  [the lava layer](http://mikehadlow.blogspot.com/2014/12/the-lava-layer-anti-pattern.html).

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
carterkozak: 1: That's exactly right, thanks!

pom.xml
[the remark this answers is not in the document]
vorburger: ACK, then I'll rework this to use AssertJ instead of Truth, no worries. (FYI I've added additional tests just now.) Expected timeframe: Coming couple of days, hopefully some time this week (but not today anymore and unlikely tomorrow).

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
carterkozak: Of course :-)
The final class means we don't have to design or review for all the ways an external consumer may decide to extend this class. This way we don't need to review as many edge cases, and we have much greater flexibility to change the class in the future without causing odd semantic breaks.

I don't believe there's any measurable performance impact (C2 should be able to optimize non-final and final methods  similarly if the receiver type is always the same. Also, bytecode manipulation allows final classes to be subclassed, e.g. Mockito + inline mock maker)

pom.xml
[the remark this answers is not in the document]
vorburger: Huh - AssertJ and Truth are 99% (`isSameInstanceAs` => `isSameAs`) API compatible! COOL. Done.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: 2. Is now all that's really left to do here (from my side); I'll look more into on one of the next few days, and then un-draft this.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: Thanks! Resolving.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: Done. Resolving this thread, and undrafting - this PR is now ready for review, from my side.

vorburger:
@carterkozak  This is "code complete" and ready for your review, from my side.

PS: I was trying to figure out why Apache RAT initially choked on my new `log4j-to-jul/src/main/resources/META-INF/services/org.apache.logging.log4j.spi.Provider` without ASL header in `mvn verify` - even though it apparently doesn't seem to mind the exact same file in `log4j-core` and `log4j-to-slf4j` (and `log4j-api` test) to not have the license... I could not see why/where, and so opted to simply add the usual header there as well, to make RAT happy.

carterkozak:
Thanks, @vorburger :-)

It looks like our pom.xml has exclusions for `META-INF/services` in some modules `<exclude>src/main/resources/META-INF/services/**/*</exclude>`, but I think your solution is cleaner.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
```diff
@@ -0,0 +1,266 @@
...
+    public JULLogger(final String name, final MessageFactory messageFactory, final Logger logger) {
+        super(name, messageFactory);
+        this.logger = requireNonNull(logger, "logger");
+        System.setProperty("jdk.logger.packages", getClass().getPackageName() + "," + getClass().getSuperclass().getPackageName());
+    }
+
+    public JULLogger(final String name, final Logger logger) {
+        super(name);
+        this.logger = requireNonNull(logger, "logger");
+        System.setProperty("jdk.logger.packages", getClass().getPackageName() + "," + getClass().getSuperclass().getPackageName());
```
carterkozak, line 52: I'm a bit worried about these, mutating system properties is a surprising side-effect in a constructor, and this will replace any value that's already set. What if somebody else modifies it later? Also no need to re-set the value on each object instantiation.

The other case we should consider is what happens if the log4j2 `log4j-slf4j-impl` module (or any other api bridge into log4j-api) is combined with `log4j-to-jul`. In the case of `log4j-slf4j-impl`, users will call `LoggerFactory.getLogger(clazz);` which returns a `org.apache.logging.slf4j.Log4jLogger` wrapping an ExtendedLogger (JULLogger).  The fqcn field on the logevent will be `org.apache.logging.slf4j.Log4jLogger` which isn't included in this list of packages.

It may be reasonable to set the location values to `null` rather than allowing potentially incorrect values to be used, as this is likely the most complex part of the implementation (and most computationally expensive). Otherwise using something like `StackLocatorUtil.calcLocation(fqcn)` in a subclass should do the trick.

vorburger: > I'm a bit worried about these

Sure, let's improve this!

> mutating system properties is a surprising side-effect in a constructor,
> (...) Also no need to re-set the value on each object instantiation.

Right. Let me move that to... where would this fit better - say right in the `JULProvider`? Or, IMHO probably better & more appropriate actually (to me), in the `JULLoggerContext`. Would you like me to do that?

> and this will replace any value that's already set.

This particular point could of course also be very easily addressed by making it just a tiny bit smarter - read the property, and add our (two) packages (unless they are already set, while we are at it). Would you like me to do that?

> What if somebody else modifies it later? 

They COULD break it yes - unless they are kind/smart enough to also do as above? 

But apparently _"[The `jdk.logger.packages` system property is consulted only once](https://hg.openjdk.java.net/jdk9/jdk9/jdk/file/ddf8af0e536a/src/java.logging/share/classes/java/util/logging/LogRecord.java#l675)."_ anyway.

FYI I tried to search how common use of `jdk.logger.packages` across GitHub, but without the JDK, is; but seem to hit some GitHub index search internal issue [with this query which says 318 results](https://github.com/search?q=%22jdk.logger.packages%22+-%22The+system+property+%7B%40code+jdk.logger.packages%7D+can+define+a+comma+separated%22+-path%3Asrc%2Fjava.base%2Fshare+-filename%3ALoggerFinderLoaderTest.java+-filename%3A%2FSimpleConsoleLogger.java&type=Code) - but does not show them (to me).

> The other case we should consider is what happens if the log4j2 `log4j-slf4j-impl` module (or any other api bridge into log4j-api) is combined with `log4j-to-jul`. In the case of `log4j-slf4j-impl`, users will call `LoggerFactory.getLogger(clazz);` which returns a `org.apache.logging.slf4j.Log4jLogger` wrapping an ExtendedLogger (JULLogger). The fqcn field on the logevent will be `org.apache.logging.slf4j.Log4jLogger` which isn't included in this list of packages.

Nice catch! Would it suffice to simply add `org.apache.logging.slf4j` as a (hard-coded?) 3rd package name to this `jdk.logger.packages` list? Are you concerned about other API bridges? Are there any in `logging-log4j2`? Or are you thinking elsewhere?

But perhaps more importantly: Why would someone combine `log4j-slf4j-impl` with `log4j-to-jul` anyway? If one wanted to build an application and depended on 3rd-party libraries which use both Log4j and Slf4j, and wanted everything to end up in JUL, wouldn't one simply use `log4j-to-jul` + [`slf4j-jdk14`](https://www.slf4j.org/manual.html#swapping)? (BTW Is there any advantage of [using `log4j-slf4j-impl`](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/) over `slf4j-log4j12`? But either of these only seem to provide value, as far as I understand, if the end-goal of the logging redirection fun is `log4j-core`... with a `log4j-to-jul`, I'm not sure why anyone would want to combine it like that.)

> (and most computationally expensive). 

I assumed that there actually should not be a performance problem in JUL about this (I think & hope), because `java.util.logging.LogRecord` only calls its private `inferCaller()` lazily if anyone actually invokes `getSourceClassName()` and `getSourceMethodName()`? (With that `needToInferCaller` thing.) Or am I missing something?

> It may be reasonable to set the location values to `null` rather than allowing potentially incorrect values to be used, as this is likely the most complex part of the implementation 

I'm happy to do that if that allows getting this merged sooner without this. Someone (myself or someone else) could then raise a follow-up PR about that, perhaps with a JIRA dedicated to it. Would you like me to do that and remove what's related to that? I'm asking to reach agreement instead of just going ahead to make changes, because it feels like we may be close to resolve this, so let's be explicit about what road we want to go.

> Otherwise using something like `StackLocatorUtil.calcLocation(fqcn)` in a subclass should do the trick.

Thanks for pointing out that utility; I ignored its existence. FYI [I actually initially explored a `LazyLogRecord`](https://github.com/apache/logging-log4j2/compare/release-2.x...vorburger:release-2.x_LOG4J2-3282_log4j-to-jul_LazyLogRecord#diff-cfaa47cd734fd2002c8729469acfa4c8c9a70f4b91d6ce9e0deff1106ddf0340), but then gave up on that because relying on that `jdk.logger.packages` system property seemed to work great and so much simpler. I'm still not convinved what's wrong with that, and I would personally prefer exploring this in a follow-up PR instead of adding this here, and set the location values to `null` for now, if you don't want to go ahead with the `jdk.logger.packages` system property approach (after minor improvement, as discussed above).

As for using `calcLocation(fqcn)`, would you be willing to educate me on the need for 340 lines of extra code (in `org.apache.logging.log4j.util.StackLocatorUtil` and `StackLocator` of `log4j-api`) for something that JUL seems to already be doing just fine on its own? That somehow... doesn't sit quite right, with me. But I certainly respect that YMMV, and background I may be missing. 

PS: There is a bit of background here that may be useful to share in this context: Something I'd like to further investigate, after this PR is accepted, at first just internally at work and possible later as a public contribution of some form (unclear, really more of a thought experiment at this very early stage), is [if it may somehow be possible to have some sort of more "minimal" `log4j-api`](https://github.com/apache/logging-log4j2/compare/release-2.x...vorburger:release-2.x_split-API), for use only with `log4j-to-jul`... if I can avoid using `StackLocatorUtil` and `StackLocator` in `log4j-to-jul`, that's already 2 classes classes from `log4j-api`.

carterkozak: > > I'm a bit worried about these
> 
> Sure, let's improve this!

:smile:

> 
> > mutating system properties is a surprising side-effect in a constructor,
> > (...) Also no need to re-set the value on each object instantiation.
> 
> Right. Let me move that to... where would this fit better - say right in the `JULProvider`? Or, IMHO probably better & more appropriate actually (to me), in the `JULLoggerContext`. Would you like me to do that?

The context would make more sense to me, but it may not always work (more on this as I answer the next question!)

> 
> > and this will replace any value that's already set.
> 
> This particular point could of course also be very easily addressed by making it just a tiny bit smarter - read the property, and add our (two) packages (unless they are already set, while we are at it). Would you like me to do that?
> 
> > What if somebody else modifies it later?
> 
> They COULD break it yes - unless they are kind/smart enough to also do as above?
> 
> But apparently _"[The `jdk.logger.packages` system property is consulted only once](https://hg.openjdk.java.net/jdk9/jdk9/jdk/file/ddf8af0e536a/src/java.logging/share/classes/java/util/logging/LogRecord.java#l675)."_ anyway.

This is a great point, I think it means that we cannot rely on setting the property anywhere in the bridge! It's probably safe to assume that any application using the log4j-to-jul bridge is using jul in other ways, so the log4j loggers may not be the first to be initialized or used. In such an environment the set of packages may already be initialized:
https://github.com/openjdk/jdk/blob/dfacda488bfbe2e11e8d607a6d08527710286982/src/java.base/share/classes/jdk/internal/logger/SimpleConsoleLogger.java#L439-L445

> 
> FYI I tried to search how common use of `jdk.logger.packages` across GitHub, but without the JDK, is; but seem to hit some GitHub index search internal issue [with this query which says 318 results](https://github.com/search?q=%22jdk.logger.packages%22+-%22The+system+property+%7B%40code+jdk.logger.packages%7D+can+define+a+comma+separated%22+-path%3Asrc%2Fjava.base%2Fshare+-filename%3ALoggerFinderLoaderTest.java+-filename%3A%2FSimpleConsoleLogger.java&type=Code) - but does not show them (to me).
> 
> > The other case we should consider is what happens if the log4j2 `log4j-slf4j-impl` module (or any other api bridge into log4j-api) is combined with `log4j-to-jul`. In the case of `log4j-slf4j-impl`, users will call `LoggerFactory.getLogger(clazz);` which returns a `org.apache.logging.slf4j.Log4jLogger` wrapping an ExtendedLogger (JULLogger). The fqcn field on the logevent will be `org.apache.logging.slf4j.Log4jLogger` which isn't included in this list of packages.
> 
> Nice catch! Would it suffice to simply add `org.apache.logging.slf4j` as a (hard-coded?) 3rd package name to this `jdk.logger.packages` list? Are you concerned about other API bridges? Are there any in `logging-log4j2`? Or are you thinking elsewhere?

I'm not terribly keen on this idea because it forces the jul adapter to have full knowledge of all logger implementations, and I'm confident they're not all public. The overlap between consumers of the jul bridge and custom logger bridges into log4j-api should be very small, but I'd like to avoid creating new idiosyncrasies if there's any chance we can avoid them.
I'm less worried about the concrete examples within this repo than I am about third-party custom bridge implementations that we're not aware of.

> 
> But perhaps more importantly: Why would someone combine `log4j-slf4j-impl` with `log4j-to-jul` anyway? If one wanted to build an application and depended on 3rd-party libraries which use both Log4j and Slf4j, and wanted everything to end up in JUL, wouldn't one simply use `log4j-to-jul` + [`slf4j-jdk14`](https://www.slf4j.org/manual.html#swapping)?

I completely agree that it's not ideal, and well-understood systems won't do this, but it should still work correctly. I've encountered more systems than I'd like which log via a complex system of adapters rather than bridging directly to the desired framework. We can't control which bridges are used, only how the components we publish behave.

> (BTW Is there any advantage of [using `log4j-slf4j-impl`](https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/) over `slf4j-log4j12`? But either of these only seem to provide value, as far as I understand, if the end-goal of the logging redirection fun is `log4j-core`... with a `log4j-to-jul`, I'm not sure why anyone would want to combine it like that.)

`slf4j-log4j12` actually targets log4j-1.2 (log4j 1.x was last released in 2012 and EOL was announced in 2015)

> 
> > (and most computationally expensive).
> 
> I assumed that there actually should not be a performance problem in JUL about this (I think & hope), because `java.util.logging.LogRecord` only calls its private `inferCaller()` lazily if anyone actually invokes `getSourceClassName()` and `getSourceMethodName()`? (With that `needToInferCaller` thing.) Or am I missing something?

Right, when the relevant methods aren't called, it's cheap (no work is done!) however when they are called, it's relatively expensive compared to the other operations the logger is responsible for. I haven't benchmarked JUL location-aware methods in a long time, but I'd expect an output format which uses them to be substantially slower than one which does not.

---
I haven't had a chance to address the remaining pieces, but I need to run for dinner half-way through! Sorry for the delay, I'll try to get back to this either tonight or tomorrow :-)

carterkozak: > I'm happy to do that if that allows getting this merged sooner without this. Someone (myself or someone else) could then raise a follow-up PR about that, perhaps with a JIRA dedicated to it. Would you like me to do that and remove what's related to that? I'm asking to reach agreement instead of just going ahead to make changes, because it feels like we may be close to resolve this, so let's be explicit about what road we want to go.

This is entirely up to you, I'm happy either way.

> As for using calcLocation(fqcn), would you be willing to educate me on the need for 340 lines of extra code (in org.apache.logging.log4j.util.StackLocatorUtil and StackLocator of log4j-api) for something that JUL seems to already be doing just fine on its own? That somehow... doesn't sit quite right, with me. But I certainly respect that YMMV, and background I may be missing.

The issue here is that JUL isn't doing this well enough for wrapped loggers -- the configuration point that JUL provides is statically initialized and cannot be (reasonably) mutated after it is configured. I believe this is based on the assumption that an application will use exactly one logging facade, however modern java services tend to pull together many libraries in a way that invalidates such a model. The JUL model can certainly work if you're able to add all the paths to jvm startup args, but setting them later doesn't give me confidence that the values will always be respected. It's unlikely that all users will correctly set values in the jvm startup args.

> PS: There is a bit of background here that may be useful to share in this context: Something I'd like to further investigate, after this PR is accepted, at first just internally at work and possible later as a public contribution of some form (unclear, really more of a thought experiment at this very early stage), is if it may somehow be possible to have some sort of more "minimal" log4j-api, for use only with log4j-to-jul... if I can avoid using StackLocatorUtil and StackLocator in log4j-to-jul, that's already 2 classes less from log4j-api.

I appreciate the context. Reducing coupling and reducing api size are reasonable goals, however I have a hard time justifying incomplete or incorrect behavior based on a thought experiment. What would prevent a future contributor from making an update which relies on an unexpected component?

vorburger: > This is a great point, I think it means that we cannot rely on setting the property anywhere in the bridge!

ACK, that now does make perfect sense to me as well (in retrospect).

>> (...) if that allows getting this merged sooner without this. (...) and remove what's related to that? 

> This is entirely up to you, I'm happy either way.

All right, I've gone gone ahead, and (for now) simply removed support for `getSourceClassName()` & `getSourceMethodName()` - they now return `null`; this can be again revisited later in a follow-up PR.

Please re-re-review (sorry, and Thank You!) and lemme know how if any other feedback, or if this LGTY as-is.

vorburger:
> It looks like our pom.xml has exclusions for `META-INF/services` in some modules `<exclude>src/main/resources/META-INF/services/**/*</exclude>`, but I think your solution is cleaner.

@carterkozak the RAT failure is back again, and I'm not sure I understand why, given I've only added 3 commits without changes to license headers, or new files. (Or is this perhaps unrelated to this PR, like in #684?) Perhaps we could make the Maven RAT plugin actually print out the offending files, instead of pointing to a file where these are listed, which is always hard to debug on a headless build server.

vorburger:
> make the Maven RAT plugin actually print out the offending files, instead of pointing to a file where these are listed

As far as I can tell this isn't something that's easily "just" possible - but I've created https://issues.apache.org/jira/browse/RAT-294 to suggest it.

vorburger:
> @carterkozak the RAT failure is back again, and I'm not sure I understand why, given I've only added 3 commits without 

Please ignore, I'm tired and somehow mixed up tabs with older and current build logs. 

This PR fails to build not because of RAT but because of a failing test, see #685 and #684.

I'll have to wait for that to be fixed, and then merge that in here and it should be green.

vorburger:
> I'll have to wait for that to be fixed, and then merge that in here and it should be green.

@carterkozak build for this now running (I expect it to be green).

carterkozak (commented):
The code generally looks good to me, but I should have time to give it a proper review + merge over the weekend. Sorry for the delay, I'm a bit swamped!

(not requesting changes, just a thought that you can take or leave) The `JULLogger` class could be package private, as the type itself is never exposed directly, only used as a cast target in tests. This would give us leniency for future refactors.
I think the same could apply to all the new classes except `JULProvider` which must be instantiable via ServiceLoader.

vorburger:
> The `JULLogger` class could be package private, as the type itself is never exposed directly, only used as a cast target in tests. This would give us leniency for future refactors. I think the same could apply to all the new classes except `JULProvider` which must be instantiable via ServiceLoader.

Done. But both the `JULProvider` AND the `JULLoggerContextFactory` must be `public`, otherwise we get a:

    ERROR StatusLogger Unable to create class org.apache.logging.log4j.tojul.JULLoggerContextFactory specified in provider URL null
     java.lang.IllegalAccessException: class org.apache.logging.log4j.LogManager cannot access a member of class org.apache.logging.log4j.tojul.JULLoggerContextFactory with modifiers "public"

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
```diff
@@ -0,0 +1,267 @@
...
+     * Level in Log4j scale.
+     * JUL Levels are mapped as follows:
+     * <ul>
+     * <li>OFF => OFF
+     * <li>SEVERE => ERROR
+     * <li>WARNING => WARN
+     * <li>INFO => INFO
+     * <li>CONFIG => INFO
+     * <li>FINE => DEBUG
+     * <li>FINER => DEBUG (inspired by https://github.com/qos-ch/slf4j/blob/6e784e4ca3a5ae9c5dc421fcd01a417af5bf5ace/jul-to-slf4j/src/main/java/org/slf4j/bridge/SLF4JBridgeHandler.java)
```
carterkozak, line 118: This doesn't agree with `convertLevel` above:

https://github.com/apache/logging-log4j2/blob/a58a06bf2365165ac5abdde931bb4ecd1adf0b3c/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java#L83-L84

Let's be consistent with the `log4j-jul` module https://github.com/apache/logging-log4j2/blob/a58a06bf2365165ac5abdde931bb4ecd1adf0b3c/log4j-jul/src/main/java/org/apache/logging/log4j/jul/DefaultLevelConverter.java#L55-L75

vorburger: Oh! Right. I'll change this (right now).

carterkozak:
I think this is ready to merge once level agreement is fixed :-) Thanks!

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/JULLogger.java
vorburger: @carterkozak Better now? Resolve if OK, our shout if I'm somehow mixing up which way we're mapping what here.

PR #654, opened 2021-12-24
define current build outputTimestamp

hboutemy:
see https://maven.apache.org/guides/mini/guide-reproducible-builds.html
having a value defined here make it follow your releases, instead of having a value fixed by inheritance from parent


garydgregory:
I'll apply the same to release-2.x.

PR #655, opened 2021-12-24
LOG4J2-3288 Interpolator non-plugin ctor includes all core lookups

carterkozak:
This adds missing lookups: `bundle`, `event`, and `sd`

There's a comment suggesting that the PluginManager could be used,
however that will require additional validation before I'm comfortable
enabling it: Logging initialization is very static and tricky,
and it's easy to produce deadlocks when referencing logging code
within initialization.


garydgregory:
Woa, aren't we missing unit tests that exercise those lookups?

carterkozak:
> Woa, aren't we missing unit tests that exercise those lookups?

I think the issue is abundance of branches.

The non-plugin constructor is used initially when the configuration is created, to interpolate the Property values themselves, then the interpolator is replaced with one which does use plugins. Existing uses of `config.getStrSubstitutor()` would work fine, while `new StrSubstitutor()` and some overloads would not.

garydgregory:
Double woa ;-) thank you for the clarification :-)

PR #657, opened 2021-12-24
Api separation documentation

rm5248:
Document the API separation of log4j2 and how to use it with other logging frameworks.

### DIFF
--- a/src/site/markdown/api-separation.md
+++ b/src/site/markdown/api-separation.md
@@ -176,7 +176,7 @@ public class Log4j2Test {
 ```
 
 We can then also route these messages to the Log4j2 core by adding in the JUL bridge,
-and setting the java.util.logging.manager property in the JVM(see the documentation on
+and setting `-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager` on the JVM (see the documentation on
 [the JUL adapter](https://logging.apache.org/log4j/2.x/log4j-jul/index.html) for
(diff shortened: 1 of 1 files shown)


src/site/markdown/api-separation.md
```diff
@@ -0,0 +1,241 @@
...
+
+    private static final Logger logger = Logger.getLogger(Log4j2Test.class.getName());
+
+    public Log4j2Test(){
+        logger.info( "Hello World!" );
+    }
+```
+
+We can then also route these messages to the Log4j2 core by adding in the JUL bridge,
+and setting the java.util.logging.manager property in the JVM(see the documentation on
```
carterkozak, line 179: It might be helpful to provide the precise arg here, otherwise folks may read over it. I've seen it done incorrectly a few places.
```suggestion
and setting `-Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager` on the JVM (see the documentation on
```

carterkozak (approved):
Looks great, thank you!

remkop:
Would it make sense to include some of the images used in the FAQ on this page as well?
For example: https://github.com/apache/logging-log4j2/blob/master/src/site/resources/images/whichjar-2.1.png and perhaps some of the other whichjar* images.

Note, this is just a suggestion for further improvement. 
The PR as it stands already adds a lot of value and I have no objection to merging it as is.

PR #658, opened 2021-12-25
Fix tests not restoring system properties

Marcono1234:
Some tests were changing system properties, but were not reverting their changes. This pull request tries to address this issue.

In some cases there were `System.setProperty(Constants.LOG4J_CONTEXT_SELECTOR, Strings.EMPTY);` calls without that system property being set explicitly. I have removed them (and replaced them with the correct property names, if any) because I assumed that they were copy and paste errors.

Maybe in the future it would be good to use extensions such as [JUnit Pioneer (JUnit 5)](https://junit-pioneer.org/docs/system-properties/) which support setting and restoring system properties, since manually doing it is rather error-prone.

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/Log4j1222Test.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/Log4j1222Test.java
@@ -34,8 +34,13 @@ public class Log4j1222Test
 	@Test
 	public void homepageRendersSuccessfully()
 	{
-        System.setProperty("log4j.configurationFile", "log4j2-console.xml");
-		Runtime.getRuntime().addShutdownHook(new ShutdownHook());
+		System.setProperty("log4j.configurationFile", "log4j2-console.xml");
+		try {
+			// TODO: Does not work correctly; when shutdown hook fails test has already finished successfully
+			Runtime.getRuntime().addShutdownHook(new ShutdownHook());
+		} finally {
+			System.clearProperty("log4j.configurationFile");
+		}
 	}
 
 	private static class ShutdownHook extends Thread {
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/appender/ConsoleAppenderTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/appender/ConsoleAppenderTest.java
@@ -63,7 +63,6 @@ public class ConsoleAppenderTest {
 
     @BeforeEach
     public void before() {
-        System.setProperty(LOG4J_SKIP_JANSI, "true");
         baos = new ByteArrayOutputStream();
     }
 
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigTest.java
@@ -33,6 +33,7 @@ import org.apache.logging.log4j.core.config.AppenderRef;
 import org.apache.logging.log4j.core.config.ConfigurationFactory;
 import org.apache.logging.log4j.core.config.DefaultConfiguration;
 import org.apache.logging.log4j.core.config.LoggerConfig;
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
@@ -43,6 +44,11 @@ public class AsyncLoggerConfigTest {
     @BeforeClass
     public static void beforeClass() {
         System.setProperty(ConfigurationFactory.CONFIGURATION_FILE_PROPERTY, "AsyncLoggerConfigTest.xml");
+    }
+
+    @AfterClass
+    public static void afterClass() {
+        System.clearProperty(ConfigurationFactory.CONFIGURATION_FILE_PROPERTY);
     }
 
     @Test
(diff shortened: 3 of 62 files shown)


log4j-core/src/test/java/org/apache/logging/log4j/core/Log4j1222Test.java
```diff
@@ -34,8 +34,13 @@
 	@Test
 	public void homepageRendersSuccessfully()
 	{
-        System.setProperty("log4j.configurationFile", "log4j2-console.xml");
-		Runtime.getRuntime().addShutdownHook(new ShutdownHook());
+		System.setProperty("log4j.configurationFile", "log4j2-console.xml");
+		try {
+			// TODO: Does not work correctly; when shutdown hook fails test has already finished successfully
```
Marcono1234, line 39: Added this comment because test might be misleading. Test "failures" would most likely not actually cause test to fail.

log4j-core/src/test/java/org/apache/logging/log4j/core/appender/ConsoleAppenderTest.java
```diff
@@ -63,7 +63,6 @@ public static void beforeClass() {
 
     @BeforeEach
     public void before() {
-        System.setProperty(LOG4J_SKIP_JANSI, "true");
```
Marcono1234, line 66: Have removed this because it is set in the `beforeClass` method above already.

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigTest.java
```diff
@@ -71,17 +77,17 @@ public void testAdditivity() throws Exception {
 
     @Test
     public void testIncludeLocationDefaultsToFalse() {
-    	final LoggerConfig rootLoggerConfig =
```
Marcono1234, line 74: Have fixed some inconsistent indentation here (mixed tabs and spaces).

Marcono1234:
The test failures seem to be unrelated to these changes; they are also occurring for `master`, e.g. [here](https://github.com/apache/logging-log4j2/runs/4628017826).

vy:
Isn't there a mechanism we can hook into to take a snapshot of existing properties before and restoring them after a test method?

garydgregory:
> Isn't there a mechanism we can hook into to take a snapshot of existing properties before and restoring them after a test method?

Yes, there's an app for that ;-) https://junit-pioneer.org/docs/system-properties/

Marcono1234:
The issue is that currently some tests are still using JUnit 4. For that there is the [System Rules](https://github.com/stefanbirkner/system-rules) project, but I did not want to add two separate extensions / rules without knowing what the migration plans are.

garydgregory:
> The issue is that currently some tests are still using JUnit 4. For that there is the [System Rules](https://github.com/stefanbirkner/system-rules) project, but I did not want to add two separate extensions / rules without knowing what the migration plans are.

I seems that should move everything to JUnit 5 first.

vy:
> Yes, there's an app for that ;-) https://junit-pioneer.org/docs/system-properties/

The reason I did not mention about `system-properties` of JUnit Pioneer is that it operates on an input list of properties, whereas I want to restore _all_ properties back to their original values. I think the former approach is more error-prone and there is no way to tell if a property was leaked.

vy:
> The issue is that currently some tests are still using JUnit 4. For that there is the [System Rules](https://github.com/stefanbirkner/system-rules) project, but I did not want to add two separate extensions / rules without knowing what the migration plans are.

As @garydgregory noted, all tests need to be migrated to JUnit 5. Given this, maybe we can leverage [system-stubs](https://github.com/webcompere/system-stubs)?

carterkozak:
Each test class in core runs in an isolated fork, otherwise we couldn’t test different values of properties which are stored into static final fields (e.g. Constants.java)

garydgregory:
> Each test class in core runs in an isolated fork, otherwise we couldn’t test different values of properties which are stored into static final fields (e.g. Constants.java)

I do not think we should rely on that Maven behavior for all styles of development, yes, it works great on the Maven command line, but it feels like a bit of a hack, leaving cruft behind. I'm not sure all IDEs are smart enough to read POMs that configure Surefire and FailSafe this way and know to do VM forks within the IDE itself. Ideally, I'd like to run all or any group of tests from an IDE and have them pass. Right now, I can only rely on Maven to do that.

garydgregory:
> > The issue is that currently some tests are still using JUnit 4. For that there is the [System Rules](https://github.com/stefanbirkner/system-rules) project, but I did not want to add two separate extensions / rules without knowing what the migration plans are.
> 
> As @garydgregory noted, all tests need to be migrated to JUnit 5. Given this, maybe we can leverage [system-stubs](https://github.com/webcompere/system-stubs)?

The system stubs warning about Java16+ is not great and it seems like it's not really doing things as simple as I'd like to see. The JUnit Pioneer is cool but too minimal in feature. What I want is something like an `@SaveSystemProperties` and `@RestoreSystemProperties` and I do not think you need to do whatever reflection system stubs does. Since sys props is a global, have the annotation also use a global would be OK. One could also imagine saving and restoring named savepoints, like in a database transaction, `@SaveSystemProperties("Savepoint1")`. Such a thing does not exist, but we can write it, just like we have custom JUnit 4 rules.

ppkarwasz:

beatngu13:
Hey @vy & @garydgregory,

JUnit Pioneer maintainer here. ✌️ I know this PR is old and the problem it was trying to solve might not relevant anymore, but regarding:

> The reason I did not mention about `system-properties` of JUnit Pioneer is that it operates on an input list of properties, whereas I want to restore _all_ properties back to their original values. I think the former approach is more error-prone and there is no way to tell if a property was leaked.

And:

>The system stubs warning about Java16+ is not great and it seems like it's not really doing things as simple as I'd like to see. The JUnit Pioneer is cool but too minimal in feature. What I want is something like an @SaveSystemProperties and @RestoreSystemProperties and I do not think you need to do whatever reflection system stubs does.

As of Pioneer 2.1.0, there is `@RestoreSystemProperties`. From the [docs](https://junit-pioneer.org/docs/system-properties/#restoresystemproperties):

>`@RestoreSystemProperties` can be used to restore changes to system properties made directly in code. While `@ClearSystemProperty` and `@SetSystemProperty` set or clear specific properties and values, they don’t allow property values to be calculated or parameterized, thus there are times you may want to directly set properties in your test code. `@RestoreSystemProperties` can be placed on test methods or test classes and will completely restore all system properties to their original state after a test or test class is complete.

PR #667, opened 2021-12-27
LOG4J2-3292 Fix ExtendedLoggerWrapper.logMessage double-logging

carterkozak:
Previously when the logger instance was location-aware and
the logger was configured to require location info, both branches
were invoked rather than only the location-aware branch.


garydgregory:
Yikes, nice find @carterkozak 👍

PR #668, opened 2021-12-27
update log4j-config.xsd fixed one default type

Tegasus9:
some PatternLayOut.java source code about "disableAnsi" in master branch like this:
If {@code "true"} (default is value of system property `log4j.skipJansi`, or `true` if undefined), do not output ANSI escape codes


garydgregory:
Hi @Kenducky 
Thank you for your PR.
You'll want to look at the branch 'release-2.x', that where the 2.x releases originate while the branch 'master' is for future 3.0 work.

ppkarwasz:

PR #669, opened 2021-12-27
fix: remove invalid badges

Xunzhuo:


vy:
It wasn't invalid, but broken; fixed. @Xunzhuo, thanks so much for the heads up!

PR #670, opened 2021-12-27
add support for FileSize in TB

ramananravi:


ramananravi:
Currently the default value of 'exceeds' attribute of 'IfAccumulatedFileSize' is Long.MAX_VALUE, which is considerably larger than 1000 TB itself. Hence adding support for specifying 'exceeds' parameter in TB (TeraBytes).

PR #671, opened 2021-12-27
[LOG4J2-3303] Add support for FileSize in TB

ramananravi:


ramananravi:
Currently the default value of 'exceeds' attribute of 'IfAccumulatedFileSize' is Long.MAX_VALUE, which is considerably larger than 1000 TB itself. Hence adding support for specifying 'exceeds' parameter in TB (TeraBytes).

vy:
@ramananravi, thanks so much for the contribution. This looks like a legit request to me. Though I find the existing `(long) (double * long)` multiplications in `FileSize` dangerous. I think we will need some `BigDecimal` spice there. I will try to spare some time for this.

vy:
@ramananravi, we think there are a few more places in the code where a similar improvement might be handy. Mind creating a JIRA ticket for this change, please?

ramananravi:
> @ramananravi, we think there are a few more places in the code where a similar improvement might be handy. Mind creating a JIRA ticket for this change, please?

@vy here's the jira ticket for reference:
https://issues.apache.org/jira/browse/LOG4J2-3303

Please let me know if there's something you want me to add as part of this PR. Thanks in advance.

ramananravi:
Edited the PR description to add reference to jira ticket (LOG4J2-3303). Could see some checks failing now (previously all checks passed). Are these intermittent failures expected and if so is there a way to re-trigger these checks? Thanks in advance.

ramananravi:
> > @ramananravi, we think there are a few more places in the code where a similar improvement might be handy. Mind creating a JIRA ticket for this change, please?
> 
> @vy here's the jira ticket for reference: https://issues.apache.org/jira/browse/LOG4J2-3303
> 
> Please let me know if there's something you want me to add as part of this PR. Thanks in advance.

@vy - please let me know if the changes from my side, on this PR, look good or if you have any suggestions for me on the same. Thanks in advance.

vy:
@ramananravi, I will do my best to pick this up as soon as possible, unless another committer steps up.

ramananravi:
> 

@vy - thanks for the reply. Just to add, my question is, is my current PR ok to be merged or do you suggest any changes on the same? If there are any suggestions, I can make changes accordingly. Hence this question. Thanks in advance.

vy:
@ramananravi, it is on `release-2.x`. I will port this to `master`.

PR #672, opened 2021-12-28
Excluded  markdown from apache-rat check

carterkozak:


jvz (approved):

vy:
@carterkozak, @jvz, why do we exclude docs? I think we just need to add license headers to them as any other file. (For the records, I've addressed this in `release-2.x` without knowing about this PR.)

jvz:
Good catch. I think the commented out license header for markdown files makes sense. We can revert the rat exclusion.

carterkozak:
This was meant to match the `README.md` file. No strong preference either way from my end, I opted for exclusion initially since it's easier to read through the file in a simple text editor without the license header.

PR #676, opened 2021-12-30
 LOG4J2-3304 - Fix issue that the initialize status of LogManager is not set when factoryClassName is present and instantiate successfully

Francis-FY:
Fix the issue that the initialize status of LogManager is not set when logger factoryClassName is present and instantiate successfully, see [LOG4J2-3304](https://issues.apache.org/jira/browse/LOG4J2-3304)

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
@@ -122,8 +122,8 @@ public class LogManager {
                         + "Please add log4j-core to the classpath. Using SimpleLogger to log to the console...");
                 factory = new SimpleLoggerContextFactory();
             }
-            LogManagerStatus.setInitialized(true);
         }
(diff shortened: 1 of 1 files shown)


garydgregory:
Hi @Francis-FY 
Thank you for your PR?
How could this be validated in a unit test?

rgoers:
I'm having a hard time understanding how this relates to Log4j's Spring Boot support.

Francis-FY:
@rgoers  The method getEvnironment of SpringEnvironmentHolder will check if LogManager is initialized through LogManagerStatus.isInitialized().  You can run a unit test to validate.
1. create file log4j2.component.properties containing  `log4j2.loggerContextFactory=org.apache.logging.log4j.core.impl.Log4jContextFactory`
2. run test `org.apache.logging.log4j.spring.boot.SpringLookupTest` or `org.apache.logging.log4j.spring.boot.SpringProfileTest`
3. the test will fail

image

PR #680, opened 2022-01-04
[LOG4J2-3326] Fix parsing filters from Log4j 1.2 properties configuration file

benjaminr-ps:
Problem: The log4j 1.x configuration file is not properly parsed, in regards of filters. Whatever is configured in file, is getting ignored, and default values are used. 
E.g. for "level range filter" the LevelMax is not properly found (eventhough properly read as property).

Root cause: The default filter prefix is used, instead of specific filter key.

### DIFF
--- a/log4j-1.2-api/src/test/resources/LOG4J2-3326.properties
+++ b/log4j-1.2-api/src/test/resources/LOG4J2-3326.properties
@@ -1,0 +1,22 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


garydgregory:
Hi @benjaminr-ps 

Thank you for your PR. Would you add a test please? 

Without a test there is no way to see that this fixes anything or that the undoing the fix would break anything (a regression).

See PropertiesConfigurationTest for an example.
TY!

garydgregory:
https://issues.apache.org/jira/browse/LOG4J2-3326

log4j-1.2-api/src/test/resources/LOG4J2-3326.properties
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+log4j.appender.CUSTOM=org.apache.log4j.CustomNoopAppender
+log4j.appender.CUSTOM.filter.1=org.apache.log4j.varia.LevelRangeFilter
+log4j.appender.CUSTOM.filter.1.levelMin=ALL
+log4j.appender.CUSTOM.filter.2=org.apache.log4j.varia.LevelRangeFilter
+
+log4j.rootLogger=trace, CUSTOM
```
garydgregory, line 22: In future PRs, please end files with an EOL.

PR #682, opened 2022-01-05
Add environment variable arbiter

MouettE-SC:
This is a simple arbiter using environment variables functionnally based on the existing system property arbiter (on which I corrected some typos).

I tried to write unit tests for it but injecting environment variables in junit is not easy. I tried [system-rules](https://stefanbirkner.github.io/system-rules/index.html#EnvironmentVariables) but it didn't work so I did not include it in this PR. I tested the arbiter in my application and it seems to work fine.


rgoers:
Thanks for this. I should be able to create the unit tests for it.

MouettE-SC:
> Thanks for this. I should be able to create the unit tests for it.

Great ! I've added the tests I wrote, based on the current system property ones with system-rules which does not work on my machine. That test structure is already done.

MouettE-SC:
Sorry, forgot to comment that I was able to create working tests for the environment arbiter using system-lambda. Hope you'll find them satisfying !

MouettE-SC:
Let me know if there is anything I can do to help, like opening an issue on Jira, writing the documentation part or anything, I'll be happy to !

Zhnz:
Wouldn't be using a PluginFactory instead of a PluginBuilderFactory way shorter? I think that would be better. I implemented my own custom EnvironmentArbiter this way (because this one still isn't merged :P)

log4j-core/src/main/java/org/apache/logging/log4j/core/config/arbiters/EnvironmentArbiter.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @PluginBuilderFactory
+    public static Builder newBuilder() {
+        return new Builder();
+    }
+
+    public static class Builder implements org.apache.logging.log4j.core.util.Builder<EnvironmentArbiter> {
+
+        public static final String ATTR_VARIABLE_NAME = "variableName";
```
zartc, line 57: Why not stick with the names used by the SystemPropertyArbiter ("propertyName" and "propertyValue") ?

jbisotti:
What is the status of the PR? Will it be merged soon? This is exactly what I need.

ppkarwasz:
@jbisotti,

I am afraid this PR fell out of our long todo list. Since we are cutting 2.20.0 right now, I am afraid it will not make it into the next release, but I'll take care of this as soon as we are done.

BTW: you can repackage the class into a separate JAR and add it to any Log4j2 release.

MouettE-SC:
I've rebased the release branch on this given the fact that it is kinda old. I didn't manage to run the whole tests yet.

PR #683, opened 2022-01-05
LOG4J2-3280 - Add support for microsecond precision in RFC5424Layout

atulpendse:
This pull request is to imlement RFE LOG4J2-3280
The change 
  - Adds new attribute timestampPrecision to Rfc5424Layout and SyslogAppender.
  - Replaces existing logic to append timestamp and uses two new fixed date formats to format timestamp.
  - Added new unit tests
  - Verified that 'mvn clean install' and 'mvn verify' are successful.


atulpendse:
I believe the failure in PropertiesConfigurationTest is unrelated to this change, and being fixed with #685 
Once that is fixed, how do I get this PR to build and run CI tests again?

garydgregory:
Hi @atulpendse
You will need to rebase this PR to pick up the latest changes.

atulpendse:
> Hi @atulpendse
> You will need to rebase this PR to pick up the latest changes.

Thanks for the info. I have updated PR with the latest from upstream. Latest commit seems to have passed all the tests.

vy:
Fantastic work @atulpendse! Thanks so much!

Mind adding `SyslogAppender` tests and a `changes.xml` entry, please?

I am also not sure if a `timestampPrecision` argument of type `String` is the right way to go. Thinking out loudly:
- Why don't we simply make the instant formatting pattern configurable?
- Doesn't the spec enforce a certain precision?
- If not, why don't we simply (automatically?) pick the most precise representation possible?

garydgregory:
I would NOT update changes.xml, this makes cherry-picking the commit to master impossible.

vy:
IMHO, cherry-picking to `master` is almost already impossible. Nevertheless, I will rephrase my comment to exclude the `changes.xml` part.

atulpendse:
@vy 
Thanks for taking time to review the PR.
Here are my thoughts about questions you raised.

> I am also not sure if a `timestampPrecision` argument of type `String` is the right way to go. Thinking out loudly:
> 
> * Why don't we simply make the instant formatting pattern configurable?

With making pattern configuration, there is risk of the message not being compliant to RFC5424 spec (if someone updates format that is not as per RFC5424)

> * Doesn't the spec enforce a certain precision?

The spec does not enforce 'a particular' precision, rather it supports three precisions

1. Without any sub-second value
2. Milliseconds
3. Microseconds


> * If not, why don't we simply (automatically?) pick the most precise representation possible?

We could, that would work for my usecase. But if someone is expecting precision to be millisecond (like my app was expecting it to be microsecond), it would break their app.

I will add a test for SyslogAppender.

atulpendse:
Commit fbccefc updates existing test for SyslogAppender

atulpendse:
@vy @garydgregory anything else that I should do to make progress on this PR?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -211,6 +214,10 @@ public String getExceptionPattern() {
             return loggerFields;
         }
 
+		public String getTimestampPrecision() {
```
garydgregory, line 217: This indentation looks messed up. Spaces, no tabs, could be the issue?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -300,6 +307,11 @@ public B setLoggerFields(final LoggerFields[] loggerFields) {
             this.loggerFields = loggerFields;
             return asBuilder();
         }
+
+		public B setTimestampPrecision(final String timestampPrecision) {
```
garydgregory, line 311: This indentation looks messed up. Spaces, no tabs, could be the issue?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -104,6 +104,9 @@
         @PluginElement("LoggerFields")
         private LoggerFields[] loggerFields;
 
+        @PluginBuilderAttribute( "timestampPrecision" )
+        private String timestampPrecision;
```
garydgregory, line 108: Shouldn't we use an enum here?

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -270,7 +278,7 @@ private static PatternParser createPatternParser(final Configuration config,
     public String toSerializable(final LogEvent event) {
         final StringBuilder buf = getStringBuilder();
         appendPriority(buf, event.getLevel());
-        appendTimestamp(buf, event.getTimeMillis());
+		appendTimestamp(buf, event.getInstant());
```
garydgregory, line 281: No tabs.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -425,6 +440,7 @@ protected SyslogAppender(final String name, final Layout<? extends Serializable>
                 .setCharsetName(charset)
                 .setExceptionPattern(exceptionPattern)
                 .setLoggerFields(loggerFields)
+                .setTimestampPrecision( timestampPrecision )
```
garydgregory, line 443: No spaces around an argument, please follow the style of the file you're in!
`.setTimestampPrecision( timestampPrecision )` -> `.setTimestampPrecision(timestampPrecision)`

garydgregory (changes requested):
I only took a cursory look this AM. Please use the style of the files you are editing: Use spaces, no tabs. The bigger question is whether we should use an enum for the microsecond vs millisecond format, @jvz any thoughts?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
jvz: I agree. An enum would make sense.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
atulpendse: Yeah.. issue due to tabs. Was due to problem with my eclipse formatter setting. I have fixed it with latest commit.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
atulpendse: I have fixed it with latest commit.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
atulpendse: I have fixed it with latest commit.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
atulpendse: I have fixed it with latest commit.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
atulpendse: I have converted it into an enum with latest commit.

atulpendse:
> I only took a cursory look this AM. Please use the style of the files you are editing: Use spaces, no tabs. The bigger question is whether we should use an enum for the microsecond vs millisecond format, @jvz any thoughts?

Thanks for taking time to review. I have fixed the concerns with latest commit. Now timestampPrecision is an enum.

garydgregory (commented):
Looks good. I approved the PR for builds here, let's see what happens, especially with the GC tests as I am not sure if this class falls into the GC free epic we had.

atulpendse:
Looks like a test is failing on MacOS. Failure doesn't look related to my changes though. I don't have access to Mac to find out why exactly it's failing.

atulpendse:
@garydgregory, please let me know how I can help to get the PR going.
I rebased the PR with latest code and submitted back with commit b425497

atulpendse:
Just FYI.. I was able to build it successfully on my Mac. Last time @garydgregory tried to submit build with this PR, it failed on Mac, not sure why.

log4j-core/revapi.json
```diff
@@ -128,6 +128,25 @@
         "annotation": "@org.apache.logging.log4j.core.config.plugins.PluginFactory",
         "justification": "LOG4J2-3317: Fix RoutingAppender backcompat while improving lookup security"
       },
+      {
+        "code": "java.method.numberOfParametersChanged",
```
garydgregory, line 132: You should deprecate the old method, not remove it.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -627,7 +608,8 @@ public static Rfc5424Layout createLayout(
             // RFC 5425
             @PluginAttribute(value = "useTlsMessageFormat") final boolean useTlsMessageFormat,
             @PluginElement("LoggerFields") final LoggerFields[] loggerFields,
-            @PluginConfiguration final Configuration config) {
+            @PluginConfiguration final Configuration config,
+            @PluginAttribute(value = "timestampPrecision") final TimestampPrecision timestampPrecision) {
```
garydgregory, line 612: Use a builder instead of breaking BC.

garydgregory:
Hello @atulpendse 
In general, you want to avoid breaking binary compatibility,  so that means not editing public and protected type names and method signatures. This is critical in the log4j-api module and less so in others, but, in this case it seems easy to do. Also, we have been switching from factory methods to builders as we go from release to release in order to avoid keeping on deprecating factory methods. So edit or add builders, don't edit factory methods is the idea. TY!

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -393,7 +408,8 @@ protected SyslogAppender(final String name, final Layout<? extends Serializable>
             final Charset charset,
             final String exceptionPattern,
             final LoggerFields[] loggerFields,
-            final boolean advertise) {
+            final boolean advertise,
+            final TimestampPrecision timestampPrecision) {
         // @formatter:on
```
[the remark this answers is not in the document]
atulpendse: I agree, but Rfc5424Layout does not have a builder which caller can use. It only supports creating instances via constructor.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -92,11 +97,12 @@
      */
     public static final String DEFAULT_MDCID = "mdc";
 
+    private static final FixedDateFormat FIXED_DATE_FORMAT_MILLI_PRECISION = FixedDateFormat
+            .create(FixedFormat.ISO8601_OFFSET_DATE_TIME_HHCMM_PERIOD);
+    private static final FixedDateFormat FIXED_DATE_FORMAT_MICRO_PRECISION = FixedDateFormat
+            .create(FixedFormat.ISO8601_OFFSET_DATE_TIME_HHCMM_PERIOD_MICRO);
+
```
ppkarwasz, line 104: I would use a (non static) `org.apache.logging.log4j.core.pattern.DatePatternConverter` to keep the layout garbage free.

rgoers:
@vy Why was this closed? To be honest, I am not sure why this hasn't been merged.

ppkarwasz:
It's a side effect of deleting the `release-2.x` branch: Github deletes the related PRs.

@atulpendse: can submit again against the `2.x` branch?

PR #684, opened 2022-01-05
Exclude Eclipse project file from RAT in root POM

vorburger:

PR #685, opened 2022-01-05
Fix broken PropertiesConfigurationTest

vorburger:
This fixes the problem seen e.g. on #684.


vorburger:
@garydgregory do you want to merge this when it passes?

garydgregory:
Note the config file has a / in it so you end up with two /
What is in the sys temp dir prop has a / at the end too IIRC in macOS at least. AFK for a few hours.

vorburger:
Indeed, now the CI failed on Mac (only) due to this.

To me this actually seems more like a bug in `org.apache.commons.lang3.SystemUtils#getJavaIoTmpDir()` - it really should be consistent between platforms.

But until that's fixed, it's probably simplest to put a big platform specific if around it...

@garydgregory it's probably simplest if I let you do whatever you deem best here to fix this, and step back and close this PR?

garydgregory:
Hi @vorburger 
I fixed the test differently in branches `release-2.x` and `master`.

PR #686, opened 2022-01-05
Use apache-rat-plugin's consoleOutput (and upgrade straggler versions)

vorburger:
See https://issues.apache.org/jira/browse/RAT-294,
and https://github.com/apache/logging-log4j2/pull/653.

### DIFF
--- a/log4j-bom/pom.xml
+++ b/log4j-bom/pom.xml
@@ -203,7 +203,7 @@
       <plugin>
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
-        <version>0.12</version>
+        <version>0.13</version>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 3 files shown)


vorburger:
This PR will fail to pass the CI build until the current known test failure is fixed, see #685.

vorburger:
@carterkozak perhaps you would like to merge this? (It's -about to- have a full green build.)

carterkozak:
Great find, thanks!

garydgregory:
I thought the RAT version was set in the parent POM through a property? I
don't think we want to override it, do we?

Gary

vorburger:
> I thought the RAT version was set in the parent POM through a property? I don't think we want to override it, do we?

This PR does not override it. If you look at the diff, you'll notice that it changes a `0.12` to a `0.13` in some places, and adds `<consoleOutput>true`. (I agree that it does look like some consolidation in the POM may be possible here, but that was not my aim & goal with this PR - I've merely changed what I found existed as-is.) Hope this makes sense?

garydgregory:
It does make perfect sense 😀 but it is not the best solution 😉 . All is well, thank you for pitching in 👍

PR #689, opened 2022-01-07
[LOG4J2-3319] Add support for log4j 1.2 API bundle in Eclipse IDE

benjaminr-ps:
This will add the "Eclipse-ExtensibleAPI" header to the manifest of "core" bundle, to allow any fragement (resp. the log4j 1.2 bridge) to extend the API of core bundle.

### DIFF
--- a/log4j-core/pom.xml
+++ b/log4j-core/pom.xml
@@ -511,6 +511,7 @@
               *
             </Import-Package>
             <Bundle-Activator>org.apache.logging.log4j.core.osgi.Activator</Bundle-Activator>
+            <Eclipse-ExtensibleAPI>true</Eclipse-ExtensibleAPI>
           </instructions>
         </configuration>
       </plugin>


garydgregory:
Should other modules include this type of change?

benjaminr-ps:
No, the change is made in **log4j-core**, only.

But have seen that several other libraries, like "log4j-appserver" or "log4j-cassandra" or ..., would benefit from this change, if developed within Eclipse IDE (resp. with their PDE tooling).

garydgregory:
https://issues.apache.org/jira/browse/LOG4J2-3319

jvz:
This looks like it should work. @garydgregory the fragments thing has to do with modules scanning for non-public bits for the most part. Generally, that means the 2.x plugin system needs the fragment hack to find and load plugins, but 3.x can rely on service definitions instead.

ppkarwasz:

PR #696, opened 2022-01-12
Strengthen test coverage of StrSubstitutor class

avandeursen:
I noticed that the central `substitute` method in the `StrSubstitutor` class that was involved in a recent CVE had some code not covered by (unit) tests. This untested logic related to cases where substitution in variables was disabled, and when no value escape delimiter (like `:\-`) was set.

I also noted some nice examples in the javadoc of `StrSubstitutor` that could be instructive test cases, and which indeed could cover some of the untested logic.

Therefore, this pull request:

- adds test cases for all examples listed in the `StrSubstitutor` javadoc
- applies small fixes to the javadoc
- adds test cases for value escape separator `:\\-` with and without a matcher for such escaped separators
- results in full statement coverage of the `substitute` method
- improves the branch coverage of the `substitute` method
- improves the test coverage of some helper methods.


carterkozak (approved):
thanks!

PR #697, opened 2022-01-13
[idea & example, request for input] Thoughts on using a code formatter?

carterkozak:
_This pr is meant only as an example for discussion_

What do you think about using a code formatter? In my experience, it makes reviews easier, avoids unnecessary style commits later on, and makes cherry-picking easier as style cannot diverge between branches that use the same formatter.
The formatter would allow us to make use of automation tooling without worrying about ruining formatting across the codebase.

Downside is that it doesn't allow for the formatter to be turned off for specific blocks of code, and doesn't support configuration.
Existing changes are likely to merge conflict, however that can be resolved by running the formatter on the change, creating a new commit based on the diff, and applying that.
The formatter requires some flags to work correctly on JDK17: https://github.com/diffplug/spotless/tree/main/plugin-maven#palantir-java-format


garydgregory:
FWIW, in Eclipse at least you can use a customizable comment to disable the
formatted for specific areas like this:

...
// @formatter:off
...
// this won't be formatted
...
// @formatter:on

Gary

rgoers:
This scares the heck out of me. 1,995 files modified means no human could review this. Gary used to (still does?) use tools to do formatting and invariably would change a piece of code and break it. For the life of me I can't recall what class that was but it happened at least 3 times.

So I get leery of tooling that creates PRs that no human could reasonably review.

carterkozak:
I'm not suggesting merging anything that formats the entire codebase in one go, this was meant as an example of the result before we commit to going down that path. Rollout is another beast, it can be done by applying constraints to files as they're modified rather than changing the world at once.

Committing to a particular formatter can be difficult because it front-loads the style bikeshedding which it eliminates later on. It may not be worthwhile, that's why I brought it up for discussion :-)

A lot of folks use the eclipse formatter, I've never found a configuration which works reasonably in all cases (lots of unnatural line breaks in unexpected places, but it may be better now).

PR #698, opened 2022-01-14
Support source class & method in to-jul (LOG4J2-3282)

vorburger:
This is a LOG4J2-3282 related follow-up to #653.

@carterkozak is this what you meant [in our conversation here](https://github.com/apache/logging-log4j2/pull/653#discussion_r778374993)? 😸  (Sorry at the time I didn't get it.)

### DIFF
--- a/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
+++ b/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
@@ -53,7 +53,7 @@ final class LazyLog4jLogRecord extends LogRecord {
         return super.getSourceMethodName();
     }
 
-    private synchronized void inferCaller() {
+    private void inferCaller() {
         StackTraceElement location = null;
         if (fqcn != null) {
             location = StackLocatorUtil.calcLocation(fqcn);
--- a/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
+++ b/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
@@ -22,15 +22,13 @@ import org.apache.logging.log4j.util.StackLocatorUtil;
 
 /**
  * Extension of {@link java.util.logging.LogRecord} with lazy get source related methods based on Log4j's {@link StackLocatorUtil#calcLocation(String)}.
- *
- * @author <a href="http://www.vorburger.ch">Michael Vorburger.ch</a> for Google
  */
-/* package-local */ final class LazyLog4jLogRecord extends LogRecord {
+final class LazyLog4jLogRecord extends LogRecord {
 
     private static final long serialVersionUID = 6798134264543826471L;
 
     // parent class LogRecord already has a needToInferCaller but it's private
-    private transient boolean stillNeedToInferCaller = true;
+    private transient boolean inferCaller = true;
 
     private final String fqcn;
 
@@ -41,7 +39,7 @@ import org.apache.logging.log4j.util.StackLocatorUtil;
 
     @Override
     public String getSourceClassName() {
-        if (stillNeedToInferCaller) {
+        if (inferCaller) {
             inferCaller();
         }
         return super.getSourceClassName();
@@ -49,16 +47,16 @@ import org.apache.logging.log4j.util.StackLocatorUtil;
 
     @Override
     public String getSourceMethodName() {
-        if (stillNeedToInferCaller) {
+        if (inferCaller) {
             inferCaller();
         }
         return super.getSourceMethodName();
     }
 
-    private void inferCaller() {
+    private synchronized void inferCaller() {
         StackTraceElement location = StackLocatorUtil.calcLocation(fqcn);
         setSourceClassName(location.getClassName());
         setSourceMethodName(location.getMethodName());
-        stillNeedToInferCaller = false;
+        inferCaller = false;
     }
 }
--- a/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
+++ b/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
@@ -21,11 +21,11 @@ import java.util.logging.LogRecord;
 import org.apache.logging.log4j.util.StackLocatorUtil;
 
 /**
- * Extension of {@link java.util.logging.LogRecord} with lazy get source related methods.
+ * Extension of {@link java.util.logging.LogRecord} with lazy get source related methods based on Log4j's {@link StackLocatorUtil#calcLocation(String)}.
  *
  * @author <a href="http://www.vorburger.ch">Michael Vorburger.ch</a> for Google
  */
-/* package-local */ final class LazyLogRecord extends LogRecord {
+/* package-local */ final class LazyLog4jLogRecord extends LogRecord {
 
     private static final long serialVersionUID = 6798134264543826471L;
 
@@ -34,7 +34,7 @@ import org.apache.logging.log4j.util.StackLocatorUtil;
 
     private final String fqcn;
 
-    LazyLogRecord(String fqcn, Level level, String msg) {
+    LazyLog4jLogRecord(String fqcn, Level level, String msg) {
         super(level, msg);
         this.fqcn = fqcn;
     }
--- a/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
+++ b/log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
@@ -54,9 +54,17 @@ final class LazyLog4jLogRecord extends LogRecord {
     }
 
     private synchronized void inferCaller() {
-        StackTraceElement location = StackLocatorUtil.calcLocation(fqcn);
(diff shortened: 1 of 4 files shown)


carterkozak:
> @carterkozak is this what you meant in our conversation here?

Yessir! Looks great :-) I probably could have described it more clearly.

Just needs a license header to appease rat.
We might consider naming the record type something more descriptive like `Log4jLogRecord`, but I defer to your judgement.

vorburger:
> Just needs a license header to appease rat.

Duh! I should know better; done!

> We might consider naming the record type something more descriptive like `Log4jLogRecord`

Done, now `LazyLog4jLogRecord` as that's really what this is (IMHO).

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,64 @@
...
+package org.apache.logging.log4j.tojul;
+
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import org.apache.logging.log4j.util.StackLocatorUtil;
+
+/**
+ * Extension of {@link java.util.logging.LogRecord} with lazy get source related methods based on Log4j's {@link StackLocatorUtil#calcLocation(String)}.
+ *
+ * @author <a href="http://www.vorburger.ch">Michael Vorburger.ch</a> for Google
```
garydgregory, line 26: Please remove author tag. You can be credit in changes.xml and/or pom.xml.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+/* package-local */ final class LazyLog4jLogRecord extends LogRecord {
```
garydgregory, line 28: This comment is not needed IMO.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private static final long serialVersionUID = 6798134264543826471L;
+
+    // parent class LogRecord already has a needToInferCaller but it's private
+    private transient boolean stillNeedToInferCaller = true;
```
garydgregory, line 33: At least stillNeedToInferCaller -> inferCaller or flip the boolean value and call it callerInitialized?

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    public String getSourceMethodName() {
+        if (stillNeedToInferCaller) {
+            inferCaller();
+        }
+        return super.getSourceMethodName();
+    }
+
+    private void inferCaller() {
```
garydgregory, line 58: Should this be synchronized to avoid extra work?

garydgregory (changes requested):
Please see my comments/questions. TY!

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
carterkozak: synchronized creates more work than it avoids! If a logrecord is passed to another thread it must either be wrapped to avoid searching for callsite data, or must be filled in prior to being passed (much like log4j createMemento)
This matches `java.util.logging.LogRecord` semantics.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: Parent class `LogRecord` in the JDK already has a field `needToInferCaller` (it's private) so this seemed consisted, to me. But I'm happy to rename it to `inferCaller` - done. I do not want to _flip the boolean value and call it callerInitialized_ because that would make it inconsistent with the parent class.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: Sure, done.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: Sure, removed.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: Sure, removed. I noticed there is already an entry for https://issues.apache.org/jira/browse/LOG4J2-3282 in `changes.xml`, so from my side we could consider this simply an improvement on that, without an additional entry. I'll raise a separate PR re. `<contributors>` in `pom.xml`.

vorburger:
> Please see my comments/questions. TY!

Done!

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public String getSourceMethodName() {
+        if (inferCaller) {
+            inferCaller();
+        }
+        return super.getSourceMethodName();
+    }
+
+    private synchronized void inferCaller() {
```
carterkozak, line 56: I don't think synchronized is necessary based on how this class is used and the thread-safety of the parent class. Synchronization here doesn't buy us anything because it only guards the method itself, not the `inferCaller` check elsewhere (which isn't volatile, nor do I think it should be) so it would only prevent additional concurrent calls from running in parallel.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
```diff
@@ -0,0 +1,62 @@
...
+        StackTraceElement location = StackLocatorUtil.calcLocation(fqcn);
+        setSourceClassName(location.getClassName());
+        setSourceMethodName(location.getMethodName());
```
carterkozak, line 59: I think we may need some null checks here in case `fqcn` is null, or the resulting `location` is null. In either case we should probably explicitly set the `sourceClassName` and `sourceMethodName` to `null`.

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: @garydgregory @carterkozak can I let you guys hash this out among ourselves and let me know your conclusion? I'm happy if you are (both) happy, but I'd prefer not going back and forth a few times between you... 😸  I believe this comment appears to request to (re-)remove `synchronized` (again, after @garydgregory suggested it be added), and also for `volatile` to be removed from `inferCaller`. Please both ACK?

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: Done, hope looks about right like that?

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
vorburger: s/ourselves/yourselves/

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
carterkozak: :+1:

garydgregory (approved):

garydgregory:
I approved the PR but as my comments are resolved. I'll let @carterkozak give the final nod.😉
TY @vorburger

carterkozak:
Thanks, Gary :-)

log4j-to-jul/src/main/java/org/apache/logging/log4j/tojul/LazyLog4jLogRecord.java
carterkozak: Let's remove the `synchronized` keyword please, sorry for the churn!

>  and also for volatile to be removed from inferCaller

I don't think `inferCaller` is volatile, no change needed there!

vorburger: OK, `synchronized` removed again now.

> I don't think `inferCaller` is volatile, no change needed there!

Duh, on TGIF the `transient` morphed into a `volatile` in my mind 😈

carterkozak (approved):
Looks good, thanks! I'll get it merged once I have a moment :-)

PR #699, opened 2022-01-14
Increase to-jul priority from 15 to 20

vorburger:
The original 15 was blindly copied from SLF4JProvider (by myself).

It's probably better to not have them be the same, and have To-JUL's higher than SLF4j's (which in turn is higher than Core's 10).


vorburger:
@carterkozak

carterkozak (approved):

PR #700, opened 2022-01-14
StatusLogger uses default status level configured from property not just for listeners but all its logging (fixes LOG4J2-3340)

vorburger:
With this StatusLogger uses default status level configured from property not just for listeners but all its logging.

This fixes https://issues.apache.org/jira/browse/LOG4J2-3340 for me.

If this is acceptable, I can raise a follow-up PR to do further clean-up / simplifcation around this topic.


garydgregory:
Anything that makes it easier to use is fine by me.

Any thoughts @rgoers ?

carterkozak:
Are the test failures related to this change?

rgoers:
I'm not sure we have that many tests on StatusLogger. This change looks like it makes sense to me.

garydgregory:
> Are the test failures related to this change?

I just started a re-run of all the jobs. Let's see if the build passes.

garydgregory:
This PR is indeed the cause of the build failure.

PR #703, opened 2022-01-15
fix deleteAction can't distinguish basePath of symbolicLink

strugerk:
fix Appenders DeleteAction didn't work when basepath is symbolicLink.
jira:https://issues.apache.org/jira/browse/LOG4J2-3342

image
image

getSortedPaths should return subFile [log4j-shix-....log],but it only return currentFile dirName [log/file]
problem:Files.walkFileTree api  in java.nio can't list File directly when file is symbolicLink ,it need do someThing to deal with it

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/AbstractPathAction.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/AbstractPathAction.java
@@ -103,7 +103,15 @@ public abstract class AbstractPathAction extends AbstractAction {
      * @return the base path (all lookups resolved)
      */
     public Path getBasePath() {
-        return Paths.get(subst.replace(getBasePathString()));
+        Path path = Paths.get(subst.replace(getBasePathString()));
+        if (Files.isSymbolicLink(path)) {
+            try {
+                path = Files.readSymbolicLink(path);
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+        return path;
     }
 
     /**


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/AbstractPathAction.java
```diff
@@ -103,7 +103,15 @@ public boolean execute(final FileVisitor<Path> visitor) throws IOException {
      * @return the base path (all lookups resolved)
      */
     public Path getBasePath() {
-        return Paths.get(subst.replace(getBasePathString()));
+        Path path = Paths.get(subst.replace(getBasePathString()));
+        if (Files.isSymbolicLink(path)) {
+            try {
+                path = Files.readSymbolicLink(path);
+            } catch (IOException e) {
+                e.printStackTrace();
```
garydgregory, line 111: Printing the stack trace is a no-no. Either recover from the problem, propagate, or log to the status logger, I'm not sure which is best here.

garydgregory: Also, should the replace call happen, before, after, or both?

strugerk: ok, I think we should throw out the exception.  
It's necessary to capture the exception , because this IOException has been caught outside.
The status logger can also be printed when handling exceptions externally

strugerk: I hava changed code，but  compile fail，do you know why? By the way，it compile normally  in my local.

ppkarwasz:

PR #704, opened 2022-01-15
Multiple small improvements in Log4j 1.x configuration parsing

ppkarwasz:
The unit tests provided by `Log4j1ConfigurationFactoryTest` unfortunately were not applied to the newer `PropertiesConfigurationFactory` and `XmlConfigurationFactory`. After applying the tests, many small errors appeared, which this PR solves:

*  the `XmlConfiguration` didn't allow for system property replacement ([LOG4J2-3328]),
* the `TTCCLayoutBuilder` lacked proper default values for its properties,
* the `RollingFileAppender` and `DailyRollingFileAppender` used incorrect file patterns,
* the `XmlLayoutBuilder` used the native Log4j 2.x `XmlLayout` instead of the compatible `Log4j1XmlLayout`,
* builders that did not inherit from `AbstractBuilder` failed to be instantiated.

For an easier cherry-pick, I tried to solve each problem in a separate commit.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
@@ -83,7 +83,7 @@ public abstract class AbstractBuilder {
     }
 
     protected String getNameAttribute(Element element) {
-        return element.getAttribute(NAME_ATTR);
+        return capitalize(element.getAttribute(NAME_ATTR));
     }
 
     protected String getValueAttribute(Element element) {
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
@@ -91,7 +91,11 @@ public abstract class AbstractBuilder {
     }
 
     public boolean getBooleanProperty(String key) {
-        return Boolean.parseBoolean(getProperty(key, Boolean.FALSE.toString()));
+        return getBooleanProperty(key, false);
+    }
+
+    public boolean getBooleanProperty(String key, boolean defaultValue) {
+        return Boolean.parseBoolean(getProperty(key, Boolean.toString(defaultValue)));
     }
 
     public int getIntegerProperty(String key, int defaultValue) {
@@ -151,4 +155,11 @@ public abstract class AbstractBuilder {
         return value == null ? null : value.toLowerCase(Locale.ROOT);
     }
 
+    protected String capitalize(final String value) {
+        if (value == null || value.length() == 0)
+            return value;
+        char[] chars = value.toCharArray();
+        chars[0] = Character.toUpperCase(chars[0]);
+        return new String(chars);
+    }
 }
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
@@ -83,7 +83,16 @@ public abstract class AbstractBuilder {
     }
 
     protected String getNameAttribute(Element element) {
-        return capitalize(element.getAttribute(NAME_ATTR));
+        return element.getAttribute(NAME_ATTR);
+    }
+
+    /**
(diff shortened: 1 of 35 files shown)


minalabib223 (approved):

garydgregory:
Hello @ppkarwasz 
Thank you for your PR. You must have missed in Jira that the ticket https://issues.apache.org/jira/browse/LOG4J2-3328 was assigned to me, so we were both working on the same issue at the same time. I committed a fix to branches `release-2.x` and `master`. The fix must be different from what you propose here because we incorrectly had the abstract builder set up to use the Log4j 2 string substitutor instead of the Log4j 1 option converter. This has been addressed. This PR also mixes in too many different changes and covers too many files (39!) IMO for a reasonable review. Granted, refactoring common code requires touching a lot of files, I understand that. Nevertheless, I think there are still valuable changes here that we could review after you rebase this PR on branch `release-2.x`.
Thank you for your efforts and patience :-)

ppkarwasz:
Sorry @garydgregory, 

I didn't plan to provide a fix to LOG4J2-3328 at first, but some of the tests the I copied from `Log4j1ConfigurationFactoryTest` were failing... :-D

I rebased my PR on top of your changes. The number of changed files is still high (30), but only 18 are Java files (I added 11 Log4j 1.x style XML configuration files) and only 14 are in the main code. The list of changes:

* the `TTCCLayoutBuilder` lacked proper default values for its properties,
* the `RollingFileAppender` and `DailyRollingFileAppender` used incorrect file patterns,
* the `XmlLayoutBuilder` used the native Log4j 2.x `XmlLayout` instead of the compatible `Log4j1XmlLayout`,
* builders that did not inherit from `AbstractBuilder` failed to be instantiated,
* the `Log4j1ConfigurationParser` used Log4j 2.x `StrSubstitutor` instead of Log4j 1.x `OptionConverter`.

Each of these issues requires only a couple of lines to fix, but I can split the PR if it helps.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
```diff
@@ -86,12 +86,25 @@ protected String getNameAttribute(Element element) {
...
 
+    /**
+     * Normalized version of the "name" attribute of the &lt;param&gt; tag.
+     * @param element
+     * @return
+     */
+    protected String getNormalizedNameAttribute(Element element) {
+        return capitalize(getNameAttribute(element));
+    }
+
```
garydgregory, line 97: Are all call sites of getNameAttribute() now call sites of this new method? If so, just update getNameAttribute() IMO.

garydgregory (commented):
Can you confirm that all new XML files are actually used from new tests?

log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
```diff
@@ -86,12 +86,25 @@ protected String getNameAttribute(Element element) {
         return element.getAttribute(NAME_ATTR);
     }
 
+    /**
+     * Normalized version of the "name" attribute of the &lt;param&gt; tag.
+     * @param element
+     * @return
```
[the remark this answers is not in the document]
ppkarwasz: Done.

rgoers:
Yeah - I would appreciate this being split into a series of PRs. I got part of that through and didn't see anything terribly concerning but I also did dig as deep as I would like to into the changes.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
ppkarwasz: No, `getNameAttribute` is used for retrieving the `"name"` attribute of both the `<appender>` and `<param>` tags. AFAIK appender names are case sensitive.

ppkarwasz:
I split this PR into 6 smaller ones: #706, #707, #708, #709, #710, #711.

gurucode007:
what is the fix for
the XmlConfiguration didn't allow for system property replacement ([LOG4J2-3328]),

gurucode007:
@ppkarwasz we are having similar issue after using 2.17.2 and 2.18.0 version.
https://issues.apache.org/jira/browse/LOG4J2-3592

Any idea why?

PR #706, opened 2022-01-16
Common unit tests for Log4j 1.x configuration factories

ppkarwasz:
`Log4j1ConfigurationFactoryTest` contains many interesting tests that fail if we apply them to the `PropertiesConfigurationFactory` and `XmlConfigurationFactory` that replaced `Log4j1ConfigurationFactory`.

This PR adapts those tests to the other two factories and provides 11 test XML configuration files. It was split from #704 for simpler analysis. Obviously many of the new tests **fail**.

### DIFF
--- a/log4j-1.2-api/src/test/java/org/apache/log4j/config/AbstractLog4j1ConfigurationTest.java
+++ b/log4j-1.2-api/src/test/java/org/apache/log4j/config/AbstractLog4j1ConfigurationTest.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


log4j-1.2-api/src/test/java/org/apache/log4j/config/AbstractLog4j1ConfigurationTest.java
```diff
@@ -0,0 +1,264 @@
...
+import org.apache.logging.log4j.core.appender.rolling.TimeBasedTriggeringPolicy;
+import org.apache.logging.log4j.core.appender.rolling.TriggeringPolicy;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.LoggerConfig;
+import org.apache.logging.log4j.core.layout.HtmlLayout;
+import org.apache.logging.log4j.core.layout.PatternLayout;
+
+public abstract class AbstractLog4j1ConfigurationTest {
+
+	abstract Configuration getConfiguration(String configResourcePrefix) throws URISyntaxException, IOException;
```
garydgregory, line 59: @ppkarwasz 
Use spaces, no tabs. Follow the style of the file you edit.

garydgregory (changes requested):
@ppkarwasz 
Use spaces, no tabs. Follow the style of the file you edit.

garydgregory:
Hi @ppkarwasz 
Failing a build is not going to be acceptable IMO. Because different team members work at different times and availability, like  me, I don't want to bring in code that breaks the build, I might have to step away and work on something else, then we would be left with a failing project build. Each PR should pass the build. If a PR fixes a bug, it should come with a test that fails without the main change.
TY.

ppkarwasz:
I'll move the tests to the PRs that actually fix bugs, then. Thanks.

PR #707, opened 2022-01-17
Allows Java-style property names in Log4j 1.x XML configuration

ppkarwasz:
Log4j 1.x allows for both capitalized (`InfoLocation`) and Java-style (`infoLocation`) component property names (`PropertySetter` calls `Introspector.decapitalize`). This PR allows for Java-style property names in the Log4j 1.x bridge. It also fixes some capitalization typos (e.g. `target` instead of `Target`).

This fixes most XML tests from #706 .

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
@@ -83,7 +83,7 @@ public abstract class AbstractBuilder {
     }
 
     protected String getNameAttribute(Element element) {
-        return element.getAttribute(NAME_ATTR);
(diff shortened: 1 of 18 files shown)


garydgregory:
Hi @ppkarwasz 
You need tests to fail without these main changes to show the main changes actually fix something. Otherwise, future changes might just reintroduce the old behavior unbeknownst to anyone.
TY!

garydgregory:
@ppkarwasz 
Thank you for your PR! It turns out I had an implementation in flight similar to this PR. Please see the latest from git. You are credited in changes.xml, as I was able to bring in all of these test changes as is.

ppkarwasz:
@garydgregory 
Looks great and since `AbstractLog4j1ConfigurationTest` is in the repository now, I can rebase the remaining PRs on top of your commit. This will reduce the size of the changes.

PR #709, opened 2022-01-17
Log4j 1.2 bridge uses the wrong default values for a TTCCLayout

ppkarwasz:
The default values used by `TTCCLayoutBuilder` were different from those used by Log4j 1.x.

This PR fixes the tests concerning the `TTCCLayout` from #706.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
@@ -109,8 +109,12 @@ public abstract class AbstractBuilder {
         return new String(chars);
     }
 
+    public boolean getBooleanProperty(final String key, final boolean defaultValue) {
+        return Boolean.parseBoolean(getProperty(key, Boolean.toString(defaultValue)));
+    }
+
     public boolean getBooleanProperty(final String key) {
-        return Boolean.parseBoolean(getProperty(key, Boolean.FALSE.toString()));
+        return getBooleanProperty(key, false);
     }
 
     public int getIntegerProperty(final String key, final int defaultValue) {
@@ -174,5 +178,4 @@ public abstract class AbstractBuilder {
         chars[0] = Character.toLowerCase(chars[0]);
         return new String(chars);
     }
-
 }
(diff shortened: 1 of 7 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/builders/AbstractBuilder.java
```diff
@@ -91,7 +91,11 @@ protected String getValueAttribute(Element element) {
     }
 
     public boolean getBooleanProperty(String key) {
-        return Boolean.parseBoolean(getProperty(key, Boolean.FALSE.toString()));
+        return getBooleanProperty(key, false);
+    }
+
+    public boolean getBooleanProperty(String key, boolean defaultValue) {
```
garydgregory, line 97: Good refactoring! :-)

log4j-1.2-api/src/main/java/org/apache/log4j/builders/layout/TTCCLayoutBuilder.java
```diff
@@ -91,10 +91,10 @@ public Layout parseLayout(Element layoutElement, XmlConfiguration config) {
 
     @Override
     public Layout parseLayout(PropertiesConfiguration config) {
-        boolean threadPrinting = getBooleanProperty(THREAD_PRINTING_PARAM);
-        boolean categoryPrefixing = getBooleanProperty(CATEGORY_PREFIXING_PARAM);
-        boolean contextPrinting = getBooleanProperty(CONTEXT_PRINTING_PARAM);
-        String dateFormat = getProperty(DATE_FORMAT_PARAM);
+        boolean threadPrinting = getBooleanProperty(THREAD_PRINTING_PARAM, true);
```
garydgregory, line 94: I can see this change matches what the defaults are indeed in Log4j 1.2 (true values) but we need a test that shows that this change works.
TY!

garydgregory (changes requested):
Hello @ppkarwasz 
Thank you for your PR.
We need a matching test that fails if the main changes are reverted, otherwise, there is no way to know why this change is "right" and the current code "wrong".
TY!

log4j-1.2-api/src/main/java/org/apache/log4j/builders/layout/TTCCLayoutBuilder.java
ppkarwasz: I added a test, which was failing before.

garydgregory:
@ppkarwasz 
Please rebase on branch `release-2.x`.

ppkarwasz:
@garydgregory: done.

PR #717, opened 2022-01-17
LOG4J2-3333: Fix ThreadContextDataInjector classloader deadlock

carterkozak:
draft, I need to validate performance related to the original reason this ran on another thread:  https://issues.apache.org/jira/browse/LOG4J2-2867

Now that the serviceloader is guarded by class init, it should be less costly, however I wonder if we can deduplicate out some parent classloaders from loaderutil.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
@@ -63,6 +63,13 @@ public class ThreadContextDataInjector {
 
     private static final List<ContextDataProvider> SERVICE_PROVIDERS = getServiceProviders();
 
+    /**
+     * Previously this method allowed ContextDataProviders to be loaded eagerly, now they
+     * are loaded when this class is initialized.
+     *
+     * @deprecated no-op
+     */
+    @Deprecated
     public static void initServiceProviders() {}
 
     private static List<ContextDataProvider> getServiceProviders() {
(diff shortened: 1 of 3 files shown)


carterkozak:
@rgoers I can't seem to find the performance test you added as `log4j-core/src/test/java/org/apache/logging/log4j/core/LoggingTest.java`
In the original issue, initialization of this component was run several times per context/configuration which is no longer the case, ideally we can avoid additional threads altogether thus avoiding deadlocks.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
```diff
@@ -62,21 +62,9 @@
     public static Collection<ContextDataProvider> contextDataProviders =
             new ConcurrentLinkedDeque<>();
 
-    private static volatile List<ContextDataProvider> serviceProviders = null;
-    private static final Lock providerLock = new ReentrantLock();
+    private static final List<ContextDataProvider> serviceProviders = getServiceProviders();
```
garydgregory, line 65: Our static finals (constants) should be in upper-case.

carterkozak: Good call, I'll update the initializer such that the list itself is unmodifiable as well.

rdotterer09:
Ok

carterkozak:
committed as fc31b44b629d9fe46bd96bfd02382519dfdb5f40

rdotterer09:
thanks for the help

PR #720, opened 2022-01-19
Expand test suite and fix NPE for StructuredDataLookup class

avandeursen:
This PR strengthens the test suite for the `StructuredDataLookup` class so that it achieves full branch coverage. The corner cases revealed the problem that a lookup of a null key with a correct event led to a null pointer exception. This is inconsistent with the javadoc ("can be null") and the other lookup classes.

The PR fixes the null pointer exception for the `key` by essentially replacing `key.equals("id")` by `"id".equals(key)`, in two places.

It also introduces explicit constants for the string literals "id" and "type".

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/StructuredDataLookup.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/StructuredDataLookup.java
@@ -27,8 +27,18 @@ import org.apache.logging.log4j.message.StructuredDataMessage;
 public class StructuredDataLookup implements StrLookup {
 
     /**
+     * Key to obtain the id of a structured message.
+     */
+    public static final String ID_KEY = "id";
+
+    /**
+     * Key to obtain the type of a structured message.
+     */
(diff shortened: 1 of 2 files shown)


avandeursen:
I doubt the failing build is related to the code change here. Is there an easier way to re-trigger the build than me pushing an empty commit?

carterkozak:
I've retriggered the jobs :-)

avandeursen:
That did the trick. Thanks!

rdotterer09:
Ok

avandeursen:
This PR seems a bit stalled, so let me re-explain why I think it's important to merge it:

- Doing a lookup in structured data with a null key is valid according to the API documentation, but yields a null pointer exception
- This is a bug in the code that is easily fixed by just swapping the arguments of two equality checks.

This  PR fixes  the bug, and provides a systematic set of  test cases for the class, exposing the bug.

vy:
Thanks so much @avandeursen! Changes look good to me, merging.

Kay-kay2019:
>Thank you 🙂

PR #721, opened 2022-01-19
Expand test suite for DateLookup

avandeursen:
This commit strengthens the test suite for the `DateLookup` class so that it achieves full branch coverage.

The javadoc for the `lookup(event, key)` method incorrectly stated that the key was ignored.
This has been updated to clarify that the date of the event is shown.

The PR replaces the hand-written `MyLogEvent` with a mockito mock class.

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
@@ -23,31 +23,42 @@ import org.apache.logging.log4j.core.LogEvent;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class DateLookupTest {
 
-
     @Test
-    public void testLookup() {
-        final StrLookup lookup = new DateLookup();
-        final LogEvent event = new MyLogEvent();
-        final String value = lookup.lookup(event, "MM/dd/yyyy");
-        assertNotNull(value);
-        assertEquals("12/30/2011", value);
+    public void testDateLookupInEvent() {
+        final LogEvent mockedEvent = mock(LogEvent.class);
+        final Calendar cal = Calendar.getInstance();
+        cal.set(2011, Calendar.DECEMBER, 30, 10, 56, 35);
+        when(mockedEvent.getTimeMillis()).thenReturn(cal.getTimeInMillis());
+
+        final String actualDate = new DateLookup().lookup(mockedEvent, "MM/dd/yyyy");
+        assertEquals("12/30/2011", actualDate);
     }
 
-    private static class MyLogEvent extends AbstractLogEvent {
-        /**
-         * Generated serial version ID.
-         */
-        private static final long serialVersionUID = -2663819677970643109L;
+    @Test
+    public void testDateLookUpNoEvent() {
+        final String dateFormat = "MM/dd/yyyy";
+        final String actualDate = new DateLookup().lookup(null, dateFormat);
+        // we don't know current time, so check length of format instead.
+        assertEquals(actualDate.length(), dateFormat.length());
+    }
 
-        @Override
-        public long getTimeMillis() {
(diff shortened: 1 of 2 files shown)


log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
```diff
@@ -23,31 +23,42 @@
...
-    private static class MyLogEvent extends AbstractLogEvent {
-        /**
-         * Generated serial version ID.
-         */
-        private static final long serialVersionUID = -2663819677970643109L;
+    @Test
+    public void testDateLookUpNoEvent() {
+        final String dateFormat = "MM/dd/yyyy";
+        final String actualDate = new DateLookup().lookup(null, dateFormat);
+        // we don't know current time, so check length of format instead.
```
vy, line 46: `lookup(null, format)` simply delegates to `formatDate(System.currentTimeMillis(), format)`, right? Hence, we know what is expected – indeed second-precision values are not reliable, but we certainly know the expected `MM/dd/yyyy` output. Mind updating this test accordingly, please?

log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
```diff
@@ -23,31 +23,42 @@
...
-        public long getTimeMillis() {
-            final Calendar cal = Calendar.getInstance();
-            cal.set(2011, Calendar.DECEMBER, 30, 10, 56, 35);
-            return cal.getTimeInMillis();
-        }
+    @Test
+    public void testNullDateLookupYieldsCurrentTime() {
+        final String currentTime = new DateLookup().lookup(null, null);
+        assertNotNull(currentTime);
+        // we don't know current time nor format, so just check length of date string.
```
vy, line 54: By following the logic in `DateLookup`, we know the current time (see my comment above) and the format: `DateFormat.getInstance()`. Mind updating this test accordingly, please?

log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
```diff
@@ -23,31 +23,42 @@
...
+        assertTrue(currentTime.length() > 0);
+    }
 
+    @Test
+    public void testInvalidFormatYieldsDefaultFormat() {
+        final String currentTime = new DateLookup().lookup(null, "bananas");
+        // we don't know current time nor format, so just check length of date string.
```
vy, line 61: By following the logic in `DateLookup`, we can guess both these – see my earlier comments. Mind improving this test too, please?

vy (changes requested):
Thanks so much for the contribution @avandeursen! I have really appreciated it. I think it will be better off if we can improve the quality of the assertions a bit rather than simple length checks. See my comments.

avandeursen:
Thanks! Based on the feedback I made a few changes.

For the test with the known format, I indeed added the test whether the returned date is today (and even added a test for tomorrow, in case the test is executed at midnight).

For the test cases falling back on the default format, most likely the time is included in that format, making this route a bit too flaky to my taste. So I decided to better document the rationale of the test instead.

While engaging with the DateLookup class, I also made some of the javadocs more accurate.

Let me know what you think!

log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
avandeursen: This would typically be at the time level of minutes -- with quite a chance of becoming flaky. So instead I better documented the rationale.

log4j-core/src/test/java/org/apache/logging/log4j/core/lookup/DateLookupTest.java
avandeursen: See above.

vy:
@avandeursen, I think the tests could be improved in many angles, but yours is certainly an improvement anyway and I have neither time, nor interest to invest on this further. As a tip for your next PR, I'd appreciate `@link` and `@code` expressions in javadocs rather than literals.

All in all, good job, thanks so much! :+1: Keep it up! :100:

PR #722, opened 2022-01-19
LOG4J2-3350: Evaluate configuration property substitutions eagerly

carterkozak:
Previously lookups between configuration properties weren't evaluated
until they were used in the configuration, which resulted in
behavior more prohibitive for RoutingAppenders than anticipated.

This attempts to pre-compute the values of top-level configuration
properties so they don't have to be re-evaluated on an as-needed
basis.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
@@ -40,13 +40,19 @@ public final class Property {
     private static final Logger LOGGER = StatusLogger.getLogger();
 
     private final String name;
+    private final String rawValue;
     private final String value;
     private final boolean valueNeedsLookup;
 
-    private Property(final String name, final String value) {
+    private Property(final String name, final String rawValue, final String value) {
         this.name = name;
+        this.rawValue = rawValue;
         this.value = value;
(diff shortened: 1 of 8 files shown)


rgoers:
Other than adding a couple of comments this looks good to me.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
```diff
@@ -57,6 +63,14 @@ public String getName() {
         return name;
     }
 
+    /**
+     * Returns the original raw property value without substitution.
+     * @return the raw value of the property, or null if it is not set.
+     */
+    public String getRawValue() {
```
carterkozak, line 70: I'm not thrilled about adding these new methods because they muddy the waters between the object model and configuration node->plugin api.

Ideally we could add a parameter to the `@PluginElement("Properties") final Property[] properties` parameter in `PropertiesPlugin` to opt out of all substitutions when that data is read, allowing us to produce the same behavior. When I looked into that, it didn't seem like something the plugin system is built to handle (contextual configuration).

rdotterer09: Ok

carterkozak:
Closing in favor of #732

PR #723, opened 2022-01-20
Fix `log4j-jakarta-web` service file

ppkarwasz:
The META-INF/services file used the old `javax` namespace, hence the `ServletContainerInitializer` was never detected.


garydgregory:
@ppkarwasz 
Please rebase on branch `release-2.x` to pick up fixes that should make the build more reliable (hopefully).

PR #726, opened 2022-01-21
 LOG4J2-3358: Fix substitutions when programmatic configuration is used

carterkozak:
The problem is that some configuration creation paths fail to
use the Interpolator constructor which loads all plugins. This
produces inconsistent behavior based on the configuration construction
path.
By merging code-paths, we can provide consistent behavior.

### DIFF
--- a/log4j-core/src/test/java/org/apache/logging/log4j/core/layout/JsonLayoutTest.java
+++ b/log4j-core/src/test/java/org/apache/logging/log4j/core/layout/JsonLayoutTest.java
@@ -33,11 +33,14 @@ import org.apache.logging.log4j.ThreadContext;
 import org.apache.logging.log4j.categories.Layouts;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.BasicConfigurationFactory;
+import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.Logger;
 import org.apache.logging.log4j.core.LoggerContext;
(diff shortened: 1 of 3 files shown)


log4j-core/src/test/java/org/apache/logging/log4j/core/layout/JsonLayoutTest.java
```diff
@@ -604,6 +608,37 @@ public void testEmptyValuesAreIgnored() {
         assertFalse(str, str.contains("\"empty\""));
     }
 
+    @Test // LOG4J2-3358
+    public void jsonLayout_should_substitute_lookups() {
```
vy, line 612:
```suggestion
    /**
     * @see <a href="https://issues.apache.org/jira/browse/LOG4J2-3358">LOG4J2-3358</a>
     */
    @Test
    public void ctx_lookup_should_work() {
```

carterkozak:
TODO: More description and context, I'll try to write something up by the end of the day

log4j-core/src/test/java/org/apache/logging/log4j/core/layout/JsonLayoutTest.java
carterkozak: Structure borrowed from the test above: https://github.com/apache/logging-log4j2/blob/e7af4e0a1ead739598a5ca96d09bc214b68cf0ec/log4j-core/src/test/java/org/apache/logging/log4j/core/layout/JsonLayoutTest.java#L598-L599

Happy to change both

PR #729, opened 2022-01-26
Fixes a StringIndexOutOfBoundsException

ppkarwasz:
`PropertiesUtilTest#partitionOnCommonPrefixes` fails if a property does not contain any dots.

The bug appears when parsing a `log4j2.properties` like in [this StackOverflow question](https://stackoverflow.com/q/70851549/11748454).

I modified the data used by `PropertiesUtilTest` to fail with a `StringIndexOutOfBoundsException`.


vy:
Thanks so much @ppkarwasz!

jvz:
Noticed this is needed in master as well in order to support the new properties format features.

PR #730, opened 2022-01-26
LOG4J2-3368 Bump Tomcat from 8.5.20 to 10.0.14 in log4j-appserver along with some dependency clean-up.

vy:


carterkozak:
I defer to @garydgregory for Tomact knowledge, but I'd question bumping a minimum dependency constraint two major versions in a patch release.

ppkarwasz:
The only Tomcat class used by Log4j is `org.apache.juli.logging.Log`, which has not been modified in ages (cf. [Github](https://github.com/apache/tomcat/blob/main/java/org/apache/juli/logging/Log.java)). It is safe to bump the dependency.

However, since dependencies are being cleaned-up, you might consider replacing `tomcat-catalina` with [`tomcat-juli`](https://mvnrepository.com/artifact/org.apache.tomcat/tomcat-juli), since that is the artifact actually used.

PR #731, opened 2022-01-27
Adds a Tomcat lookup

ppkarwasz:
The `TomcatLookup` can be used to retrieve the engine, host and context names under which the application is running. These are also provided by the original Tomcat JULI implementation (cf. [ClassLoaderLogManager#replaceWebApplicationProperties](https://github.com/apache/tomcat/blob/a05603de0fc5ae1ee60d41ea8677f96818a9b11f/java/org/apache/juli/ClassLoaderLogManager.java#L654)).

It also provides three helper properties `*.logger` that resolve to the logger name used by `ServletContext#log`.

### DIFF
--- a/log4j-appserver/src/main/java/org/apache/logging/log4j/appserver/tomcat/TomcatLookup.java
+++ b/log4j-appserver/src/main/java/org/apache/logging/log4j/appserver/tomcat/TomcatLookup.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


log4j-appserver/src/main/java/org/apache/logging/log4j/appserver/tomcat/TomcatLookup.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String SERVICE_LOGGER_FORMAT = "org.apache.catalina.core.ContainerBase.[%s]";
+    private static final String HOST_LOGGER_FORMAT = "org.apache.catalina.core.ContainerBase.[%s].[%s]";
+    private static final String CONTEXT_LOGGER_FORMAT = "org.apache.catalina.core.ContainerBase.[%s].[%s].[%s]";
+
+    @Override
+    public String lookup(String key) {
+        final ClassLoader cl = Thread.currentThread().getContextClassLoader();
+        if (cl instanceof WebappProperties) {
+            final WebappProperties props = (WebappProperties) cl;
+            switch (key) {
```
avandeursen, line 41: The `StrLookup` interface explicitly allows the `key` parameter of the `lookup` method to be null. This would trigger a `NullPointerException` here. Perhaps add a condition to check that `key != null`?

ppkarwasz: Corrected, thanks.

vy:
I am against introducing components for 3rd party software integrations. These better be developed and maintained by the community; Log4j provides first-class extensibility mechanisms for this very reason. The core is complicated enough; there is a never-ending backlog of issues and improvements that can take the lifetime of the entire team. I think maintainers should rather focus on those.

PR #732, opened 2022-01-27
LOG4J2-3350: Lookup recursion is not allowed, except through config properties

carterkozak:
Previously lookups between configuration properties weren't evaluated
until they were used in the configuration, which resulted in
behavior more prohibitive for RoutingAppenders than anticipated.

This attempts to provide a safer, more compatible lookup strategy
with consistent behavior between users.

**TODO(ckozak): needs additional test coverage, but I'd like to share the concept first**

## How does property substitution work on `release-2.x`?

```xml
<Configuration>
  <Properties>
    <Property name="path">/tmp/</Property>
    <Property name="name">test.${lower:LOG}</Property>
    <Property name="filename">${path}${name}</Property>
    <Property name="pattern">%m%n</Property>
  </Properties>
  <Appenders>
    <RollingFile name="rolling" fileName="${filename}">
      <PatternLayout>
        <pattern>$${filename} ${pattern}</pattern>
      </PatternLayout>
    </RollingFile>
  </Appenders>
</Configuration> 
```

When this configuration is read, we end up with a Substitutor based on the properties defined in the top of the file.
It is build using standard PluginBuilder components, so the plugin value component undergoes substitution using the initial StrSubstitutor, however it's not yet aware of other properties because we're still parsing them. So, we evaluate the properties in order:

* `path`: `/tmp/` (unmodified)
* `name`: `test.${lower:LOG}` -> `test.log` due to `Interpolator.lookup("lower:LOG")` -> `LowerLookup.lookup("LOG")` -> `"log"`
* `filename`: `${path}${name}` (unmodified) the substitutor knows nothing of `path` or `name` when it calls `PropertiesPlugin.configureSubstitutor` and cannot find replacements.
* `pattern`: `%m%n` (unmodified)

This leaves us in an odd state, some data in the properties map is fully interpolated, while other data is incomplete. A reasonable user would expect the value of `filename` to be `/tmp/test.log`, but we're pretty far off.

This presents itself in 2.17.1, where  we limited runtime lookup recursion, because the escaped `$${filename}` component in our pattern is unescaped when the PluginBuilder renders the pattern value as `${filename} %m%n`, which results in `LiteralPatternConverter` being used to evaluate the runtime lookup for `${filename}` for each log line.
Given `log.info("Hello");` we get output `${path}${name} Hello` where the user would want `/tmp/test.log Hello`, which was the behavior prior to 2.17.1 because the resulting lookups where evaluated.

## How does this PR improve matters?

Given the same configuration as above, we handle properties differently from _all_ other lookups: Properties themselves can be evaluated recursively, while the results of other lookups cannot. Properties are defined in a way that's already aware of lookups, so this is safe, while environment variables, jvm system properties, etc are not designed with lookups in mind.

Given the example configuration above, we construct our lookups using the raw values with no substitution. This is important because any amount of substitution prior to the final evaluation will taint values such that we no longer know whether the value contains a lookup that we should evaluate, or referenced untrusted data which may or may not contain lookups.
* `path`: `/tmp/` (unmodified)
* `name`: `test.${lower:LOG}`  (unmodified)
* `filename`: `${path}${name}`  (unmodified)
* `pattern`: `%m%n` (unmodified)

Now, this data is also an unexpected state, but it cannot be observed because of changes to the way we're handling lookup recursion:
In 2.17.1 we eliminated runtime recursive evaluation, here we're eliminating configuration-time recursive evaluation as well, but allowing the results of property lookups to be recursively evaluated.

Given the `log.info("Hello");` example from earlier, we once again get output `/tmp/test.log Hello` because `${filename}` -> `${path}${name}` -> `/tmp/${name}` -> `/tmp/test.${lower:LOG}` -> `/tmp/test.log`

## What about Spring Boot `sys:NAME` lookups?

The [spring boot configuration](https://github.com/spring-projects/spring-boot/blob/8125b46ed5b904801153eab8394272888c09b611/spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/log4j2/log4j2.xml) uses properties which can be overridden by setting jvm system properties `java -DCONSOLE_LOG_PATTERN=<value>`

```xml
<Configuration status="WARN">
	<Properties>
		<Property name="LOG_EXCEPTION_CONVERSION_WORD">%xwEx</Property>
		<Property name="LOG_LEVEL_PATTERN">%5p</Property>
		<Property name="LOG_DATEFORMAT_PATTERN">yyyy-MM-dd HH:mm:ss.SSS</Property>
		<Property name="CONSOLE_LOG_PATTERN">%clr{%d{${sys:LOG_DATEFORMAT_PATTERN}}}{faint} %clr{${sys:LOG_LEVEL_PATTERN}} %clr{%pid}{magenta} %clr{---}{faint} %clr{[%15.15t]}{faint} %clr{%-40.40c{1.}}{cyan} %clr{:}{faint} %m%n${sys:LOG_EXCEPTION_CONVERSION_WORD}</Property>
		<Property name="FILE_LOG_PATTERN">%d{${LOG_DATEFORMAT_PATTERN}} ${LOG_LEVEL_PATTERN} %pid --- [%t] %-40.40c{1.} : %m%n${sys:LOG_EXCEPTION_CONVERSION_WORD}</Property>
	</Properties>
	<Appenders>
		<Console name="Console" target="SYSTEM_OUT" follow="true">
			<PatternLayout pattern="${sys:CONSOLE_LOG_PATTERN}" charset="${sys:CONSOLE_LOG_CHARSET}"/>
		</Console>
	</Appenders>
   <!-- loggers, etc -->
</Configuration>
```

These weren't impacted by the 2.17.1 changes, but they _could_ be impacted by this PR.
In the default configuration, there's no change because whether or not recursion is allowed on values depends on the source of the value, not the structure of the lookup (e.g. `Interpolator` failing to find a `sys` value for `CONSOLE_LOG_PATTERN` will fall back to the default `PropertiesLookup`). However, there's an oddity in which overriding a property with a jvm system property, with exactly the same value, will prevent nested lookups from being evaluated, when they would be in all versions up to 2.17.1. It would be fair to classify this as a bug, the behavior we provide isn't meant to recursively evaluate through system properties, environment variables, etc, but we cannot prove that no users rely on it.

## Other Risks and Caveats

Primary risk is that changes in edge-case behavior impact the entire configuration surface, and are not isolated to the more obscure runtime consumers.

For example, a property defined thusly (note the escaped lookup)
`<property name="foo">Date is $${date:YYYY-MM-dd}</property>`
When used via `${foo}` would result in `Date is ${date:YYYY-MM-dd}`
rather than `Date is 2022-01-27`, because initially when
the configuration is parsed, lookups aren't evaluated on properties,
which previously stripped one level of escapement.
The new behavior is easier to understand, but may result in some
friction for consumers.

**Temporal Risk**: rather than simple lookups executing immediately when the configuration is loaded, they may be re-evaluated later on, expanding through the property each time the property is read, where it's possible that the value may change in ways that were previously memoized by substitution within `<property/>` plugin object construction.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/PropertiesPlugin.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/PropertiesPlugin.java
@@ -16,22 +16,24 @@
  */
 package org.apache.logging.log4j.core.config;
 
-import java.util.HashMap;
-import java.util.Map;
-
+import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginConfiguration;
 import org.apache.logging.log4j.core.config.plugins.PluginElement;
 import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.lookup.Interpolator;
+import org.apache.logging.log4j.core.lookup.LookupResult;
 import org.apache.logging.log4j.core.lookup.PropertiesLookup;
 import org.apache.logging.log4j.core.lookup.StrLookup;
+import org.apache.logging.log4j.core.lookup.StrSubstitutor;
 
 /**
  * Handles properties defined in the configuration.
  */
 @Plugin(name = "properties", category = Node.CATEGORY, printObject = true)
 public final class PropertiesPlugin {
+
+    private static final StrSubstitutor UNESCAPING_SUBSTITUTOR = createUnescapingSubstitutor();
 
     private PropertiesPlugin() {
     }
@@ -45,15 +47,62 @@ public final class PropertiesPlugin {
     @PluginFactory
     public static StrLookup configureSubstitutor(@PluginElement("Properties") final Property[] properties,
                                                  @PluginConfiguration final Configuration config) {
-        if (properties == null) {
-            return new Interpolator(config.getProperties());
+        // For backwards compatibility, we unescape all escaped lookups when properties are parsed.
+        // This matches previous behavior for escaped components which were meant to be executed later on.
+        Property[] unescapedProperties = new Property[properties == null ? 0 : properties.length];
+        for (int i = 0; i < unescapedProperties.length; i++) {
+            unescapedProperties[i] = unescape(properties[i]);
         }
-        final Map<String, String> map = new HashMap<>(config.getProperties());
+        return new Interpolator(
+                new PropertiesLookup(unescapedProperties, config.getProperties()),
+                config.getPluginPackages());
+    }
 
-        for (final Property prop : properties) {
-            map.put(prop.getName(), prop.getValue());
+    private static Property unescape(Property input) {
+        return Property.createProperty(
+                input.getName(),
+                unescape(input.getRawValue()),
+                input.getValue());
+    }
+
+    // Visible for testing
+    static String unescape(String input) {
+        return UNESCAPING_SUBSTITUTOR.replace(input);
+    }
+
+    /**
+     * Creates a new {@link StrSubstitutor} which is configured with no lookups and does not handle
+     * defaults. This allows it to unescape one level of escaped lookups without any further processing
+     * or removing replacing {@code ${ctx:foo:-default}} with {@code default}.
+     */
+    private static StrSubstitutor createUnescapingSubstitutor() {
+        StrSubstitutor substitutor = new StrSubstitutor(NullLookup.INSTANCE);
+        substitutor.setValueDelimiter(null);
+        substitutor.setValueDelimiterMatcher(null);
+        return substitutor;
+    }
+
+    private enum NullLookup implements StrLookup {
+        INSTANCE;
+
+        @Override
+        public String lookup(String key) {
+            return null;
         }
 
-        return new Interpolator(new PropertiesLookup(map), config.getPluginPackages());
+        @Override
+        public String lookup(LogEvent event, String key) {
+            return null;
+        }
+
+        @Override
+        public LookupResult evaluate(String key) {
+            return null;
+        }
+
+        @Override
+        public LookupResult evaluate(LogEvent event, String key) {
+            return null;
+        }
     }
 }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/LookupResult.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/LookupResult.java
@@ -14,33 +14,22 @@
  * See the license for the specific language governing permissions and
  * limitations under the license.
  */
+
 package org.apache.logging.log4j.core.lookup;
 
-/**
- * A default lookup for others to extend.
- *
- * @since 2.1
- */
-public abstract class AbstractLookup implements StrLookup {
+public interface LookupResult {
+
+    /** Value of the lookup result. Never null. */
+    String value();
 
     /**
-     * Calls {@code lookup(null, key)} in the implementation.
-     *
-     * @see StrLookup#lookup(LogEvent, String)
+     * True if the {@link #value()} should be re-evaluated for other lookups.
+     * This is used by {@link PropertiesLookup} to allow properties to be evaluated against other properties,
+     * because the configuration properties are completely trusted and designed with lookups in mind. It is
+     * unsafe to return true in most cases because it may allow unintended lookups to evaluate other lookups.
      */
-    @Override
-    public String lookup(final String key) {
-        return lookup(null, key);
-    }
-
-    /**
-     * Calls {@code evaluate(null, key)} in the implementation.
-     *
-     * @see StrLookup#evaluate(LogEvent, String)
-     */
-    @Override
-    public LookupResult evaluate(final String key) {
-        return evaluate(null, key);
+    default boolean isLookupEvaluationAllowedInValue() {
+        return false;
     }
 
 }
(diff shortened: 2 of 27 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/PropertiesPlugin.java
```diff
@@ -45,15 +46,14 @@ private PropertiesPlugin() {
     @PluginFactory
     public static StrLookup configureSubstitutor(@PluginElement("Properties") final Property[] properties,
                                                  @PluginConfiguration final Configuration config) {
-        if (properties == null) {
-            return new Interpolator(config.getProperties());
-        }
+        final Property[] props = properties == null ? Property.EMPTY_ARRAY : properties;
         final Map<String, String> map = new HashMap<>(config.getProperties());
```
carterkozak, line 50: TODO + test: the configuration context-properties (`config.getProperties()`) are distinct from the configuration properties (`Property[] properties`), and should not allow recursive evaluation.
The PropertiesLookup should take both maps, and return values which don't support recursive evaluation for config context props.

log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/LookupResult.java
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * True if the result should be re-evaluated for other lookups.
+     * This is used by {@link PropertiesLookup} to allow properties to be evaluated against other properties,
+     * because the configuration properties are completely trusted and designed with lookups in mind. It is
+     * unsafe to return true in most cases because it may allow unintended lookups to evaluate other lookups.
+     */
+    default boolean allowRecursiveLookups() {
+        return false;
+    }
```
carterkozak, line 33: I don't think the name of this method is obvious enough. Perhaps `allowLookupEvaluationInValue`?

garydgregory:
Maybe this is the wrong place to ask but it feels to me we are going about this backward or I am missing some information. I would find it helpful to work from the documentation down to the code instead of the other way around. IOW, let's define first what it is the feature set is: What we can configure where and why (we don't want infinite recursion), what is the behavior, then implementation can follow. For example, where do we allow recursive lookups? When do they happen? How deep? And so on. I guess what I am looking for is some written agreement on what it is we want to provide to users before we release 2.17.2 and create more behavior differences and confusion since the behavior has already changed enough in 2.17.1 to break existing user configurations. With 2.17.2 (or 2.18.0), we are introducing a 3rd behavior set, so let's try and be done with it ;-)

carterkozak:
This experiment came out of a discussion with Ralph, and deserves more context in this PR. This is still marked as a draft, and needs a better description of intent and potential impact.

> I would find it helpful to work from the documentation down to the code instead of the other way around

This is tricky. The documentation describes but a sliver of the full breadth of functionality, much of the existing functionality I think is defacto api because it's already used, regardless of whether it's known or understood by any of us.

For what it's worth, this does introduce a behavior change, and that worries me. I think the result of that change puts us in a much better place, but a design along these lines is safer in terms of potential impact to existing configurations: #722

carterkozak:
I have updated the description with several more words :-)

carterkozak:
I’m having a bit of an internal struggle with this — in its current state, it changes the behavior of `$${ctx:foo}` used in properties because today we build properties using substitution, such that the escapement is dropped when the property is built: `${ctx:foo}`, then when the property is referenced, recursive evaluation will then evaluate `${ctx:foo}` (assuming recursive evaluation is allowed)
My change modifies the behavior so that substitution is allowed through properties recursively, but no other lookups, so when the Property is initially constructed, it retains the `$${ctx:foo}` value, and when it’s referenced, the value is expanded to `${ctx:foo}`.
I am convinced the new behavior is better, and easier to reason about. However, changes impacting config are bad. That said, this fixes an issue in which lookup-shaped data in sources used by the configuration are unexpectedly evaluated (e.g. `env:FOO` where FOO includes something like `${java:version}`)

We can play some dirty tricks to minimize impact, however that sort of thing will more than likely come back to haunt us. I can do a `value.replace("$${", "${")` on the raw value to simulate the old behavior, such that old configs continue to work, however there are edge cases in which literal `$${` won’t work as one might expect.

carterkozak:
I've added a bit of code to unescape escaped lookups, which will allow existing properties to work as expected. Rather than `value.replace("$${", "${")`, I'm using a `StrSubstitutor` with a lookup that always returns `null` and doesn't understand defaults, this way original values are retained, but escaped components are unescaped so that there's no regression from 2.17.0 and earlier.

We had discussed the idea of making such a feature opt-in, however I feel pretty strongly that it should be on-by-default, or shouldn't exist at all. Reason being: In cases where it's needed, it would require a similar level of effort to unescape lookups in logging configurations as to turn on the feature.

At this point, the question is:
Are more concerned with easy upgrades that don't break functionality, or providing the cleanest, most straightforward design at the cost of requiring some configuration changes?

What do you think?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/PropertiesPlugin.java
```diff
@@ -45,15 +47,62 @@ private PropertiesPlugin() {
     @PluginFactory
     public static StrLookup configureSubstitutor(@PluginElement("Properties") final Property[] properties,
                                                  @PluginConfiguration final Configuration config) {
-        if (properties == null) {
-            return new Interpolator(config.getProperties());
+        // For backwards compatibility, we unescape all escaped lookups when properties are parsed.
+        // This matches previous behavior for escaped components which were meant to be executed later on.
+        Property[] escapedProperties = new Property[properties == null ? 0 : properties.length];
```
carterkozak, line 52:
```suggestion
        Property[] unescapedProperties = new Property[properties == null ? 0 : properties.length];
```

PR #733, opened 2022-01-28
[LOG4J2-3341] Support logger level and appender refs shorthand

jvz:
This adds a properties format feature where the property names "rootLogger" and "logger.<key>" can be set to a level and one or more appender refs.

https://issues.apache.org/jira/browse/LOG4J2-3341

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
@@ -17,11 +17,8 @@
 
 package org.apache.logging.log4j.core.config.properties;
 
-import java.util.Map;
-import java.util.Properties;
-import java.util.concurrent.TimeUnit;
-
 import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.ConfigurationException;
@@ -42,8 +39,14 @@ import org.apache.logging.log4j.core.config.builder.api.ScriptComponentBuilder;
 import org.apache.logging.log4j.core.config.builder.api.ScriptFileComponentBuilder;
 import org.apache.logging.log4j.core.filter.AbstractFilter.AbstractFilterBuilder;
 import org.apache.logging.log4j.core.util.Builder;
+import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.util.PropertiesUtil;
 import org.apache.logging.log4j.util.Strings;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Helper builder for parsing properties files into a PropertiesConfiguration.
@@ -53,6 +56,7 @@ import org.apache.logging.log4j.util.Strings;
 public class PropertiesConfigurationBuilder extends ConfigurationBuilderFactory
     implements Builder<PropertiesConfiguration> {
 
+    private static final Logger LOGGER = StatusLogger.getLogger();
     private static final String ADVERTISER_KEY = "advertiser";
     private static final String STATUS_KEY = "status";
     private static final String SHUTDOWN_HOOK = "shutdownHook";
@@ -170,17 +174,20 @@ public class PropertiesConfigurationBuilder extends ConfigurationBuilderFactory
                 }
             }
         } else {
-            final Map<String, Properties> loggers = PropertiesUtil
-                    .partitionOnCommonPrefixes(PropertiesUtil.extractSubset(rootProperties, "logger"));
+            final Properties context = PropertiesUtil.extractSubset(rootProperties, "logger");
+            final Map<String, Properties> loggers = PropertiesUtil.partitionOnCommonPrefixes(context);
             for (final Map.Entry<String, Properties> entry : loggers.entrySet()) {
                 final String name = entry.getKey().trim();
                 if (!name.equals(LoggerConfig.ROOT)) {
-                    builder.add(createLogger(name, entry.getValue()));
+                    final Properties loggerProps = entry.getValue();
+                    useSyntheticLevelAndAppenderRefs(name, loggerProps, context);
+                    builder.add(createLogger(name, loggerProps));
                 }
             }
         }
 
         final Properties props = PropertiesUtil.extractSubset(rootProperties, "rootLogger");
+        useSyntheticLevelAndAppenderRefs("rootLogger", props, rootProperties);
         if (props.size() > 0) {
             builder.add(createRootLogger(props));
         }
@@ -188,6 +195,23 @@ public class PropertiesConfigurationBuilder extends ConfigurationBuilderFactory
         builder.setLoggerContext(loggerContext);
 
         return builder.build(false);
+    }
+
+    private static void useSyntheticLevelAndAppenderRefs(final String propertyName, final Properties loggerProps, final Properties context) {
+        final String propertyValue = (String) context.remove(propertyName);
+        if (propertyValue != null) {
+            final String[] parts = propertyValue.split(",");
+            if (parts.length > 0) {
+                final String level = parts[0].trim();
+                loggerProps.setProperty("level", level);
+                LOGGER.debug("Using log level '{}' for logger '{}'", level, propertyName);
+                for (int i = 1; i < parts.length; i++) {
+                    final String appenderRef = parts[i].trim();
+                    LOGGER.debug("Adding synthetic appender ref '{}' for logger '{}'", appenderRef, propertyName);
+                    loggerProps.setProperty("appenderRef.$" + i + ".ref", appenderRef);
+                }
+            }
+        }
     }
 
     private ScriptComponentBuilder createScript(final Properties properties) {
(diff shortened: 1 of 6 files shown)


log4j-core/src/test/resources/LoggerLevelAppenderTest.properties
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the license for the specific language governing permissions and
+# limitations under the license.
+#
+
+status = all
+name = LoggerLevelAppenderTest
+logger.0.name = org.apache.logging.log4j.core.config.properties
+logger.0 = INFO,first,second
```
Apache9, line 21: Good. I think this is enough for hadoop and hbase, although it maybe a pain that we need to maintain the index. It will be easier to specify appender.first.type = List. But anyway, it works.

Another question is how to specify leven and appenders at once in xml configuration file?

jvz: The indexed bit is how the current properties format works. I think a separate feature could be added here to automatically set the `name` attribute to the property key when otherwise absent.

rgoers: Matt and I only worked out how to do it for properties as that was what was requested. Doing it for XML could be possible, but it would be more difficult in YAML and JSON. I'm not even sure what the point of that would be.

For XML the syntax would have to look like

<Logger name="org.apache.logging.log4j.core.config.properties" level="INFO" appenderRefs="first,second"/>

That is actually not "normal" for XML to have a list specified that way. And there would definitely be no reason to combine the level and appenderRefs into a single attribute.
Also, looking at the Log4j1 DTD the xml had to be specified as 

```
  <root>
    <priority value ="trace" />
    <appender-ref ref="A1" />
    <appender-ref ref="A2" />
  </root>
```

Note that the priority (which could also be specified as level) had to be an XML element whereas Log4j 2 allows it to be an element or an attribute. 
So the bottom line is, I don't know why we would add the support to XML since it doesn't buy anything in the way of compatibility with Log4j 1.

jvz: I've added a way to do implicit appender names, too. As for the XML config variant, that would need to add the CompoundProperty plugin.

log4j-core/src/test/resources/RootLoggerLevelAppenderTest.properties
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the license for the specific language governing permissions and
+# limitations under the license.
+#
+
+status = off
+name = RootLoggerLevelAppenderTest
+rootLogger = INFO,app
+appender.app.type = List
```
jvz, line 21: Now this properties format is getting less shitty to use, too! I like this.

rgoers (approved):

ppkarwasz:
@jvz: maybe you can add a default for all component names: `name = key.replaceAll("-", ".")`? This way you'll allow also to configure a logger in one property:

`logger.org-apache-logging-log4j=INFO,app1,app2`

log4j-core/src/test/resources/LoggerLevelAppenderTest.properties
Apache9: OK, got it. So this is a properties file specific way.

Apache9 (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -207,7 +221,12 @@ private ScriptFileComponentBuilder createScriptFile(final Properties properties)
     }
 
     private AppenderComponentBuilder createAppender(final String key, final Properties properties) {
-        final String name = (String) properties.remove(CONFIG_NAME);
+        final String name;
+        if (properties.containsKey(CONFIG_NAME)) {
+            name = (String) properties.remove(CONFIG_NAME);
+        } else {
+            name = key;
+        }
```
carterkozak, line 229: I think we may be slightly better off retaining the original `remove` logic, only traversing the map once:
```suggestion
        final String name = Objects.toString(properties.remove(CONFIG_NAME), key);
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -190,6 +193,17 @@ public PropertiesConfiguration build() {
         return builder.build(false);
     }
 
+    private static void useSyntheticLevelAndAppenderRefs(final String propertyName, final Properties loggerProps, final Properties context) {
+        final String propertyValue = (String) context.remove(propertyName);
+        if (propertyValue != null) {
+            final String[] parts = propertyValue.split(",");
+            loggerProps.setProperty("level", parts[0]);
```
carterkozak, line 200: We should check that `parts` is not empty to avoid an `ArrayIndexOutOfBoundsException`. If memory serves, `",".split(",")` will result in an empty array.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -190,6 +193,17 @@ public PropertiesConfiguration build() {
...
+            for (int i = 1; i < parts.length; i++) {
+                loggerProps.setProperty("appenderRef.$" + i + ".ref", parts[i]);
```
carterkozak, line 202: It may be worthwhile to add some StatusLogger debug logging here, I defer to your judgement :-)

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -190,6 +193,17 @@ public PropertiesConfiguration build() {
         return builder.build(false);
     }
 
+    private static void useSyntheticLevelAndAppenderRefs(final String propertyName, final Properties loggerProps, final Properties context) {
+        final String propertyValue = (String) context.remove(propertyName);
+        if (propertyValue != null) {
+            final String[] parts = propertyValue.split(",");
```
carterkozak, line 199: Do we need to trim each segment of `parts` to avoid whitespace issues? I suspect the level piece may handle it gracefully, but I'm less sure about the appenderRefs. Could be worth a test.

carterkozak (commented):
Very nifty!

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
jvz: I think that split would result in a two value array of two empty strings.

jvz:
> @jvz: maybe you can add a default for all component names: `name = key.replaceAll("-", ".")`? This way you'll allow also to configure a logger in one property:
> 
> `logger.org-apache-logging-log4j=INFO,app1,app2`

This is an interesting idea. The only potential issue here is that dashes are technically allowed in logger names, though I doubt they're used. Could probably be added with this whole update.

I'll have docs updates to go with this before it's ready to merge.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
jvz: Yes, that's a good point.

- [ ] Trim appender refs

log4j-core/src/test/resources/RootLoggerLevelAppenderTest.properties
jvz: Updated.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
jvz: Good idea; added.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
carterkozak: It's subtle and counter-intuitive. Alas, this test passes:
```java
@Test
public void split() {
    assertThat(",".split(",")).isEmpty();
}
```

carterkozak (approved):
Looks good to me. Thanks, Matt!

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
jvz: Indeed it is! I only remembered that because I checked the javadocs for `split` before using it here. At first, I was limiting the split to 1 match, but then I added the multiple refs thing.

jvz:
I've squashed what's been reviewed. Will add more commits for docs and changelog.

ppkarwasz:
@jvz: #729 drops previously invalid entries like `logger.core-config-properties=INFO,app1,app2`. You should probably add a test for that.

jvz:
I have some code in PropertiesConfigurationBuilder that manually includes the invalid keys left behind.

PR #735, opened 2022-01-29
Adds a shorthand method to specify components names

ppkarwasz:
As an extension of #733, this PR allows to specify the name of any component in its key. Dots in components names are escaped using dashes `-`. Since this is not the only way to specify a components name, there is **no** shorthand for names with dashes.

After this change, only one property is needed to specify a logger, e.g.:

`logger.org-apache-logging-log4j=DEBUG, Console`

Some changes were necessary in two helper functions in `PropertiesUtil`:

 - `extractSubset` didn't extract the property **equal** to the given prefix, although the Javadoc says it should,
 - `partitionOnCommonPrefixes` ignored properties that didn't contain a dot (or threw an exception before #729 ). Now it detects also dotless properties.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -524,8 +524,8 @@ public final class PropertiesUtil {
         final List<String> keys = new ArrayList<>();
 
         for (final String key : properties.stringPropertyNames()) {
-            if (key.startsWith(prefixToMatch)) {
-                subset.setProperty(key.substring(prefixToMatch.length()), properties.getProperty(key));
+            if (key.startsWith(prefixToMatch) || key.equals(prefix)) {
+                subset.setProperty(key.substring(Math.min(key.length(), prefixToMatch.length())), properties.getProperty(key));
                 keys.add(key);
             }
         }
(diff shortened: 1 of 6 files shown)


garydgregory:
Do we really want to open the door here? What about underscore? Any valid Java name separators? What about mixing separators?

carterkozak:
I agree with Gary. I’m concerned about the user experience, when an operator sees logger.org-apache-foo it’s not going to be obvious that is equivalent to org.apache.foo.
Relatedly, it’s not clear when hyphens are meant to replace dots as opposed to referencing hyphens.

ppkarwasz:
What about a more common way of escaping the dots, like `\.`? Or maybe quoting/double-quoting the logger name? Although it might not be intuitive why the part after `logger.` must be quoted.

What I mostly dislike in the properties format is its verbosity. Unfortunately users apparently like it (judging by the amount of `log4j.properties -> log4j2.properties` questions on StackOverflow) and the format might benefit from fixing those component prefixes to something meaningful.

carterkozak:
I could see something like this working:
`logger.org.apache.foo.level = INFO, console`

If the property name begins with a known value `logger.` and ends with `.level`, so there's no ambiguity about which components are part of the name. This avoids the `logger.X.name = org.apache.foo`. The `level = <level>, <appender...>` component may be a bit odd, I think it only creates ambiguity if level names can include commas, I haven't checked if that's allowed, but should be unlikely.

ppkarwasz:
@carterkozak: this could work, but maybe let's use another suffix for the condensed property. Is `.config` Ok?

Basically it could work like this:
 - the implicit name of an appender is from `appender.` to the **first** dot. The same could apply to filters and internal components,
 - the implicit name of a property is the part following `property.` (this is already true),
 - the implicit name of a logger is from `logger`. to the **last** dot. We add `.config` to Matt's shorthand notation (#733)

@jvz : what do you think?

An example config file:

```
property.log.file.name=myapp
appender.ROLLING.type=RollingFile
appender.ROLLING.fileName=logs/${log.file.name}.log
appender.ROLLING.filePattern=logs/${log.file.name}.%i.log
appender.ROLLING.SIZE=SizeBasedTriggeringPolicy
logger.org.apache.logging.log4j.config=INFO, ROLLING
```

carterkozak:
It sounds like we're discussing improving several different parts of the properties configuration scheme at once -- It's great to have an idea of where we want to go before we get started, but I think it would also be helpful to decomp the pieces and approach them separately.

1. Support logger configuration without a `.name = <NAME>` using `logger.<NAME>.<suffix> = <value>` e.g. `logger.org.apache.foo.level = INFO`
2. Support setting logger level and appenderRef at the same time, perhaps with a new suffix e.g. `logger.myAppender.config = INFO, console`
3. Similar improvements to other components (appenders/filters)

I think **1** is the most complex, and we'll have to think a bit about interactions with components that can be nested within the logger config. Would we need to ban nested component references when this is used? I suppose it depends how the parser works. It is a substantial departure from all other node-based configurations, and may be difficult to support as features are added (since propertiesconfiguration may need to be aware of logger suffixes).

**2** I don't have a strong opinion between `.level` and `.config` suffixes, the initial `level` suggestion might be easier for users to understand without spending a great deal of time reading the documentation. The `.config` shorthand suffix would be clearer though.

I'd like to hear what other folks think.

PR #737, opened 2022-01-30
Adds support for missing Log4j 1.x properties and fixes default values

ppkarwasz:
The properties of the generated appenders are not always the same as in Log4j 1.x. This PR:

 - adds default values to all Log4j 1.x configuration builders,
 - adds a couple of properties that were missing in the builders,
 - expands the test suite to include the case of all properties unspecified and all properties set to a value different from the default.

Problems with default values cause issues as in [this SO question](https://stackoverflow.com/q/70895631/11748454).


vy:
@garydgregory, would you be interested in taking this on as a part of your ongoing `log4j-1.2-bridge` crusade?

garydgregory:
@vy I might be able to take a look this weekend.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/ConsoleAppenderBuilder.java
```diff
@@ -107,6 +110,24 @@ public Appender parseAppender(final Element appenderElement, final XmlConfigurat
...
+                            }
+                            break;
+                        }
+                        case IMMEDIATE_FLUSH_PARAM: {
+                            String value = getValueAttribute(currentElement);
+                            if (value == null) {
+                                LOGGER.warn("No value supplied for ImmediateFlush parameter. Using default of {}", true);
+                            } else if (!Boolean.getBoolean(name)) {
+                                LOGGER.warn("The value {} for ImmediateFlush parameter is not supported.", false);
+                            }
```
garydgregory, line 128: FYI : This seems incomplete, where is the value used?

ppkarwasz: Since the Log4j2 `ConsoleAppender` does not use the `immediateFlush` parameter (cf. [`ConsoleAppender.Builder#build`](https://github.com/apache/logging-log4j2/blob/0de646bde36d2c426988970314428d68b8858a5a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/ConsoleAppender.java#L218)), I only issue a warning that the value supplied by the user will be ignored.
Alternatively the Log4j2 `ConsoleAppender` could issue a warning if `immediateValue`, `bufferedIO` or `bufferSize` differ from the default values.

PR #601, opened 2021-11-22
LOG4J2-3193 Configuration precedence fix

david-burke-securian:
For https://issues.apache.org/jira/browse/LOG4J2-3193

I can't get the project to build locally, even with a clean copy of `master`. I will work on that (could be user error--I have more experience with Gradle than Maven) and I also intend to add unit tests for my change.

PR #742, opened 2022-02-06
LOG4J2-3366 Fixes order of property sources

ppkarwasz:
This PR fixes the order of the property sources to:

1. Spring Boot properties with lowest numerical priority (overrides all others). In a Spring Boot application with `log4j-spring` present it will use in order: the servlet context init parameters, JNDI environment entries (from `java:comp/env` exclusively), Java system properties and system environment variables (among others, cf. [the complete list](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config)).
2. Java system properties,
3. OS environment variables,
4. `log4j2.component.properties` to allow devs to provide some defaults.

Due to some issues with the previous implementation, the logic of the `Environment` class has been rewritten:

- the `normalized` map maps **any** key to the value of its normalization in the highest priority property source. E.g.: the key `log4j.configurationFile` will contain the value of the `log4j2.configurationFile` property (remark the `2`) in Java's system properties or the value of the `LOG4J_CONFIGURATION_FILE` environment variable. Previously it contained two separate mappings: `log4j2.configurationFile` -> value in Java's system properties, `LOG4J_CONFIGURATION_FILE` -> value of the environment variable. This caused a miss for 95% of the `PropertiesUtil#getStringProperty` calls, which usually use the legacy property name.
- the three caches (`normalized`, `literal` and `tokenized`) are filled in two steps: first we recover all the keys from **enumerable** sources and then we query **all** sources to retrieve the appropriate values. Previously non-enumerable sources didn't get a chance to provide a value in these caches (the Spring Boot property source is non-enumerable).

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -25,8 +25,10 @@ import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.ResourceBundle;
 import java.util.ServiceLoader;
@@ -61,7 +63,7 @@ public final class PropertiesUtil {
      * @param props the Properties to use by default
      */
     public PropertiesUtil(final Properties props) {
-        this.environment = new Environment(new PropertiesPropertySource(props));
+        this(new PropertiesPropertySource(props));
     }
 
     /**
@@ -75,7 +77,15 @@ public final class PropertiesUtil {
     }
 
     private PropertiesUtil(final String propertiesFileName, final boolean useTccl) {
-        this.environment = new Environment(new PropertyFilePropertySource(propertiesFileName, useTccl));
+        this(new PropertyFilePropertySource(propertiesFileName, useTccl));
+    }
+
+    /**
+     * Constructs a PropertiesUtil for a give property source as source of additional properties.
+     * @param source a property source
+     */
+    PropertiesUtil(final PropertySource source) {
+        this.environment = new Environment(source);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


ppkarwasz:
This PR should also solve the problem of PR #601.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -435,17 +452,18 @@ private Environment(final PropertySource propertySource) {
...
-					for (final PropertySource source : ServiceLoader.load(PropertySource.class, classLoader)) {
-						sources.add(source);
-					}
-				} catch (final Throwable ex) {
-					/* Don't log anything to the console. It may not be a problem that a PropertySource
-					 * isn't accessible.
-					 */
-				}
-			}
+            for (final ClassLoader classLoader : LoaderUtil.getClassLoaders()) {
```
ppkarwasz, line 455: This was indented with tabs.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -454,28 +472,36 @@ private synchronized void reload() {
...
                 });
-            }
-        }
-
-        private static boolean hasSystemProperty(final String key) {
-            try {
-                return System.getProperties().containsKey(key);
-            } catch (final SecurityException ignored) {
-                return false;
-            }
```
ppkarwasz, line 478: System properties are resolved by the `SystemPropertiesPropertySource`.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
```diff
@@ -34,7 +38,7 @@
 
     /**
      * Returns the order in which this PropertySource has priority. A higher value means that the source will be
-     * applied later so as to take precedence over other property sources.
+     * searched later and can be overridden by other property sources.
```
ppkarwasz, line 41: The documentation actually says the opposite: sources with lower numerical priority override the others.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
garydgregory: Never tabs, only spaces in this project; -)

Kay-kay2019: Okay, thank you so much.

ppkarwasz: @garydgregory, would you mind replacing tabs with spaces in `log4j-core`? IIRC you already did it for `log4j-1.2-api`.

adwsingh:
This change along with https://github.com/apache/logging-log4j2/pull/975 broke one of our systems on upgrading to 2.19.

In our applications we had both `LOG4J_CONFIGURATION_FILE` environment variable as well as `log4j.configurationFile` system property set. 

In version 2.17.2 `log4j.configurationFile` gets priority vs in 2.19 `LOG4J_CONFIGURATION_FILE` gets priority. This also breaks the contract mentioned in the [documentation](https://logging.apache.org/log4j/2.x/manual/configuration.html#SystemProperties)

ppkarwasz:
Hi @adwsingh,
Old pre-2.10 properties such as `log4j.configurationFile` are overridden by any property in normalized form (e.g. the `log4j2.configurationFile` Java system property or the `LOG4J_CONFIGURATION_FILE` environment variable). Changing from `log4j.configurationFile` to `log4j2.configurationFile` should work for you.

If you believe this to be a bug, please submit an issue on [JIRA](https://issues.apache.org/jira).

adwsingh:
I have created a bug here, https://issues.apache.org/jira/browse/LOG4J2-3621

PR #745, opened 2022-02-08
[LOG4J2-3362] Adds a SmtpManager compatible with Jakarta EE 9

ppkarwasz:
Since `javax.mail` and `jakarta.mail` use different package names, as single `SmtpAppender` can use both through different managers.

Unfortunately the implementations `com.sun.mail:javax.mail` and `com.sun.mail:jakarta.mail` use the same class names, so it is not
possible to test both implementations at the same time. A separate `test` execution was configured for the tests using `jakarta.*` classes.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SmtpAppender.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SmtpAppender.java
@@ -18,6 +18,7 @@
 package org.apache.logging.log4j.core.appender;
 
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 
 import org.apache.logging.log4j.core.Appender;
 import org.apache.logging.log4j.core.Core;
@@ -36,11 +37,18 @@ import org.apache.logging.log4j.core.config.plugins.PluginElement;
 import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
 import org.apache.logging.log4j.core.config.plugins.validation.constraints.ValidPort;
 import org.apache.logging.log4j.core.filter.ThresholdFilter;
+import org.apache.logging.log4j.core.layout.AbstractStringLayout.Serializer;
 import org.apache.logging.log4j.core.layout.HtmlLayout;
+import org.apache.logging.log4j.core.layout.PatternLayout;
+import org.apache.logging.log4j.core.net.MailManager;
+import org.apache.logging.log4j.core.net.MailManager.FactoryData;
+import org.apache.logging.log4j.core.net.MailManagerFactory;
 import org.apache.logging.log4j.core.net.SmtpManager;
 import org.apache.logging.log4j.core.net.ssl.SslConfiguration;
 import org.apache.logging.log4j.core.util.Booleans;
 import org.apache.logging.log4j.core.util.Integers;
+import org.apache.logging.log4j.util.ServiceLoaderUtil;
+import org.apache.logging.log4j.util.Strings;
 
 /**
  * Send an e-mail when a specific logging event occurs, typically on errors or
@@ -66,12 +74,16 @@ public final class SmtpAppender extends AbstractAppender {
     private static final int DEFAULT_BUFFER_SIZE = 512;
 
     /** The SMTP Manager */
-    private final SmtpManager manager;
+    private final MailManager manager;
 
     private SmtpAppender(final String name, final Filter filter, final Layout<? extends Serializable> layout,
-            final SmtpManager manager, final boolean ignoreExceptions, final Property[] properties) {
+            final MailManager manager, final boolean ignoreExceptions, final Property[] properties) {
         super(name, filter, layout, ignoreExceptions, properties);
         this.manager = manager;
+    }
+
+    public MailManager getManager() {
+        return manager;
     }
 
     /**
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -687,10 +691,43 @@
...
+        <artifactId>jakarta.activation-api</artifactId>
+        <version>${jakarta.activation.version}</version>
+        <optional>true</optional>
+      </dependency>
+      <dependency>
+        <groupId>com.sun.mail</groupId>
+        <artifactId>jakarta.mail</artifactId>
+        <version>${jakarta.mail.version}</version>
+        <scope>runtime</scope>
+        <optional>true</optional>
```
ppkarwasz, line 730: Same philosophy as `logback-classic`: use APIs as (optional) compile dependencies, while the implementation is an (optional) runtime dependency.

log4j-core/src/test/java/org/apache/logging/log4j/core/net/SmtpManagerTest.java
```diff
@@ -39,14 +39,15 @@
 public class SmtpManagerTest {
 
     @Test
-    void testCreateManagerName() {
-        String managerName = SmtpManager.createManagerName("to", "cc", null, "from", null, "LOG4J2-3107",
-                "proto", "smtp.log4j.com", 4711, "username", false, "filter");
+    public void testCreateManagerName() {
```
ppkarwasz, line 42: These tests were disabled (non-public).

garydgregory:
IMO, anything new that depends on other dependencies should be in a new Maven module.

ppkarwasz:
@garydgregory: I think of this as a nice application of the separation between appenders and managers, which mitigates the problems of the jakartification.

However I see you point: the `SmtpAppender` and `JmsAppender` probably should be moved out of `log4j-core` and a jakartified version should be provided separately.

jvz:
> However I see you point: the `SmtpAppender` and `JmsAppender` probably should be moved out of `log4j-core` and a jakartified version should be provided separately.

Already prepared as separate modules in the 3.0.0 branch: https://github.com/apache/logging-log4j2/tree/master

However, those two modules (log4j-jms and log4j-smtp) would still need corresponding jakarta variants.

ppkarwasz:
@jvz: is there any sense in using `javax.*` in the 3.0 release? Spring Boot 6.0 with a baseline of Java 17 and Jakarta EE 9 is due at the end of this year.

jvz:
I think as long as the maintenance burden is low enough, we can keep support for the old APIs indefinitely. Our Java baseline there is 11, so not as aggressive as Spring's baseline. That is a good point, though.

ppkarwasz:
@garydgregory: I moved the `SmtpManager` to a new `log4j-jakarta-smtp` module.

The choice between the `javax.*` and `jakarta.*` version is automatic, based on the classpath, but can be forced with a new `useJakarta` attribute.

log4j-core/src/test/java/org/apache/logging/log4j/core/net/SmtpManagerTest.java
andrei-ivanov: JUnit Jupiter doesn't need tests to be public, I think it's actually recommended to use the default visibility

astappiev:
Any progress on this or release date?

ppkarwasz:
@astappiev,
As far as I am concerned, this could end up in the next release. You can check if it works for you.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SmtpAppender.java
```diff
@@ -261,9 +277,21 @@ public SmtpAppender build() {
...
+            final Serializer subjectSerializer = PatternLayout.newSerializerBuilder()
+                    .setConfiguration(getConfiguration())
+                    .setPattern(subject)
+                    .build();
+            final FactoryData data = new FactoryData(to, cc, bcc, from, replyTo, subject, subjectSerializer,
+                    smtpProtocol, smtpHost, smtpPort, smtpUsername, smtpPassword, smtpDebug, bufferSize,
+                    sslConfiguration, getFilter().toString());
+            final MailManager smtpManager = AbstractManager.getManager(data.getManagerName(), FACTORY, data);
+            if (smtpManager == null) {
+                LOGGER.error("Unabled to instantiate SmtpAppender named {}", getName());
```
jvz, line 292: Should `return null` here as well to avoid further invalid state when constructing the appender below.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/MailManager.java
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public abstract void add(LogEvent event);
+
+    /**
+     * Send the contents of the cyclic buffer as an e-mail message.
+     * 
+     * @param layout      The layout for formatting the events.
+     * @param appendEvent The event that triggered the send.
+     */
+    public void sendEvents(final Layout<?> layout, final LogEvent appendEvent) {
```
jvz, line 210: Shouldn't this also be `abstract`?

jvz (commented):
General approach looks good to me. I especially like that this didn't go the Spring route of encapsulating every possible email-related API we might have been using and instead simply uses plugins as plugins. :)

romainmoreau:
Is it done on purpose not adding `org.apache.logging.log4j:log4j-jakarta-smtp` to `org.apache.logging.log4j:log4j-bom`? For instance, in a Spring Boot project which uses the BOM, we have to set the version by hand.

ppkarwasz:
@romainmoreau: I just forgot to add it, thanks for noticing.

PR #751, opened 2022-02-11
check method.getName() is not null

gomezabajo:


garydgregory:
Needs a test, otherwise the behavior can be reverted. When can this happen? See https://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Method.html#getName--

PR #753, opened 2022-02-14
Issues with multiple Log4j 1.x filters

ppkarwasz:
This PR fixes a couple of issues concerning filters in the Log4j 1.x bridge:

 - there is an endless loop in `FilterAdapter#filter` that hangs the program if more than one filter per appender is configured,
 - filters are executed multiple times per log message.

The second problem comes from chains of Log4j 1.x filters: if we have three legacy filters chained `filter1 -> filter2 -> filter3` (`->` denotes the _filter1.getNext() == filter2_ relation), we end up with `CompositeFilter[FilterAdapter[filter1 -> filter2 -> filter3], FilterAdapter[filter2 -> filter3], FilterAdapter[filter3]`, which executes `filter3` three times.

In order to preserve the behavior of `FilterAdapter` the following restrictions have been applied to the generated filters:

 1. no chains of Log4j 1.x filters are generated by the configuration factories; a list of Log4j 1.x filters is represented as a wrapped `CompositeFilter` of `FilterAdapter`s (or native Log4j 2.x filters),
 2. the factories don't generate any `FilterWrapper`s of `FilterAdapter`s nor `FilterAdapter`s of `FilterWrapper`s.


log4j-1.2-api/src/main/java/org/apache/log4j/bridge/FilterAdapter.java
```diff
@@ -37,14 +81,14 @@ public Result filter(LogEvent event) {
         LoggingEvent loggingEvent = new LogEventAdapter(event);
         Filter next = filter;
         while (next != null) {
-            switch (filter.decide(loggingEvent)) {
+            switch (next.decide(loggingEvent)) {
```
ppkarwasz, line 84: Infinite loop.

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/CompositeFilter.java
```diff
@@ -62,9 +62,9 @@ public CompositeFilter addFilter(final Filter filter) {
         if (filter instanceof CompositeFilter) {
             final int size = this.filters.length + ((CompositeFilter) filter).size();
             final Filter[] copy = Arrays.copyOf(this.filters, size);
-            final int index = this.filters.length;
+            int index = this.filters.length;
             for (final Filter currentFilter : ((CompositeFilter) filter).filters) {
-                copy[index] = currentFilter;
+                copy[index++] = currentFilter;
```
ppkarwasz, line 67: Small bug fix in `log4j-core`. Probably does not merit its own PR.

ppkarwasz:
Regarding the inclusion in version 2.17.2 (cf. [this thread](https://lists.apache.org/thread/fk538mcvyf6n8wsbxpfjrlnj9kk71hj8)), the two commented snippets of code are easy to verify and a must have.

The rest of the PR is only a performance optimization: in the worst case scenario of n chained legacy filters (each returning `NEUTRAL`), O(n^2) filters are executed per log message. In the more common scenario of n native Log4j 2.x filters, only n filters are evaluated (they "lose" the chain when the `AppenderWrapper` is unwrapped). In any case the evaluation of the filters gives always the correct result.

If a deadline is approaching, I wouldn't rush to include the whole PR.

PR #755, opened 2022-02-15
trim MaxBackupIndex if last char is blank

marcushoja:
using log4j1.2-api-2.x bridge with log4j.properties a blank after MaxBackupIndex=<value> stops creating the appender and logging is disabled


marcushoja:
an invisible NumberFormatException is thrown

garydgregory:
Just a second... would it not be better to look for all call sites and refactor this kind of convertion into an internal utility?

ppkarwasz:
@vy: shouldn't `max` be of type `Integer`, so that the injection framework takes care of the conversion?

This would break the API, so can probably only be done in the `master` branch.

vy:
@ppkarwasz Good point. I have tried to see if it is something quick to fix. I have started with `min`, `max`, and `compressionLevel` in `DefaultRolloverStrategy`. Then it spread to `RollingRandomAccessFileAppender`. And then to `DirectWriteRolloverStrategy`. Hence, it is not a quick fix. I will return my back to my own TODO list. Though be my guest to submit a PR.

PR #756, opened 2022-02-16
[LOG4J2-3404] Creates default layouts using the available Configuration

ppkarwasz:
When a default `PatternLayout` is created without a `Configuration` a new `DefaultConfiguration` is created and abandoned. This causes a leak of `ConsoleManager`s.

This PR provides a `Configuration` argument whenever possible.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/PatternLayout.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/PatternLayout.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import org.apache.logging.log4j.core.Layout;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.Configuration;
-import org.apache.logging.log4j.core.config.DefaultConfiguration;
 import org.apache.logging.log4j.core.config.Node;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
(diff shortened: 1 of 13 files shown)


carterkozak (commented):
Change looks good!

I think that `new DefaultConfiguration()` in `PatternLayout.Builder.build()` is a bug, perhaps the layout itself should handle a null configuration gracefully? Not necessarily a requirement for this PR as we're ensuring we're passing the correct reference everywhere, but I think the `new DefaultConfiguration()` that doesn't come from a `ConfigurationFactory` has high likelihood of causing similar problems again in the future.

ppkarwasz:
@carterkozak: I'll check if we can leave the `Configuration` as `null` and update the PR.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/PatternLayout.java
```diff
@@ -763,10 +762,7 @@ public Builder withFooter(final String footer) {
 
         @Override
         public PatternLayout build() {
-            // fall back to DefaultConfiguration
-            if (configuration == null) {
-                configuration = new DefaultConfiguration();
-            }
+            // should work with a null configuration
```
ppkarwasz, line 765: This requires more explanations: the `PatternParser` does not use a configuration directly (or has null checks), but allows one to be injected into the `PatternConverter`s.

Looking at the available `PatternConverter`s, the color converters, "enc", "equals*", "highlight", "maxLength", "replace", "style", "ex*" require a configuration to create a `PatternParser`.

I didn't find any converters except `LiteralPatternConverter` that actually use `Configuration`. `LiteralPatternConverter` has appropriate null checks.

ppkarwasz:
@carterkozak: that should do it.

As far as I can see, no NPEs will result from providing `null` as configuration in the `PatternLayout`. I added a null check for the only case I found.

carterkozak (approved):
Great work, thank you for the contribution!

PR #758, opened 2022-02-17
LOG4J2-3388 custom levels configured using log4j2.properies is giving number format exception.

jnjanardhannaidu:
LOG4J2-3388  if we have below properties configured in log4j2.properies files, when we will hit NumberFormatException or there will be no logs for custom level.  This issue is happening because we are trying to convert string "VERBOSE" into integer which is not expected.  It is expected to covert the values which is there in property called intLevel only. 

customLevels = V
customLevel.V.name = VERBOSE
customLevel.V.intLevel = 525


jnjanardhannaidu:
LOG4J2-3388

PR #761, opened 2022-02-18
LOG4J2-3407 Log4j 1.2 bridge Check for non-existent appender when parsing properties

kennymacleod:
This change will make `PropertiesConfiguration` tolerant of attempts to resolve a non-existent Appender. See JIRA ticket for details.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfiguration.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfiguration.java
@@ -436,6 +436,10 @@ public class PropertiesConfiguration extends Log4j1Configuration {
         final String layoutPrefix = prefix + ".layout";
         final String filterPrefix = APPENDER_PREFIX + appenderName + ".filter.";
         final String className = OptionConverter.findAndSubst(prefix, props);
+        if (className == null) {
+            LOGGER.debug("Appender \"" + appenderName + "\" does not exist.");
+            return null;
+        }
         appender = manager.parseAppender(appenderName, className, prefix, layoutPrefix, filterPrefix, props, this);
         if (appender == null) {
             appender = buildAppender(appenderName, className, prefix, layoutPrefix, filterPrefix, props);
@@ -505,7 +509,7 @@ public class PropertiesConfiguration extends Log4j1Configuration {
         final ErrorHandler eh = newInstanceOf(errorHandlerClass, "ErrorHandler");
         final String[] keys = new String[] {
             // @formatter:off
-            errorHandlerPrefix + "." + ROOT_REF, 
+            errorHandlerPrefix + "." + ROOT_REF,
             errorHandlerPrefix + "." + LOGGER_REF,
             errorHandlerPrefix + "." + APPENDER_REF_TAG};
             // @formatter:on
--- a/log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
+++ b/log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
@@ -106,7 +106,7 @@ public class PropertiesConfigurationTest extends AbstractLog4j1ConfigurationTest
             assertTrue(filter.getFilter() instanceof NeutralFilterFixture);
         }
     }
-    
+
     @Test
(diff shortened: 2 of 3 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfiguration.java
```diff
@@ -436,6 +436,10 @@ public Appender parseAppender(final Properties props, final String appenderName)
         final String layoutPrefix = prefix + ".layout";
         final String filterPrefix = APPENDER_PREFIX + appenderName + ".filter.";
         final String className = OptionConverter.findAndSubst(prefix, props);
+        if (className == null) {
+            LOGGER.debug("Appender \"" + appenderName + "\" does not exist.");
+            return null;
+        }
```
kennymacleod, line 442: Without this null-check, there will be a NPE from the next line down.

log4j-1.2-api/src/test/resources/LOG4J2-3407.properties
```diff
@@ -0,0 +1,26 @@
...
+#
+
+log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
+log4j.appender.CONSOLE.filter.1=org.apache.log4j.config.NeutralFilterFixture
+log4j.appender.CONSOLE.filter.1.onMatch=neutral
+log4j.appender.CONSOLE.Target=System.out
+log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
+log4j.appender.CONSOLE.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p %c{1}:%L - %m%n
+
+log4j.logger.org.apache.log4j.xml=trace, A1
```
kennymacleod, line 25: Here we have a logger referring to an appender (A1) that does not exist

log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
```diff
@@ -125,6 +125,15 @@ public void testConsoleAppenderLevelRangeFilter() throws Exception {
       }
     }
 
+    @Test
+    public void testConfigureAppenderDoesNotExist() throws Exception {
+        // Verify that we tolerate a logger which specifies an appender that does not exist.
+        try (LoggerContext loggerContext = TestConfigurator.configure("target/test-classes/LOG4J2-3407.properties")) {
+            final Configuration configuration = loggerContext.getConfiguration();
+            assertNotNull(configuration);
+        }
```
kennymacleod, line 134: This test fails with an NPE without the fix

kerysgang937 (approved):

log4j-1.2-api/src/test/resources/LOG4J2-3407.properties
```diff
@@ -0,0 +1,26 @@
...
+log4j.rootLogger=trace, CONSOLE
```
kerysgang937, line 26: Kerstintdurden:release-2.2Ftownship.onelink.me

kerysgang937, line 26: log4j-1.2-api/src/test/resources/LOG4J2-3407.properties

log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfiguration.java
ppkarwasz: I would move this behavior to the `BuilderManager` (so we solve the problem at the source). The `BuilderManager` should return `new AppenderWrapper(null)` so that line 445 does not throw.

Remark also that Log4j 1.x behaves differently

- whenever **any** component has an empty property the error "Could not find value for key ..." is logged (cf. [source code](https://github.com/apache/logging-log4j1/blob/v1_2_17/src/main/java/org/apache/log4j/helpers/OptionConverter.java#L120).
- then there is a message "Could not instantiate appender named ..." (cf. [source code](https://github.com/apache/logging-log4j1/blob/de9f0ea504db9d1178db2cf45323d8e182a5df0a/src/main/java/org/apache/log4j/PropertyConfigurator.java#L790)). This should probably be checked just before the return statement. Remark that appender instantiation can also fail if some other properties are missing (e.g. `File` for the file appender).

PR #762, opened 2022-02-18
Trim property values

ppkarwasz:
Log4j 1.x almost always trimmed the values found in the configuration, the Log4j 1.x bridge should do the same.

Since `OptionConverter` already does most of the trimming, only class name trimming must be added. The (obsolete?) `Log4j1ConfigurationFactory` also fails to trim spaces in the _"level and appenders"_ specification.


log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
```diff
@@ -71,7 +71,7 @@ public Appender parseAppender(String name, String className, String prefix, Stri
             String filterPrefix, Properties props, PropertiesConfiguration config) {
         Objects.requireNonNull(plugins, "plugins");
         Objects.requireNonNull(className, "className");
-        PluginType<?> plugin = plugins.get(className.toLowerCase());
+        PluginType<?> plugin = plugins.get(className.toLowerCase().trim());
```
garydgregory, line 74: This could cause an NPE, see the bull check on the line below.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
```diff
@@ -118,7 +119,8 @@ public Layout parseLayout(String className, Element layoutElement, XmlConfigurat
         return null;
     }
     public Layout parseLayout(String className, String layoutPrefix, Properties props, PropertiesConfiguration config) {
-        PluginType<?> plugin = plugins.get(className.toLowerCase());
+        Objects.requireNonNull(className, "className");
+        PluginType<?> plugin = plugins.get(className.toLowerCase().trim());
```
garydgregory, line 123: NPE

log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
ppkarwasz: AFAIK `toLowerCase()` does not return nulls. :-) Besides, `className` is protected by an `Objects.requireNonNull`.

log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationParser.java
```diff
@@ -159,7 +159,7 @@ private void warn(final String string) {
     }
 
     private void buildAppender(final String appenderName, final String appenderClass) {
-        switch (appenderClass) {
+        switch (appenderClass.trim()) {
```
ppkarwasz, line 162: @garydgregory: I double checked that this private method is never called with a null `appenderClass`, but can you check again?

log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
garydgregory: I am sorry @ppkarwasz I misread the line, I though the trim was applied to the result of the get, my mistake.

garydgregory:
Hi @ppkarwasz 
I did this a bit differently in release-2.x, please verify your use case. 
TY!

ppkarwasz:
@garydgregory: I tested the `Log4j1ConfigurationConverter` with a configuration full of spaces and it worked correctly, thanks.

PR #764, opened 2022-02-20
Adds Log4j 1.x global threshold

ppkarwasz:
The `log4j.threshold` configuration key was not translated into its Log4j 2.x equivalent global filter.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationParser.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationParser.java
@@ -25,8 +25,10 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.TreeMap;
 
+import org.apache.log4j.LogManager;
 import org.apache.log4j.helpers.OptionConverter;
 import org.apache.logging.log4j.Level;
+import org.apache.logging.log4j.core.Filter.Result;
 import org.apache.logging.log4j.core.appender.ConsoleAppender;
 import org.apache.logging.log4j.core.appender.FileAppender;
 import org.apache.logging.log4j.core.appender.NullAppender;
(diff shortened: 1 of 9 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationParser.java
```diff
@@ -103,6 +107,18 @@
             if (Boolean.parseBoolean(debugValue)) {
                 builder.setStatusLevel(Level.DEBUG);
             }
+            // if log4j.reset=true then reset hierarchy
+            final String reset = getLog4jValue("reset");
+            if (reset != null && OptionConverter.toBoolean(reset, false)) {
```
garydgregory, line 112: I do not see this reset setting tested in this PR. Did I miss it? Also, elsewhere we call the reset API from a doConfigure() method, it seems odd to call it here. Is it then called twice or should this call be moved to the site when the configuration is actually applied? @rgoers WDYT?

ppkarwasz: Yes, you are right. I copy-pasted the "reset" part from the `PropertiesConfiguration`, without checking exactly what it does. I'll remove it for now (besides the `XmlConfiguration` does not recognize this option).

ppkarwasz: Looking at the Log4j 1.2 code, neither `Log4j1ConfigurationParser` nor any `Configuration` class should check the _"log4j.reset"_ key. This should be done in `PropertyConfigurator#configure` to decide whether to replace the old configuration or to make a `CompositeConfiguration` out of them.

Looking at [this SO question](https://stackoverflow.com/q/71087714/11748454) I have the impression that calling `PropertyConfigurator#configure` multiple times without resetting the old config, was a common practice.

garydgregory:
@ppkarwasz 
Thank you for your PR. 
If you are looking for more fixes to investigate, please have a look at the disabled tests in `org.apache.log4j.xml.DOMTestCase`.

PR #767, opened 2022-02-21
Provide a uniform Log4j 1.x message factory

ppkarwasz:
This PR provides a factory method that decides the type of message based on the runtime type of the logged object (cf. [LOG4J2-3080](https://issues.apache.org/jira/browse/LOG4J2-3080)).

Before the implementations of `maybeLog` and `forcedLog` created different types of messages.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
@@ -307,17 +307,28 @@ public class Category implements AppenderAttachable {
         }
     }
 
+    private static Message createMessage(Object message) {
+        if (message instanceof String) {
+            return new SimpleMessage((String) message);
+        }
+        if (message instanceof CharSequence) {
+            return new SimpleMessage((CharSequence) message);
+        }
+        if (message instanceof Map) {
+            return new MapMessage<>((Map<String, ?>) message);
+        }
+        if (message instanceof Message) {
+            return (Message) message;
+        }
+        return new ObjectMessage(message);
+    }
+
     public void forcedLog(final String fqcn, final Priority level, final Object message, final Throwable t) {
         final org.apache.logging.log4j.Level lvl = org.apache.logging.log4j.Level.toLevel(level.toString());
+        final Message msg = createMessage(message);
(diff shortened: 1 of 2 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
```diff
@@ -309,15 +309,10 @@ private void fireRemoveAppenderEvent(final Appender appender) {
...
+        final Message msg = LegacyMessageFactory.INSTANCE.newMessage(message);
         if (logger instanceof ExtendedLogger) {
-            @SuppressWarnings("unchecked")
-            final Message msg = message instanceof Message ? (Message) message
-                : message instanceof Map ? new MapMessage((Map) message) : new ObjectMessage(message);
             ((ExtendedLogger) logger).logMessage(fqcn, lvl, null, msg, t);
         } else {
-            final ObjectRenderer renderer = get(message.getClass());
-            final Message msg = message instanceof Message ? (Message) message
-                : renderer != null ? new RenderedMessage(renderer, message) : new ObjectMessage(message);
```
ppkarwasz, line 320: The `RenderedMessage` case is omitted, since this would require a stateful `MessageFactory` (per category).

Anyway there is currently no way to add elements to the `RendererMap`.

log4j-api/src/main/java/org/apache/logging/log4j/util/SortedArrayStringMap.java
```diff
@@ -128,6 +128,21 @@
...
     private transient boolean iterating;
 
+    /**
+     * Transforms a generic map to a {@link StringMap} by mapping the keys to
+     * {@link String}.
+     * 
+     * @param map a generic map
+     * @return a {@link StringMap}
+     */
+    public static SortedArrayStringMap fromGenericMap(final Map<?, ?> map) {
```
garydgregory, line 138: I don't think this is needed, would you check your tests against HEAD, without any main changes?

ppkarwasz: @garydgregory: after your fix it is not needed anymore.

PR #768, opened 2022-02-21
Improves behavior of failing Log4j 1.x builders

ppkarwasz:
If a Builder fails due to a missing property, it returns `null`. This activates the fallback _"instantiate by class name"_ mechanism that will fail with `NoClassDefFoundException` in the best scenario or will instantiate a native Log4j 1.x component in the worst scenario.

A better approach, IMHO, is to return an empty wrapper, which deactivates the fallback mechanism, but will be unwrapped to `null` soon after and removed from the configuration. If the Log4j 1.x builder does not check for a condition that is fatal to the Log4j 2.x component, an empty wrapper will be returned anyway.

This PR also increases the level of status logger messages to `ERROR` if the condition prevents the instantiation of the component.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/BuilderManager.java
@@ -24,6 +24,10 @@ import java.util.function.Function;
 
 import org.apache.log4j.Appender;
 import org.apache.log4j.Layout;
+import org.apache.log4j.bridge.AppenderWrapper;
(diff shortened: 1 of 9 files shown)


ppkarwasz:
This should prevent frustrating debugging like in [this StackOverflow question](https://stackoverflow.com/q/71149004/11748454), where the user just got:

```
2022-02-16 21:04:20,396 main ERROR Unable to create Appender org.apache.log4j.RollingFileAppender due to ClassNotFoundException:org.apache.log4j.RollingFileAppender
```

PR #769, opened 2022-02-21
Slf4j binding minor cleanups

dfa1:
Minor cleanups in SLF4J bindings

Including removal of [SLF4JLoggingException](https://github.com/apache/logging-log4j2/commit/d993a4faf17c1e5380e0eb87c02ee9d69318eb68)


log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
```diff
@@ -52,8 +52,7 @@ public void clear() {
     }
 
     @Override
-    @SuppressWarnings("unchecked") // nothing we can do about this, restricted by SLF4J API
-    public void setContextMap(@SuppressWarnings("rawtypes") final Map map) {
+    public void setContextMap(final Map<String, String> map) {
```
garydgregory, line 55: This will break source compatibility, does it matter?

dfa1: thanks for your time! :-) 

Given that SLF4J interface is: 
```
    public void setContextMap(Map<String, String> contextMap);
```

garydgregory: Seems ok then. Thank you for the info.

PR #770, opened 2022-02-21
Initial support for detached markers in SLF4J binding

dfa1:
Hello all,

please accept this small patch from a long time user of log4j: it will enable using log4j 2.x in several applications here at @six-group.com!

This proposal allows creating detached `Markers` as specified by the `IMarkerFactory` interface. A detached marker allows the caller to inject non-constant data that can be used to implement structured logging (my use-case). Because of this, they cannot be cached in `Log4jMarkerFactory.markerMap` or converted.
 
Please let me know if you require additional modifications (e.g. unit tests perhaps?)

### DIFF
--- a/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
+++ b/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
@@ -136,8 +136,7 @@ public class Log4jMarkerFactory implements IMarkerFactory {
 
     @Override
     public Marker getDetachedMarker(final String name) {
-        final org.apache.logging.log4j.Marker log4jMarker = MarkerManager.getMarker(name);
-        return new Log4jMarker(log4jMarker);
+        return new Log4jMarker(new MarkerManager.Log4jMarker(name));
     }
 
 }
--- a/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
+++ b/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
@@ -19,8 +19,10 @@ package org.apache.logging.slf4j;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Iterator;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArraySet;
 
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.MarkerManager;
@@ -36,9 +38,11 @@ public class Log4jMarkerFactory implements IMarkerFactory {
     private static final Logger LOGGER = StatusLogger.getLogger();
 
     private final ConcurrentMap<String, Marker> markerMap = new ConcurrentHashMap<>();
+    private final Set<String> detachedMarkers = new CopyOnWriteArraySet<>();
 
     /**
      * Returns a Log4j Marker that is compatible with SLF4J.
+     *
      * @param name The name of the Marker.
      * @return A Marker.
      */
@@ -63,26 +67,33 @@ public class Log4jMarkerFactory implements IMarkerFactory {
 
     /**
      * Returns a Log4j Marker converted from an existing custom SLF4J Marker.
+     * If the marker has been detached (see {@link Log4jMarkerFactory#detachMarker(String)}),
+     * returns the unmodified marker.
+     *
      * @param marker The SLF4J Marker to convert.
-     * @return A converted Log4j/SLF4J Marker.
+     * @return Same input Marker is it has been detached, a converted Log4j/SLF4J Marker otherwise.
      * @since 2.1
      */
     public Marker getMarker(final Marker marker) {
         if (marker == null) {
             throw new IllegalArgumentException("Marker must not be null");
         }
-        final Marker m = markerMap.get(marker.getName());
+        final String name = marker.getName();
+        if (detachedMarkers.contains(name)) {
+            return marker;
+        }
+        final Marker m = markerMap.get(name);
         if (m != null) {
             return m;
         }
-        return addMarkerIfAbsent(marker.getName(), convertMarker(marker));
+        return addMarkerIfAbsent(name, convertMarker(marker));
     }
 
     private static org.apache.logging.log4j.Marker convertMarker(final Marker original) {
         if (original == null) {
             throw new IllegalArgumentException("Marker must not be null");
         }
-        return convertMarker(original, new ArrayList<Marker>());
+        return convertMarker(original, new ArrayList<>());
     }
 
     private static org.apache.logging.log4j.Marker convertMarker(final Marker original,
@@ -105,6 +116,7 @@ public class Log4jMarkerFactory implements IMarkerFactory {
 
     /**
      * Returns true if the Marker exists.
+     *
      * @param name The Marker name.
      * @return {@code true} if the Marker exists, {@code false} otherwise.
      */
@@ -113,26 +125,19 @@ public class Log4jMarkerFactory implements IMarkerFactory {
         return markerMap.containsKey(name);
     }
 
-    /**
-     * Log4j does not support detached Markers. This method always returns false.
-     * @param name The Marker name.
-     * @return {@code false}
-     */
     @Override
     public boolean detachMarker(final String name) {
-        return false;
+        if (name == null) {
+            return false;
+        }
+        detachedMarkers.add(name);
+        return true;
     }
 
-    /**
-     * Log4j does not support detached Markers for performance reasons. The returned Marker is attached.
-     * @param name The Marker name.
-     * @return The named Marker (unmodified).
-     */
     @Override
     public Marker getDetachedMarker(final String name) {
-        LOGGER.warn("Log4j does not support detached Markers. Returned Marker [{}] will be unchanged.", name);
-        return getMarker(name);
+        final org.apache.logging.log4j.Marker log4jMarker = MarkerManager.getMarker(name);
+        return new Log4jMarker(log4jMarker);
     }
-
 
 }
--- a/log4j-slf4j-impl/src/test/java/org/apache/logging/slf4j/Log4jMarkerFactoryTest.java
+++ b/log4j-slf4j-impl/src/test/java/org/apache/logging/slf4j/Log4jMarkerFactoryTest.java
@@ -129,7 +129,7 @@ class Log4jMarkerFactoryTest {
         Marker nonDetachedMarker = sut.getMarker("name");
         Marker detachedMarker = sut.getDetachedMarker("name");
 
-        // Then, the original marker has been converted (and cached)
+        // Then
         assertNotSame(nonDetachedMarker, detachedMarker);
         assertTrue(nonDetachedMarker.equals(detachedMarker));
         assertTrue(detachedMarker.equals(nonDetachedMarker));
--- a/log4j-slf4j-impl/src/test/java/org/apache/logging/slf4j/Log4jMarkerFactoryTest.java
+++ b/log4j-slf4j-impl/src/test/java/org/apache/logging/slf4j/Log4jMarkerFactoryTest.java
@@ -1,0 +1,174 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.slf4j;
+
+import org.junit.jupiter.api.Test;
+import org.slf4j.Marker;
+
+import java.util.Collections;
+import java.util.Iterator;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotSame;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class Log4jMarkerFactoryTest {
+
+    @Test
+    void getMarker() {
+        // Given
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // When
+        Marker firstResult = sut.getMarker("FOO");
+        Marker secondResult = sut.getMarker("FOO");
+
+        // Then, second result is cached
+        assertSame(firstResult, secondResult);
+        assertTrue(sut.exists("FOO"));
+    }
+
+    @Test
+    void getDetachedMarker() {
+        // Given
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // When
+        Marker firstResult = sut.getDetachedMarker("FOO");
+        Marker secondResult = sut.getDetachedMarker("FOO");
+
+        // Then, second result is not cached (a detached marker)
+        assertNotSame(firstResult, secondResult);
+        assertFalse(sut.exists("FOO"));
+    }
+
+    @Test
+    void detachMarker() {
+        // Given
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // When
+        boolean result = sut.detachMarker("WHATEVER");
+
+        // Then
+        assertTrue(result);
+    }
+
+    @Test
+    void detachedMarkerIsUntouched() {
+        // Given
+        StructuredArg structuredArg = new StructuredArg(42);
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // we ask to detach markers with this same
+        sut.detachMarker(structuredArg.getName());
+
+        // When, we call the internal method getMarker (not part of the SLF4J org.slf4j.IMarkerFactory interface)
+        Marker result = sut.getMarker(structuredArg);
+
+        // Then, the original marker is not converted
+        assertSame(structuredArg, result);
+    }
+
+    @Test
+    void detachedMarkerIsNotCached() {
+        // Given
+        StructuredArg first = new StructuredArg(42);
+        StructuredArg second = new StructuredArg(10);
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // we ask to detach markers with this same
+        sut.detachMarker(first.getName());
+
+        // When
+        Marker firstResult = sut.getMarker(first);
+        Marker secondResult = sut.getMarker(second);
+
+        // Then, the original marker is not converted
+        assertSame(firstResult, first);
+        assertSame(secondResult, second);
+        assertNotSame(firstResult, secondResult);
+    }
+
+    @Test
+    void nonDetachedMarkerIsConverted() {
+        // Given
+        StructuredArg structuredArg = new StructuredArg(42);
+        Log4jMarkerFactory sut = new Log4jMarkerFactory();
+
+        // When
+        Marker result = sut.getMarker(structuredArg);
+
+        // Then, the original marker has been converted (and cached)
+        assertNotSame(structuredArg, result);
+        assertTrue(result instanceof Log4jMarker); // converted, not the original marker anymore
+    }
+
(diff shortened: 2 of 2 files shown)


garydgregory:
Hi @dfa1 
Thank you for the PR.
You will need to add tests to this PR to prove the code does what you say it does.

dfa1:
@garydgregory  thank you for the review... please have a look now, I wrote few unit tests

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
```diff
@@ -36,9 +38,11 @@
     private static final Logger LOGGER = StatusLogger.getLogger();
 
     private final ConcurrentMap<String, Marker> markerMap = new ConcurrentHashMap<>();
+    private final Set<String> detachedMarkers = new CopyOnWriteArraySet<>();
```
garydgregory, line 41: Doesn't tracking the markers defeat the purpose of them being "detached"?

dfa1:
@garydgregory is there something that you would like to change/improve in this PR? please let me know :-)

garydgregory:
I'm going to be quite busy for a couple of days but I will try to take a look. It might be only next week though. Thank you for your patience! That should not prevent someone else on the team to look at this of course. Do take note that we are currently taking time to validate a release candidate for 2.17.2.

dfa1:
@garydgregory hello 👋 would it be possible to have another look? 😊
Let me know how to improve this code to be included in a log4j release!

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
```diff
@@ -113,26 +125,19 @@ public boolean exists(final String name) {
...
-    /**
-     * Log4j does not support detached Markers for performance reasons. The returned Marker is attached.
-     * @param name The Marker name.
-     * @return The named Marker (unmodified).
-     */
     @Override
     public Marker getDetachedMarker(final String name) {
-        LOGGER.warn("Log4j does not support detached Markers. Returned Marker [{}] will be unchanged.", name);
-        return getMarker(name);
+        final org.apache.logging.log4j.Marker log4jMarker = MarkerManager.getMarker(name);
```
carterkozak, line 139: This `MarkerManager.getMarker(name)` invocation causes the `name` to be tracked in the `org.apache.logging.log4j.MarkerManager.MARKERS` collection, the proposed change only avoids the slf4j adapters cache of markers converted into the slf4j api.

For this to work, log4j2 would also need to be aware of detached markers, which would also require changes to `MarkerManager.Log4jMarker.isInstanceOf(Marker marker)` which currently does a reference equality check rather than checking the string value.

dfa1: @carterkozak thanks for the review! :) 

Yes, this is exactly the point: *I tried to keep Log4j unaware of detached markers*. 

Since Log4j doesn't support detached markers, while SLF4J does, the proposed implementation is wrapping a Lo4j's marker (that is cached internally as you mentioned) with a fresh instance of SLF4J's marker. 

Unless I'm missing something big, there are also few newly written unit tests that can be used to disproof my understanding of the problem.

rgoers: 1. I've never understood the point of a detached Marker. A marker is essentially nothing more than a String to be compared. Having multiple instances of the String makes no sense to me (which is why Log4j 2 didn't implement it).
2. My fear with this is that it will lead to a bug report that SLF4J detached markers aren't detached. Saying we don't support them sends a clearer message.

dfa1: thanks @rgoers for your time! In my use case a `Marker` contains health check data, a POJO that gets serialized in JSON for structured logging.  

The only advantage of detached markers is that they can be garbage collected as soon as they not needed anymore... but you're right, this method is not fulfilling this promise (because it is delegating to the `MarkerFactory` that is caching it). 

Let me try to fix that :-)

dfa1: @carterkozak @rgoers I pushed an additional commit to create fully detached markers (just to satisfy the SLF4J's contract). Is that enough? do you see any additional changes? 

Thanks for your time!

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
```diff
@@ -113,26 +125,18 @@ public boolean exists(final String name) {
...
-    /**
-     * Log4j does not support detached Markers for performance reasons. The returned Marker is attached.
-     * @param name The Marker name.
-     * @return The named Marker (unmodified).
-     */
     @Override
     public Marker getDetachedMarker(final String name) {
-        LOGGER.warn("Log4j does not support detached Markers. Returned Marker [{}] will be unchanged.", name);
-        return getMarker(name);
+        return new Log4jMarker(new MarkerManager.Log4jMarker(name));
```
carterkozak, line 139: I don't think this will work correctly when compared against a standard non-detached marker that log4j uses in filters because they use an instance comparison, not a string value comparison.

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
```diff
@@ -113,26 +125,18 @@ public boolean exists(final String name) {
...
-     * @return {@code false}
-     */
     @Override
     public boolean detachMarker(final String name) {
-        return false;
+        if (name == null) {
+            return false;
+        }
+        detachedMarkers.add(name);
+        return true;
```
carterkozak, line 134: I thought the point of detached markers was so that we can create arbitrarily many without worrying about the impact on heap -- if we hold a collection `detachedMarkers` strings, that is no longer true, and likely results in a memory leak.
I would suggest against supporting the `boolean detachMarker(final String name) {` method, at least until we have a well understood use-case for it.

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
carterkozak: So I think my point is, we may need a different log4j marker implementation than `MarkerManager.Log4jMarker` which handles string-value comparisons rather than identity comparisons.

dfa1: I start to see the problem... let me try to implement such value based marker :)

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
dfa1: I do have a use case :-) 

I will call this method *once* with "STRUCTURE_ARG". This is needed to avoid caching of the incoming `Marker` instance by name (all my markers are value-based classes). 

In `Log4jLogger` there is the following code:

```java
@Override
public void info(final Marker marker, final String s) {
    logger.logIfEnabled(FQCN, Level.INFO, getMarker(marker), s);
}
```

the ultimate goal of `detachedMarkers` is to skip the last line of the `getMarker` method :

```
    public Marker getMarker(final Marker marker) {
        if (detachedMarkers.contains(name)) {
            return marker;
        }
        final Marker m = markerMap.get(name);
        if (m != null) {
            return m;
        }
        return addMarkerIfAbsent(name, convertMarker(marker));  <-- I want to skip this processing
```

PS
Of course, it could be abused by calling it with millions of unique strings but that is true also for the non-detached version...

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
dfa1: @carterkozak looks like that `MarkerManager.Log4jMarker` is already string-value:

```java 
final Marker marker = (Marker) o;
return name.equals(marker.getName());
```

and Log4J implementation of SLF4J marker is doing:

```
final Log4jMarker other = (Log4jMarker) obj;
		if (!Objects.equals(marker, other.marker)) {
			return false;
		}
		return true;
```

did I miss something?

dfa1: @carterkozak created unit test:

```java
    @Test
    void comparisonDetachedAndNonDetachedMarkers() {
        // Given
        Log4jMarkerFactory sut = new Log4jMarkerFactory();

        // When
        Marker nonDetachedMarker = sut.getMarker("name");
        Marker detachedMarker = sut.getDetachedMarker("name");

        // Then
        assertNotSame(nonDetachedMarker, detachedMarker);
        assertTrue(nonDetachedMarker.equals(detachedMarker));
        assertTrue(detachedMarker.equals(nonDetachedMarker));
    }
```

does that looks good for you? (test is green)

carterkozak: Most checks don't use `.equals(`, rather `.isInstanceOf(`, for example https://github.com/apache/logging-log4j2/blob/f72100df0decc9bda96b4d769822c4e48b2848fc/log4j-core/src/main/java/org/apache/logging/log4j/core/filter/MarkerFilter.java#L70-L72

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
carterkozak: I think the `Log4jLogger` marker conversion fallback (when the slf4j marker implementation is not a Log4jMarker) should use a simple wrapper around the input instance without any caching. Logging would incur object allocation, but this is a special case where markers don't come from the marker factory.
https://github.com/apache/logging-log4j2/blob/f72100df0decc9bda96b4d769822c4e48b2848fc/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java#L379-L388

What do you think?

rgoers:
@dfa1 In your comment above you said you do have a use case (one would hope so for you to be implementing this) but you didn't explain your use case. All you really said is you want detached markers. Why? I can't think of a single thing that would be useful for that would be better done some other way. I do not want to add support for this without fully understanding how it is useful.

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
dfa1: @carterkozak yes, I'm already doing this with custom `org.slf4j.impl.StaticMarkerBinder` that is wrapping the incoming marker with `Log4jMarker` (to go via line 383 instead of line 386). 

Problem is that now our applications are sensitive to the order of the jars in the classpath: my binding must come *before* log4j-slf4j-impl. Do you see another way perhaps? (and thank you for the nice review by the way!)

dfa1:
@rgoers yes, of course :-)  Let me try to explain clearly: at @SIX-Group we have an internal requirement to use a structured logging format (called unified log event) for all our applications. This format is quite rich and detailed: there are predefined events for auditing, security incidents and so on. 

The initial implement was simply `JsonLayout` + fluent bit configuration (hundreds of lines of configuration): but that is not enough to express certain combination of logs (and quite complex to understand and hard to test).

Then, after some struggling with fluent bit, I tried to write a custom plugin in log4j2 to have a `UnifiedLogEventLayout`. The experiment was successful : now it is possible to unit test all the possible events and I must say was also a learning experience (i.e. log4j plugins are awesome!). 

And here the main point: I'm using custom SLF4J `Marker` instances to pass extra information like structured arg or predefined events. I don't expect to create thousands of markers per second, but usually we create few per day per application node (i.e. I'm not worried about performance at all).

Example:
```java
HealthCheck pojo = ...;
LOGGER.info(StructuredArg.of(healthPojo), "health check status");
```

`StructuredArg` implements `Marker` and `getName` yields always `STRUCTURED_ARG` but data is always different, so it cannot be cached by "name" (the default behaviour we have in log4j 2.17.x):

https://github.com/apache/logging-log4j2/blob/f72100df0decc9bda96b4d769822c4e48b2848fc/log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jLogger.java#L379-L388

My first attempt to workaround the problem was to wrap my marker with `Log4jMarker` (to go via line 383). To implement this workaround I used a custom `org.slf4j.impl.StaticMarkerBinder`: problem is that now all application have 2 `StaticMakerBinders` and the order matters.

My second attempt was this one (in the hope that someone else finds this useful too!).

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
dfa1: @carterkozak but that seems to be fine as well:

```java
        assertTrue(((Log4jMarker) detachedMarker).getLog4jMarker().isInstanceOf("name"));
        assertTrue(((Log4jMarker) nonDetachedMarker).getLog4jMarker().isInstanceOf("name"));
```

dfa1: implementation is simply: 

```java
        @Override
        @PerformanceSensitive({"allocation", "unrolled"})
        public boolean isInstanceOf(final String markerName) {
            requireNonNull(markerName, "A marker name is required");
            if (markerName.equals(this.getName())) {
                return true;
            }
```

rgoers:
Ok. So your big mistake here is trying to do this with the SLF4J API. It wasn't designed to support structured event logging. I fought for that years ago and gave up and that was one of the reasons I created Log4j 2. You really want to use a MapMessage or extend it. I have teams that are doing exactly this to capture data to display in Kibana dashboards. WE still use Markers, but the Marker simply identifies it a) as an event and b) the type of event. The second could have been captured in the event itself. 

This is the code that logs generates the event.

```
    public void recordInbound(Inbound in, Vendors vendor) {
        InboundLogEntry inboundLogEntry = new InboundLogEntry();
        inboundLogEntry.setVendor(vendor.name());
        inboundLogEntry.setInbound(true);
        inboundLogEntry.setDestination(in.getDestination());
        inboundLogEntry.setRefId(in.getId());
        inboundLogEntry.setMediaUrls(in.getMediaUrls());
        inboundLogEntry.setSource(in.getSource());
        inbound.ogEntry.setMessageId(in.getMessageId());
        inboundLogEntry.logEvent();
    }
```
and here is the logEvent method.
```
    public void logEvent() {
        StackTraceElement element = StackLocatorUtil.getStackTraceElement(2);
        LOGGER.atInfo().withLocation(element).withMarker(INBOUND_EVENT).log(this);
    }
```
Note that a detached marker is not required as all the variable data is in the event.

Finally, in the logging configuration we use JsonTemplateLayout and include in the template
```
  "event.action": {
    "$resolver": "marker",
    "field": "name"
  },
  "event.data": {
    "$resolver": "map",
    "stringified": true
  },
```
So the Marker will be logged as the event.action field and all the data will be included as attributes under event.data - such as event.data.vendor, event.data.destination, etc.

So this is exactly why I am not in favor of merging this PR as the Log4j API is much richer than SLF4J. You do not need to convert your whole application to use SLF4J but you should modify anything generating structured events.

garydgregory:
I have a product at work that does almost exactly what @rgoers describes, it works great, it's clean, and easy enough to understand. I encourage you as well to move those parts of your code to Log4j 2! 👍😀

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
carterkozak: I think the string method is fine, the overload which takes another marker is not

dfa1:
@rgoers @garydgregory thank you for the feedback ;)

Actually we already have a lot of internal libraries and frameworks written against `slf4j-api`, and we do the actual binding on per application basis, some of them even uses quarkus + jboss-logging. The `StructuredArg` marker is needed for one of those intermediate libraries. But I can't disagree with you: the current approach is far from being perfect but sometimes we have to deal with what you have!

@rgoers what we have is indeed very similar to log4j-audit! :-) (maybe a bit less generic)

dfa1:
@rgoers  @carterkozak @garydgregory 

*Current state*
- this PR is implementing the SLF4J's contract is a reasonable way (my humble opinion)
    - test coverage is good 
    - it is possible to create directly detached markers 
    - it is possible to detach an existing `Marker` (my use case with `StructuredArg`) to avoid any processing by log4j
    - detached markers can be filtering by name  (checked with @carterkozak  in `MarkerFilter`) 
    - `Marker.isInstanceOf(Marker)` is not working properly for detached markers because of identify comparison (this could be a problem for people using detached markers with log4j)
    
I will be more than happy to:
- finish this PR, adding anything you think is important
- add these changes in `log4j-slf4j18-impl` too
- provide feedback, testing and bug fixing after merging
- help you to maintain this code 

I don't see the point in using my time to maintain my "private" workarounds, I would be happy (and honored too) to use my time to collaborate with you on log4j 2.x instead! :)

My 2 cents,
D.

rgoers:
@dfa1 I do understand your dilemma. You already have existing code that has implemented this in the only way possible using SLF4J, which is a hack, However, I have a real problem implementing something that encourages doing things the wrong way. I don't want to see you have to maintain your private workaround. I'd rather see you spend time improving what Log4j does to make doing it the right way even easier. This change doesn't do that. The question I would ask is if in the long run you would be better off maintaining your own fork of the slf4j-log4j bridge or in spending the effort to do it the right way.

FWIW, I did a search of Jira and only one other issue (LOG4J2-585) mentions detached Markers and that was to agree they should not be implemented in Log4j 2. To be honest, a surprisingly low number of users use Markers despite the utility they provide. But you are the first to have ever requested support for SLF4J's detached markers. 

I guess I am at the point with this PR that I personally don't want to merge it. But if another committer disagrees I won't veto it if they do.

dfa1:
@rgoers thanks for yours the honest feedback! :) Maintaining my own fork(s) of SLF4J is what I wanted to avoid by proposing this PR!

After this long discussion, I realized that what **I really** need is just a way to avoid:

```java
return addMarkerIfAbsent(name, convertMarker(marker));
```

do you see a clean way to skip this conversion? maybe a feature toggle (config)?

rgoers:
That would be back to the same thing. Now you would have getMarker returning a detached marker. If I thought detached markers were a good idea then detachedMarker would be where it belongs.
But remember, Log4jMarker has a public constructor. You are free to use ServiceLoader to locate the SLF4J Service Provider and get the Marker factory and then create Markers to your hearts content.

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
ppkarwasz: What about a `Map<String, WeakReference<Marker>>`?

dfa1:
@rgoers I'm already using public ctor in `Log4jMarker`: but in order to do that I need to "override" the code coming from log4j binding and now my application are sensitive to the order of jars in the classpath.  And that is the workaround I would like to  remove with this PR.

log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/Log4jMarkerFactory.java
dfa1: @ppkarwasz thank you for the review :) maybe the name of the field is misleading: should be "detachedMarkerNames"! For my use case, the marker name is always "STRUCTURED_ARG" but every instance has different data (in my use case, health check data). 

Basically, log4j is designed to use markers as "glorified strings", instead I would like to use them as "data carrier" (after my discussion with @rgoers, see below). I hope it is clear enough!

dfa1:
@rgoers tried also your last suggestion too:

> You are free to use ServiceLoader to locate the SLF4J Service Provider and get the Marker factory and then create Markers to your hearts content." 

but unfortunately it is sensitive to classpath order since SLF4J is picking up the first provider: 

```java
List<SLF4JServiceProvider> providersList = findServiceProviders();
reportMultipleBindingAmbiguity(providersList);
if (providersList != null && !providersList.isEmpty()) {
    PROVIDER = providersList.get(0); <-- only the first one is considered
    // SLF4JServiceProvider.initialize() is intended to be called here and nowhere else.
    PROVIDER.initialize();
```

so far, my workaround is to provide my JAR with my custom layout implementation *before* any other jar in the classpath (this is done for about 20 Dockerfiles, it feels really dirty).

dfa1:
ok,  I'm going to close this pull request since it is not making any progress since March. Thank you for your feedback!

**To future users, reading this long discussion**:
after a bit of experiments with logback I concluded that is trivial to use `Marker` for structured logging as it does not introduce any conversion/stringification like log4j2.x.

PR #771, opened 2022-02-23
Improves validation of Log4j 2.x components

ppkarwasz:
Since most configurations use the `PluginBuilder` to build components, many constraints are not enforced, when the same components are built programmatically.

This PR provides a `Builder#validate()` method that performs annotation based validation and includes it in a couple of appender builders.

This way constraint validation of programmatically configured components will be on par with those created by `PluginBuilder` at the cost of double validation for the latter.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/validation/validators/NotBlankValidator.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/validation/validators/NotBlankValidator.java
@@ -16,41 +16,32 @@
  */
 package org.apache.logging.log4j.core.config.plugins.validation.validators;
 
-import java.util.Collection;
-import java.util.Map;
-
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.config.plugins.validation.ConstraintValidator;
-import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
-import org.apache.logging.log4j.core.util.Assert;
+import org.apache.logging.log4j.core.config.plugins.validation.constraints.NotBlank;
 import org.apache.logging.log4j.status.StatusLogger;
+import org.apache.logging.log4j.util.Strings;
 
 /**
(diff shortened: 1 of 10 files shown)


ppkarwasz:
`log4j-layout-template-json` uses a different semantic for when builders fail (they throw instead of returning `null`). I should look into that.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Builder.java
```diff
@@ -42,4 +44,17 @@
      * object.
      */
     T build();
+
+    public default boolean isValid() {
```
jvz, line 48: Nit: `public` isn't needed on interfaces as everything is public.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/validation/validators/NotBlankValidator.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.logging.log4j.status.StatusLogger;
+import org.apache.logging.log4j.util.Strings;
+
+/**
+ * Validator that checks if a {@link CharSequence} is not entirely composed of
+ * whitespace.
+ *
+ * @since 2.17.2
+ */
+public class NotBlankValidator implements ConstraintValidator<Required> {
```
jvz, line 31:
```suggestion
public class NotBlankValidator implements ConstraintValidator<NotBlank> {
```

Plus the changes to this class that go with it.

jvz (changes requested):
Looks pretty good, got a couple notes.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/validation/validators/NotBlankValidator.java
ppkarwasz: Fixed, thanks.

jvz (approved):

PR #774, opened 2022-02-25
[LOG4J2-3413] Fixes resolution of non-Log4j properties

ppkarwasz:
For properties that does not start with `log4j` or `org.apache.logging.log4j` the loosed token-based lookup should not be
applied. There are a couple of exceptions to this rule: the properties that start with `AsyncLogger` and those related to
`disableThreadContext`.

This PR prevents properties such as `level` to be used as a substitute for `log4j2.level`

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
@@ -18,6 +18,7 @@ package org.apache.logging.log4j.util;
 
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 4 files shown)


jvz (approved):
Nice update!

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
```diff
@@ -100,9 +101,19 @@ public int compare(final PropertySource o1, final PropertySource o2) {
...
+                Pattern.CASE_INSENSITIVE);
+        private static final Pattern PROPERTY_TOKENIZER = Pattern.compile("([A-Z]*[a-z0-9]+|[A-Z0-9]+)[-._/]?");
         private static final Map<CharSequence, List<CharSequence>> CACHE = new ConcurrentHashMap<>();
+        static {
+            // Add legacy properties without Log4j prefix
+            CACHE.put("disableThreadContext", Arrays.asList("disable", "thread", "context"));
+            CACHE.put("disableThreadContextStack", Arrays.asList("disable", "thread", "context", "stack"));
+            CACHE.put("disableThreadContextMap", Arrays.asList("disable", "thread", "context", "map"));
+            CACHE.put("isThreadContextMapInheritable", Arrays.asList("is", "thread", "context", "map", "inheritable"));
+        }
```
ppkarwasz, line 116: These are here just to satisfy `LegacyPropertiesCompatibilityTest`, which verifies that the tokenized version of the legacy and normalized properties are the same.

From the `PropertiesUtil` perspective they are not strictly necessary, since `getStringProperty("disableThreadContext")` will always look up at least the exact string "disableThreadContext" and it will retrieve the value before falling back to token-based comparison.

PR #775, opened 2022-02-26
[LOG4J2-3415] Support JSON configuration using embedded JSON parser

jvz:
* Moves `JsonReader` from log4j-layout-json-template to log4j-api.
* Adds support for comments in parsing JSON for compatibility with existing Jackson usage.
* Adds new default `JsonConfigurationFactory` implementation using this.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/JsonReaderTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/JsonReaderTest.java
@@ -14,7 +14,7 @@
  * See the license for the specific language governing permissions and
  * limitations under the license.
  */
-package org.apache.logging.log4j.util;
+package org.apache.logging.log4j.core.util;
 
 import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/util/JsonReaderTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/util/JsonReaderTest.java
@@ -14,7 +14,7 @@
  * See the license for the specific language governing permissions and
  * limitations under the license.
  */
-package org.apache.logging.log4j.layout.template.json.util;
+package org.apache.logging.log4j.util;
 
 import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
@@ -24,6 +24,7 @@ import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.Map;
 
 class JsonReaderTest {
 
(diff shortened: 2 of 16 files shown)


garydgregory:
Why does any parser need to be in the API module when all configurations
take place in the Core module? I don't think we should fatten up the API
with parsers.

Gary

jvz:
For convenience with the PropertiesUtil enhancements Ralph proposed. Otherwise, it can certainly go in core instead.

garydgregory:
> For convenience with the PropertiesUtil enhancements Ralph proposed. Otherwise, it can certainly go in core instead.

It seems like a bad idea to me to put a parser of any kind in a module that I would hope to be minimal in breadth and scope, as well as only focused on logging.

jvz:
That's a good point. The `PropertiesUtil` enhancement will likely need some additional inversion of control or something to avoid having the JSON parsing done at that level. I'll move this to core, though I also need to revise the parser a bit as I think Jackson might also allow for duplicate keys (otherwise I'm not exactly sure how to represent a configuration for things like arbiters).

rgoers:
@jvz You are correct. Accomplishing what I want to with Properties will be much simpler if the JsonReader is a utility class in the api module. I will absolutely need JSON support for that. But to be honest, I am not sure that PropertiesUtil really needs to be in the API. Or if we have one it could probably be greatly simplified since it controls only a very few system properties.

ppkarwasz:
@rgoers: isn't the properties format enough for your properties enhancement? `log4j-core` just needs to provide a `PropertySource` for the JSON format.

IMHO `PropertiesUtil` should stay in the API. This way `log4j2.loggerContextFactory` (or its equivalent) does not need to be set using Java system properties. Ditto for the simple logger implementation.

jvz:
I think one possibility would be to add an interface in log4j-api for loading properties and the general LoggerContext-specific properties. Then we can move the parsing logic to core.

We could also try to split up properties related to controlling the API versus controlling the Core.

jvz:
I'm going to merge this as-is. We can figure out how to integrate JSON parsing with PropertiesUtil later.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/JsonReaderTest.java
```diff
@@ -375,6 +376,11 @@ public void test_nesting() {
                                 }})));
     }
 
+    @Test
```
vy, line 379: I'd prefer here a `@ParametrizedTest` touching different corner cases. Further, the following cases are not tested:
1. multi-line comment blocks (should pass)
1. not closed comment blocks (should fail)
1. not opened comment blocks (should fail)

jvz: Added in https://github.com/apache/logging-log4j2/commit/e03f1832ed0791e5c489a65983de9ee707145ccd

PR #777, opened 2022-02-28
[LOG4J2-3418] Fix Log4j2CloudConfigLoggingSystem registration

ppkarwasz:
This PR adds support for the `spring.factories` registration method of the `Log4j2CloudConfigLoggingSystem` to replace the old Java system properties injection method.

A Log4j 2.x `log4j2.disableCloudConfigLoggingSystem` property is added to enable a choice between the `Log4j2CloudConfigLoggingSystem` and `Log4j2LoggingSystem`.


membersound:
Could you add documentation how to activate this feature in future?

ppkarwasz:
> Could you add documentation how to activate this feature in future?

The behavior of the artifact does not change: you add it to the classpath and Spring Boot detects it. What changes is the way this is done: before Java system properties were used, so **all** Spring Boot applications in the JVM tried to use this plugin. After this PR only the one with `log4j-spring-boot` will use it.

PR #780, opened 2022-03-02
LOG4J2-3423 - Prevent unclosed JAR file containing Log4j configuration

rtlusty:


rtlusty:
Better option for the fix has been provided. Closing.

PR #783, opened 2022-03-04
LOG4J2-3424: Correctly handle property attribute values

carterkozak:
This fixes handling of properties in .properties configurations
as well as those defined using `<Property name="key" value="value"/>`
which are parsed slightly differently from those which use the
`<Property name="key">value</Property>` syntax.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginConfiguration;
 import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.config.plugins.PluginValue;
 import org.apache.logging.log4j.core.lookup.StrSubstitutor;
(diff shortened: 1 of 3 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
```diff
@@ -118,15 +119,36 @@ public static Property createProperty(final String name, final String value) {
...
+     * @param rawValue The value without any substitution applied.
+     * @param configuration configuration used to resolve the property value from the rawValue
+     * @return A Property.
+     */
     @PluginFactory
     public static Property createProperty(
             @PluginAttribute("name") final String name,
             @PluginValue(value = "value", substitute = false) final String rawValue,
-            @PluginValue("value") final String value) {
+            @PluginConfiguration final Configuration configuration) {
```
carterkozak, line 144: Attribute are consumed as they're read, this way we consume the value once, and process it using the configuration substitutor.

carterkozak:
planning to clean this up and fix revapi tomorrow

log4j-core/src/main/java/org/apache/logging/log4j/core/config/Property.java
```diff
@@ -118,15 +119,38 @@ public static Property createProperty(final String name, final String value) {
      * @param value The value.
      * @return A Property.
      */
+    public static Property createProperty(
+            final String name,
+            final String rawValue,
+            final String value) {
+        if (name == null) {
+            LOGGER.error("Property name cannot be null");
```
garydgregory, line 127: If it's an error and we continue as if nothing happened, then either this should not be logged as an ERROR or some error-like behavior should happen. It's just weird to log an ERROR and then continue as if nothing happened.

carterkozak: Agree that this is odd, it matches the behavior of the previous plugin-based constructor. I have to imagine anything that receives a Property instance with a null key will throw, so it's probably best to throw in this case. I can update all the constructors

carterkozak:
Merged + changelog

PR #785, opened 2022-03-04
Remove the revapi plugin

carterkozak:
We should enable it again, however at the moment it's causing
all builds to fail.


garydgregory:
FWIW, over at Apache Commons, we use JApiCmp.

carterkozak:
I'll check it out :-)
Not planning to block this commit on investigating other options, hoping to get builds green to unblock bugfixes in the interim

garydgregory:
Sounds good to me.

PR #786, opened 2022-03-06
[LOG4J2-3426] Refactors wrapping code

ppkarwasz:
Refactors the wrapping logic into single `wrap` or `adapt` methods to be sure we always perform the correct logic.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
@@ -376,7 +376,7 @@ public class Category implements AppenderAttachable {
         if (appender == null && LogManager.isLog4jCorePresent()) {
             final org.apache.logging.log4j.core.Appender coreAppender = CategoryUtil.getAppenders(logger).get(name);
             if (coreAppender != null) {
-                addAppender(appender = new AppenderWrapper(coreAppender));
+                addAppender(appender = AppenderWrapper.adapt(coreAppender));
             }
         }
         return appender;
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/FilterWrapper.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/FilterWrapper.java
@@ -27,6 +27,27 @@ public class FilterWrapper extends Filter {
 
     private final org.apache.logging.log4j.core.Filter filter;
 
+    /**
+     * Adapts a Log4j 2.x filter into a Log4j 1.x filter. Applying this method to
+     * the result of {@link FilterAdapter#adapt(Filter)} should return the original
+     * Log4j 1.x filter.
+     * 
+     * @param filter a Log4j 2.x filter
+     * @return a Log4j 1.x filter or {@code null} if the parameter is {@code null}
+     */
+    public static Filter adapt(org.apache.logging.log4j.core.Filter filter) {
+        if (filter instanceof Filter) {
+            return (Filter) filter;
+        }
+        if (filter instanceof FilterAdapter) {
+            return ((FilterAdapter) filter).getFilter();
+        }
+        if (filter != null) {
+            return new FilterWrapper(filter);
+        }
+        return null;
+    }
+
     public FilterWrapper(org.apache.logging.log4j.core.Filter filter) {
         this.filter = filter;
     }
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LayoutAdapter.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LayoutAdapter.java
@@ -29,10 +29,32 @@ import org.apache.logging.log4j.core.layout.ByteBufferDestination;
 public class LayoutAdapter implements org.apache.logging.log4j.core.Layout<String> {
     private Layout layout;
 
-    public LayoutAdapter(Layout layout) {
+    /**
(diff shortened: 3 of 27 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/FileAppenderBuilder.java
```diff
@@ -125,21 +125,16 @@ public Appender parseAppender(final String name, final String appenderPrefix, fi
...
-        } else if (layout != null) {
-            fileLayout = new LayoutAdapter(layout);
-        }
         final org.apache.logging.log4j.core.Filter fileFilter = buildFilters(level, filter);
         if (fileName == null) {
             LOGGER.error("Unable to create FileAppender, no file name provided");
             return null;
         }
-        return new AppenderWrapper(FileAppender.newBuilder()
+        return AppenderWrapper.wrap(FileAppender.newBuilder()
```
ppkarwasz, line 137: This and similar changes actually modify the function's behavior: if the argument of `AppenderWrapper.wrap` is `null`, then `null` is returned. However this should be counteracted by #768 .

log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
```diff
@@ -376,7 +376,7 @@ public Appender getAppender(final String name) {
         if (appender == null && LogManager.isLog4jCorePresent()) {
             final org.apache.logging.log4j.core.Appender coreAppender = CategoryUtil.getAppenders(logger).get(name);
             if (coreAppender != null) {
-                addAppender(appender = new AppenderWrapper(coreAppender));
+                addAppender(appender = AppenderWrapper.wrap(coreAppender));
```
ppkarwasz, line 379: This is the main change: if `coreAppender` is an `AppenderAdapter` and **no** filters are configured, then the wrapped object is returned.

@garydgregory: can you check if this is Ok?

pooja-achievement:
@ppkarwasz,

Hi Piotr,

I am not building log4j2 1.x bridge jars but downloading the jar. I wanted to test with this Pull Request changes, is it possible for you to provide log4j1.x bridge.jar log4j2-core.jar and log4j2-api.jar built with this change ??

ppkarwasz:
Hi @pooja-achievement ,
I am waiting for a second opinion on these changes: there is not much sense in testing something that will not end up in the next release. If everything is alright, we can generate a snapshot.

log4j-1.2-api/src/main/java/org/apache/log4j/bridge/FilterWrapper.java
```diff
@@ -27,6 +27,16 @@
 
     private final org.apache.logging.log4j.core.Filter filter;
 
+    public static Filter wrap(org.apache.logging.log4j.core.Filter filter) {
+        if (filter instanceof FilterAdapter) {
```
garydgregory, line 31: Test if the input is already a FilterWrapper and return it?

log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LayoutAdapter.java
```diff
@@ -29,10 +29,23 @@
 public class LayoutAdapter implements org.apache.logging.log4j.core.Layout<String> {
     private Layout layout;
 
-    public LayoutAdapter(Layout layout) {
+    public static org.apache.logging.log4j.core.Layout<?> adapt(Layout layout) {
+        if (layout instanceof LayoutWrapper) {
```
garydgregory, line 33: Test if input is already a LayoutAdapter and return it?

log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LayoutWrapper.java
```diff
@@ -26,6 +26,16 @@
 
     private final org.apache.logging.log4j.core.Layout<?> layout;
 
+    public static Layout wrap(org.apache.logging.log4j.core.Layout<?> layout) {
+        if (layout instanceof LayoutAdapter) {
```
garydgregory, line 30: Test if input is already a LayoutWrapper and return it?

log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/SyslogAppenderBuilder.java
```diff
@@ -127,20 +127,15 @@ private Appender createAppender(final String name, final Log4j1Configuration con
...
-        } else if (layout != null) {
-            appenderLayout = new LayoutAdapter(layout);
-        } else {
+        org.apache.logging.log4j.core.Layout<? extends Serializable> appenderLayout = LayoutAdapter.adapt(layout);
+        if (appenderLayout == null) {
             appenderLayout = SyslogLayout.newBuilder()
                     .setFacility(Facility.toFacility(facility))
-                    .setConfiguration(configuration)
-                    .build();
+                    .setConfiguration(configuration).build();
```
garydgregory, line 134: I like the existing formatting better.

garydgregory (changes requested):
- I like the refactoring in general, but:
- Let's stick to one term: adapt instead of wrap, since sometimes, we "unwap" or do nothing.
- Add some minimal Javadoc to new public methods to help maintainers.
- Adapter methods should check if the input is already "adapted" and return it for consistency and avoid multi-layer wrapping by accident.
- See my comments.

garydgregory (approved):
LGTM.

PR #787, opened 2022-03-06
[LOG4J2-3427] Add a ServiceLoader tool

ppkarwasz:
This adds a `ServiceLoader` which:

 * automatically discards service classes, which can not be instantiated (e.g. classes in the server's classloader of a servlet container), from the list of returned services,
 * implements a common mechanism to override the service found by `ServiceLoader`,
 * adds warnings if we only need one service, but multiple candidates are found.


jochenw:
Shouldn't the dependency junit-pioneer have test scope?

ppkarwasz:
@jochenw, yes and it has a `test` scope.
The scope of dependencies is inherited from the parent POM.

PR #788, opened 2022-03-08
Update documentation 

phreed:
Documentation corrections an updates as mentioned in https://issues.apache.org/jira/b…rowse/LOG4J2-3429
Corrected typo in builder method 'withtFilter' -> 'withFilter'

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
@@ -215,7 +215,7 @@ public class LoggerConfig extends AbstractFilterable implements LocationAware {
         }
 
         /**
-         * @deprecated Use {@link #withFilter()} instead which performs the same function.
+         * @deprecated Use {@link #withFilter(Filter)} instead which performs the same function.
          */
         @Deprecated
         public B withtFilter(Filter filter) {
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
@@ -893,6 +893,15 @@ public class LoggerConfig extends AbstractFilterable implements LocationAware {
                 return filter;
             }
 
+            /**
+             * @deprecated Use {@link #withFilter(Filter)} instead which performs the same function.
+             */
+            @Deprecated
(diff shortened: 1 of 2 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -214,7 +214,7 @@ public Filter getFilter() {
             return filter;
         }
 
-        public B withtFilter(Filter filter) {
+        public B withFilter(Filter filter) {
```
ppkarwasz, line 217: I have mixed feelings about this change: yes, it would be nice to have a correctly spelled method, but the other one was already released and it is API.

Don't change the name of the method, deprecate it and add another one with the correct spelling.

phreed: I deprecated the misspelled method and created a new one with the correct spelling as suggested.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -884,7 +893,7 @@ public Filter getFilter() {
                 return filter;
             }
 
-            public B withtFilter(Filter filter) {
+            public B withFilter(Filter filter) {
```
ppkarwasz, line 905: Also copy and deprecate.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
phreed: There were two such cases.
What is the rule for naming methods?
When should the name be `setFilter` and when `withFilter`?
I just noticed that the appender was changed from  `setFilter` and when `withFilter`

garydgregory: > I have mixed feelings about this change: yes, it would be nice to have a correctly spelled method, but the other one was already released and it is API.
> 
> Don't change the name of the method, deprecate it and add another one with the correct spelling.

Just add a new method with the correct name and deprecate the bogus one. Simple ;-)

garydgregory:
The "with" style is old, the "set" style is new. We already deprecated a bunch of "with" methods in favor of "set" methods.

ppkarwasz:

PR #789, opened 2022-03-09
[LOG4J2-3419] Adds support for custom Log4j 1.x levels

ppkarwasz:
Adds support for custom Log4j 1.x levels in the form `level_name#class_name`.

Levels defined directly in Log4j 2.x can be used in the bridge using the form `level_name#org.apache.logging.log4j.Level`.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Level.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Level.java
@@ -23,6 +23,7 @@ import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.Locale;
 
+import org.apache.log4j.helpers.OptionConverter;
 import org.apache.logging.log4j.util.Strings;
 
 /**
@@ -48,50 +49,50 @@ public class Level extends Priority implements Serializable {
      * The <code>OFF</code> has the highest possible rank and is
      * intended to turn off logging.
      */
-    public static final Level OFF = new Level(OFF_INT, "OFF", 0);
+    public static final Level OFF = new Level(OFF_INT, "OFF", 0, org.apache.logging.log4j.Level.OFF);
 
     /**
      * The <code>FATAL</code> level designates very severe error
      * events that will presumably lead the application to abort.
      */
-    public static final Level FATAL = new Level(FATAL_INT, "FATAL", 0);
+    public static final Level FATAL = new Level(FATAL_INT, "FATAL", 0, org.apache.logging.log4j.Level.FATAL);
 
     /**
      * The <code>ERROR</code> level designates error events that
      * might still allow the application to continue running.
      */
-    public static final Level ERROR = new Level(ERROR_INT, "ERROR", 3);
+    public static final Level ERROR = new Level(ERROR_INT, "ERROR", 3, org.apache.logging.log4j.Level.ERROR);
 
     /**
      * The <code>WARN</code> level designates potentially harmful situations.
      */
-    public static final Level WARN = new Level(WARN_INT, "WARN", 4);
+    public static final Level WARN = new Level(WARN_INT, "WARN", 4, org.apache.logging.log4j.Level.WARN);
 
     /**
      * The <code>INFO</code> level designates informational messages
      * that highlight the progress of the application at coarse-grained
      * level.
      */
-    public static final Level INFO = new Level(INFO_INT, "INFO", 6);
+    public static final Level INFO = new Level(INFO_INT, "INFO", 6, org.apache.logging.log4j.Level.INFO);
 
     /**
      * The <code>DEBUG</code> Level designates fine-grained
      * informational events that are most useful to debug an
      * application.
      */
-    public static final Level DEBUG = new Level(DEBUG_INT, "DEBUG", 7);
+    public static final Level DEBUG = new Level(DEBUG_INT, "DEBUG", 7, org.apache.logging.log4j.Level.DEBUG);
 
     /**
      * The <code>TRACE</code> Level designates finer-grained
      * informational events than the <code>DEBUG</code> level.
      */
-    public static final Level TRACE = new Level(TRACE_INT, "TRACE", 7);
+    public static final Level TRACE = new Level(TRACE_INT, "TRACE", 7, org.apache.logging.log4j.Level.TRACE);
 
     /**
      * The <code>ALL</code> has the lowest possible rank and is intended to
      * turn on all logging.
      */
-    public static final Level ALL = new Level(ALL_INT, "ALL", 7);
+    public static final Level ALL = new Level(ALL_INT, "ALL", 7, org.apache.logging.log4j.Level.ALL);
 
     /**
      * Serialization version id.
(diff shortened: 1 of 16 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/Priority.java
```diff
@@ -96,6 +96,7 @@
     transient int level;
     transient String levelStr;
     transient int syslogEquivalent;
+    transient org.apache.logging.log4j.Level version2Equivalent;
```
garydgregory, line 99: How about "version2Level" instead?

log4j-1.2-api/src/test/java/org/apache/log4j/helpers/UtilLoggingLevelTest.java
```diff
@@ -17,29 +17,40 @@
...
      * Test toLevel("fiNeSt").
      */
     public void testToLevelFINEST() {
-        assertSame(UtilLoggingLevel.FINEST, UtilLoggingLevel.toLevel("fiNeSt"));
+        assertEquals(UtilLoggingLevel.FINEST, UtilLoggingLevel.toLevel("fiNeSt"));
+    }
+
+    static Stream<Arguments> namesAndLevels() {
+        @SuppressWarnings("unchecked")
+        final List<UtilLoggingLevel> levels = UtilLoggingLevel.getAllPossibleLevels();
```
garydgregory, line 45: Do we really need a local variable here?

garydgregory, line 45: Do we really need a local variable here?

log4j-1.2-api/src/test/java/org/apache/log4j/helpers/OptionConverterLevelTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void testCustomLog4j2Levels() {
+        final int infoDebug = (StandardLevel.INFO.intLevel() + StandardLevel.DEBUG.intLevel()) / 2;
+        final org.apache.logging.log4j.Level v2Level = org.apache.logging.log4j.Level.forName("INFO_DEBUG", infoDebug);
+        Level v1Level = OptionConverter.toLevel("INFO_DEBUG#" + org.apache.logging.log4j.Level.class.getName(), null);
+        assertNotNull(v1Level);
+        assertEquals(v2Level, v1Level.getVersion2Level());
+        final int expectedLevel = (Priority.INFO_INT + Priority.DEBUG_INT) / 2;
+        assertEquals(expectedLevel, v1Level.toInt());
+        // Non-existent level
+        v1Level = OptionConverter.toLevel("WARN_INFO#" + org.apache.logging.log4j.Level.class.getName(), null);
```
garydgregory, line 118: This assignment seems superfluous (IMO)

garydgregory, line 118: This assignment seems superfluous (IMO)

log4j-1.2-api/src/main/java/org/apache/log4j/Priority.java
```diff
@@ -148,6 +149,14 @@ final int getSyslogEquivalent() {
         return syslogEquivalent;
     }
 
+    /**
+     * Returns the Log4j 2.x level associated with this priority
```
garydgregory, line 153: For getter methods, I write Javadocs that say "Gets ...". IOW, a getter "Gets...", a setter "Sets..." and so on.

garydgregory, line 153: For getter methods, I write Javadocs that say "Gets ...". IOW, a getter "Gets...", a setter "Sets..." and so on.

log4j-1.2-api/src/main/java/org/apache/log4j/Level.java
```diff
@@ -48,50 +49,50 @@
...
      */
-    public static final Level DEBUG = new Level(DEBUG_INT, "DEBUG", 7);
+    public static final Level DEBUG = new Level(DEBUG_INT, "DEBUG", 7, org.apache.logging.log4j.Level.DEBUG);
 
     /**
      * The <code>TRACE</code> Level designates finer-grained
      * informational events than the <code>DEBUG</code> level.
      */
-    public static final Level TRACE = new Level(TRACE_INT, "TRACE", 7);
+    public static final Level TRACE = new Level(TRACE_INT, "TRACE", 7, org.apache.logging.log4j.Level.TRACE);
```
garydgregory, line 89: I like this solution because it makes the correspondence between version 1 and 2 levels obvious, especially because of the surprising difference in the level number order.

garydgregory, line 89: I like this solution because it makes the correspondence between version 1 and 2 levels obvious, especially because of the surprising difference in the level number order.

log4j-1.2-api/src/main/java/org/apache/log4j/helpers/OptionConverter.java
```diff
@@ -75,55 +100,82 @@ public CharMap(final char key, final char replacement) {
...
+    public static Level convertLevel(final org.apache.logging.log4j.Level level) {
+        final Level l = toLevel(level.name(), null);
+        return l != null ? l : Level.ERROR;
     }
 
     public static org.apache.logging.log4j.Level convertLevel(final String level,
             final org.apache.logging.log4j.Level defaultLevel) {
         final Level l = toLevel(level, null);
-        return l != null ? convertLevel(l) : defaultLevel;
+        return l != null ? l.getVersion2Level() : defaultLevel;
```
garydgregory, line 178: I am really not a fan of a variable called "l" (lower-case L) which can look like a one "1" depending on your font. I prefer to make names obvious. Here maybe "actualLevel"? Or something nor "l" ;-)

garydgregory, line 178: I am really not a fan of a variable called "l" (lower-case L) which can look like a one "1" depending on your font. I prefer to make names obvious. Here maybe "actualLevel"? Or something nor "l" ;-)

garydgregory (changes requested):
Hi @ppkarwasz 
Good stuff I think :-) See my few requests SVP.
Version 2 also supports custom levels, it would be nice to note in comments if verstion 2 custom levels matter in different parts of this PR.
TY!

log4j-1.2-api/src/main/java/org/apache/log4j/Level.java
ppkarwasz: Since a level conversion is performed every time a Log4j 1.x logger is used, this should also provide the fastest conversion we can offer.
The inverse conversion (Log4j 2.x level to Log4j 1.x level) uses a longer path (a lookup in a static map), but it is only necessary if someone wants to use native Log4j 1.x appenders or filters together with Log4j 2.x custom levels. Very unlikely.

ppkarwasz:
@garydgregory: I added some Javadoc to the methods where Log4j 2.x and Log4j 1.x levels interact. Basically the `Level(int, String, int)` constructor also creates new Log4j 2.x levels and `OptionConverter#toLevel(String, String, Level)` can create Log4j 1.x equivalents of custom Log4j 2.x levels.

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
```diff
@@ -675,24 +675,13 @@ private void parseLevel(Element element, LoggerConfig logger, boolean isRoot) {
             }
         } else {
             String className = subst(element.getAttribute(CLASS_ATTR));
+            final Level level;
             if (EMPTY_STR.equals(className)) {
-                logger.setLevel(OptionConverter.convertLevel(priStr, org.apache.logging.log4j.Level.DEBUG));
+                level = OptionConverter.toLevel(priStr, Level.DEBUG);
```
garydgregory, line 680: THere are several places where we specify a default level (Level.DEBUG) in this method and likely elsewhere. It would be nicer to have a default level defined in a constant IMO; this would make it obvious that this is a "default" as opposed to "in this case, we have level Foo". WDYT?

garydgregory (changes requested):
Looks, one (last?) comment in the code.

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
ppkarwasz: I'll look into that. There are actually two levels used throughout the code: `Level.DEBUG` in places where we expect an operation to fail and `Level.ERROR` where a conversion failure is unexpected (like in the `convertLevel` methods).

garydgregory (approved):
Looks good now👍😀

PR #794, opened 2022-03-14
[LOG4J2-3419] Add a Log4j 1.x level name pattern converter

ppkarwasz:
When custom levels are in play the names of Log4j 1.x custom levels (which are not unique) and those of Log4j 2.x custom levels (which must be unique) do not match.

There is a need therefore for a pattern converter specific to Log4j 1.x.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/pattern/Log4j1LevelPatternConverter.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/pattern/Log4j1LevelPatternConverter.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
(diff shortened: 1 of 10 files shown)


garydgregory (commented):
Shouldn't we document %v1Level on the site? Or is it only used internally?

ppkarwasz:
@garydgregory, it's rather used internally.
For custom Log4j 1.x levels it will give the level name without the class name suffix. For custom Log4j 2.x levels it should give the same result as `%p`.

log4j-1.2-api/src/test/java/org/apache/log4j/config/AbstractLog4j1ConfigurationTest.java
```diff
@@ -276,7 +276,7 @@ private void testDailyRollingFileAppender(final String configResource, final Str
 
     public void testConsoleEnhancedPatternLayout() throws Exception {
         final PatternLayout layout = (PatternLayout) testConsole("config-1.2/log4j-console-EnhancedPatternLayout");
-        assertEquals("%d{ISO8601} [%t][%c] %-5p %properties %ndc: %m%n", layout.getConversionPattern());
+        assertEquals("%d{ISO8601} [%t][%c] %-5v1Level %properties %ndc: %m%n", layout.getConversionPattern());
```
garydgregory, line 279: I think we need a comment before this line and one like it because I am guessing that "v1Level" was converted by us when users say "level" in their 1.2 configuration files?

log4j-1.2-api/src/main/java/org/apache/log4j/pattern/Log4j1LevelPatternConverter.java
```diff
@@ -0,0 +1,50 @@
...
+        return INSTANCE;
+    }
+
+    private Log4j1LevelPatternConverter() {
+        super("Log4j1Level", "v1Level");
+    }
+
+    @Override
+    public void format(LogEvent event, StringBuilder toAppendTo) {
+        toAppendTo.append(
```
garydgregory, line 45: Why the weird formatting? I just use one line but that's just me...

garydgregory (approved):
LGTM.

PR #797, opened 2022-03-15
LOG4J2-3393 Improve JsonTemplateLayout performance.

vy:
* Replace hardcoded `8 * 1024` private value used for `byteBufferSize` in `StringBuilderEncoder()` ctor with `Constants.ENCODER_BYTE_BUFFER_SIZE`.
* Replace `StringBuilderEncoder` (uses an internal thread-local) and `LockingStringBuilderEncoder` (uses locks) with custom `JsonTemplateLayout.StringBuilderEncoder` class implementing `Encoder<StringBuilder>` interface. This was made possible by making `TextEncoderHelper.encodeText()` public. This makes use of JTL recyclers and provides better allocation performance. For instance, if JTL is configured with thread-local recycler, each encoding request will trigger a single TLA, whereas previously it needed to do two: one for `JsonTemplateLayout.Context` and `StringBuilderEncoder`.
* Add `CharSequence` specialization in `MessageResolver`.
* Improve troubleshooting experience by replacing lambdas with classes in `TemplateResolvers`.
* Improve locality and branching in `TemplateResolvers`.
* Improve JMH tests.
* Thanks so much to @carterkozak for being the awesome sparring partner.


log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/MessageResolver.java
```diff
@@ -134,7 +134,10 @@ private static void resolveString(
             jsonWriter.writeObjectStart();
             jsonWriter.writeObjectKey(fallbackKey);
         }
-        if (message instanceof StringBuilderFormattable) {
+        if (message instanceof CharSequence) {
+            final CharSequence sequence = (CharSequence) message;
+            jsonWriter.writeString(sequence);
```
ppkarwasz, line 139: Is the local variable necessary?

vy: Nope, I just find it easier to read and wrap my mind around when there is one operation per line.

garydgregory: I agree with @ppkarwasz, this just makes the code more verbose than it has to be IMO.
Type casting is not an "operation", it does not do anything.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/MessageResolver.java
```diff
@@ -134,7 +134,10 @@ private static void resolveString(
             jsonWriter.writeObjectStart();
             jsonWriter.writeObjectKey(fallbackKey);
         }
-        if (message instanceof StringBuilderFormattable) {
+        if (message instanceof CharSequence) {
```
carterkozak, line 137: Is this an optimization for SimpleMessage? I wonder if there's something we can do better in `jsonWriter.writeString(StringBuilderFormattable)`? in the SimpleMessage/ReusableSimpleMessage case, I suspect we would do better using `getFormattedMessage()` as `StringBuilder.append(String)` is faster than iterating over a charsequence. Note that this would allocate in the case that the message value is a non-string charsequence.

StringBuilderFormattable handling writes from the original message to an intermediate StringBuilder, which is escaped into the final StringBuilder. I think we could probably optimize this process, for instance I think we always call `quoteString(formattableBuffer, 0, length);` which uses `StringBuilder.append(cs,start,end)` rather than `StringBuilder.append(cs)` -- unfortunately only the latter appears to be special-cased for AbstractStringBuilder.

carterkozak (approved):
Thanks, Volkan, the change looks great! We can always iterate through additional performance optimizations later on :-)

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/MessageResolver.java
vy: Fixed in 5a37bafb.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/MessageResolver.java
vy: You have a point @carterkozak. This said, I think we better address this in a future iteration to get this PR rolling.

PR #799, opened 2022-03-15
Update CustomConfiguration.java

DavidChinedu23:
Added 2020 to Apache license.


garydgregory:
Why? In general, do not touch the license headers unless it is broken. There is no such URL anyway.

rgoers:
This change would cause our RAT check to fail, would not match ASF guidelines documented at https://www.apache.org/legal/src-headers.html, and as Gary noted the URL is invalid.

PR #804, opened 2022-03-18
[LOG4J2-3427] Replace ServiceLoader calls with ServiceLoaderUtil

ppkarwasz:
In an attempt to uniformize the calls to `ServiceLoader` and work around its limitations, this replaces the `ServiceLoader#load` call sites with `ServiceLoaderUtil#loadServices`.

`ServiceLoaderUtil` is also modified to return `Stream` instead of a `List`, hence allowing for a more comfortable filtering and sorting of the results.

### DIFF
--- a/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
+++ b/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
@@ -16,13 +16,11 @@
  */
 package org.apache.logging.log4j.util.java9;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import java.lang.invoke.MethodHandles;
-import java.util.Collections;
 import java.util.List;
-import java.util.ServiceConfigurationError;
 import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.util.PropertySource;
@@ -35,43 +33,21 @@ public class ServiceLoaderUtilTest {
 
     @Test
     public void testServiceResolution() {
-        // Run only if we are a module
-        if (ServiceLoaderUtil.class.getModule().isNamed()) {
-            List<Object> services = Collections.emptyList();
-            // Service from test module
-            try {
-                services = ServiceLoaderUtil.loadServices(Service.class, MethodHandles.lookup())
-                        .collect(Collectors.toList());
-            } catch (ServiceConfigurationError e) {
-                fail(e);
-            }
-            assertEquals(2, services.size(), "Service services");
-            // BetterService from test module
-            services.clear();
-            try {
-                services = ServiceLoaderUtil.loadServices(BetterService.class, MethodHandles.lookup())
-                        .collect(Collectors.toList());
-            } catch (ServiceConfigurationError e) {
-                fail(e);
-            }
-            assertEquals(1, services.size(), "BetterService services");
-            // PropertySource from org.apache.logging.log4j module from this module
-            services.clear();
-            try {
-                services = ServiceLoaderUtil.loadServices(PropertySource.class, MethodHandles.lookup())
-                        .collect(Collectors.toList());
-            } catch (ServiceConfigurationError e) {
-                fail(e);
-            }
-            assertEquals(0, services.size(), "PropertySource services");
-            // PropertySource from within org.apache.logging.log4j module
-            services.clear();
-            try {
-                services = PropertySource.loadPropertySources().collect(Collectors.toList());
-            } catch (ServiceConfigurationError e) {
-                fail(e);
-            }
-            assertEquals(2, services.size(), "PropertySource services");
-        }
+        List<Object> services;
+        // Service from test module
+        services = assertDoesNotThrow(() -> ServiceLoaderUtil.loadServices(Service.class, MethodHandles.lookup())
+                .collect(Collectors.toList()));
+        assertThat(services).hasSize(2);
+        // BetterService from test module
+        services = assertDoesNotThrow(() -> ServiceLoaderUtil.loadServices(BetterService.class, MethodHandles.lookup())
+                .collect(Collectors.toList()));
+        assertThat(services).hasSize(1);
+        // PropertySource from org.apache.logging.log4j module from this module
+        services = assertDoesNotThrow(() -> ServiceLoaderUtil.loadServices(PropertySource.class, MethodHandles.lookup())
+                .collect(Collectors.toList()));
+        assertThat(services).hasSize(0);
+        // PropertySource from within org.apache.logging.log4j module
+        services = assertDoesNotThrow(() -> PropertySource.loadPropertySources().collect(Collectors.toList()));
+        assertThat(services).hasSize(2);
     }
 }
--- a/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
+++ b/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the license for the specific language governing permissions and
+ * limitations under the license.
+ */
+package org.apache.logging.log4j.util.java9;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.lang.invoke.MethodHandles;
+import java.util.Collections;
+import java.util.List;
+import java.util.ServiceConfigurationError;
+import java.util.stream.Collectors;
+
+import org.apache.logging.log4j.util.PropertySource;
+import org.apache.logging.log4j.util.ServiceLoaderUtil;
+import org.apache.logging.log4j.util.java9.test.BetterService;
+import org.apache.logging.log4j.util.java9.test.Service;
+import org.junit.jupiter.api.Test;
+
+public class ServiceLoaderUtilTest {
+
+    @Test
+    public void testServiceResolution() {
+        // Run only if we are a module
+        if (ServiceLoaderUtil.class.getModule().isNamed()) {
+            List<Object> services = Collections.emptyList();
+            // Service from test module
+            try {
+                services = ServiceLoaderUtil.loadServices(Service.class, MethodHandles.lookup())
+                        .collect(Collectors.toList());
+            } catch (ServiceConfigurationError e) {
+                fail(e);
+            }
+            assertEquals(2, services.size(), "Service services");
+            // BetterService from test module
+            services.clear();
+            try {
+                services = ServiceLoaderUtil.loadServices(BetterService.class, MethodHandles.lookup())
+                        .collect(Collectors.toList());
+            } catch (ServiceConfigurationError e) {
+                fail(e);
+            }
+            assertEquals(1, services.size(), "BetterService services");
+            // PropertySource from org.apache.logging.log4j module from this module
+            services.clear();
+            try {
+                services = ServiceLoaderUtil.loadServices(PropertySource.class, MethodHandles.lookup())
+                        .collect(Collectors.toList());
+            } catch (ServiceConfigurationError e) {
+                fail(e);
+            }
+            assertEquals(0, services.size(), "PropertySource services");
+            // PropertySource from within org.apache.logging.log4j module
+            services.clear();
+            try {
+                services = PropertySource.loadPropertySources().collect(Collectors.toList());
+            } catch (ServiceConfigurationError e) {
+                fail(e);
+            }
+            assertEquals(2, services.size(), "PropertySource services");
+        }
+    }
+}
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
@@ -17,6 +17,7 @@
 package org.apache.logging.log4j.util;
 
 import java.io.IOException;
+import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.security.AccessController;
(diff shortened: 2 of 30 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
```diff
@@ -120,6 +120,22 @@ public ClassLoader run() {
         return classLoaders.toArray(EMPTY_CLASS_LOADER_ARRAY);
     }
 
+    /**
+     * Gets the the give classloader and its ancestors.
```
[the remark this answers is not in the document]
ppkarwasz: Yes, that didn't come right.

log4j-api/src/main/java/org/apache/logging/log4j/util/ProviderUtil.java
```diff
@@ -63,13 +63,10 @@
...
-        for (final ClassLoader classLoader : LoaderUtil.getClassLoaders()) {
-            try {
-                loadProviders(classLoader);
-            } catch (final Throwable ex) {
-                LOGGER.debug("Unable to retrieve provider from ClassLoader {}", classLoader, ex);
-            }
-        }
+        ServiceLoaderUtil.loadServices(Provider.class)
+                .filter(provider -> validVersion(provider.getVersions()) && !PROVIDERS.contains(provider))
+                .forEach(PROVIDERS::add);
```
carterkozak, line 68: No need to check contains, add will return false and no-op if the element already exists

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
```diff
@@ -47,10 +57,20 @@ private ServiceLoaderUtil() {
      * 
      * @param <T>         The service type.
      * @param serviceType The class of the service.
-     * @return List of available services.
+     * @return Stream of available services.
      */
-    public static <T> List<T> loadServices(final Class<T> serviceType) {
-        return loadServices(serviceType, null);
+    public static <T> Stream<T> loadServices(final Class<T> serviceType) {
```
carterkozak, line 62: This is an api/abi break moving from `List<T>` to `Stream<T>`. The Stream allows us to avoid unused service-loader lookups, but should be added as a new method where we deprecate the original.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
```diff
@@ -60,43 +80,48 @@ private ServiceLoaderUtil() {
...
-            addOsgiServices(serviceType, services);
+            try {
+                return Stream.concat(services, loadOsgiServices(serviceType));
+            } catch (Throwable e) {
+                if (verbose) {
+                    StatusLogger.getLogger().error("Unable to load OSGI services for service {}", serviceType, e);
+                }
+            }
         }
         return services;
```
carterkozak, line 126: I wonder if we want something like `return services.unique()` here to avoid loading the same instance from multiple related classloaders?

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
```diff
@@ -110,28 +135,64 @@ private ServiceLoaderUtil() {
...
-    private static <T> void addOsgiServices(final Class<T> serviceType, final Collection<T> services) {
+    private static <T> Stream<? extends T> loadOsgiServices(final Class<T> serviceType) {
         final Iterable<? extends T> iterable = org.glassfish.hk2.osgiresourcelocator.ServiceLoader
                 .lookupProviderInstances(serviceType);
-        if (iterable != null) {
-            iterable.forEach(services::add);
+        return iterable != null ? StreamSupport.stream(iterable.spliterator(), false) : Stream.empty();
+    }
+
+    private static class ServiceLoaderSpliterator<S> implements Spliterator<S> {
```
carterkozak, line 149: jdk9+ introduced similar functionality with https://docs.oracle.com/javase/9/docs/api/java/util/ServiceLoader.html#stream--

The upside is that `Provider.get()` (instantiating a default instance) can fail (classpath/linker issues) without breaking the entire call.
Probably not worth adding jdk11 mr-jar code to handle, but we should add a comment describing that the feature should be used when jdk11 is the minimum supported version

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
ppkarwasz: I added the `ServiceLoaderUtil` class after the 2.17.2 release, so there is no API break for now.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
carterkozak: `.distinct()` rather (a real stream method!)

ppkarwasz: I'll go for a single instance per class for now. The `Provider` service overrides `equals/hashCode`, but the other service does not.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
ppkarwasz: I'll add `ServiceLoader#stream()` to the `ServiceLoaderUtil` in master, when this version reaches its final form.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
```diff
@@ -110,28 +139,64 @@ private ServiceLoaderUtil() {
...
+
+        private final Iterator<S> serviceIterator;
+        private final ClassLoader classLoader;
+        private final Logger logger;
+        private final String serviceName;
+
+        public ServiceLoaderSpliterator(final Class<S> serviceType, final ClassLoader classLoader,
+                final boolean verbose) {
+            this.serviceIterator = ServiceLoader.load(serviceType, classLoader).iterator();
```
rgoers, line 162: This code will not work for applications using the module path. Currently, all calls to ServiceLoader happen in the jar that needs the service. However, according to https://openjdk.java.net/projects/jigsaw/spec/sotms/#services, if an automatic module declares a service it behaves as if it had a corresponding provides clause.  Note that log4j-api has a module-info.java, so it will act as an explicitly named module. Currently, it declares uses clauses for the services log4j-api needs. This change will mean it will require a uses clauses for any use of ServiceLoader, which makes no sense and certainly won't work as expected.

rgoers (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
ppkarwasz: @rgoers,
Thanks for the explanations.

In `master` you use a lambda that calls [`ServiceLoader#load(ModuleLayer, Class)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/ServiceLoader.html#load(java.lang.ModuleLayer,java.lang.Class)) to work around the caller semantics. This obviously can not work in 2.x, since `ModuleLayer` does not exist and we would need a Java9 version of each class that uses the `ServiceLoaderUtil`. Do you know any tricks that can be applied to Java 8? Would adding a `MethodHandles.Lookup` parameter work?

If it works, we could have a

```lang-java
ServiceLoaderUtil#loadServices(Class<T> serviceType, MethodHandles.Lookup lookup, boolean useTccl)
```
signature and look for services in the caller's classloader and the thread context classloader (if `useTccl=true`).

@jvz, any suggestions?

rgoers: @ppkarwasz No matter how you slice it, so long as the call originates from the log4j-api module it won't work in JPMS. I would suggest that you use ServiceLoader.load(Class<S> service, ClassLoader classLoader) from the lambda. You will end up calling it once for each ClassLoader you have. In master I don't have to do this since anything running on the module path is going to have a ModuleLayer. 
Another option, of course, is to create a Java 9 specific version of this that supports using ModuleLayer. Then the code you currently have would work. That is probably the route I would go.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
```diff
@@ -110,28 +78,73 @@ private ServiceLoaderUtil() {
...
         }
+        return Collections.emptyList();
+
     }
 
-    private static <T> void addOsgiServices(final Class<T> serviceType, final Collection<T> services) {
-        final Iterable<? extends T> iterable = org.glassfish.hk2.osgiresourcelocator.ServiceLoader
-                .lookupProviderInstances(serviceType);
-        if (iterable != null) {
-            iterable.forEach(services::add);
```
ppkarwasz, line 134: This is not the right way to find OSGI services in Log4j: Log4j artifacts register OSGI services properly, so no need for the (probably) unsupported `osgi-resource-locator` or other tricks.

log4j-api/src/test/java/org/apache/logging/log4j/util/ServiceLoaderUtilTest.java
```diff
@@ -86,42 +85,4 @@ public void testBrokenServiceFile() {
         assertEquals(warnings + 2, counter.get());
     }
 
-    @Test
-    public void testMultipleServicesPresentWarning() {
-        final int warnings = counter.get();
```
ppkarwasz, line 91: The methods used in these tests were unused and removed.

log4j-api/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
ppkarwasz: I used the `stream()` method in the Java 9 specific class, so we don't forget.

ppkarwasz:
After several fixes, this should work now. A summary of changes:

- I removed `LoaderServiceUtil#getService` that I introduced in a previous commit, since it's logic is not really needed,
- OSGI support was also removed from `LoaderServiceUtil`, because it uses an artifact (`osgi-resource-locator`) that seems to be abandoned (it is still supported by JAXB API and similar, but I found no other references). Log4j registers (and probably should register) its services as native OSGI services.
- all `ServiceLoader#load` call sites have been refactored to use `ServiceLoaderUtil`,
- the number of `ServiceLoader#load` calls has been reduced to at most 2: one for the caller's classloader and possibly one for the thread context classloader,
- most `ServiceLoaderUtil#loadServices` call sites store the result in a static variable. To prevent memory leaks those calls don't use the TCCL.
- a Java 9 version of `ServiceLoaderUtil` was introduced to take advantage of the new `stream()` method.

ppkarwasz:
@jvz,
I added an `OsgiServiceLocator` to retrieve the services that are registered as OSGI services. This also works for services declared through _Declarative Services_ in other bundles, although adding DS in Log4j should not be necessary: we register the services manually.

log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.logging.log4j.util.java9.test.BetterService;
+import org.apache.logging.log4j.util.java9.test.Service;
+import org.junit.jupiter.api.Test;
+
+public class ServiceLoaderUtilTest {
+
+    @Test
+    public void testServiceResolution() {
+        // Run only if we are a module
+        if (ServiceLoaderUtil.class.getModule().isNamed()) {
```
jvz, line 39: You can use `assumeTrue()` from JUnit here.

log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+            List<Object> services = Collections.emptyList();
+            // Service from test module
+            try {
+                services = ServiceLoaderUtil.loadServices(Service.class, MethodHandles.lookup())
+                        .collect(Collectors.toList());
+            } catch (ServiceConfigurationError e) {
```
jvz, line 45: Nit: these types of tests can use `assertDoesNotThrow(() -> ...)`

log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+                fail(e);
+            }
+            assertEquals(2, services.size(), "Service services");
```
jvz, line 48: If you're going to add assertion messages, I'd suggest just using `assertThat()` from AssertJ to get a nice output message when assertions fail. For example, this could be `assertThat(services).hasSize(2)`.

log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public static <T> Stream<T> loadServices(final Class<T> serviceType, final Lookup lookup, final boolean verbose) {
+        final Bundle bundle = FrameworkUtil.getBundle(lookup.lookupClass());
+        if (bundle != null) {
+            final BundleContext ctx = bundle.getBundleContext();
+            try {
+                @SuppressWarnings("unchecked")
+                final ServiceReference<T>[] serviceRefs = (ServiceReference<T>[]) ctx
+                        .getAllServiceReferences(serviceType.getName(), null);
```
jvz, line 60: Can't we use the typed APIs here rather than the old untyped ones?

jvz (commented):
Looks generally good to me! I left some small suggestions, though feel free to merge this when you're ready.

log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
ppkarwasz: Oh, somehow I missed that, probably because the new method is not called `getAllServiceReferences`. Thanks.

log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/ServiceLoaderUtilTest.java
ppkarwasz: I am afraid I will have to leave it for now, since the module runs tests using Surefire 2.13 and JUnit 3. Upgrading the Surefire plugin is impossible due to [SUREFIRE-2073](https://issues.apache.org/jira/browse/SUREFIRE-2073).

jvz: Huh, oh well. If this is running via JUnit 3, shouldn't the test extend TestCase or something like that?

ppkarwasz: My mistake, Surefire is running the tests as [POJO tests](https://maven.apache.org/surefire/maven-surefire-plugin/examples/pojo-test.html). Since these also use the `JUnit3Provider`, I thought that JUnit 3 was involved. It isn't.

However these tests are not run by JUnit 5, which wasn't even supported in version 2.13. I'll try some hacks (like compiling the tests using Java 8 and `module-info.java` using Java 9) and maybe it will work on the latest Surefire release.

rgoers:
This looks very good to me. I much prefer the MethodHandles approach to what I was doing in 3.0.

PR #805, opened 2022-03-20
ternary operator condition needs to be updated

Suji04:


vy:
Thanks for the heads up @Suji04! I've applied your changes in 6f282e6.

PR #807, opened 2022-03-24
[LOG4J2-3359] Adds Log4j 1.x specific syslog layout

ppkarwasz:
Log4j 1.x always used the BSD syslog layout, but allowed for an internal layout to specify the _MESSAGE_ format.

This PR adds a variant of the Log4j 2.x `SyslogLayout`, that emulates Log4j 1.x behavior.


garydgregory:
@rgoers 
Ping: Does this look good to you?

ppkarwasz:
I added a third parameter to `DatePatternConverter`, so that I can use it with a fixed English locale. I am not documenting it in the manual, because for now it does not work with the `FixedDateFormat`.
By replacing `FastDateFormat` from the initial commit with `DatePatternConverter` I want to fix two problems: the generation of garbage, even if it is not needed and the possibility to cache the formatted date.

rgoers (approved):

PR #808, opened 2022-03-26
Make ConfigurationFactory injectable

jvz:
This allows for a ConfigurationFactory binding to be registered as well as generally participate in dependency injection.

Signed-off-by: Matt Sicker <mattsicker@apache.org>


log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -137,71 +119,15 @@ public ConfigurationFactory() {
...
-
     private static final String[] PREFIXES = {"log4j2.", "log4j2.Configuration."};
 
-    private static volatile AuthorizationProvider authorizationProvider;
-
-    /**
-     * Returns the ConfigurationFactory.
-     * @return the ConfigurationFactory.
-     */
+    @Deprecated(since = "3.0.0", forRemoval = true)
```
jvz, line 124: Being invoked by some Spring Boot class even though our override doesn't invoke it. Not sure what's going on, but it's also causing a test failure for me.

PR #810, opened 2022-03-31
Update Spring to 5.3.18 which fixes CVE-2022-22965 aka "Spring4shell".

loganaden:
See details: https://tanzu.vmware.com/security/cve-2022-22965


garydgregory:
Dependabot takes care of this for us.

jvz:
We're updated to Spring 5.3.20.

PR #811, opened 2022-04-01
fix typo in method name getRequesteApiVersion()

controllerface:
SLF4JServiceProvider defines the `getRequestedApiVersion()` method, but this class has a typo, the method is incorrectly named `getRequesteApiVersion()` which causes the following error message to be reported when using this library:

```
Unexpected problem occured during version sanity check
Reported exception:
java.lang.AbstractMethodError: Receiver class org.apache.logging.slf4j.SLF4JServiceProvider does not define or inherit an implementation of the resolved method 'abstract java.lang.String getRequestedApiVersion()' of interface org.slf4j.spi.SLF4JServiceProvider.
```


ppkarwasz:
Hello @controllerface,
That is not a typo, it is the method name SLF4J 1.8.x uses: see [LOG4J2-3370](https://issues.apache.org/jira/browse/LOG4J2-3370).

controllerface:
ok. wow. that's just... really confusing. I will close the PR.

PR #812, opened 2022-04-01
[LOG4J2-3462] Duplicate dependency in pom.xml

clara0:
https://issues.apache.org/jira/browse/LOG4J2-3462


vy:
Thanks so much @clara0!

PR #813, opened 2022-04-03
[LOG4J2-3440] Synchronize Log4j 1.x and Log4j 2.x appenders

ppkarwasz:
Forwards the `addAppender`, `callAppenders`, `getAllAppenders` and `getAppender` calls the the corresponding Log4j 2.x objects if Log4j 2.x Core is present.

The `getAllAppenders` method only returns those appenders, which are attached to homonymous logger config and are native Log4j 1.x appenders.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
@@ -16,13 +16,18 @@
  */
 package org.apache.log4j;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Map;
 import java.util.ResourceBundle;
 import java.util.Vector;
 import java.util.concurrent.ConcurrentMap;
+import java.util.stream.Collectors;
 
+import org.apache.log4j.bridge.AppenderAdapter;
 import org.apache.log4j.bridge.AppenderWrapper;
+import org.apache.log4j.bridge.LogEventWrapper;
 import org.apache.log4j.helpers.AppenderAttachableImpl;
 import org.apache.log4j.helpers.NullEnumeration;
 import org.apache.log4j.legacy.core.CategoryUtil;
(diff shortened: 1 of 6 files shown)


log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
```diff
@@ -353,14 +369,26 @@ public boolean getAdditivity() {
...
-     * {@link NullEnumeration} is returned.
+     * Get all the Log4j 1.x appenders contained in this category as an
+     * {@link Enumeration}. Log4j 2.x appenders are omitted.
      *
      * @return Enumeration An enumeration of the appenders in this category.
      */
     @Override
-    @SuppressWarnings("rawtypes")
-    public Enumeration getAllAppenders() {
+    public Enumeration<Appender> getAllAppenders() {
```
garydgregory, line 378: Using a generic here will cause compile errors in some cases

log4j-1.2-api/src/main/java/org/apache/log4j/legacy/core/CategoryUtil.java
```diff
@@ -69,6 +79,22 @@ public static LoggerContext getLoggerContext(final Logger logger) {
...
+    }
+
+    /**
+     * Gets the {@code LoggerConfig} corresponding to this logger or an empty
+     * optional if this logger does not have a {@code LoggerConfig}
+     * 
+     * @param logger the target logger,
+     * @return a logger config.
+     */
+    public static Optional<LoggerConfig> getLoggerConfig(final Logger logger) {
```
garydgregory, line 94: Please check that we only make public what we must.

garydgregory (changes requested):
Hi @ppkarwasz 
Just a few comments. Ty.

garydgregory (approved):
Hi @ppkarwasz 
Thank you for reworking the PR. Should be OK, but we can double-check with @rgoers. @rgoers ?

PR #815, opened 2022-04-07
[LOG4J2-2872] Normalizes level names

ppkarwasz:
Calling `Level.forName` with a non upper case argument, can create a custom level that can not be retrieved using `Level.toLevel`.

This PR fixes to upper case the keys in the `Level.LEVELS` map.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/Level.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/Level.java
@@ -17,7 +17,6 @@
 package org.apache.logging.log4j;
 
 import java.io.Serializable;
-import java.util.Collection;
 import java.util.Locale;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/Level.java
```diff
@@ -259,15 +258,23 @@ public String toString() {
      * @throws java.lang.IllegalArgumentException if the name is null or intValue is less than zero.
      */
     public static Level forName(final String name, final int intValue) {
-        final Level level = LEVELS.get(name);
```
garydgregory, line 262: Above we use Strings.isEmpty(name), shouldn't we use the same here?

log4j-api/src/main/java/org/apache/logging/log4j/Level.java
```diff
@@ -276,9 +283,13 @@ public static Level forName(final String name, final int intValue) {
      *
      * @param name The name of the Level.
      * @return The Level or null.
+     * @throws java.lang.IllegalArgumentException if the name is null.
      */
     public static Level getLevel(final String name) {
-        return LEVELS.get(name);
+        if (name == null) {
```
garydgregory, line 289: See comment above.

log4j-api/src/main/java/org/apache/logging/log4j/Level.java
ppkarwasz: Yes, that make sense, although the main reason for this check is to prevent an NPE in the `ConcurrentHashMap`.

PR #818, opened 2022-04-13
[LOG4J2-3475] Add missing message parameterization in RegexFilter

jjlin:


ppkarwasz:
@jjlin: this looks good to me. Can you add a unity test that fails on the currently released code?

jjlin:
@ppkarwasz unit test added

vy:
Note that `changes.xml` needs to updated too.

PR #819, opened 2022-04-13
LOG4J2-3477 JsonLayout template doesn't contain the context stack

filipc:
There is a typo in definition of context stack resolver. Should be **mdc** what stands for Mapped Diagnostic Context, was **ndc** by mistake.

### DIFF
--- a/log4j-layout-template-json/src/main/resources/JsonLayout.json
+++ b/log4j-layout-template-json/src/main/resources/JsonLayout.json
@@ -45,7 +45,7 @@
     }
   },
   "contextStack": {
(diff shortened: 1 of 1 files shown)


vy:
No, `contextStack` pointing to `ndc` is correct, though you have a point: MDC (`contextData`) is indeed missing. I have created [LOG4J2-3477](https://issues.apache.org/jira/browse/LOG4J2-3477) and fixed this in 041648a6 (`release-2.x`) and 53009e0f (`master`). Let me know if the fix doesn't address your problem. Thanks so much for the heads up! :bowing_man:

PR #821, opened 2022-04-14
LOG4J2-3476 ignore JUL ApiLogger.setLevel without error

remkop:
This PR implements the change discussed in https://issues.apache.org/jira/browse/LOG4J2-3476:

When JUL ApiLogger.setLevel is called, do not throw an UnsupportedOperationException, but instead log an error in the status logger.

### DIFF
--- a/log4j-jul/src/main/java/org/apache/logging/log4j/jul/ApiLogger.java
+++ b/log4j-jul/src/main/java/org/apache/logging/log4j/jul/ApiLogger.java
@@ -93,7 +93,7 @@ public class ApiLogger extends Logger {
     @Override
     public void setLevel(final Level newLevel) throws SecurityException {
         StatusLogger.getLogger().error("Cannot set JUL log level through log4j-api: " +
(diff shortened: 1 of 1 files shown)


log4j-jul/src/main/java/org/apache/logging/log4j/jul/ApiLogger.java
```diff
@@ -91,7 +92,8 @@ public String getName() {
 
     @Override
     public void setLevel(final Level newLevel) throws SecurityException {
-        throw new UnsupportedOperationException("Cannot set level through log4j-api");
+        StatusLogger.getLogger().error("Cannot set JUL log level through log4j-api: " +
+                "ignoring call to Logger.setLevel(" + newLevel + ")");
```
ppkarwasz, line 96: I would suggest a parameterized message, instead of string joining.

remkop: @ppkarwasz Thank you for the suggestion. Done.

ppkarwasz (approved):

PR #824, opened 2022-04-19
LOG4J2-3472 support custom WaitStrategy configuration

remkop:
This PR implements the change discussed in https://issues.apache.org/jira/browse/LOG4J2-3472:

Allow the configuration of custom `WaitStrategy` implementations by specifying a `org.apache.logging.log4j.core.async.AsyncWaitStrategyFactory` implementation.
Configuration looks similar to the below:

```xml
<Configuration>
  <AsyncWaitStrategyFactory
      class="my.package.MyWaitStrategyFactory" />
```


log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigDisruptor.java
```diff
@@ -166,10 +166,13 @@ private void notifyIntermediateProgress(final long sequence) {
     private EventFactory<Log4jEventWrapper> factory;
     private EventTranslatorTwoArg<Log4jEventWrapper, LogEvent, AsyncLoggerConfig> translator;
     private volatile boolean alreadyLoggedWarning;
+    private final AsyncWaitStrategyFactory asyncWaitStrategyFactory;
+    WaitStrategy waitStrategy; // package-protected for testing
```
garydgregory, line 171: How about a package private getter instead?

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
```diff
@@ -51,14 +53,17 @@ class AsyncLoggerDisruptor extends AbstractLifeCycle {
 
     private volatile Disruptor<RingBufferLogEvent> disruptor;
     private String contextName;
+    private final Supplier<AsyncWaitStrategyFactory> waitStrategyFactorySupplier;
 
     private boolean useThreadLocalTranslator = true;
     private long backgroundThreadId;
     private AsyncQueueFullPolicy asyncQueueFullPolicy;
     private int ringBufferSize;
+    WaitStrategy waitStrategy; // package-protected for testing
```
garydgregory, line 62: How about a package private getter instead?

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactory.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * This interface allows users to configure a custom Disruptor WaitStrategy used for
+ * Async Loggers and Async LoggerConfigs.
+ *
+ * @since 2.17.3
+ */
+public interface AsyncWaitStrategyFactory {
+    /**
+     * Returns a non-null implementation of the LMAX Disruptor's WaitStrategy interface.
+     * This WaitStrategy will be used by Log4j Async Loggers and Async LoggerConfigs.
```
garydgregory, line 31: "Returns" -> "Creates"

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DefaultAsyncWaitStrategyFactory.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Override
+    public WaitStrategy createWaitStrategy() {
+        final String strategy = PropertiesUtil.getProperties().getStringProperty(propertyName, "TIMEOUT");
+        LOGGER.trace("DefaultAsyncWaitStrategyFactory property {}={}", propertyName, strategy);
+        final String strategyUp = Strings.toRootUpperCase(strategy);
+        // String (not enum) is deliberately used here to avoid IllegalArgumentException being thrown. In case of
+        // incorrect property value, default WaitStrategy is created.
+        switch (strategyUp) {
```
garydgregory, line 32: Make an enum from these case values?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/Configuration.java
```diff
@@ -179,6 +180,16 @@ default StrSubstitutor getConfigurationStrSubstitutor() {
      */
     AsyncLoggerConfigDelegate getAsyncLoggerConfigDelegate();
 
+    /**
+     * Returns the {@code AsyncWaitStrategyFactory} defined in this Configuration;
```
garydgregory, line 184: I like Javadocs for getters to say "Gets..."

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryIncorrectConfigGlobalLoggersTest.java
```diff
@@ -0,0 +1,80 @@
...
+    public static void beforeClass() {
+        System.setProperty(Constants.LOG4J_CONTEXT_SELECTOR,
+                AsyncLoggerContextSelector.class.getName());
+        System.setProperty(ConfigurationFactory.CONFIGURATION_FILE_PROPERTY,
+                "AsyncWaitStrategyIncorrectFactoryConfigGlobalLoggerTest.xml");
+    }
+
+    @AfterClass
+    public static void afterClass() {
+        System.setProperty(Constants.LOG4J_CONTEXT_SELECTOR, Strings.EMPTY);
```
garydgregory, line 55: Also reset CONFIGURATION_FILE_PROPERTY.

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryIncorrectConfigGlobalLoggersTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+
+    private AsyncLoggerDisruptor extractAsyncLoggerDisruptor(Logger logger) throws NoSuchFieldException, IllegalAccessException {
+        assertThat("logger is AsyncLogger", logger instanceof AsyncLogger);
+        Field f = AsyncLogger.class.getDeclaredField("loggerDisruptor");
+        f.setAccessible(true);
+        AsyncLoggerDisruptor delegate = (AsyncLoggerDisruptor) f.get(logger);
```
garydgregory, line 63: Can we use a package private getter here?

log4j-core/src/test/resources/AsyncWaitStrategyFactoryConfigGlobalLoggerTest.xml
```diff
@@ -0,0 +1,17 @@
...
+  <Appenders>
+    <List name="WaitStrategyAppenderList" />
+  </Appenders>
+  
+  <Loggers>
+    <Root level="info">
+      <AppenderRef ref="WaitStrategyAppenderList"/>
+    </Root>
+  </Loggers>
+</Configuration>
```
garydgregory, line 17: Files should end with an EOL.

garydgregory (changes requested):
Hi @remkop 
I scattered some comments.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerConfigDisruptor.java
remkop: Makes sense. Done.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
remkop: Makes sense. Done.

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryIncorrectConfigGlobalLoggersTest.java
remkop: Good catch. Done.

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryIncorrectConfigGlobalLoggersTest.java
remkop: Absolutely. Done.

log4j-core/src/test/resources/AsyncWaitStrategyFactoryConfigGlobalLoggerTest.xml
remkop: Good catch. Done.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/Configuration.java
remkop: My impression is that the convention used in the JDK is to start javadoc comments with "Returns..." for methods whose main function is to return a value. So I would prefer to the current phrasing.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactory.java
remkop: Ok, I made it "Creates and returns ..."

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DefaultAsyncWaitStrategyFactory.java
remkop: (This is actually existing code, moved into a new class, but that is just an aside.) :-)
I do love enums myself but in this case it would not be useful to create one.

In case the user specifies an undefined value like "BLAHBLAH" we need to handle that by falling back to the default WaitStrategy. If we have an enum we would first need to convert the user-specified String value into an enum value, and then catch the IllegalArgumentException for the undefined value and fall back to the default.

Using an enum would give no benefit, but the resulting code would be more clumsy than the current code.
That is why the implementation comment says "String (not enum) is deliberately used here (...)".
So I prefer to leave it as is.

remkop:
@garydgregory Awesome fast feedback Gary! Super appreciated!

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DefaultAsyncWaitStrategyFactory.java
garydgregory: Ok sounds good.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/Configuration.java
garydgregory: There is no such convention in the JDK, for example in System, we see "Gets" for https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getSecurityManager-- and also "Determines" and "Returns". I just settled on "Gets".

garydgregory:
> @garydgregory Awesome fast feedback Gary! Super appreciated!

@remkop 
You're welcome! 😀👍

remkop:
@garydgregory Thanks for your replies! Are you okay with the PR as it is now? (The github review status still says "changes requested"...)

garydgregory (approved):
LGTM.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryConfig.java
```diff
@@ -0,0 +1,96 @@
...
+        public B asBuilder() {
+            return (B) this;
+        }
+    }
+
+    public AsyncWaitStrategyFactory createWaitStrategyFactory() {
+        try {
+            @SuppressWarnings("unchecked")
+            final Class<? extends AsyncWaitStrategyFactory> klass = (Class<? extends AsyncWaitStrategyFactory>) Loader.loadClass(factoryClassName);
+            return klass.newInstance();
```
carterkozak, line 89: Perhaps we should add a runtime check to ensure we're not instantiating arbitrary unexpected classes?

```java
if (AsyncWaitStrategyFactory.class.isAssignableFrom(klass)) {
    return klass.newInstance();
}
LOGGER.error("Factory '{}' is not assignable to AsyncWaitStrategyFactory", factoryClassName);
return null;
```

log4j-core/src/test/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryConfigTest.java
```diff
@@ -0,0 +1,83 @@
...
+    @Test
+    @LoggerContextSource("AsyncWaitStrategyFactoryConfigTest.xml")
+    public void testWaitStrategy(final LoggerContext context) throws Exception {
+
+        org.apache.logging.log4j.Logger logger = context.getRootLogger();
+
+        AsyncLoggerConfig loggerConfig = (AsyncLoggerConfig) ((org.apache.logging.log4j.core.Logger) logger).get();
+        AsyncLoggerConfigDisruptor delegate = (AsyncLoggerConfigDisruptor) loggerConfig.getAsyncLoggerConfigDelegate();
+        assertEquals(YieldingWaitStrategy.class, delegate.waitStrategy.getClass());
+        assertThat("waitstrategy is YieldingWaitStrategy", delegate.waitStrategy instanceof com.lmax.disruptor.YieldingWaitStrategy);
```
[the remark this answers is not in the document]
remkop: Oops yes I missed that. Fixed now. Thank you!

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryConfig.java
remkop: Good idea. Done.

remkop:
@carterkozak and @ppkarwasz Thank you for your reviews!

remkop:
Rats, it turns out that I need to add the license text to some files:

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncWaitStrategyFactoryConfig.java
log4j-core/src/main/java/org/apache/logging/log4j/core/async/DefaultAsyncWaitStrategyFactory.java

I will do that when I get home later today.

ppkarwasz (approved):
LGTM

PR #827, opened 2022-04-20
(doc) Update mongodb dependencies

rozza:
Update the mongo java driver versions listed on the site to those in the pom.xml:

<mongodb3.version>3.12.11</mongodb3.version>
<mongodb4.version>4.5.0</mongodb4.version>


rozza:
Updated again due to a recent change in #841 - @ppkarwasz could you review?

PR #836, opened 2022-05-09
[LOG4J2-1376] Define eid as String instead of int to allow for Oid according to RFC5424

mneundorfer:
As discussed in https://issues.apache.org/jira/browse/LOG4J2-1376, the enterprise ID is actually not an `int`, but an OID. Since converting directly to an OID is a non-trivial task, this PR proposes a first step in the right direction by using a `String`, which allows for RFC5424-conform behavior, since now an OID can be passed as enterprise ID.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
@@ -139,11 +139,26 @@ public class StructuredDataId implements Serializable, StringBuilderFormattable 
      * @param enterpriseNumber The enterprise number.
      * @param required The list of keys that are required for this id.
      * @param optional The list of keys that are optional for this id.
+     * @deprecated Use {@link #StructuredDataId(String, String, String[], String[])} instead.
+     */
+    @Deprecated
+    public StructuredDataId(final String name, final int enterpriseNumber, final String[] required,
+                            final String[] optional) {
+        this(name, String.valueOf(enterpriseNumber), required, optional, MAX_LENGTH);
+    }
+
+    /**
+     * A Constructor that helps conformance to RFC 5424.
+     *
+     * @param name The name portion of the id.
+     * @param enterpriseNumber The enterprise number.
+     * @param required The list of keys that are required for this id.
+     * @param optional The list of keys that are optional for this id.
      * @param maxLength The maximum length of the StructuredData Id key.
      * @since 2.9
      */
     public StructuredDataId(final String name, final String enterpriseNumber, final String[] required,
-            final String[] optional, final int maxLength) {
+                            final String[] optional, final int maxLength) {
         if (name == null) {
             throw new IllegalArgumentException("No structured id name was supplied");
         }
@@ -161,6 +176,23 @@ public class StructuredDataId implements Serializable, StringBuilderFormattable 
         }
         this.required = required;
         this.optional = optional;
+    }
+
+    /**
+     * A Constructor that helps conformance to RFC 5424.
+     *
+     * @param name The name portion of the id.
+     * @param enterpriseNumber The enterprise number.
+     * @param required The list of keys that are required for this id.
+     * @param optional The list of keys that are optional for this id.
+     * @param maxLength The maximum length of the StructuredData Id key.
+     * @since 2.9
+     * @deprecated Use {@link #StructuredDataId(String, String, String[], String[], int)} instead.
+     */
+    @Deprecated
+    public StructuredDataId(final String name, final int enterpriseNumber, final String[] required,
+            final String[] optional, final int maxLength) {
+        this(name, String.valueOf(enterpriseNumber), required, optional, maxLength);
     }
 
     /**
@@ -201,6 +233,19 @@ public class StructuredDataId implements Serializable, StringBuilderFormattable 
         }
 
         return new StructuredDataId(id, anEnterpriseNumber, req, opt);
+    }
+
+    /**
+     * Creates an id based on the current id.
+     *
+     * @param defaultId The default id to use if this StructuredDataId doesn't have a name.
+     * @param anEnterpriseNumber The enterprise number.
+     * @return a StructuredDataId.
+     * @deprecated Use {@link StructuredDataId#makeId(String, String)} instead
+     */
+    @Deprecated
+    public StructuredDataId makeId(final String defaultId, final int anEnterpriseNumber) {
+        return makeId(defaultId, String.valueOf(anEnterpriseNumber));
     }
 
     /**
@@ -245,7 +290,7 @@ public class StructuredDataId implements Serializable, StringBuilderFormattable 
      * @return true if the id uses the reserved enterprise number, false otherwise.
      */
     public boolean isReserved() {
-        return enterpriseNumber.equals(RESERVED);
+        return RESERVED.equals(enterpriseNumber);
     }
 
     @Override
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
@@ -47,14 +47,14 @@ public class StructuredDataId implements Serializable, StringBuilderFormattable 
     /**
      * Reserved enterprise number.
      */
-    public static final int RESERVED = -1;
+    public static final String RESERVED = "-1";
 
-    private static final long serialVersionUID = 9031746276396249990L;
+    private static final long serialVersionUID = -8252896346202183738L;
     private static final int MAX_LENGTH = 32;
     private static final String AT_SIGN = "@";
 
     private final String name;
-    private final int enterpriseNumber;
+    private final String enterpriseNumber;
     private final String[] required;
     private final String[] optional;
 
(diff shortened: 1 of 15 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -71,9 +71,9 @@
 public final class Rfc5424Layout extends AbstractStringLayout {
 
     /**
-     * Not a very good default - it is the Apache Software Foundation's enterprise number.
+     * The default example enterprise number from RFC5424.
      */
-    public static final int DEFAULT_ENTERPRISE_NUMBER = 18060;
+    public static final String DEFAULT_ENTERPRISE_NUMBER = "32473";
```
ppkarwasz, line 76: I totally agree, `32473` should be the default.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -610,8 +610,8 @@ public static Rfc5424Layout createLayout(
             // @formatter:off
             @PluginAttribute(value = "facility", defaultString = "LOCAL0") final Facility facility,
             @PluginAttribute("id") final String id,
-            @PluginAttribute(value = "enterpriseNumber", defaultInt = DEFAULT_ENTERPRISE_NUMBER)
-            final int enterpriseNumber,
+            @PluginAttribute(value = "enterpriseNumber", defaultString = DEFAULT_ENTERPRISE_NUMBER)
+            final String enterpriseNumber,
```
ppkarwasz, line 614: I think it is about time to deprecate this method and create a `Rfc5424Layout.Builder` class. The builder should validate the `enterpriseNumber` parameter in its `build()` method.

The signature of the method should remain the same.

log4j-core/src/test/java/org/apache/logging/log4j/core/appender/TlsSyslogAppenderTest.java
```diff
@@ -94,7 +94,7 @@ private SyslogAppender createAppender() {
         }
 
         return SyslogAppender.createAppender("localhost", PORTNUM, "SSL", sslConfiguration, 0, -1, true, "Test", true,
-            false, Facility.LOCAL0, "Audit", 18060, true, "RequestContext", null, null, includeNewLine, null,
+            false, Facility.LOCAL0, "Audit", "18060", true, "RequestContext", null, null, includeNewLine, null,
```
ppkarwasz, line 97: No need to modify the tests if backward compatibility is maintained.

ppkarwasz (changes requested):
@mneundorfer, thank you for the contribution.

I like the proposal, but you should take care to maintain backward compatibility. Can you also:

- add some tests for `Rfc5424Layout` and `SyslogAppender`,
- add an entry in `changes.xml`?

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
mneundorfer: Deprecating this constructor and introducing a Builder feels like it could be a separate MR? In case I still should give it a shot, what do you mean by "validate" the parameter? What kind o f validation would you expect there?

log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
```diff
@@ -142,15 +142,15 @@ public StructuredDataId(final String name, final int enterpriseNumber, final Str
      * @param maxLength The maximum length of the StructuredData Id key.
      * @since 2.9
      */
-    public StructuredDataId(final String name, final int enterpriseNumber, final String[] required,
+    public StructuredDataId(final String name, final String enterpriseNumber, final String[] required,
```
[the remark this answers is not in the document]
mneundorfer: I addressed this to all public signatures - kept the old one, deprecated it and added a new method with the parameter as `String`

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -634,11 +640,142 @@ public static Rfc5424Layout createLayout(
             includes = null;
         }
 
-        return new Rfc5424Layout(config, facility, id, enterpriseNumber, includeMDC, newLine, escapeNL, mdcId,
+        return new Rfc5424Layout(config, facility, id, String.valueOf(enterpriseNumber), includeMDC, newLine, escapeNL, mdcId,
                 mdcPrefix, eventPrefix, appName, msgId, excludes, includes, required, StandardCharsets.UTF_8,
                 exceptionPattern, useTlsMessageFormat, loggerFields);
     }
 
+    public static class Rfc5424LayoutBuilder {
```
mneundorfer, line 648: Please take a look if this Builder meets your expectations

log4j-api/src/main/java/org/apache/logging/log4j/message/StructuredDataId.java
```diff
@@ -139,11 +139,30 @@ public StructuredDataId(final String name, final String enterpriseNumber, final
...
+                            final String[] optional) {
+        this(name, String.valueOf(enterpriseNumber), required, optional, MAX_LENGTH);
+    }
+
+    /**
+     * A Constructor that helps conformance to RFC 5424.
+     *
+     * Deprecated:
+     * enterpriseNumber should be a String, use StructuredDataId(final String name, final String enterpriseNumber, final String[] required,
+     * final String[] optional, final int maxLength)
```
ppkarwasz, line 155: The "deprecated" comment shouldn't be in the Javadoc of the other constructor?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
```diff
@@ -226,6 +242,11 @@ public B setEnterpriseNumber(final String enterpriseNumber) {
             return asBuilder();
         }
 
+        public B setEnterpriseNumber(final int enterpriseNumber) {
```
ppkarwasz, line 245: `@Deprecated`

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
ppkarwasz: Looks good to me.

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -634,11 +640,142 @@ public static Rfc5424Layout createLayout(
...
+            return this;
+        }
+
+        public Rfc5424Layout build() {
+            if (includes != null && excludes != null) {
+                LOGGER.error("mdcIncludes and mdcExcludes are mutually exclusive. Includes wil be ignored");
+                includes = null;
+            }
+
+            if (ein != null && !ENTERPRISE_ID_PATTERN.matcher(ein).find()) {
```
ppkarwasz, line 770: I would use `matches()` instead of `find()`, since the entire enterprise id must match the regular expression.

Samsung72: > I would use `matches()` instead of `find()`, since the entire enterprise id must match the regular expression.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/SyslogAppender.java
mneundorfer: Thanks, added it

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
mneundorfer: Very valid, I changed it

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
rgoers: Note that when implementing the builder the constructor it calls should be private.

log4j-core/src/test/java/org/apache/logging/log4j/core/appender/TlsSyslogAppenderTest.java
rgoers: You mean - existing tests should not be modified. New ones should be created to test the new functionality.

rgoers (approved):

ppkarwasz (approved):
Looks good to me, thank you for your contribution.

Can you rebase the PR so we can merge it?

log4j-core/src/test/java/org/apache/logging/log4j/core/layout/Rfc5424LayoutTest.java
```diff
@@ -527,4 +532,61 @@ public void testParameterizedMessage() {
...
+    @Test
+    void testLayoutBuilder() {
+        for (final Appender appender : root.getAppenders().values()) {
+            root.removeAppender(appender);
+        }
+
+        final AbstractStringLayout layout = new Rfc5424Layout.Rfc5424LayoutBuilder()
+                .setFacility(Facility.LOCAL0)
+                .setId("Event")
+                .setEin("36aaa92")
```
ppkarwasz, line 545: Probably a typo, which will cause `layout` to be `null`.

mneundorfer: Indeed! I fixed this and rebased the branch

PR #850, opened 2022-05-23
Increase to 50 dependabot proposals

ppkarwasz:
We have 38 upgradeable dependencies. In order to limit the annoyance from Dependabot PRs, I propose to limit the number of `test` scope upgrade proposals to 10 and to give a larger number (50) for the remaining ones.

Another solution would be to prioritize the number of upgrade proposals for `log4j-api` and `log4j-core` and limit those for the remaining modules.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,9 +16,17 @@
 #
 version: 2
 updates:
+# No more than 10 PRs for dependencies with a `test` scope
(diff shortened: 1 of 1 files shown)


vy:
LGTM. I am against using `dependabot` only for `-api` and `-core`. I do my best to take care `dependabot` PRs for `-jackson-` and `-layout-template-json`.

PR #916, opened 2022-06-05
[LOG4J2-2921] Parallel execution for JUnit5 tests

ppkarwasz:
Unit tests in `release-2.x` are executed sequentially in per-test JVMs. While parallel execution in a single JVM requires a lot of work, this PR makes a step in this direction: the tests based on JUnit5 in `log4j-core` (about 50% of the tests) are executed in per-test JVMs, but in parallel.

This requires only a private/temporary directory for each test to store its logs (@PrivateDir extension), so that tests using the same file names don't clash.

As a preliminary step to phase 2 (running multiple JVMs in parallel and reusing them for multiple tests), this PR also introduces a thread-bound property source/lookup that can be retrieved with the @TestProperties annotation (or @SetTestProperty).

### DIFF
--- a/log4j-core/pom.xml
+++ b/log4j-core/pom.xml
@@ -444,6 +444,30 @@
             <org.apache.activemq.SERIALIZABLE_PACKAGES>*</org.apache.activemq.SERIALIZABLE_PACKAGES>
           </systemPropertyVariables>
         </configuration>
+        <executions>
+            <execution>
+                <id>default-test</id>
+                <phase>test</phase>
+                <goals>
+                  <goal>test</goal>
+                </goals>
+                <configuration>
+                   <includeJUnit5Engines>junit-jupiter</includeJUnit5Engines>
+                   <forkCount>1C</forkCount>
+                   <runOrder>random</runOrder>
+                </configuration>
+            </execution>
+            <execution>
+                <id>junit4-test</id>
+                <phase>test</phase>
+                <goals>
+                  <goal>test</goal>
+                </goals>
+                <configuration>
+                    <excludeJUnit5Engines>junit-jupiter</excludeJUnit5Engines>
+                </configuration>
+            </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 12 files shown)


jvz:
This looks like a great approach! I like that this can enable more parallelism than the equivalent system property approach since those are global. I also like that this is such a small change compared to modifying every existing test. While I've managed to get the API and plugins modules to work in parallel+concurrency (single process multi threaded option rather than forking), what I've been doing there won't scale to core tests without additional changes such as in this PR. As such, this PR most likely applies to master with minimal changes, too!

log4j-core/pom.xml
```diff
@@ -444,6 +444,30 @@
...
+            <execution>
+                <id>default-test</id>
+                <phase>test</phase>
+                <goals>
+                  <goal>test</goal>
+                </goals>
+                <configuration>
+                   <includeJUnit5Engines>junit-jupiter</includeJUnit5Engines>
+                   <forkCount>1C</forkCount>
+                   <runOrder>random</runOrder>
```
jvz, line 457: I've taken to using `<runOrder>balanced</runOrder>` when parallelizing tests. I did set them to random recently in master because I was sick of seeing the same slow tests at the beginning and wanted to change things up. :)

ppkarwasz: I set the run order to random, to empirically check if two test classes do not use the same log files. Hopefully if they do, once in a while we will have some meaningful test failures.

jvz: Ah, that's another goal of mine whenever I set the ordering to random. I'm trying to see if there are any test interferences that would otherwise have gone unnoticed locally.

ppkarwasz:
@rgoers,
Can I merge this before the 2.18.0 release or should I wait to be safe?

PRINSIS:
Duplicate of #Open
[[LOG4J2-2921] Parallel execution for JUnit5 tests](https://github.com/apache/logging-log4j2/pull/916#)
#916
[ppkarwasz](https://github.com/ppkarwasz) wants to merge 1 commit into [apache:release-2.x](https://github.com/apache/logging-log4j2) from [ppkarwasz:partial-parallel-tests](https://github.com/ppkarwasz/logging-log4j2/tree/partial-parallel-tests)

PR #924, opened 2022-06-09
[LOG4J2-3534] Fix LevelRangeFilterBuilder to align with log4j1's behavior

yueki1993:
https://issues.apache.org/jira/browse/LOG4J2-3534


log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
```diff
@@ -326,4 +326,9 @@ public void testEnhancedRollingFileAppender() throws Exception {
         }
     }
 
+    @Override
```
yueki1993, line 329: Should I add test in `Log4j1ConfigurationFactoryTest`, too?
I don't understand the difference of `PropertiesConfigurationTest` and it.

log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
```diff
@@ -129,6 +129,28 @@ public void testConsoleAppenderLevelRangeFilter() throws Exception {
...
+            assertEquals(0, legacyAppender.getEvents().size());
+            // denied???
+            logger.warn("WARN");
+            assertEquals(0, legacyAppender.getEvents().size());
+            // denied???
+            logger.error("ERROR");
+            assertEquals(0, legacyAppender.getEvents().size());
+            // denied???
+            logger.fatal("FATAL");
+            assertEquals(0, legacyAppender.getEvents().size());
```
yueki1993: this test fails in ba48d2b, so I took a look here.
Under this setting, this rejects all output logs.

I believe we expect here we get all messages accepted because we set `levelMin=ALL`.
(see: https://issues.apache.org/jira/browse/LOG4J2-2315)

log4j-1.2-api/src/main/java/org/apache/log4j/builders/filter/LevelRangeFilterBuilder.java
```diff
@@ -83,19 +86,23 @@ public Filter parse(PropertiesConfiguration config) {
...
-            max = OptionConverter.toLevel(levelMax, org.apache.log4j.Level.FATAL).getVersion2Level();
+            max = OptionConverter.toLevel(levelMax, org.apache.log4j.Level.OFF).getVersion2Level();
         }
         if (levelMin != null) {
-            min = OptionConverter.toLevel(levelMin, org.apache.log4j.Level.DEBUG).getVersion2Level();
+            min = OptionConverter.toLevel(levelMin, org.apache.log4j.Level.ALL).getVersion2Level();
         }
         org.apache.logging.log4j.core.Filter.Result onMatch = acceptOnMatch
                 ? org.apache.logging.log4j.core.Filter.Result.ACCEPT
                 : org.apache.logging.log4j.core.Filter.Result.NEUTRAL;
```
yueki1993, line 100: Is it better to output warning log if `max <  min` here?
Current released implementation should work if we swap `levelMin` and `levelMax` in `log4j.properties`.
This is degradation for people who configure log4j like that.

log4j-1.2-api/src/test/java/org/apache/log4j/config/PropertiesConfigurationTest.java
ppkarwasz: `Log4j1ConfigurationFactory` is an old factory that does not use builders. No need to add it there.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/filter/LevelRangeFilterBuilder.java
ppkarwasz: There was no warning in Log4j 1.x, so I wouldn't add it here. You can add a warning in the Log4j2 factory `LevelRangeFilter#createFilter` (using the `StatusLogger`).

ppkarwasz (approved):
@yueki1993,

Thank You for your contribution. This looks good to me, just remember to add an entry in the `src/changes/changes.xml` file.

log4j-1.2-api/src/main/java/org/apache/log4j/builders/filter/LevelRangeFilterBuilder.java
yueki1993: OK, I won't add warning log in Log4j2 factory too because if someone misconfigures LogLevelRangeFilter, no logs will be output - they should be able to notice misconfiguration easily.

yueki1993:
Hello @ppkarwasz,
Thank you for your review. I pushed `src/changes/changes.xml`.
Is it better to squash my commits?

PR #933, opened 2022-06-15
LOG4J2-3538 Added support for 24 colors in highlighting

PashaTurok:
My first pull request. Please, check.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
@@ -325,7 +325,7 @@ public enum AnsiEscape {
                 }
                 first = false;
                 String hexColor = null;
-                final String trimmedName = name.trim();
+                final String trimmedName = name.trim().toUpperCase(Locale.ENGLISH);
                 if (trimmedName.startsWith("#")) {
                     sb.append("38");
                     sb.append(SEPARATOR.getCode());
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
@@ -22,6 +22,7 @@ import java.util.Locale;
 import java.util.Map;
 
 import org.apache.logging.log4j.core.util.Patterns;
+import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.util.EnglishEnums;
 import org.apache.logging.log4j.util.Strings;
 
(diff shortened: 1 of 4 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
```diff
@@ -320,12 +321,43 @@ public static String createSequence(final String... names) {
...
+                    hexColor = trimmedName.substring(3);
+                }
+                if (hexColor != null) {
+                    Color color = Color.decode(hexColor);
+                    sb.append(color.getRed());
+                    sb.append(SEPARATOR.getCode());
+                    sb.append(color.getGreen());
+                    sb.append(SEPARATOR.getCode());
+                    sb.append(color.getBlue());
+                    //no separator at the end
```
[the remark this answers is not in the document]
PashaTurok: @ppkarwasz Please, check.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
```diff
@@ -320,12 +320,42 @@ public static String createSequence(final String... names) {
...
         for (final String name : names) {
             try {
-                final AnsiEscape escape = EnglishEnums.valueOf(AnsiEscape.class, name.trim());
                 if (!first) {
                     sb.append(AnsiEscape.SEPARATOR.getCode());
                 }
                 first = false;
-                sb.append(escape.getCode());
+                String hexColor = null;
+                final String trimmedName = name.trim();
```
ppkarwasz, line 328:
```suggestion
                final String trimmedName = name.trim().toUpperCase(Locale.ENGLISH);
```

since the other keywords are also matched in a case insensitive way.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/AnsiEscape.java
```diff
@@ -320,12 +320,42 @@ public static String createSequence(final String... names) {
...
+                    sb.append(SEPARATOR.getCode());
+                    sb.append(Integer.valueOf(hexColor.substring(5, 7), 16));//b
+                    //no separator at the end
+                } else {
+                    final AnsiEscape escape = EnglishEnums.valueOf(AnsiEscape.class, trimmedName);
+                    sb.append(escape.getCode());
+                }
             } catch (final Exception ex) {
                 // Ignore the error.
             }
```
ppkarwasz, line 361:
```suggestion
            } catch (final Exception ex) {
                StatusLogger.getLogger().warn("The style attribute {} is incorrect.", name, ex);
            }
```

The current convention is to log configuration errors.

ppkarwasz (approved):
Apart the two remarks below, this looks good to me.

If other committers don't have remarks I'll merge it in a couple of days, thank you.

PashaTurok:
@ppkarwasz Could you take a look why it failed? It seems the problem is that we added event logging for exception.

PR #947, opened 2022-06-29
add unit tests for ConsoleAppender direct mode

sullis:


sullis:
WDYT? @rgoers

sullis:
what do you think of this PR? @rgoers

ppkarwasz:
Hi @sullis,

I admit I don't really see the point of just adding a test case, unless it solves a problem. Besides your test does not check if the appender uses a `FileOutputStream` attached to the file descriptor 1 and 2, it just shows that `System.out` and `System.err` are not used.

PR #948, opened 2022-06-29
run benchmark.yml workflow

sullis:


rgoers (changes requested):
I don't understand the purpose of this. Why do we want to run on a push or pull and do it against a non-asf repo?

sullis:
> I don't understand the purpose of this. Why do we want to run on a push or pull and do it against a non-asf repo?

Oops. Sorry.  I pushed this to the wrong repo.  I'll close this PR.

PR #955, opened 2022-07-11
fix SystemPropertyArbiter

DongjianPeng:
It looks like this variable was incorrectly assigned.


rgoers:
This looks good but we need a Jira issue for it.

DongjianPeng:
> This looks good but we need a Jira issue for it.

Ok, here it is: https://issues.apache.org/jira/browse/LOG4J2-3550

PR #959, opened 2022-07-12
[LOG4J2-3548] Don't replace null password with default if provided

kristofarkas:
Password-less key stores require passing null as password. A recent change prevented that, replacing null password with a default value. This PR reverts that change, allowing null password to be used again.

https://issues.apache.org/jira/browse/LOG4J2-3548


ppkarwasz:
Hi @kristofarkas,

I would appreciate if you could add some tests to prevent a regression in the future. I imagine we would like to check these three cases:

 * a PKCS12 keystore with a non-empty password,
 * a PKCS12 keystore with an empty password,
 * a keystore, which requires both password and location to be `null` (e.g. `Windows-MY` and `Windows-ROOT`).

While your PR fixes the usage of `Windows-MY/ROOT` inside a Log4j2 configuration file, there is another case that you could consider fixing: the `SslConfigurationFactory` contains an `SslConfiguration` used to download a Log4j2 configuration file itself over HTTPS. Unless I am mistaken, this configuration will not work with `Windows-MY`, since it requires the property "log4j2.keyStoreLocation" to be non-null.

ppkarwasz (approved):

PR #960, opened 2022-07-13
log4j2 inquiry

philly706869:
Can you please allow me to set the output format (pattern) for each level for better readability?

I used a translator so it might be a bit weird


ppkarwasz:
@philly706869: for questions and feature requests, please use the [`log4j-user` mailing list](https://logging.apache.org/log4j/2.x/mail-lists.html).

PR #970, opened 2022-07-21
[LOG4J2-3360] Pin codeql-action/init to a full length commit SHA

marcwrobel:
As part of LOG4J2-3360 GitHub Action versions were replaced by commit checksums, except for github/codeql-action/init. This was probably an oversight.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -49,7 +49,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v2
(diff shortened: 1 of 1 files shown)


vy:
Thanks for the heads up @marcwrobel! Fixed.

marcwrobel:
You are welcome @vy. I think the version you put in your comment is not correct. You put 2.10.1, which is the CodeQL CLI version, not the github/codeql-action version. Last version is 2.1.16, as you can see on https://github.com/github/codeql-action/tags.

vy:
:man_facepalming: Doh! Is it okay now?

marcwrobel:
Looks good to me 👍

PR #974, opened 2022-07-23
LOG4J2-3560: Logger$PrivateConfig.filter(Level, Marker, String) varargs does not allocate

schlosna:
Fixes [LOG4J2-3560](https://issues.apache.org/jira/browse/LOG4J2-3560)

While reviewing some JFR profiles, I notice significant allocations of `Object[]` due to use of some `isEnabled` conditions that ends up using empty varargs invoking `org.apache.logging.log4j.core.Filter#filter(Logger, Level, Marker, String, Object...)`, for example the following call trace:

```
at org.apache.logging.log4j.core.Logger$PrivateConfig.filter(Level, Marker, String)
at org.apache.logging.log4j.core.Logger.isEnabled(Level, Marker, String)
```

We can avoid allocation on this hot path when invoking `Filter#filter(Logger, Level, Marker, String, Object...)` by reusing an empty object array for the varargs.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
@@ -53,9 +53,6 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
 
     private static final long serialVersionUID = 1L;
 
-    // Used to avoid allocations for empty varargs
-    private static final Object[] EMPTY_VARARGS = new Object[] {};
-
     /**
      * Config should be consistent across threads.
      */
@@ -436,7 +433,7 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
         boolean filter(final Level level, final Marker marker, final String msg) {
             final Filter filter = config.getFilter();
             if (filter != null) {
-                final Filter.Result r = filter.filter(logger, level, marker, msg, EMPTY_VARARGS);
+                final Filter.Result r = filter.filter(logger, level, marker, msg, (Throwable) null);
                 if (r != Filter.Result.NEUTRAL) {
                     return r == Filter.Result.ACCEPT;
                 }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
@@ -19,6 +19,7 @@ package org.apache.logging.log4j.core;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -172,7 +173,7 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
 
     @Override
     public boolean isEnabled(final Level level, final Marker marker, final String message) {
-        return privateConfig.filter(level, marker, message, (Throwable) null);
+        return privateConfig.filter(level, marker, message);
     }
 
     @Override
@@ -299,7 +300,7 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
     public Iterator<Filter> getFilters() {
         final Filter filter = privateConfig.loggerConfig.getFilter();
         if (filter == null) {
-            return new ArrayList<Filter>().iterator();
+            return Collections.emptyIterator();
         } else if (filter instanceof CompositeFilter) {
             return ((CompositeFilter) filter).iterator();
         } else {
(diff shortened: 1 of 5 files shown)


schlosna:
@carterkozak I'd be interested in your thoughts on this

log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
```diff
@@ -53,6 +53,9 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
 
     private static final long serialVersionUID = 1L;
 
+    // Used to avoid allocations for empty varargs
+    private static final Object[] EMPTY_VARARGS = new Object[0];
```
garydgregory, line 57: Use can can use the short cut {} on the right-hand side.

schlosna: Thanks for the quick review! updated

ppkarwasz:
I wonder if the `Filter` interface isn't just missing a `filter(Logger, Level, Marker, String, Throwable)` method.

Alternatively we can call the `filter(Logger, Level, Marker, Object, Throwable)` method instead of the one with the `Object` array.

ppkarwasz (approved):
Looks good to me.

A test would be nice, but testing for lack of garbage is difficult. Can you add an entry in `changes.xml`?

carterkozak (approved):
Thanks!

garydgregory:
We have tests that explicitly check for generated garbage already, so a similar test would be nice.

schlosna:
I added a unit test that covers the filter code path, but I'm not quite sure how to hook into the garbage generation tests. Do `GcFreeAsynchronousLoggingTest` and `GcFreeSynchronousLoggingTest` automatically pick these up or do tests/test class need to be tagged in some fashion? Do I need to use `GcFreeLoggingTestUtil` directly?

carterkozak:
I'll take care of getting this integrated, and verify the efficacy of the provided test. I should have time later today -- thanks @schlosna :-)

I agree with @ppkarwasz that we may be missing an overload on the filter interface, which should be easier to add now that we support java 8 compilation target with default methods. I'll look into adding this API as a separate follow-up to this bugfix.

schlosna:
Thanks @carterkozak !

src/changes/changes.xml
```diff
@@ -33,6 +33,9 @@
       <action issue="LOG4J2-3550" dev="rgoers" type="fix" due-to="DongjianPeng">
         SystemPropertyAribiter was assigning the value as the name.
       </action>
+      <action issue="LOG4J2-3560" dev="schlosna" type="fix">
```
carterkozak, line 36: I'll take care of this update prior to merging, the dev attribute reflects the committer who has signed off on the change while `due-to` provides attribution for the work if it wasn't generated by the committer.
```suggestion
      <action issue="LOG4J2-3560" dev="ckozak" type="fix" due-to="David Schlosnagle">
```

carterkozak:
I've pushed a couple minor edits, including an update to `GcFreeLoggingTestUtil` which I've verified on `release-2.x` catches this bug, and applied to this branch passes as expected.

Test failure on release-2.x matches our expectations:
```
org.opentest4j.AssertionFailedError: pattern mismatch at line 1: FATAL o.a.l.l.c.GcFreeSynchronousLoggingTest [main] value1 {aKey=value1, key2I just allocated the object [Ljava.lang.Object;@32b260fa of type java/lang/Object whose size is 16 ==> 
Expected :true
Actual   :false
	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:210)
	at org.apache.logging.log4j.core.GcFreeLoggingTestUtil.lambda$runTest$1(GcFreeLoggingTestUtil.java:241)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:647)
	at org.apache.logging.log4j.core.GcFreeLoggingTestUtil.runTest(GcFreeLoggingTestUtil.java:226)
	at org.apache.logging.log4j.core.GcFreeSynchronousLoggingTest.testNoAllocationDuringSteadyStateLogging(GcFreeSynchronousLoggingTest.java:33)
...etc
```

src/changes/changes.xml
schlosna: Sounds good!

schlosna:
Excellent, GC test updates look good to me, thanks for adding @carterkozak for pushing this through.

jamesjames1122 (commented):
@

PR #1006, opened 2022-08-20
LOG4J2-3579 Add privileged execution for ServiceLoading

boris-unckel:
Fixes https://issues.apache.org/jira/browse/LOG4J2-3579

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -20,11 +20,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
 import java.nio.charset.Charset;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


ppkarwasz:
Hi @boris-unckel, thank You for noticing and solving this `SecurityManager` problem.

The idea behind the `ServiceLoaderUtil` is to work around the caller sensitivity of the `ServiceLoader` methods by using the `MethodHandles.Lookup` parameter to create a lambda in the caller's domain.

Since `AccessController#doPrivileged` is another caller sensitive method, I think that your PR should do the same and call it through `MethodHandles.Lookup`.

boris-unckel:
@ppkarwasz 

> Since `AccessController#doPrivileged` is another caller sensitive method, I think that your PR should do the same and call it through `MethodHandles.Lookup`.

Thanks for your feedback. I have changed this accordingly.

I have serious doubts the current fix is a good idea in general. Obviously addressing Java 8 and Java 11 consumers the SecurityManager is a must have, causing a regression between 2.17.2 and 2.18. I would expect a major version jump indicating a "no support" situation including explicit documentation. But:

The class is available to all the user land: It makes it possible to run code (service loading) with privileged rights (in terms of WildFly: server side - "AllPermission"). Now "all world" can load any Service (even remote?). I have not tested it yet, but I would assume it is better to fix it at all current usage points of the public methods. Then only fixed Services can be loaded privileged, but not any. (For WildFly it would be only for classes visible for the log4j2 module loader).

boris-unckel:
@ppkarwasz I have prepared a caller based solution: https://github.com/apache/logging-log4j2/pull/1007 I think it's better to avoid a general purpose privileged execution utility.

ppkarwasz:
@boris-unckel, I totally share your concerns about the usage of `AccessController.doPrivileged`, but I am not sure if #1007 solves them.

If we were to introduce additional `AccessController.doPrivileged`, we need to audit our code to make sure the services we obtain through privileged calls can not be abused by unprivileged callers. E.g. I think that the `SystemPropertiesPropertySource` class can be used by anyone to obtain the value of privileged system properties, e.g. "log4j2.keystorePassword".

I started a thread (cf. [https://lists.apache.org/thread/2j0dyccsm0ddjy60ydd3f19flotjfjxd](https://lists.apache.org/thread/2j0dyccsm0ddjy60ydd3f19flotjfjxd)) on our mailing list to discuss the problem. Personally I almost never use a `SecurityManager` in my day job and I suppose this also applies to most of the team: it would be nice if you could expose your ideas on the mailing list.

jamezp:
I'm a +1 for the second (#1007) approach compared to this. It is more of a pain to ensure it's always used in a privileged action, but it's definitely more secure. There may be cases where an implementation or user might *not* want it privileged.

WRT the `SecurityManager` as a whole. Yes it's deprecated in Java 17 and proposed for removal. However, there are some users that are required to use a `SecurityManager` for compliance reasons.

boris-unckel:
I'm closing this PR for the reasons described here and as discussed on the mailing list. Please check https://github.com/apache/logging-log4j2/pull/1007 for the ongoing review.

PR #1007, opened 2022-08-21
LOG4J2-3579 Add privileged execution for ServiceLoading

boris-unckel:
Add privileged actions on caller side

Fixes https://issues.apache.org/jira/browse/LOG4J2-3579


jamezp:
I prefer this approach as opposed to #1006.

log4j-api/src/main/java/org/apache/logging/log4j/util/EnvironmentPropertySource.java
```diff
@@ -38,21 +38,11 @@ public int getPriority() {
...
 	public void forEach(final BiConsumer<String, String> action) {
 		final Map<String, String> getenv;
-		try {
-			getenv = System.getenv();
-		} catch (final SecurityException e) {
-			logException(e);
-			return;
-		}
+		getenv = System.getenv();
+
```
ppkarwasz, line 45: `forEach` is an optional method for a property source: Log4j2 does **not** need access to all environment variables, but if it does, it can cache the variable values.

I would leave the `catch`, so that Log4j2 users do not need to give the `getenv.*` permission to `log4j-api`.

log4j-api/src/main/java/org/apache/logging/log4j/util/EnvironmentPropertySource.java
```diff
@@ -75,32 +65,17 @@ public CharSequence getNormalForm(final Iterable<? extends CharSequence> tokens)
 
 	@Override
 	public Collection<String> getPropertyNames() {
-		try {
 			return System.getenv().keySet();
-		} catch (final SecurityException e) {
-			logException(e);
-			return PropertySource.super.getPropertyNames();
-		}
```
ppkarwasz, line 83: Same as above: the `getPropertyNames` is used for caching, but is not necessary for Log4j2.

I would leave the `catch`.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -446,22 +447,36 @@ private static class Environment {
...
+            PropertyFilePropertySource sysProps = new PropertyFilePropertySource(LOG4J_SYSTEM_PROPERTIES_FILE_NAME,
+                false);
+            sysProps.forEach((key, value) -> {
+                if (System.getProperty(key) == null) {
+                    System.setProperty(key, value);
+                }
+            });
             sources.add(propertySource);
+            sources.add(new SystemPropertiesPropertySource());
+            sources.add(new EnvironmentPropertySource());
```
ppkarwasz, line 459: I wouldn't hard code these property sources. Isn't looking up services in the same protection domain always allowed?

log4j-api/src/main/java/org/apache/logging/log4j/util/SystemPropertiesPropertySource.java
```diff
@@ -39,17 +39,8 @@ public int getPriority() {
...
-		try {
-			properties = System.getProperties();
-		} catch (final SecurityException e) {
-			// (1) There is no status logger.
-			// (2) LowLevelLogUtil also consults system properties ("line.separator") to
-			// open a BufferedWriter, so this may fail as well. Just having a hard reference
-			// in this code to LowLevelLogUtil would cause a problem.
-			// (3) We could log to System.err (nah) or just be quiet as we do now.
-			return;
-		}
```
ppkarwasz, line 52: This is not fatal, the catch&ignore is fine.

log4j-api/src/main/java/org/apache/logging/log4j/util/SystemPropertiesPropertySource.java
```diff
@@ -71,20 +62,12 @@ public CharSequence getNormalForm(final Iterable<? extends CharSequence> tokens)
 
 	@Override
 	public Collection<String> getPropertyNames() {
-		try {
 			return System.getProperties().stringPropertyNames();
-		} catch (final SecurityException e) {
-			return PropertySource.super.getPropertyNames();
-		}
```
ppkarwasz, line 78: Also non fatal.

ppkarwasz (changes requested):

boris-unckel:
Closed, solution is done with https://github.com/apache/logging-log4j2/pull/1008

PR #1008, opened 2022-08-21
LOG4J2-3579 Calls ServiceLoader with caller's security context

ppkarwasz:
When running with a security manager, `ServiceLoaderUtil` will call `ServiceLoader` with the privileges of the caller.

This probably requires further testing to be sure it works with both JPMS and a SecurityManager.

### DIFF
--- a/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
+++ b/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/ServiceLoaderUtil.java
@@ -17,10 +17,13 @@
 
 package org.apache.logging.log4j.util;
 
+import java.lang.invoke.CallSite;
+import java.lang.invoke.LambdaMetafactory;
 import java.lang.invoke.MethodHandle;
 import java.lang.invoke.MethodHandles.Lookup;
 import java.lang.invoke.MethodType;
-import java.util.Arrays;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.ServiceConfigurationError;
@@ -31,9 +34,6 @@ import java.util.stream.Stream;
 import org.apache.logging.log4j.status.StatusLogger;
 
 public final class ServiceLoaderUtil {
-
-    private static final MethodType LOAD_CLASS_CLASSLOADER = MethodType.methodType(ServiceLoader.class, Class.class,
-            ClassLoader.class);
 
     private ServiceLoaderUtil() {
     }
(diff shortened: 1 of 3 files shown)


ppkarwasz:
@boris-unckel, does this change solve your problem?

The `AccessControl.doPrivileged` is called from the caller's domain using a `PrivilegedAction` also in the caller's domain.

As a result all calls to `ServiceLoaderUtil.loadServices` from inside `log4j-api` and `log4j-core` succeed, but a call from inside the Wildfly test code does not (as expected).

This prevents from using this tool from an unprivileged domain.

rgoers:
The only thing I don't like about this is its heavy reliance on SecurityManager stuff. This is fine in release-2.x but in master I'd want to skip all this if there is no SecurityManager enabled.

boris-unckel:
@jamezp My reproducer shows this works. Could you have a look at it, please?

jamezp:
My only concern with this would be is it exposes the privileged action. Essentially anyone that uses this can get around the security manager using the protection domain from log4j-api.

ppkarwasz:
@jamezp, can you provide more details? My understanding is that if `foo.jar` calls `ServiceLoaderUtil.loadServices` the call will be made as if `foo.jar` called `AccessController.doPrivileged(() -> ServiceLoader.load(...)` directly. My tests seem to confirm it: the `AccessControlContext` cached by the `ServiceLoader` contains just the domain of the caller, not `log4j-api.jar`.

jamezp:
> @jamezp, can you provide more details? My understanding is that if `foo.jar` calls `ServiceLoaderUtil.loadServices` the call will be made as if `foo.jar` called `AccessController.doPrivileged(() -> ServiceLoader.load(...)` directly. My tests seem to confirm it: the `AccessControlContext` cached by the `ServiceLoader` contains just the domain of the caller, not `log4j-api.jar`.

@ppkarwasz Yes, that is correct that the `AccessControlContext` will be that of `foo.jar`. I could just be overly paranoid really :)

ppkarwasz:
@rgoers, I don't see the heavy reliance on `SecurityManager`: I do use a `PrivilegedAction`, but just as synonym for `Supplier`. The part the really relies on the `SecurityManager` is 3 lines long.

Nevertheless, should I cherry-pick this to `master` or we do not support the `SecurityManager` in 3.x?

jvz:
> Nevertheless, should I cherry-pick this to `master` or we do not support the `SecurityManager` in 3.x?

Considering Java 11 still has that API, go for it. We can consider removal of SecurityManager stuff when and if it becomes a compatibility issue going forward.

PR #1019, opened 2022-08-26
Update BUILDING.md

wu-yy:


vy:
Thanks for the fix @wu-yy!

PR #1022, opened 2022-08-26
SLF4J2

terrettaz:
- Bump SLF4J 2.0.0
  - Avoid error reported in [LOG4J2-3139](https://issues.apache.org/jira/browse/LOG4J2-3139)
  - Fix test about [SLF4J-421](https://jira.qos.ch/browse/SLF4J-421)

### DIFF
--- a/log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
+++ b/log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
@@ -16,7 +16,9 @@
  */
 package org.apache.logging.slf4j;
 
+import java.util.ArrayDeque;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.Map;
 
 import org.apache.logging.log4j.ThreadContext;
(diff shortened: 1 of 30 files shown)


ppkarwasz:
Hi @terrettaz,

Thank you for you contribution. Since the SLF4J binding should go in the `log4j-slf4j20-impl` module, I have merged the `slf4j-2.0` branch into `release-2.0`. Can you rewrite your contribution in the new Maven module?

Regarding the stack-valued MDC implementation, there is [an ongoing thread](https://lists.apache.org/thread/pzmqt2ypjgcttj93hconqgg16thyf81o) on the `dev` mailing list regarding the proper course of action. This is something that requires a consensus among devs and you are more than welcome to express your opinion.

rgoers:
I have two concerns with this:

1. Because this upgrades log4j-slf4j18-impl to SLF4J 2.0 anyone using SLF4J 1.8 might have problems. That said, 1.8 never had a GA release. So I believe we should release the 2.0 module and remove the 1.8 module.
2. From what I can tell the NDC queue is private to Log4jMdcAdapter. Thus log4j has no knowledge of it at all, which means it might as well have been implemented as a no-op.

terrettaz:
Hi, thanks for the comments!
I agree this is much better to use the new module and probably delete the slf4j18, I can include the deletion in this PR.

I've checked the code of slf4j-reload4j or logback and I didn't really see the use of these new methods added in MDC. So in order to benefit from this new SLF4J api, I added the option to wire the top of the stack to the MDC of log4j. I added a unit-test to emphase the idea.
I also prepare and option to activate this behaviour but didn't find the way to get options from the config. Any suggestion from the core developers on how to properly configure log4j-slf4j20-impl

log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
```diff
@@ -27,57 +29,229 @@
...
+        public void clearDequeByKey(String key) {
+            threadLocalMapOfDeques.clearDequeByKey(key);
+        }
+    }
+
+    /**
+     * A simple implementation of ThreadLocal backed Map containing values of type
+     * Deque<String>. This class is inspired from SLF4J version 2.0.0
+     */
+    private static final class ThreadLocalMapOfStacks {
```
ppkarwasz, line 192: I wouldn't copy this class, but just use the one Ceki provides in `slf4j-api`.

log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
```diff
@@ -27,57 +29,229 @@
...
+        @Override
+        public void clearDequeByKey(String key) {
+            // NOP
+        }
+    }
+
+    /**
+     * Wires the top of the stack to the MDC current value
+     */
+    private static final class ExtendedMDCAdaptor extends StandardMDCAdaptor {
```
ppkarwasz, line 151: Personally I am fine with this implementation being the only one.

log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
```diff
@@ -27,57 +29,229 @@
...
+     * A simple implementation of ThreadLocal backed Map containing values of type
+     * Deque<String>. This class is inspired from SLF4J version 2.0.0
+     */
+    private static final class ThreadLocalMapOfStacks {
+
+        private final ThreadLocal<Map<String, Deque<String>>> tlMapOfStacks = new ThreadLocal<>();
+
+        void pushByKey(String key, String value) {
+            if (key == null)
+                return;
```
ppkarwasz, line 198: If the `key` is `null` I would delegate to `ThreadContext.push` to take advantage of our NDC.

log4j-slf4j20-impl/src/main/java/org/apache/logging/slf4j/Log4jMDCAdapter.java
```diff
@@ -27,57 +29,229 @@
...
+                map = new HashMap<>();
+                tlMapOfStacks.set(map);
+            }
+
+            Deque<String> deque = map.get(key);
+            if (deque == null) {
+                deque = new ArrayDeque<>();
+            }
+            deque.push(value);
+            map.put(key, deque);
```
ppkarwasz, line 212: We could issue a warning if the `key` is already used in the classical MDC (`ThreadContext.get() != null`) (cf. [SLF4J-531](https://jira.qos.ch/browse/SLF4J-531) and return.

There is really no reason to allow users to use the same `key` in both the classical and stack-based MDC.

ppkarwasz (changes requested):
@terrettaz,

I think Ceki didn't implement the stack-valued MDC in Logback yet.

You could add a [ContextDataProvider](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/util/ContextDataProvider.html) to your PR that injects the top of every deque to the context map of a log event. This way you can use the `pushByKey` method with the `%X` pattern converter.

I believe that in the long term (i.e. as soon as a user requests it) we will add a stack-based MDC to the Log4j2 API, but I like this PR for the time being.

TWiStErRob:
As a user [agree with @rgoers's 1)](https://github.com/apache/logging-log4j2/pull/1022#issuecomment-1228845543). I found found the "18" artifact useful for migration, as it works nicely with SLF4J2, but it's confusing, because there was no such release of SLF4J.

Feedback: I ended up here, because I wanted to PR a doc fix for this page:
https://logging.apache.org/log4j/2.x/log4j-slf4j-impl/index.html
again, it's confusing it refers to "1.8" when there's no such [stable release](https://mvnrepository.com/artifact/org.slf4j/slf4j-api). I believe this should be also included in this PR or a following one.

ppkarwasz:
@terrettaz,

I modified your original PR to work with the current pattern converters (#1025). What name should I put in the release notes?

terrettaz:
> @terrettaz,
> 
> I modified your original PR to work with the current pattern converters (#1025). What name should I put in the release notes?

I am still working on this PR to include all your comment, should be finish soon.

About my name: Pierrick Terrettaz
Thanks

ppkarwasz:
@terrettaz, if the rest of the team agrees, we'll merge #1025 instead (which contains your initial contribution and my remarks), so no need to improve this PR.

Taking account of the recent changes the `log4j-slf4j20-impl` bridge is almost complete, but the `log4j-to-slf4j` bridge probably needs an SLF4J2 version (module) to profit from the new SLF4J 2.0 features.

terrettaz:
@ppkarwasz Thanks you for your comments!
I still pushed one commit in my branch you might be interested in. I added several unit tests to cover different corner cases especially when Log4j NDC is used along with MDC.pushByKey(null, "value")

PR #1050, opened 2022-09-12
(doc) Improve documentation for logging methods with Object param

Marcono1234:
Logging methods which only have an `Object` parameter do not log the exception strack trace when the argument is a `Throwable`. As mentioned [here](https://issues.apache.org/jira/browse/LOG4J2-86?focusedCommentId=17199071&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17199071) this is quite error-prone and I would be much happier if that behavior was changed. But at the very least the documentation should highlight this (even though I am afraid not many users read the documentation in detail).

Please let me know if this PR should be retargeted against the `release-2.x` branch.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/LogBuilder.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/LogBuilder.java
@@ -17,12 +17,13 @@
 package org.apache.logging.log4j;
 
 import org.apache.logging.log4j.message.Message;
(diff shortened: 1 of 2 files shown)


Marcono1234:
It looks like the test failure is unrelated to these changes.

jvz:
This should be targeted against the release-2.x branch if you'd like to include this in 2.20.0. Otherwise, this would target 3.0.0.

ppkarwasz:

PR #1051, opened 2022-09-12
LOG4J2-2947 - Make LogBuilder methods abstract

Marcono1234:
The current approach of having no-op `default` methods is pretty error-prone because it is easy to forget to override a method when a class implements the interface. Additionally the documentation stating that the interface method does nothing is probably quite confusing for users, who don't care about the implementation details of Log4j 2.


Marcono1234:
It looks like the test failure is unrelated to these changes.

ppkarwasz:
@Marcono1234,
Yes, they are unrelated, but you probably want to submit a PR against the current stable release `release-2.x`. The `master` branch is for the unreleased 3.x series.

Marcono1234:
In theory this might be a breaking change though in case users implement their own `LogBuilder` and (erronously?) had not overridden all of the methods before.

ppkarwasz:

PR #1058, opened 2022-09-16
Add triggerpolicy support for using current time

wuqian0808:
SizeBasedTrgiggerPolicy always use current time as compressed file name. However the crontrigger policy and timebasepolicy always use the last cycle start time. We want to add suport to keep their behavior same. So add support to use current time for other triggerpolices(cron and time). Test case also added. Pls help to review


wuqian0808:
Kindly help to advise

rgoers:
I disagree with both the premise and the implementation of this change.

1. By design size-based rollovers increment the index. They only update the time if no time-based triggering policy is configured. Time-based rollovers update the time, they reset the index to 1.
2. When direct-write is in use the file is always created at the start of the interval, so the file name should align with what the default rollover policy will currently create.
3. I don't normally use a date pattern on a size based file without a time-based trigger as the date/index won't make much sense together but if I did I would expect it to be from when the file is created. If that is not the case I could see correcting that.

PR #1062, opened 2022-09-16
Sort release 2.x POM files

ppkarwasz:
This PR performs two cleanup actions:

* it moves the `<scope>` of dependencies from the `<dependencyManagement>` and adds it wherever the dependencies are used. This has some beneficial effect: e.g. `log4j-api` **had** a transitive dependency of `javax.inject:javax.inject` in the **provided** scope because of the interaction between `maven-core` and our `<dependencyManagement>`. Now the dependency is in the **test** scope.
* it uses `src/tools/sort-pom.xslt` to sort all dependencies and plugins according to: the scope (import > provided > system > compile > runtime > text), the artifact id (except Log4j2 artifacts which come always first) and group id.

**Warning**: do not merge, until all dependency scope changes are accounted for.

### DIFF
--- a/log4j-1.2-api/pom.xml
+++ b/log4j-1.2-api/pom.xml
@@ -34,6 +34,13 @@
     <maven.doap.skip>true</maven.doap.skip>
   </properties>
   <dependencies>
+    <!-- Used for JMS appenders (needs an implementation of course) -->
+    <dependency>
+      <groupId>javax.jms</groupId>
+      <artifactId>javax.jms-api</artifactId>
+      <scope>provided</scope>
+      <optional>true</optional>
+    </dependency>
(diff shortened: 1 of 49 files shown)


ppkarwasz:
The difference in the dependency list before and after this change is minimal:

 * the transitive dependency `javax.inject` was before forced in the `provided` scope. It is now in the same scope as the artifact that depends on it,
 * there are some version changes in the transitive dependencies due to the order change: guava (test scope) is bumped to `29.0-jre` in `log4j-core`, whereas asm is bumped from `7.0` to `7.2`,
 * `commons-io`, `xz` and `netty-all` are not necessarily in the `test` scope.

PR #1063, opened 2022-09-18
Delegate plugin versions to ASF parent

ppkarwasz:
This PR removes all explicitly versioned plugins that are also defined in the Apache ASF parent POM (`org.apache:apache`).

This has the following effect on plugin versions:

 * `maven-assembly-plugin` bumped to 3.4.1
 * `maven-jar-plugin` bumped to 3.2.2
 * `maven-javadoc-plugin` bumped to 3.4.0
 * `maven-project-info-reports-plugin` bumped to 3.3.0
 * `maven-release-plugin` bumped to 3.0.0-M6
 * `maven-scm-plugin` bumped to 1.13.0
 * `maven-site-plugin` bumped to 3.12.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
   <parent>
     <groupId>org.apache.logging</groupId>
     <artifactId>logging-parent</artifactId>
-    <version>5</version>
+    <version>7</version>
     <relativePath/>
   </parent>
   <description>Apache Log4j 2</description>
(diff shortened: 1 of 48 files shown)


ppkarwasz:
`./mvnw site` also succeeded locally.

pom.xml
```diff
@@ -318,20 +311,19 @@
...
+  <repositories>
+    <repository>
+      <id>apache.snapshots</id>
+      <name>Apache Snapshot Repository</name>
+      <url>https://repository.apache.org/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+    </repository>
+  </repositories>
```
ppkarwasz, line 325: Of course this is not part of the PR. It is only necessary to resolve `logging-parent:6-SNAPSHOT`.

ppkarwasz:
The main point of this change are:

 * Dependabot stops looking at upgrades of the most common Maven plugins,
 * the builds are reproducible.

PR #1064, opened 2022-09-18
[LOG4J2-3601] log4j-slf4j2-impl: depend on log4j-core scope=runtime

afs:
https://issues.apache.org/jira/browse/LOG4J2-3601

This makes the dependency structure the same between log4j-slf4j-impl and log4j-slf4j2-impl. 

Then user applications only need a dependency on log4j-slf4j2-impl and it is then the same pattern as for log4j-slf4j-impl.

### DIFF
--- a/log4j-slf4j2-impl/pom.xml
+++ b/log4j-slf4j2-impl/pom.xml
@@ -50,12 +50,12 @@
     </dependency>
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
-      <artifactId>log4j-api-test</artifactId>
-      <scope>test</scope>
+      <artifactId>log4j-core</artifactId>
+      <scope>runtime</scope>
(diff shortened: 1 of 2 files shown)


ctubbsii (approved):

ppkarwasz:
@afs,

I merged it into `release-2.x` and `master`. Thanks

ctubbsii (commented):
:+1: for this. Looking forward to seeing this released. Any idea when 2.20 might be ready for users?

ppkarwasz:
@ctubbsii,

The release date mostly depends on how much time we have outside our dayjobs. You can follow [Ralph's thread on `dev@logging.apache.org`](https://lists.apache.org/thread/m3qnk4mtf2j0fkszdljxdv91tw1q694l) and help by testing and voting on the release candidates.

PR #1107, opened 2022-10-10
LOG4J2-3584 Make StatusConsoleListener use SimpleLogger internally.

vy:


log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -75,14 +75,23 @@ public final class StatusLogger extends AbstractLogger {
...
 
     private static final String DEFAULT_STATUS_LEVEL = PROPS.getStringProperty(DEFAULT_STATUS_LISTENER_LEVEL);
 
+    static final String DATE_FORMAT = PROPS.getStringProperty(STATUS_DATE_FORMAT);
+
+    static final boolean DATE_FORMAT_PROVIDED = Strings.isNotBlank(DATE_FORMAT);
+
+    static final boolean DEBUG_ENABLED = PropertiesUtil
+            .getProperties()
+            .getBooleanProperty(Constants.LOG4J2_DEBUG, false, true);
```
ppkarwasz, line 90: Can we put these into a `createSimpleLogger()` factory method? It would certainly ease up testing.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusConsoleListener.java
```diff
@@ -52,6 +59,17 @@ public StatusConsoleListener(final Level level, final PrintStream stream) {
...
+                "StatusConsoleListener",
+                level,
+                false,
+                true,
+                StatusLogger.DATE_FORMAT_PROVIDED,
+                false,
+                StatusLogger.DATE_FORMAT,
+                ParameterizedNoReferenceMessageFactory.INSTANCE,
+                StatusLogger.PROPS,
+                stream);
```
ppkarwasz, line 72: We could refactor this call of the `SimpleLogger` constructor and the one in `StatusLogger` into a common factory method.

ppkarwasz (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Fixed in 7877defe18ca5fb820da095c022dc3065a02c67a. Mind checking, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusConsoleListener.java
vy: Fixed in 7877defe18ca5fb820da095c022dc3065a02c67a. Mind checking, please?

ppkarwasz (approved):

PR #1111, opened 2022-10-12
fix(sec): upgrade org.apache.kafka:kafka-clients to 2.7.2

eurrio:
### What happened？
There are 1 security vulnerabilities found in org.apache.kafka:kafka-clients 1.1.1
- [CVE-2021-38153](https://www.oscs1024.com/hd/CVE-2021-38153)

### What did I do？
Upgrade org.apache.kafka:kafka-clients from 1.1.1 to 2.7.2 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How was this patch tested?
Run `mvn compile` failed locally, couldn't complete the build process.
Run `mvn clean test` failed locally, unit-test couldn't pass.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,8 +14,7 @@
   ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
   ~ See the license for the specific language governing permissions and
   ~ limitations under the license.
-  -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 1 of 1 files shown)


ppkarwasz:
@eurrio,

IMHO and according to [MvnRepository](https://mvnrepository.com/artifact/org.apache.kafka/kafka-clients) that CVE does not apply to any Kafka client (as opposed to [Kafka servers](https://mvnrepository.com/artifact/org.apache.kafka/kafka)). There is no security reason to upgrade our `kafka-clients` version.

Due to the stability of Kafka's Producer API the `KafkaAppender` can use **any** Kafka client version from 1.1.1 onwards: check our [Kafka tests](https://github.com/apache/logging-log4j2/actions/workflows/log4j-kafka-test.yml).

So the only question that remains is: which version should we advertise in our POM file. @rgoers: any ideas? I don't want to "upgrade" to the latest version to show people that we also support older revisions.

PR #1128, opened 2022-10-26
[LOG4J2-3581] Provide a JCL to Log4j2 API binary weaver

ppkarwasz:
Similarly to what the [Tomcat Migration Tools for Jakarta EE](https://github.com/apache/tomcat-jakartaee-migration) does, this PR provides a binary classfile weaver that replaces JCL loggers and methods with Log4j2 API.

The supported JCL methods include the whole `Log` interface  and those `LogFactory` methods that produce (directly or indirectly) loggers, cf. `JclConverterExample/JclConverterTest` for the current limits of the weaver.

This PR most notably **lacks**:

- a command line interface,
- a converter to refactor source code,
- more converters to transform entire JARs or WARs (though we can borrow these from Tomcat).

### DIFF
--- a/log4j-migration/pom.xml
+++ b/log4j-migration/pom.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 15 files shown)


ppkarwasz:
Will be resubmitted to `logging-log4j-transform`.

PR #1136, opened 2022-11-07
Refactor API initialization into common LoggingSystem class

jvz:
This refactors ProviderUtil, LogManager, and ThreadContextMapFactory to all rely on a common LoggingSystem class for initialization. This should make it easier to continue refactoring the API to avoid using static state. This also provides a potential integration point for more sophisticated PropertyEnvironment implementations.

This also updates the API version to 3.0.0, but that's only tangentially related.

Edit: related to https://cwiki.apache.org/confluence/display/LOGGING/API+Initialization


log4j-api/src/main/java/org/apache/logging/log4j/ThreadContext.java
```diff
@@ -414,7 +395,7 @@ public static ContextStack getImmutableStack() {
      * @param stack The stack to use.
      */
     public static void setStack(final Collection<String> stack) {
-        if (stack.isEmpty() || !useStack) {
```
jvz, line 417: This was a premature optimization as the two lines afterwards are no-ops anyways.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggingSystem.java
```diff
@@ -0,0 +1,349 @@
...
+         * @see ThreadContextMap
+         * @see ReadOnlyThreadContextMap
+         * @see org.apache.logging.log4j.ThreadContext
+         */
+        public ThreadContextMap createContextMap() {
+            final PropertyEnvironment environment = PropertiesUtil.getProperties();
+            // TODO(ms): make property lookup lazier here to avoid eager init
+            GarbageFreeSortedArrayThreadContextMap.init(environment);
+            CopyOnWriteSortedArrayThreadContextMap.init(environment);
+            DefaultThreadContextMap.init(environment);
```
jvz, line 315: These used to be initialized by `ThreadContextMapFactory::init`, but they're here now. Ideally, those classes should use a `PropertyEnvironment` constructor argument or similar, though I've deferred refactoring all of that until a later commit as this one was already getting big enough.

log4j-api/src/main/java/org/apache/logging/log4j/util/Activator.java
```diff
@@ -101,64 +100,52 @@ private String toStateString(final int state) {
...
         }
-        final List<URL> urls = bundleWiring.findEntries("META-INF", "log4j-provider.properties", 0);
-        for (final URL url : urls) {
-            ProviderUtil.loadProvider(url, bundleWiring.getClassLoader());
+    }
+
+    private List<Provider> loadProviders(final BundleWiring bundleWiring) {
+        final List<Provider> providers = new ArrayList<>();
+        final ClassLoader classLoader = bundleWiring.getClassLoader();
+        final Iterator<Provider> iterator = ServiceLoader.load(Provider.class, classLoader).iterator();
```
jvz, line 116: This approach seems to be the easiest way to support OSGi without requiring extra metadata or similar to make it work. Only disadvantage, as I mentioned on the dev list, is that this requires a couple admin permissions in OSGi to obtain different bundles' ClassLoader and similar private-ish objects. On the other hand, this makes it so we don't need to use anything outside the core OSGi API (which includes bundles and services basically, the latter we don't need to support anymore).

I'll note that I removed support for loading legacy providers here as they'd be unlikely to work properly in OSGi anyways, and we bumped the API compatibility version to 3.0.0 anyways.

log4j-api/src/main/java/org/apache/logging/log4j/util/Activator.java
```diff
@@ -101,64 +100,52 @@ private String toStateString(final int state) {
...
-            loadProvider(bundle);
-        }
-        unlockIfReady();
-    }
-
-    private void unlockIfReady() {
-        if (lockingProviderUtil && !ProviderUtil.PROVIDERS.isEmpty()) {
-            ProviderUtil.STARTUP_LOCK.unlock();
-            lockingProviderUtil = false;
-        }
```
jvz, line 147: Turns out that all this logic was redundant since it's impossible (?) for a Log4j Provider to have been installed before installing `log4j-api` due to dependencies. Thus, we don't need to scan bundles for Log4j Provider services until after the API has been installed.

In theory, you could try to upgrade the API or provider bundles at runtime, but that likely wouldn't work anyways, so there's not much point in scanning what's already installed here.

Note that the unlocking logic has also been simplified to unlock once we actually register a `Provider` with `ServiceRegistry`.

log4j-core/src/main/java/org/apache/logging/log4j/core/osgi/Activator.java
```diff
@@ -71,7 +70,6 @@ public int getOrder() {
         final ClassLoader classLoader = bundle.adapt(BundleWiring.class).getClassLoader();
         registry.registerBundleServices(InjectorCallback.class, bundleId, List.of(injectorCallback));
         registry.loadServicesFromBundle(PluginService.class, bundleId, classLoader);
-        registry.loadServicesFromBundle(Provider.class, bundleId, classLoader);
```
jvz, line 74: Logic moved to the API activator class instead.

ppkarwasz:
The latest build failure is a `ConcurrentModificationException` in `ServiceRegistry`, so this can be related to the change.

jvz:
Oh damn, that means I need to re-add the code for fixing `ServiceRegistry` to use a reentrant form of `computeIfAbsent` since `ConcurrentHashMap` doesn't support reentrancy.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggingSystem.java
jvz: Fixing this issue in a subsequent commit.

ppkarwasz (approved):
LGTM

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggingSystem.java
```diff
@@ -0,0 +1,363 @@
...
+
+    /**
+     * Acquires a lock on the initialization of locating a logging system provider. This lock should be
+     * {@linkplain #releaseInitializationLock() released} once the logging system provider is loaded. This lock is
+     * provided to allow for lazy initialization via frameworks like OSGi to wait for a provider to be installed
+     * before allowing initialization to continue.
+     *
+     * @see <a href="https://issues.apache.org/jira/browse/LOG4J2-373">LOG4J2-373</a>
+     */
+    public void acquireInitializationLock() {
```
rgoers, line 91: This class is not marked internal in any way. Exposing locks publicly like this is very dangerous.  It would make more sense to me to provide a method that accepts a Lambda to performed while the lock is held.

jvz: Fixed. Thanks for catching that!

PR #1138, opened 2022-11-09
Add minimal style check

ppkarwasz:
This PR adds a check for formatting problems in files from the last 32 commits. These include:
* differences in the license,
* trailing whitespace,
* incorrect file endings,
* indentation with tabs,
* incorrect ordering of imports.

Problems can be corrected by running `mvn spotless:apply`.


jvz (approved):
Seems reasonable. I like consistency!

PR #1139, opened 2022-11-13
Clean up static config in SimpleLogger/StatusLogger

jvz:


jvz:
I was kind of thinking of some sort of way to better represent this sort of property name to type map binding, though in the meantime, I figured I'd at least make the use of configuration beans more explicit. One idea I had was to support interfaces with annotations declaring the property name. The interface could be dynamically proxied to return data from the appropriate property environment. Alternatively, if there were a way to generate code from some annotated interfaces, then we could have compile-time proxies, too, to make it simpler to update how properties work for https://cwiki.apache.org/confluence/display/LOGGING/Properties+Enhancement

ppkarwasz (commented):
LGTM

PR #1140, opened 2022-11-14
[LOG4J2-3634] move invocation of propagateLogLevels()

mknet:
See https://issues.apache.org/jira/browse/LOG4J2-3634


ppkarwasz (changes requested):
@mknet,

Thank You for your contribution. Except two small synchronization issues, it looks good to me.

While this is an issue specific to Tomcat, which fails to properly configure the root logger if `Logger.getLogger("")` is called by any handler constructor, I am afraid that many other `LogManager` implementations may suffer from this bug. Therefore you solution is the cleanest we can get.

log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
```diff
@@ -1,319 +1,321 @@
...
+    public void publish(LogRecord record) {
+        if (record == null) {    // silently ignore null records
+            return;
+        }
+
+        if (this.installAsLevelPropagator) {
+            @SuppressWarnings("resource")    // no need to close the AutoCloseable ctx here
+            LoggerContext context = LoggerContext.getContext(false);
+            context.addPropertyChangeListener(this);
+            propagateLogLevels(context.getConfiguration());
```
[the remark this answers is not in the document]
mknet: Makes sense to me

log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
```diff
@@ -1,319 +1,321 @@
...
+    private static final org.apache.logging.log4j.Logger SLOGGER = StatusLogger.getLogger();
+
+    // the caller of the logging is java.util.logging.Logger (for location info)
+    private static final String FQCN = java.util.logging.Logger.class.getName();
+    private static final String UNKNOWN_LOGGER_NAME = "unknown.jul.logger";
+    private static final java.util.logging.Formatter julFormatter = new java.util.logging.SimpleFormatter();
+
+    private boolean doDebugOutput = false;
+    private String julSuffixToAppend = null;
+    private boolean installAsLevelPropagator = false;
```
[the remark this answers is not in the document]
mknet: I don't get why the topic serialization would into play here. On the other hand I don't have any worries about it. So, I will do as you wish.

mknet:
> @mknet,
> 
> Thank You for your contribution. Except two small synchronization issues, it looks good to me.
> 
> While this is an issue specific to Tomcat, which fails to properly configure the root logger if `Logger.getLogger("")` is called by any handler constructor, I am afraid that many other `LogManager` implementations may suffer from this bug. Therefore you solution is the cleanest we can get.

Thanks for your reply and feedback, @ppkarwasz

mknet:
@ppkarwasz It seems to me, the code format wasn't correct. I re-ran spotless and committed the change. Sorry for the frustration.

ppkarwasz (approved):
@mknet, thanks.

I'll leave this PR open for a couple of days to give a chance to other maintainers to review it and then I'll merge it.

log4j-jul/src/main/java/org/apache/logging/log4j/jul/Log4jBridgeHandler.java
```diff
@@ -1,319 +1,322 @@
...
+
+        if (this.installAsLevelPropagator) {
+            synchronized(this) {
+                @SuppressWarnings("resource")    // no need to close the AutoCloseable ctx here
+                LoggerContext context = LoggerContext.getContext(false);
+                context.addPropertyChangeListener(this);
+                propagateLogLevels(context.getConfiguration());
+                // note: java.util.logging.LogManager.addPropertyChangeListener() could also
+                // be set here, but a call of JUL.readConfiguration() will be done on purpose
+                this.installAsLevelPropagator = false;
```
ppkarwasz, line 186: By looking at some established practices for lazy initialization (e.g. [Commons Lang](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/concurrent/LazyInitializer.java)), there should be a second `if (this.installAsLevelPropagator)` inside the synchronized block to ensure that `propagateLogLevels` is called only once.

In the `master` branch we do have a generic [LazyUtil](https://github.com/apache/logging-log4j2/blob/master/log4j-api/src/main/java/org/apache/logging/log4j/util/LazyUtil.java), but it is probably not worth backporting it for a single callsite.

jvz: It depends if other lazy initialized code could be refactored in this branch, too. Can look for volatile fields for starters.

carterkozak: `installAsLevelPropagator` must also be volatile

mknet: Uhhhh, now it makes sense to me as well. I will modify the code tomorrow

mknet: @ppkarwasz @carterkozak `volatile` is applied. Anything else? Thx!

ppkarwasz: @mknet: a second check for `installAsLevelPropagator` is needed so that if two thread enter in the first conditional block one will update the levels and the other will wait until it is done:
```lang-java
if (this.installAsLevelPropagator) {
    synchronized(this) {
        if (this.installAsLevelPropagator) {
```

mknet: Check. Will come back to you in the next couple of ours.

ppkarwasz (approved):

ppkarwasz:
@mknet,

Thank you for finding and fixing this bug.

mknet:
It was a pleasure!

PR #1141, opened 2022-11-14
(doc) Use Gradle platform to import Log4j BOM in Gradle 5.0+

grossws:


src/site/markdown/maven-artifacts.md.vm
```diff
@@ -149,14 +149,29 @@ file. When you specify the version identifier in this section, you don't have to
...
+to align dependency versions.
+
+`build.gradle`
+
+```
+dependencies {
+  implementation platform('org.apache.logging.log4j:log4j-bom:${Log4jReleaseVersion}')
+
+  implementation 'org.apache.logging.log4j:log4j-api'
+  implementation 'org.apache.logging.log4j:log4j-core'
```
ppkarwasz, line 162: I would use `runtime` as scope for `log4j-core`.

carterkozak: Do you think this documentation should describe the common case, or advanced component development configurations?

grossws: I'm more inclined to describe common case. If it's ok with you, I'll change it to runtimeOnly.

ppkarwasz: @grossws, the documentation you cited might be misleading: if someone develops custom plugins, `log4j-core` needs to be a compile dependency because custom plugins must reference implementation classes. As it is written right now someone might think that adding Log4j2 Core as compile dependency improves startup.

I would remove that remark from the docs.

grossws: Updated with note for Gradle 2.8-3.3 about compile/runtime instead of implementation/runtimeOnly

rgoers (approved):
LGTM

PR #1142, opened 2022-11-14
(doc) Add log4j-slf4j2-impl section to maven-artifacts

grossws:


src/site/markdown/maven-artifacts.md.vm
```diff
@@ -190,10 +190,19 @@ then add the following but do not remove any Commons Logging 1.x dependencies.
...
+If existing components use SLF4J 2.0.x and you want to have this logging routed to Log4j 2, then add the
 following but do not remove any SLF4J dependencies.
+Ensure that all SLF4J artifacts have consistent version.
+
+#dependencies(['log4j-slf4j2-impl'])
+
+$h3 SLF4J 1.7.x Bridge
+
+If existing components use SLF4J 1.7.x and you want to have this logging routed to Log4j 2, then either
+update SLF4J to 2.0.x, or add the following but do not remove any SLF4J dependencies.
```
vy, line 204:
```suggestion
If existing components use SLF4J 1.7.x and you want to have this logging routed to Log4j 2, then add the
following but do not remove any SLF4J dependencies.
```

_"either update SLF4J to 2.0.x, or"_ phrase can be misread as _"updating to SLF4J 2.0.x would make Log4j work magically"_, hence I removed it.

src/site/markdown/maven-artifacts.md.vm
```diff
@@ -190,10 +190,19 @@ then add the following but do not remove any Commons Logging 1.x dependencies.
 
 #dependencies(['log4j-jcl'])
 
-$h3 SLF4J Bridge
+$h3 SLF4J 2.0.x Bridge
 
-If existing components use SLF4J and you want to have this logging routed to Log4j 2, then add the
+If existing components use SLF4J 2.0.x and you want to have this logging routed to Log4j 2, then add the
 following but do not remove any SLF4J dependencies.
+Ensure that all SLF4J artifacts have consistent version.
```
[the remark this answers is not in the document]
grossws: I mean that if downstream user already has for example `org.slf4j:jcl-over-slf4j` and doesn't wish to switch to `log4j-jcl` than `jcl-over-slf4j` version should be consistent with `slf4j-api`

src/site/markdown/maven-artifacts.md.vm
grossws: Fixed

src/site/markdown/maven-artifacts.md.vm
```diff
@@ -190,10 +190,19 @@ then add the following but do not remove any Commons Logging 1.x dependencies.
...
+
+#dependencies(['log4j-slf4j2-impl'])
```
gemmellr, line 199: This is the same 'add single dependency' style documentation as given below / previously for the SLF4J 1.x bridge, suggesting similar behaviour. However, the SLF4J 2.x -> Log4J 2.19 bridge currently behaves differently than prior SLF4J -> Log4j (1.x and 2.x) bridges, also requiring users add their own explicit dependency on log4j-core, because the log4j-slf4j2-impl bridge dependency doesnt actually provide it, unlike log4j-slf4j-impl.

It would be good to also fix the behaviour so its as generally expected and previously was, then this documentation would appear accurate.

Covered in https://issues.apache.org/jira/browse/LOG4J2-3601 with PR at https://github.com/apache/logging-log4j2/pull/1064

vy:
I think the changes imply that all used logging APIs should first be bridged to SLF4J and then to Log4j, whereas ideally all logging APIs should be directly bridged to Log4j.

PR #1165, opened 2022-12-15
[LOG4J2-3642] Consider Osgi versions older than 1.5 as not available.

adwsingh:
We use the method [FrameworkUtil.getBundle](https://github.com/apache/logging-log4j2/blob/release-2.x/log4j-core/src/main/java/org/apache/logging/log4j/core/osgi/BundleContextSelector.java#L59) from OSGi, but this method was added [only in OSGi 1.5](https://github.com/osgi/osgi/blob/main/org.osgi.framework/src/org/osgi/framework/FrameworkUtil.java#L224), this causes NoSuchMethodError if older version of OSGi is used.

I would suggest to consider OSGi being unavailable for OSGi version < 1.5 or when that method is not available.

### DIFF
--- a/log4j-api-test/pom.xml
+++ b/log4j-api-test/pom.xml
@@ -99,16 +99,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.felix</groupId>
-      <artifactId>org.apache.felix.framework</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.eclipse.tycho</groupId>
-      <artifactId>org.eclipse.osgi</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>uk.org.webcompere</groupId>
       <artifactId>system-stubs-jupiter</artifactId>
       <scope>test</scope>
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/util/ServiceLoaderUtilTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/util/ServiceLoaderUtilTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.logging.log4j.util;
 
+import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.util.ArrayList;
 import java.util.List;
@@ -66,4 +67,11 @@ public class ServiceLoaderUtilTest {
                 ServiceConfigurationError.class, ServiceConfigurationError.class, ClassFormatError.class);
     }
 
+    @Test
+    public void testOsgiUnavailable() {
+        // OSGI classes are present...
+        assertDoesNotThrow(() -> Class.forName("org.osgi.framework.FrameworkUtil"));
+        // ...but we don't run in an OSGI container
+        assertThat(OsgiServiceLocator.isAvailable()).as("Running in OSGI container").isFalse();
+    }
 }
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
@@ -30,12 +30,17 @@ public class OsgiServiceLocator {
 
     private static boolean checkOsgiAvailable() {
         try {
-            Class.forName("org.osgi.framework.Bundle");
-            return true;
-        } catch (final ClassNotFoundException | LinkageError e) {
+            /*
+             * OSGI classes of any version can still be present even if Log4j2 does not run in
+             * an OSGI container, hence we check if this class is in a bundle.
+             */
+            final Class< ? > clazz = Class.forName("org.osgi.framework.FrameworkUtil");
+            return clazz.getMethod("getBundle", Class.class)
+                    .invoke(null, OsgiServiceLocator.class) != null;
+        } catch (final ClassNotFoundException | NoSuchMethodException | LinkageError e) {
             return false;
         } catch (final Throwable e) {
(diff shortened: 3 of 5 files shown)


ppkarwasz:
Hi @adwsingh,

Since `log4j-api` has a non-optional import on `org.osgi.framework;version="[1.8,2)"` (OSGI Release 6), it shouldn't work on older OSGI releases (Framework API 1.5 is part of Release 4, Version 4.2).

We don't really use any method specific to Framework API 1.8, but the activator of the bundle uses `Bundle#adapt` that appeared in OSGI Release 4, Version 4.3.

Do you think of any reason we should lower our baseline? The current Release 6 baseline is the same as the baseline of OPS4J Pax Logging 1.11.x. As mentioned above we can lower it to the baseline of the 10 years old OPS4J Pax Logging 1.7.x (Release 4, Version 4.3), but not lower.

adwsingh:
@ppkarwasz Thanks for the feedback.

I wanted to add a check ensure we are not loading an OSGi with which we cannot work with. I am moving the check to use `Bundle#adapt`

adwsingh:
@ppkarwasz did you get a chance to take a look at the updated PR?

ppkarwasz:
Hi @adwsingh,

The check is correct, but I am still not convinced if we need this PR. Does it solve a concrete problem you encountered? Does it prevent a potential problem? Due to the bundle descriptor in `logj4-api`, an OSGI container prior to Release 6 will not load the bundle.

rgoers:
@adwsingh The Jira issue says you got a NoSuchMethodError when trying to load the bundle. How does this happen in your environment?

adwsingh:
@rgoers This is the stack trace we have get in one of our unit tests in a CI build environment.

```
  [junit] java.lang.NoSuchMethodError: org.osgi.framework.FrameworkUtil.getBundle(Ljava/lang/Class;)Lorg/osgi/framework/Bundle;
    [junit] 	at org.apache.logging.log4j.util.OsgiServiceLocator.loadServices(OsgiServiceLocator.java:52)
    [junit] 	at org.apache.logging.log4j.util.ServiceLoaderUtil.loadServices(ServiceLoaderUtil.java:91)
    [junit] 	at org.apache.logging.log4j.util.PropertiesUtil$Environment.<init>(PropertiesUtil.java:471)
    [junit] 	at org.apache.logging.log4j.util.PropertiesUtil$Environment.<init>(PropertiesUtil.java:444)
    [junit] 	at org.apache.logging.log4j.util.PropertiesUtil.<init>(PropertiesUtil.java:88)
    [junit] 	at org.apache.logging.log4j.util.PropertiesUtil.<init>(PropertiesUtil.java:80)
    [junit] 	at org.apache.logging.log4j.util.PropertiesUtil.<clinit>(PropertiesUtil.java:56)
```

We get the `NoSuchMethodError` while attempting to load the Osgi itself.

> Due to the bundle descriptor in logj4-api, an OSGI container prior to Release 6 will not load the bundle.

@ppkarwasz can you point me to the code in `log4j-api` which does this check?

ppkarwasz:
@adwsingh,

The restriction is not in the code, but in the manifest (which we generate with `maven-bundle-plugin`):

```
Import-Package:
 org.apache.logging.log4j;version="[2.19,3)",
 org.apache.logging.log4j.internal,
 org.apache.logging.log4j.message;version="[2.19,3)",
 org.apache.logging.log4j.simple;version="[2.19,3)",
 org.apache.logging.log4j.spi;version="[2.19,3)",
 org.apache.logging.log4j.status;version="[2.19,3)",
 org.apache.logging.log4j.util;version="[2.19,3)",
 org.osgi.framework;version="[1.8,2)",
 org.osgi.framework.wiring;version="[1.2,2)",
 sun.reflect;resolution:=optional
```

An OSGI framework should not resolve such a bundle if all requirements are not satisfied.

adwsingh:
But the `NoSuchMethodError` is being triggered during static initialization of `PropertiesUtil` (check the stacktrace I shared above) while trying to invoke `Framework.getBundle` 

https://github.com/apache/logging-log4j2/blob/518c3f32dd2e3cc347ac5d4c066b7f34cf7462d4/log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java#L52

I am not sure if OsGi is even involved at this point.

ppkarwasz:
OSGI resolves a bundle's dependencies **before** it loads any class from the bundle. Running the tests in `log4j-osgi` with `-Dfelix.version=4.4.1` (which contains OSGI Framework API 1.7) I get an error like this:

```
org.osgi.framework.BundleException: Unresolved constraint in bundle org.apache.logging.log4j.api [11]: Unable to resolve 11.0: missing requirement [11.0] osgi.wiring.package; (&(osgi.wiring.package=org.osgi.framework)(version>=1.8.0)(!(version>=2.0.0)))
```

Probably, what is happening in your test is that `log4j-api` is not initialized inside the OSGI container, but is contained in the tests runtime. Log4j2 is initialized before the OSGI framework detect the version incompatibility.
 the
Amore stable solution to this problem would be to set the `OSGI_AVAILABLE` field from the bundle's activator. This way we have a garantee that `OSGI_AVAILABLE` is set if and only if the activator of the bundle has been called.

ppkarwasz:
@adwsingh,

I reverted back to your first idea of checking for `FrameworkUtil#getBundle`: if the method exists (OSGI 4.2) and returns `null`, we are not running in an OSGI container and `OsgiServiceLocator` should be disabled regardless of the OSGI version on the classpath.

You can test this by running `mvn verify -pl log4j-api-test -Dosgi.api.version=X` with values of X in `4.0.0`, `4.1.0` and `4.2.0`.

What name should we put in the changelog for this PR and #1127 ?

adwsingh:
You can use Adwait Kumar Singh as the name.

PR #1167, opened 2022-12-19
add JDK 19 to benchmark workflow

sullis:


sullis:
wdyt? @jvz

jvz:
Sounds like a good idea. @vy this seem reasonable? Makes sense to include 11, 17, and latest.

jvz (approved):

vy:
`benchmark.yml` is the very first (and early!) building block of the continuous performance test bed I have been unsuccessfully trying to work on for the last 1-2 years. I am not keen on adding non-LTS releases there. I can hardly find time to work on it, I don't want to maintain non-LTS release versions.

PR #1170, opened 2022-12-25
jmh 1.36

sullis:
JMH 1.36
https://mail.openjdk.org/pipermail/jmh-dev/2022-November/003553.html


ppkarwasz:
@sullis:

Thank you for the information. We certainly will upgrade to version 1.36, but we have Dependabot configured to remind us of new dependency versions: #1151.

PR #1171, opened 2022-12-27
[LOG4J2-3645] Minimal changes to allow flow tracing with LogBuilder

ppkarwasz:
This PR adds a minimal amount of methods to greatly simplify flow tracing with `LogBuilder`. E.g.:

```
logger.traceEntry("a={}, b={}", 1, 42);
```

can be now rewrittent as:

```
logger.atTrace()
        .withMarker(AbstractLogger.ENTRY_MARKER)
        .logAndGet(() -> {
            LoggingSystem.getFlowMessageFactory()
                    .newEntryMessage("a={}, b={}", 1, 42);
        });
```

The latter can be easily perfomed by a bytecode manipulation tool (cf. #1147).

It also refactors the factory methods that create `EntryMessage/ExitMessage`s to be low-garbage.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/LoggerTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/LoggerTest.java
@@ -16,15 +16,6 @@
  */
 package org.apache.logging.log4j;
 
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.hamcrest.CoreMatchers.endsWith;
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.startsWith;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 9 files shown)


log4j-api-test/src/test/java/org/apache/logging/log4j/LoggerTest.java
```diff
@@ -203,11 +191,18 @@ public String get() {
...
+        assertThat(results.get(1)).startsWith("EXIT[ FLOW ] TRACE Exit").contains("doFoo(a=1, b=2): 3");
+    }
+
+    @Test
+    public void flowTracingNoFormat() {
+        logger.traceEntry(null, 1, "2", new ObjectMessage(3));
+        logger.traceExit((String) null, 4);
+        assertThat(results).hasSize(2);
+        assertThat(results.get(0)).isEqualTo("ENTER[ FLOW ] TRACE Enter with params: [1, 2, 3]");
+        assertThat(results.get(1)).isEqualTo("EXIT[ FLOW ] TRACE Exit with: 3");
```
ppkarwasz, line 205: The previous code printed "Enter param(1, 2, 3)", but used a format that depended on the number of parameters. The current code uses a `ParameterizedMessage("params: {}", (Object) params)`.

rgoers (approved):
Looks OK to me.

garydgregory:
I've never been a fan of using "And" in method names, but no great  alternative comes to mind. 

Shouldn't new public  methods have Java since tags?

ppkarwasz:
@garydgregory,

This is not a breaking change, so I added `@since 2.20` tags.

log4j-api/src/main/java/org/apache/logging/log4j/message/FlowMessageFactory.java
```diff
@@ -30,6 +40,25 @@ public interface FlowMessageFactory {
...
 
+    /**
+     * Creates a new exit message based on a return value and a forma string.
+     *
+     * @param format a format string
+     * @param result the return value
+     * @return the new exit message
+     * @since 2.20
+     */
+    ExitMessage newExitMessage(String format, Object result);
```
jvz, line 51: Should these be default methods, or is it ok to break the API here?

ppkarwasz: I am not sure what the default implementation should be.

Anyway a minor version bump allows to break a SPI.

jvz: That sounds reasonable. Sometimes, a default implementation could be to either return `null` or just `throw new UnsupportedOperationException()` with appropriate checks on the consuming end as appropriate.

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
```diff
@@ -1686,6 +1687,14 @@ void fatal(String message, Object p0, Object p1, Object p2, Object p3, Object p4
      */
     <MF extends MessageFactory> MF getMessageFactory();
 
+    /**
+     * Gets the flow message factory used to convert messages into flow messages.
+     *
+     * @return the flow message factory, as an instance of {@link FlowMessageFactory}.
+     * @since 2.20
+     */
+    FlowMessageFactory getFlowMessageFactory();
```
jvz, line 1696: Just wanted to say thanks for handling this here as I was also adding this field in 3.x, so that simplified things for me!

PR #1173, opened 2022-12-30
LOG4J2-1631 Honor timezone in file name pattern

dnsmkl:
Given pattern "app-%d{yyyy-MM-dd}{timezone}.log.gz" treat second pair of curly braces as timezone option.

If timezone option is not set, then use default timezone as before.
If timezone option is set, but it's parsing fails, then use GMT+0.
(same DatePatternConverter behavior as in other places)

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
@@ -16,9 +16,11 @@
  */
 package org.apache.logging.log4j.core.appender.rolling;
 
-import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneId;
 import java.util.Calendar;
-import java.util.Date;
 import java.util.Locale;
 import java.util.TimeZone;
 
@@ -34,8 +36,10 @@ import static org.junit.jupiter.api.Assertions.*;
  */
 public class PatternProcessorTest {
 
-    private String format(final long time) {
-        return new SimpleDateFormat("yyyy/MM/dd HH:mm:ss.SSS").format(new Date(time));
+    private static Instant parseLocalDateTime(String text) {
+        return LocalDateTime.parse(text)
+                .atZone(ZoneId.systemDefault())
+                .toInstant();
     }
 
     @Test
@@ -56,30 +60,22 @@ public class PatternProcessorTest {
     @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
     public void testGetNextTimeHourlyReturnsFirstMinuteOfNextHour() {
         final PatternProcessor pp = new PatternProcessor("logs/app-%d{yyyy-MM-dd-HH}.log.gz");
-        final Calendar initial = Calendar.getInstance();
-        initial.set(2014, Calendar.MARCH, 4, 10, 31, 59); // Tue, March 4, 2014, 10:31
-        final long actual = pp.getNextTime(initial.getTimeInMillis(), 1, false);
+        Instant initial = parseLocalDateTime("2014-03-04T10:31:59");
+        final long actual = pp.getNextTime(initial.toEpochMilli(), 1, false);
 
-        // expect Wed, March 4, 2014, 11:00
-        final Calendar expected = Calendar.getInstance();
-        expected.set(2014, Calendar.MARCH, 4, 11, 0, 0);
-        expected.set(Calendar.MILLISECOND, 0);
-        assertEquals(format(expected.getTimeInMillis()), format(actual));
+        Instant expected = parseLocalDateTime("2014-03-04T11:00:00");
+        assertEquals(expected, Instant.ofEpochMilli(actual));
     }
 
     @Test
     @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
     public void testGetNextTimeHourlyReturnsFirstMinuteOfNextHour2() {
         final PatternProcessor pp = new PatternProcessor("logs/app-%d{yyyy-MM-dd-HH}.log.gz");
-        final Calendar initial = Calendar.getInstance();
-        initial.set(2014, Calendar.MARCH, 4, 23, 31, 59); // Tue, March 4, 2014, 23:31
-        final long actual = pp.getNextTime(initial.getTimeInMillis(), 1, false);
+        Instant initial = parseLocalDateTime("2014-03-04T23:31:59");
+        final long actual = pp.getNextTime(initial.toEpochMilli(), 1, false);
 
-        // expect Wed, March 5, 2014, 00:00
-        final Calendar expected = Calendar.getInstance();
-        expected.set(2014, Calendar.MARCH, 5, 0, 0, 0);
-        expected.set(Calendar.MILLISECOND, 0);
-        assertEquals(format(expected.getTimeInMillis()), format(actual));
+        Instant expected = parseLocalDateTime("2014-03-05T00:00:00");
+        assertEquals(expected, Instant.ofEpochMilli(actual));
     }
 
     @Test
@@ -89,15 +85,11 @@ public class PatternProcessorTest {
         // during winter GMT-6
         // during summer GMT-5
         final PatternProcessor pp = new PatternProcessor("logs/app-%d{yyyy-MM-dd-HH}{America/Chicago}.log.gz");
-        final Calendar initial = Calendar.getInstance(TimeZone.getTimeZone("GMT-6"));
-        initial.set(2014, Calendar.MARCH, 9, 1, 31, 59); // Tue, March 9, 2014, 01:31
-        final long actual = pp.getNextTime(initial.getTimeInMillis(), 1, false);
+        Instant initial = OffsetDateTime.parse("2014-03-09T01:31:59-06:00").toInstant();
+        final long actual = pp.getNextTime(initial.toEpochMilli(), 1, false);
 
-        // expect Wed, March 9, 2014, 02:00
-        final Calendar expected = Calendar.getInstance(TimeZone.getTimeZone("GMT-6"));
-        expected.set(2014, Calendar.MARCH, 9, 2, 0, 0);
-        expected.set(Calendar.MILLISECOND, 0);
-        assertEquals(format(expected.getTimeInMillis()), format(actual));
+        Instant expected = OffsetDateTime.parse("2014-03-09T02:00:00-06:00").toInstant();
+        assertEquals(expected, Instant.ofEpochMilli(actual));
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
```diff
@@ -83,6 +87,11 @@ void formatToBuffer(final Instant instant, final StringBuilder destination) {
         public String toPattern() {
             return fastDateFormat.getPattern();
         }
+
+        @Override
+        public TimeZone getTimeZone() {
```
dnsmkl, line 92: Expose parsed timezone option.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessor.java
```diff
@@ -160,9 +164,9 @@ public long getNextTime(final long currentMillis, final int increment, final boo
         if (frequency == null) {
             throw new IllegalStateException("Pattern does not contain a date");
         }
-        final Calendar currentCal = Calendar.getInstance();
+        final Calendar currentCal = Calendar.getInstance(timeZone);
         currentCal.setTimeInMillis(currentMillis);
-        final Calendar cal = Calendar.getInstance();
+        final Calendar cal = Calendar.getInstance(timeZone);
```
dnsmkl, line 169: Use timezone option that was parsed by the pattern parser.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
```diff
@@ -52,6 +52,10 @@ private abstract static class Formatter {
         public String toPattern() {
             return null;
         }
+
+        public TimeZone getTimeZone() {
+            return null;
+        }
```
ppkarwasz, line 58: I would return `TimeZone.getDefault()` here to remove all null checks.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessor.java
```diff
@@ -343,6 +347,15 @@ private RolloverFrequency calculateFrequency(final String pattern) {
...
     }
 
+    private static TimeZone calculateTimeZone(DatePatternConverter dateConverter) {
+        TimeZone timeZone = dateConverter.getTimeZone();
+        if (timeZone != null) {
+            return timeZone;
+        } else {
+            return Calendar.getInstance().getTimeZone();
+        }
+    }
```
ppkarwasz, line 357: If `getTimeZone()` didn't return `null`, this would be unnecessary.

ppkarwasz (changes requested):
@dnsmkl,

Thank You for your contribution. Other than the remark below it looks good to me.

Regarding the unit tests, could you approach the problem of "daylight saving time" too? It didn't appear with your PR, but was not testable before.

Java time classes support daylight saving time changes if you use a named timezone like `America/Chicago` instead of a GMT offset (`GMT-6:00`). With such a timezone I would expect (and test):

 * that a daily rotation the day DST begins occurs after 23 hours,
 * that a daily rotation the day DST ends occurs after 25 hours,
 * that an hourly rotation at 1 am the day DST begins  is followed by a rotation at 3 am (after 1 hour),
 * that an hourly rotation at 2 am the day DST ends is followed by a rotation at 4 am (after **2** hours).

ppkarwasz:
BTW: run `mvn spotless:apply -pl log4j-core` to fix the formatting. Due to our current configuration whoever touches a file, has to format it entirely.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessor.java
dnsmkl: Ok.

dnsmkl:
I would like to highlight that `Calendar.getInstance().getTimeZone()` is different from `TimeZone.getDefault()`.
So pattern "logs/app-%d{yyyy-MM-dd}.log.gz" has slight change.

Before this PR, timezone used was `Calendar.getInstance().getTimeZone()`
After this PR, for same pattern timezone will be `TimeZone.getDefault()`.

Is such change acceptable?

From Calendar.java:

    private static TimeZone defaultTimeZone(Locale l) {
        TimeZone defaultTZ = TimeZone.getDefault();
        String shortTZID = l.getUnicodeLocaleType("tz");
        return shortTZID != null ?
            TimeZoneNameUtility.convertLDMLShortID(shortTZID)
                .map(TimeZone::getTimeZone)
                .orElse(defaultTZ) :
            defaultTZ;
    }

rgoers:
Where are you seeing getTimeZone() implemented like that? It certainly doesn't look like that in Java 8 from what I can see. It pretty much just uses Timezone.getDefault().

dnsmkl:
@ppkarwasz

I have a nagging wish to convert `PatternProcessorTest` to use `java.time` and use date formats `ISO_ZONED_DATE_TIME`, `ISO_INSTANT`.
Would that be useful?
(this would reduce some verbosity; it could be done as separate "migration" commit)

dnsmkl:
> Where are you seeing getTimeZone() implemented like that?

In jdk-11.

> If they changed getTimeZone to work differently in Java 11 then I don't think us directly using TimeZone.getDefault() matters since 2.x is based on Java 8 support.

Great.

ppkarwasz:
@dnsmkl,

Thanks for the remark, I didn't know it worked like this in Java 11. Experimenting a little bit with this feature I conclude that:

 * you can add timezone data to a locale: e.g. `Locale.forLanguageTag("pl-PL-u-tz-uschi")`, which uses `America/Chicago` as timezone,
 * no sane OS sets a timezone data for the "usual" locales,
 * our date formatters **always** use `TimeZone.getDefault()` if no timezone is set explicitly,
 * our rotation logic **should** do the same (we don't want timestamps in Warsaw time, but rotation on Chicago time),
 * fortunately there are only 5 call sites of `Calendar.getInstance()` in Log4j2: the two that you replaced with the correct `Calendar.getInstance(TimeZone)`, two irrelevant in `CronExpression` and one in `AbsoluteDateFormat`.

ppkarwasz:
> I have a nagging wish to convert `PatternProcessorTest` to use `java.time` and use date formats `ISO_ZONED_DATE_TIME`, `ISO_INSTANT`. Would that be useful? (this would reduce some verbosity; it could be done as separate "migration" commit)

You can do whatever you want in test classes.

In the main classes like `PatternProcessor` however, we strive to achieve a garbage-free code, so AFAIK we are stuck with `Calendar`. @rgoers, is it safe to assume that calls to `PatternProcessor#getNextTime` are synchronized and we can replaced the two `Calendar.getInstance()` calls with cached instances? It certainly seems so in the current code.

dnsmkl:
> ... could you approach the problem of "daylight saving time" too?

Implementation works just as you described.
Tests added.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
dnsmkl: Done as suggested

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
```diff
@@ -274,4 +310,171 @@ public void testGetNextTimeWeeklyReturnsFirstWeekInYear_US() {
...
+        // expect Wed, March 5, 2014, 00:00
+        final Calendar expected = Calendar.getInstance(TimeZone.getTimeZone("GMT-6"));
+        expected.set(2014, Calendar.MARCH, 5, 0, 0, 0);
+        expected.set(Calendar.MILLISECOND, 0);
+        assertEquals(format(expected.getTimeInMillis()), format(actual));
+    }
+
+    @Test
+    @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
+    @ResourceLock(value = Resources.TIME_ZONE, mode = ResourceAccessMode.READ)
```
ppkarwasz, line 347: The access mode should be `READ_WRITE` (the default if you omit `mode`).

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
```diff
@@ -274,4 +310,171 @@ public void testGetNextTimeWeeklyReturnsFirstWeekInYear_US() {
...
+            expected.set(Calendar.MILLISECOND, 0);
+            assertEquals(format(expected.getTimeInMillis()), format(actual));
+        } finally {
+            TimeZone.setDefault(old);
+        }
+    }
+
+    @Test
+    @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
+    @ResourceLock(value = Resources.TIME_ZONE, mode = ResourceAccessMode.READ)
```
ppkarwasz, line 369: Same as above.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
```diff
@@ -274,4 +310,171 @@ public void testGetNextTimeWeeklyReturnsFirstWeekInYear_US() {
...
+        // expect Mon, November 3, 2014, 00:00
+        final Calendar expected = Calendar.getInstance(TimeZone.getTimeZone("GMT-6")); // offset changed
+        expected.set(2014, Calendar.NOVEMBER, 3, 0, 0, 0);
+        expected.set(Calendar.MILLISECOND, 0);
+        assertEquals(format(expected.getTimeInMillis()), format(actual));
+    }
+
+    @Test
+    @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
+    @ResourceLock(value = Resources.TIME_ZONE, mode = ResourceAccessMode.READ)
```
ppkarwasz, line 461: `READ_WRITE`

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
```diff
@@ -274,4 +310,171 @@ public void testGetNextTimeWeeklyReturnsFirstWeekInYear_US() {
...
+            expected.set(Calendar.MILLISECOND, 0);
+            assertEquals(format(expected.getTimeInMillis()), format(actual));
+        } finally {
+            TimeZone.setDefault(old);
+        }
+    }
+
+    @Test
+    @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
+    @ResourceLock(value = Resources.TIME_ZONE, mode = ResourceAccessMode.READ)
```
ppkarwasz, line 391: In this test and the remaining ones I don't think we need to lock `Resources.TIME_ZONE`: the result of the test should **not** depend on the default timezone.

ppkarwasz (approved):
Can you double-check the `@ResourceLock` annotations? Otherwise it looks good to me, thank You. If there are no objections from other committers I'll merge it in a couple of days.

What name should I put in the release notes?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/PatternProcessorTest.java
```diff
@@ -56,155 +60,142 @@ public void testDontInterpretBackslashAsEscape() {
...
+    }
+
+    @Test
+    @ResourceLock(value = Resources.LOCALE, mode = ResourceAccessMode.READ)
+    public void testGetNextTimeHourlyReturnsFirstMinuteOfNextHourDstStart() {
+        // America/Chicago 2014 - DST start - Mar 9 02:00
+        // during winter GMT-6
+        // during summer GMT-5
+        final PatternProcessor pp = new PatternProcessor("logs/app-%d{yyyy-MM-dd-HH}{America/Chicago}.log.gz");
+        Instant initial = OffsetDateTime.parse("2014-03-09T01:31:59-06").toInstant();
```
ppkarwasz, line 88:
```suggestion
        Instant initial = OffsetDateTime.parse("2014-03-09T01:31:59-06:00").toInstant();
```

JRE 8 does not support all ISO 8601 offset formats.

dnsmkl:
I'll squash everything into 2 commits:

1. honor timezone option
2. Test migration to java.time

Ok?


> What name should I put in the release notes?

Danas Mikelinskas

ppkarwasz:
> I'll squash everything into 2 commits:
> 
>     1. honor timezone option
> 
>     2. Test migration to java.time
> 
> Ok?

Yes, please rebase and squash.

ppkarwasz:
@dnsmkl, thank you for your contribution.

PR #1195, opened 2023-01-09
Optimize Lazy::relaxed

jvz:
This fixes the relaxed Lazy<T> implementation to use the witness value from compareAndExchangeRelease as a potential return value rather than doing a second getAcquire to see the existing value. This adds some tests that are in Kotlin for its equivalent lazy implementations.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/util/LazyTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/util/LazyTest.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache license, Version 2.0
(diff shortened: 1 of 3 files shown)


rgoers (commented):
Looks Good to me

jvz:
I've realized recently that this class could use some better internal docs. I had to chase a lot of academic pointers to find out what the hell release/acquire memory ordering semantics means since the `VarHandle` API docs essentially tell you that "it works basically the same way as some feature in C and C++" which itself is a fairly recent addition to those language standards (C11 and C++11, though of course prior versions had non-standard extensions in various compilers to support similar features since they're relevant to implementing multithreaded runtimes). I hope that the API docs are sufficiently clear, though, on the reasoning behind using `Lazy::lazy` (the "strict" lazy) versus `Lazy::relaxed` (the "ok to compute multiple times as long as we all get the same published reference in the end" lazy) versus the other one (the "ok to compute multiple times as we always get the same result, so any thread can have any published value" lazy). Kotlin's own lazy class supports less than this, but they don't seem to use Java 11 yet.

PR #1196, opened 2023-01-09
Apply spotless automatically

jvz:
This adds executions during process-sources and process-test-sources to ensure that modified files are formatted properly.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1655,6 +1655,19 @@
               <goal>check</goal>
             </goals>
           </execution>
+          <execution>
+            <id>apply-spotless</id>
+            <goals>
+              <goal>apply</goal>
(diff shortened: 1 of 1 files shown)


ppkarwasz:
@jvz,

Wouldn't `process-test-sources` be enough? The idea is great: if you run tests on a PR, the code will be automatically formatted. However we have also to disable this in the CI, since it renders the `spotless:check` run useless.

jvz:
Oh shit that's a good point on the CI side. And I wasn't sure about the phase; I thought both would be required, but I suppose it depends on the particular plugin (I'm used to configuring the compiler and surefire plugins so much that I get explicit on the phases to attach to).

PR #1197, opened 2023-01-09
Serialize cached location

ppkarwasz:
The current code does not copy the location information if `includeLocation` is false.

The primary goal was to prevent an expensive call to `StackLocatorUtil`, but it also erases location obtained through other means (e.g. `LogBuilder#withLocation`). This PR fixes it.


jvz (approved):
LGTM. Note that I've proposed removal of Serializable in 3.0.0, so this is only relevant to 2.x.

PR #1199, opened 2023-01-11
LOG4J2-3228 - Remove support for java.io.Serializable

jvz:
This removes Serializable from various APIs including Message, Layout, LogEvent, Logger, and ReadOnlyStringMap. Java serialization is a security hazard. Messages and LogEvents already have numerous alternatives for generic serialization which should be used instead.


jvz:
Alright, @garydgregory, I've removed the unnecessary generic parameter. However, this raises a cleanup issue: we now have `AbstractLayout` and `AbstractStringLayout` which logically should be merged, but this isn't really that easy to do without making weird classes like `AbstractStringLayout.Serializer2` or breaking backward compatibility. Given the existing tests we have for `EcsLayout`, there's clearly some desire for binary compatibility with basic Log4j2 plugins, and the abstract layout classes are seemingly one of those. What do you think? Should I try to pull up whatever members we can that maintain binary compatibility, or should I just leave it as is?

log4j-1.2-api/src/main/java/org/apache/log4j/spi/LocationInfo.java
```diff
@@ -16,24 +16,21 @@
...
 
-import java.io.Serializable;
 import java.util.Objects;
 
 /**
  * The internal representation of caller location information.
  *
  * @since 0.8.3
  */
-public class LocationInfo implements Serializable {
```
ppkarwasz, line 27: I am not sure we ca remove these, since we need to maintain compatibility with Log4j 1.2.

log4j-api/src/main/java/org/apache/logging/log4j/message/ObjectMessage.java
```diff
@@ -115,28 +108,14 @@ private boolean equalObjectsOrStrings(final Object left, final Object right) {
 
     @Override
     public int hashCode() {
-        return obj != null ? obj.hashCode() : 0;
+        return obj.hashCode();
```
ppkarwasz, line 111: Are we certain nobody calls `new ObjectMessage(null)`?

jvz:
```
    public ObjectMessage(final Object obj) {
        this.obj = obj == null ? "null" : obj;
    }
```

Doesn't matter if they do.

ppkarwasz (approved):
I scrolled through about half of the modified files, but it looks good to me.

jvz:
I'm currently running the tests to make sure the revert of `log4j-1.2-api` code is ok with the rest of this PR before I push that up. Then I can merge from there after a PR run.

PR #1203, opened 2023-01-13
[LOG4J2-3647] Add global filter support to `LogBuilder`

ppkarwasz:
The way the `LogBuilder` works right now makes the following calls equivalent:

```lang-xml
logger.atTrace().withMarker(marker).log("Hello");
if (logger.isTraceEnabled()) {
    logger.trace(marker, "Hello");
}
```

If the user configures a global filter and sets a level more specific than `TRACE`, the code will not log anything, regardless of the result of the filter.

This patch:

 * introduces an appropriate `isEnabled(...)` check to `LogBuilder`, so that messages are filtered even if the `at*` methods were to return a real builder,
 * overrides the `at*` methods in Log4j2 Core: if no global filter is present the builder works the same way as before. However if a global filter is present, the `at*` methods always return **real** builder instances instead of no-ops,
 * overrides the logger in `log4j-to-slf4j`: if the backend is Logback (which also supports global filters), no filtering is performed by the Log4j2 API. Filtering is delegated to SLF4J.

# Benchmark results

Running the `LogBuilderMarkerFilterBenchmark` on a single thread of a 3.2 GHz Intel i7 gives the following results:

| Benchmark | Global Filter | Mode | Cnt | Score | Error | Units |
| - | - | - | - | - | - | - |
| logBuilderInfoNoMarker  | false  | thrpt | 25 |    401955,662 |    4452,326 | ops/s |
| loggerInfoNoMarker      | false  | thrpt | 25 |    401901,138 |    4962,817 | ops/s |
| logBuilderInfoNoMarker  |  true  | thrpt | 25 |    397392,140 |    5714,547 | ops/s |
| loggerInfoNoMarker      |  true  | thrpt | 25 |    401216,060 |    5157,227 | ops/s |
| logBuilderTraceNoMarker | false  | thrpt | 25 | 481090144,928 | 1047566,448 | ops/s |
| loggerTraceNoMarker     | false  | thrpt | 25 | 696911075,263 | 3323230,926 | ops/s |
| logBuilderTraceNoMarker |  true  | thrpt | 25 | 230876086,487 | 3658691,602 | ops/s |
| loggerTraceNoMarker     |  true  | thrpt | 25 | 545905032,453 | 3454089,549 | ops/s |
| logBuilderTraceMarker   | false  | thrpt | 25 | 479616595,520 | 2696417,835 | ops/s |
| loggerTraceMarker       | false  | thrpt | 25 | 604741975,172 | 4017689,657 | ops/s |
| logBuilderTraceMarker   |  true  | thrpt | 25 |    401843,967 |    2733,869 | ops/s |
| loggerTraceMarker       |  true  | thrpt | 25 |    401009,737 |    3303,312 | ops/s |

The logger was configured with a threshold of `INFO` and the global marker filter accepts all marked messages.

These figures suggest that:

 * when the logger is enabled, there is not statistical difference between `LogBuilder` and `Logger` or the presence/absence of a global filter,
 * when the logger is disabled, there is a 20% penalty for using a `LogBuilder` if no-op log builders are returned and a 50% penalty if a real log builder is returned. Anyway each operation takes no more than 15 CPU cycles, which I find acceptable.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/internal/DefaultLogBuilder.java
@@ -16,6 +16,8 @@
  */
 package org.apache.logging.log4j.internal;
 
+import java.util.Arrays;
+
 import org.apache.logging.log4j.BridgeAware;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogBuilder;
@@ -23,11 +25,11 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.Marker;
 import org.apache.logging.log4j.message.Message;
 import org.apache.logging.log4j.message.SimpleMessage;
+import org.apache.logging.log4j.spi.ExtendedLogger;
 import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.util.LambdaUtil;
 import org.apache.logging.log4j.util.StackLocatorUtil;
 import org.apache.logging.log4j.util.Supplier;
-
 
 /**
  * Collects data for a log event and then logs it. This class should be considered private.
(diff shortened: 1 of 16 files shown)


rgoers:
This makes no sense to me. Commenting on the Jira issue.

PR #1214, opened 2023-01-17
This change fixes incorrect behavior of stack elements cache.

alex-dubrouski:
Change of data structure from Stack to Deque (LIFO to FIFO) in this commit https://github.com/apache/logging-log4j2/commit/97f3153b73ab071a69c2e61e0130fd0104d45407 broke cache logic. This bug causes a major performance regression. 
Screen Shot 2023-01-11 at 6 31 58 PM
Screen Shot 2023-01-11 at 6 32 55 PM

I created a small local project to reproduce the issue. I can attach it if needed.

P.S. I signed ICLA and got a confirmation, but haven't got any response to my request to create a JIRA account, so can't create a bug.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
@@ -85,7 +85,7 @@ class ThrowableProxyHelper {
             stackLength = stackTrace.length;
         }
         final ExtendedStackTraceElement[] extStackTrace = new ExtendedStackTraceElement[stackLength];
-        Class<?> clazz = stack.isEmpty() ? null : stack.peek();
+        Class<?> clazz = stack.isEmpty() ? null : stack.peekLast();
         ClassLoader lastLoader = null;
         for (int i = stackLength - 1; i >= 0; --i) {
             final StackTraceElement stackTraceElement = stackTrace[i];
@@ -98,8 +98,8 @@ class ThrowableProxyHelper {
                 final CacheEntry entry = toCacheEntry(clazz, true);
                 extClassInfo = entry.element;
                 lastLoader = entry.loader;
-                stack.pop();
-                clazz = stack.isEmpty() ? null : stack.peek();
+                stack.pollLast();
+                clazz = stack.isEmpty() ? null : stack.peekLast();
             } else {
                 final CacheEntry cacheEntry = map.get(className);
                 if (cacheEntry != null) {
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
@@ -99,7 +99,7 @@ class ThrowableProxyHelper {
                 extClassInfo = entry.element;
                 lastLoader = entry.loader;
                 stack.pollLast();
-                clazz = stack.isEmpty() ? null : stack.peekLast();
+                clazz = stack.peekLast();
             } else {
                 final CacheEntry cacheEntry = map.get(className);
                 if (cacheEntry != null) {
--- a/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/StackLocatorTest.java
+++ b/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/StackLocatorTest.java
@@ -16,16 +16,16 @@
  */
 package org.apache.logging.log4j.util.java9;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertSame;
-
 import java.util.Deque;
 import java.util.Stack;
 
 import org.apache.logging.log4j.util.StackLocator;
 import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
 
 public class StackLocatorTest {
 
@@ -133,8 +133,15 @@ public class StackLocatorTest {
          */
         final StackTraceElement element = new Foo().foo();
         assertEquals("org.apache.logging.log4j.util.java9.StackLocatorTest$Foo", element.getClassName());
-        // The line number below may need adjustment if this file is changed. 
+        // The line number below may need adjustment if this file is changed.
         assertEquals(100, element.getLineNumber());
+    }
+
+    @Test
+    public void testTopElementInStackTrace() {
+        final StackLocator stackLocator = StackLocator.getInstance();
+        final Deque<Class<?>> classes = stackLocator.getCurrentStackTrace();
+        assertSame(StackLocator.class, classes.getFirst());
     }
 
     @Test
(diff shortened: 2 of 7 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
```diff
@@ -85,7 +85,7 @@ static ExtendedStackTraceElement[] toExtendedStackTrace(
             stackLength = stackTrace.length;
         }
         final ExtendedStackTraceElement[] extStackTrace = new ExtendedStackTraceElement[stackLength];
-        Class<?> clazz = stack.isEmpty() ? null : stack.peek();
+        Class<?> clazz = stack.isEmpty() ? null : stack.peekLast();
```
ppkarwasz, line 88: [`Deque#peekLast()`](https://docs.oracle.com/javase/8/docs/api/java/util/Deque.html#peekLast--) already returns `null` if the deque is empty, so the check is not necessary.

ppkarwasz (changes requested):
Hi @alex-dubrouski,

Thank you for your contribution. I think that the change that reversed the order of the stack trace is this one:

```lang-java
     static Deque<Class<?>> getCurrentStackTrace() {
         final Class<?>[] array = SECURITY_MANAGER.getClassContext();
         final Deque<Class<?>> classes = new ArrayDeque<>(array.length);
-        for (final Class<?> clazz : array) {
-            classes.push(clazz);
-        }
+        Collections.addAll(classes, array);
         return classes;
     }
```
in commit [b457cf74](https://github.com/apache/logging-log4j2/commit/b457cf74f4054b514c2f280928fc3280be6411e5). In a `Deque`, `add` is a synonym for `addLast`, while `push` a synonym for `addFirst`.

Can you also:
 * update the comment in `PrivateSecurityManagerStackTraceUtil#getCurrentStackTrace` to point out that the order is actually reversed,
 * fix `StackLocator#getCurrentStackTrace()` to use `addLast` instead of `push`,
 * add a regression test,
 * add a changelog entry in `src/changelog/.2.x.x`. The naming convention is `issueId_mnemonic`, e.g. `1214_fix_stacktrace_order`.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java
alex-dubrouski: Screen Shot 2023-01-17 at 2 43 45 PM
Screen Shot 2023-01-17 at 2 46 02 PM
Order of elements in stack did not change between 2.17.1 and 2.19.0 at least from the index number point of view.
I might have misunderstood you, sorry about that. Would like me to change the order of stack elements and preserve the current code in `ThrowableProxyHelper` or keep my changes to helper and don't change `StackLocator` as I assume these changed are mutually exclusive.
I might need some time to develop a unit test as there is no tests for helper class so far

ppkarwasz: Yes, I was referring to the broken slow path, which still uses `push` instead of `addLast`.

Also:

* `stack.isEmpty() ? null : stack.peekLast()` can be replaced with `stack.peekLast()`,
* the comment in `org.apache.logging.log4j.util.PrivateSecurityManagerStackTraceUtil` states _"The size of the Deque is the number of methods on the execution stack. The first element is the class that started execution on this thread..."_ and is clearly not up-to-date.

Testing the slow path is not easy: you need to create a test named `*IT` that has a security manager that does not allow the "createSecurityManager" permission. See `EnvironmentPropertySourceSecurityManagerIT` for inspiration.

ppkarwasz:
`StackLocator` is a complex feature with several different paths, so maybe I'll add a cheatsheet for the unity tests:

 * the [`StackLocator#getCurrentStackTrace`](https://github.com/apache/logging-log4j2/blob/543bc664a88558f5a0bb523cbe7bbbc448e8d0e5/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java#L84) version in `log4j-api-java9` is used when running Java 9+. The tests are in the same module.
 * the [`StackLocator#getCurrentStackTrace`](https://github.com/apache/logging-log4j2/blob/543bc664a88558f5a0bb523cbe7bbbc448e8d0e5/log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocator.java#L197) version in `log4j-api` is used when running Java 8. It uses two implementations:
   * the _fast path_ provided by [`PrivateSecurityManagerStackTraceUtil`](https://github.com/apache/logging-log4j2/blob/543bc664a88558f5a0bb523cbe7bbbc448e8d0e5/log4j-api/src/main/java/org/apache/logging/log4j/util/PrivateSecurityManagerStackTraceUtil.java#L65) needs to be tested in the `log4j-api-test` module (in a "normal" `*Test` class),
   * the _slow path_ requires running the test with a security manager enabled. The test should be in the `log4j-api-test` module and be named `*IT`. The reason behind this is that "normal" tests in `log4j-api-test` run in parallel. Setting a security manager is a very disruptive operation, so we have a second test execution that runs `*IT` classes with a JVM-per-test configuration.

alex-dubrouski:
I had quite hard time developing unit test for this class as it contains only static and private static methods. Final test verifies items in cache after executing "toExtendedStackTrace" method. Please provide your feedback and ideas.
I will work on integration test for StackLocator

ppkarwasz:
@alex-dubrouski,

Feel free to change `private` methods to package access. I think that we just need to test that `StackLocatorUtil#getCurrentStackTrace()` returns a `Deque` that has the calling class as first element.

alex-dubrouski:
> 

Are you sure calling class should be the first? Screenshots from test project I attached before have calling class as the last element (index 26)

ppkarwasz:
Yes, I would expect the class that called `StackLocatorUtil` at the lowest index (or at least near the top), while the class that initiated execution on the thread (JUnit?) at the end.

alex-dubrouski:
> Yes, I would expect the class that called `StackLocatorUtil` at the lowest index (or at least near the top), while the class that initiated execution on the thread (JUnit?) at the end.

Java9 StackLocator uses "add" method which is for ArrayDeque equals to addLast, so it works as expected. Plus it has a unit test that verifies order of elements already (https://github.com/apache/logging-log4j2/blob/release-2.x/log4j-api-java9/src/test/java/org/apache/logging/log4j/util/java9/StackLocatorTest.java#L60). It also uses WALKER by default as "PrivateSecurityManagerStackTraceUtil.isEnabled()" always returns False. I assume it means we can skip additional tests for Java9 code

ppkarwasz (approved):
@alex-dubrouski,

This looks good to me. Unless other committers have remarks, I'll merge it in a couple of days.
Thank You for your work.

alex-dubrouski:
Hi @ppkarwasz 
I am sorry for bothering, is it possible to merge this PR and release a new patch version? We would like to upgrade our applications.

jitheshtr:
Hi maintainer team, we are blocked on the availability of new log4j version with this patch for our service upgrade due to transitive dependency on slf4j-api 2.0.6 (slf4j-api 2.0 is released in Aug 2022, and hence should require log4j > 2.18 to work, and notice that we have recent log4j patches from January to accommodate slf4j). Do you know when the next log4j version release would be available for consumption?

ppkarwasz:
@jitheshtr,

You should ask on the [`log4j-user`](https://logging.apache.org/log4j/2.x/mail-lists.html) mailing list.

As far as SLF4J-to-Log4j2 API support goes, nothing relevant happened since the release of `2.19.0`. [SLF4J-560](https://jira.qos.ch/browse/SLF4J-560) should be of no concern to end users.

ppkarwasz:
@alex-dubrouski,

Thanks for your contribution. Our snapshots are generated after every CI run on `https://repository.apache.org/content/groups/snapshots/`. Can you check if `2.19.1-SNAPSHOT` solves your problem?

alex-dubrouski:
@ppkarwasz,

We ran JMH benchmarks and results are promising, but we can't use snapshots in Production applications. That's why we have to wait till next GA release.
Thanks

ppkarwasz:
@alex-dubrouski,

Sure, I was just double-checking that the next release will fix the problem.

We are currently in the process of wrapping version `2.20.0` up, so that we can concentrate on rendering `StackLocator` obsolete (see apache/logging-log4j-transform#1) and replace it with compile-time weaving. That should boost performance even further.

davemessink:
Hey @alex-dubrouski - Great job on this!  Thank you.

PR #1218, opened 2023-01-21
Use changelog plugin

rgoers:
Use log4j changelog plugin instead of exec plugin. No changelog entry is required since the changelog feature is new to this release.

Note - this uses the unreleased SNAPSHOT of log4j-tools so the CI builds will fail until log4j-tools is released and the version is changed here, so this should NOT be merged until that happens.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -287,8 +287,7 @@
     <docker-maven-plugin.version>0.40.2</docker-maven-plugin.version>
     <exam-maven-plugin.version>4.13.5</exam-maven-plugin.version>
     <findsecbugs-plugin.version>1.12.0</findsecbugs-plugin.version>
-    <log4j-maven-changelog-plugin.version>0.2.0-SNAPSHOT</log4j-maven-changelog-plugin.version>
-    <log4j-tools.version>0.1.0</log4j-tools.version>
+    <log4j-tools.version>0.2.0-SNAPSHOT</log4j-tools.version>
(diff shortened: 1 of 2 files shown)


ppkarwasz (approved):
The changelog itself works nicely, but I find the generated changelog a little bit verbose: our names are all over the place.

Instead of:

 * Add support for timezones in RollingFileAppender date pattern (for [LOG4J2-1631](https://issues.apache.org/jira/browse/LOG4J2-1631) by Piotr P. Karwasz, Danas Mikelinskas)

 I would prefer something like:

 * [LOG4J2-1631](https://issues.apache.org/jira/browse/LOG4J2-1631): Add support for timezones in RollingFileAppender date pattern. Thanks to Danas Mikelinskas. (pkarwasz)

In the next `log4j-changelog` version, we might add a `<committer>` tag or author's `role` so that we can distinguish between "Contributed by" (PR author) and "Thanks to" (issue reporter).

rgoers:
I am going to have to close this PR and do it again. With the changes Volkan made to the pom fixing the merge conflicts is too difficult.

rgoers:
@ppkarwasz While I agree with your comments they really belong as an issue int the log4j-tools repo. The template can't be changed here.

ppkarwasz:
@rgoers,

The template is in this repo as a hidden file: https://github.com/apache/logging-log4j2/blob/release-2.x/src/changelog/.changelog-entries.adoc.ftl

Issue #1229, opened 2023-01-27
JMX should be disabled by default


jvz:
## Description

JMX is an older API from Java for recording various metrics and publishing them along with other remote management features. It seems as though the [JDK Flight Recorder](https://docs.oracle.com/en/java/javase/11/docs/api/jdk.jfr/module-summary.html) API would be more appropriate to use in 3.x, though in the meantime, it would be nice to make JMX usage opt-in as it increases the time for a cold app start and is otherwise less commonly used these days thanks to things like JFR.

If disabling JMX entirely seems too drastic, then the JMX functionality that exists in Log4j should be broken up into more fine-grained feature flags to allow for selective use of it.

sdavids:
https://github.com/apache/logging-log4j2/pull/2468

PR #1259, opened 2023-02-02
[#1229] Disable write access by default

ppkarwasz:
This PR introduces a `log4j2.jmxReadOnly` property with a default value of `true` that disables all JMX setters. This solves #1229.


ppkarwasz:
This might arguably be useless, since JMX already provides a setting to switch JMX write access off: https://docs.oracle.com/en/java/javase/17/management/monitoring-and-management-using-jmx-technology.html#GUID-D4CBA2D6-2E24-4856-A7D8-62B3DFFB76EA

PR #1266, opened 2023-02-05
Improves changelog template

ppkarwasz:
This changes the changelog template to:

 * add issue links at the beginning,
 * strip Log4j2 committers from the credit line,
 * add a discrete "(<committer_id>)" at the end of the description line.

### DIFF
--- a/src/changelog/.changelog-entries.adoc.ftl
+++ b/src/changelog/.changelog-entries.adoc.ftl
@@ -14,35 +14,58 @@
    See the License for the specific language governing permissions and
    limitations under the License.
 -->
+<#function isCommitter id>
+  <#switch id>
+    <#case "bbrouwer">
+    <#case "ckozak">
+    <#case "ggregory">
+    <#case "mattsicker">
+    <#case "mikes">
+    <#case "nickwilliams">
+    <#case "pkarwasz">
+    <#case "ppkarwasz">
+    <#case "rgoers">
+    <#case "rgrabowski">
+    <#case "rgupta">
+    <#case "rpopma">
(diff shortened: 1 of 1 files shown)


ppkarwasz:
@rgoers,

What do you think about this template? The old one always credited committers in the by-line. I think this is too verbose since:
 * we are already credited elsewhere,
 * we shouldn't be credited for PRs that we just merged (there is no way to distinguish PRs from issue reports).

In the future I would to add metadata to the changelog format to distinguish between:
 * people that reported an issue ("Thanks to"),
 * people that reported an issue and contributed a solution ("Contributed by"),
 * committers that approved and take responsibility for the code change.

vy:
## Before

image

## After

image

Note that I needed to adopt the changelog-type headers to match the fix I have pushed earlier today.

## Feedback

I don't agree with this change for the following reasons.

* A brief summary of the change should come first, that is what matters most, not the ticket ID. Many projects (Keep-a-Changelog, Spring, etc.) follow this convention.
* When a change is associated with multiple tickets, the reading flow will be broken since ticket IDs prefix the entry.
* Why would anybody want to see an ID instead of the actual name?
* I am strictly against author types; committer, reviewer, triage professional, reporter, etc. For one, it is a trap to come up with an exhaustive list, because nobody cares. The names are there simply to get the credits right. Those who are interested in who did what, can simply click on the ticket linked.
* The current version reads natively, i.e., _"&lt;description&gt; (for &lt;ticket&gt; by &lt;author&gt;)"_, whereas the new one doesn't. That is, the readability is degraded, in particular, given both almost occupy the screen real estate.

PR #1268, opened 2023-02-07
Making legacy normalized properties from the same source have higher …

adwsingh:
Fix for https://issues.apache.org/jira/browse/LOG4J2-3621

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/util/PropertiesUtilOrderTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/util/PropertiesUtilOrderTest.java
@@ -14,14 +14,7 @@
  * See the license for the specific language governing permissions and
  * limitations under the license.
  */
-
 package org.apache.logging.log4j.util;
-
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
```diff
@@ -71,6 +71,20 @@ default CharSequence getNormalForm(Iterable<? extends CharSequence> tokens) {
...
+     * This is not supposed to be implemented by newer implementations, but only exists to add compatibility support
+     * for existing implementations.
+     *
+     * @param tokens list of property name tokens
+     * @return a normalized legacy property name using the given tokens
+     */
+    default CharSequence getLegacyNormalForm(Iterable<? extends CharSequence> tokens) {
+        return null;
+    }
+
```
ppkarwasz, line 87: There is no legacy normal form.

Legacy properties have an incoherent naming policy (cf. [documentation](https://logging.apache.org/log4j/2.x/manual/configuration.html#SystemProperties)), that is why they were replaced with the new normalized version.

However Log4j2 code uses **only** legacy properties internally, so you can be sure that the parameter of an `Environment#get(String)` call is a legacy property.

ppkarwasz (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
adwsingh: Updated the PR.

ppkarwasz (approved):
@adwsingh,

Looks good to me.

Could you add an entry in `src/changelog`?

PR #1276, opened 2023-02-16
Improving error messages in PathCondition

### LINKED ISSUE #1231: Improve validation of `PathCondition`s
## Description

As reported in [this StackOverflow question](https://stackoverflow.com/q/75091192/11748454) if the configuration does not contain all **required** properties, at least the `IfLastModified` fails with a `NullPointerException` instead of a **proper error message**.

We should probably check all components in `org.apache.logging.log4j.core.appender.rolling` for similar issues.

adwsingh:
Improving error messages in PathCondition. Fixes #1231

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfAll.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfAll.java
@@ -36,7 +36,7 @@ public final class IfAll implements PathCondition {
     private final PathCondition[] components;
 
     private IfAll(final PathCondition... filters) {
-        this.components = Objects.requireNonNull(filters, "filters");
+        this.components = Objects.requireNonNull(filters, "At-least 1 PathCondition is required in IfAll.");
     }
 
     public PathCondition[] getDeleteFilters() {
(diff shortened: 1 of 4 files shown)


ppkarwasz:
@adwsingh,

In #1231 I was thinking about something slightly different. There are two ways to instantiate a Log4j2 component like `IfLastModified`:

 * you can call `IfLastModified#createAgeCondition` in your code. We don't recommend such a programmatic approach, so if you pass `null` as `age` parameter, IMHO you deserve the NPE with a standard message,
 * you can declare it in your configuration. In such a case the component is instantiated by [`PluginBuilder`](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/config/plugins/util/PluginBuilder.html) that validates all constraints and logs the validation errors appropriately.
 
I think that the path conditions just lack an appropriate [`@Required`](https://logging.apache.org/log4j/2.x/log4j-core/apidocs/org/apache/logging/log4j/core/config/plugins/validation/constraints/Required.html) annotation on those parameters that are required.

We also have some complex constraints that are not supported by `PluginBuilder`: e.g. `IfFileName` requires a glob **or** a regex. To deal with this case `IfFileName#createNameCondition` should check if exactly one of `glob` and `regex` is `null`. If this is not the case the method should log an error (`StatusLogger`) and return `null`.

vy:
@adwsingh, we have renamed `release-2.x` and `master` branches to `2.x` and `main`, respectively. As an unexpected outcome, all PRs targeting those branches are closed. Would you mind re-creating your PR against these new branches, please?

PR #1277, opened 2023-02-16
Moving away from PARAM_VALUE_ESCAPE_PATTERN in Rfc5424Layout

### LINKED ISSUE #1224: `Rfc5424Layout` shouldn't match regular expressions against user input
`escapeSDParams()` called by `appendMap()` matches `PARAM_VALUE_ESCAPE_PATTERN` expressions against MDC data. This can be improved by a manual comparison.

adwsingh:
Moving away from PARAM_VALUE_ESCAPE_PATTERN in Rfc5424Layout to manually checking for es-capable characters. Fixes #1224 

## Checklist

* `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `spotless:apply` and retry)
* Changes contain an entry file in the `src/changelog/.2.x.x` directory


log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -570,7 +570,15 @@ private void appendMap(final String prefix, final Map<String, String> map, final
     }
 
     private String escapeSDParams(final String value) {
-        return PARAM_VALUE_ESCAPE_PATTERN.matcher(value).replaceAll("\\\\$0");
+        StringBuilder output = new StringBuilder();
```
vy, line 573: @adwsingh, could you implement the following changes as well, please?
1. Remove the `PARAM_VALUE_ESCAPE_PATTERN` constant (reorganize the `import`s)
2. Don't instantiate the `StringBuilder` until you encounter one of the target characters. If nothing to be escaped, i.e., `output == null`, you can simply return the `value` as is.
3. Please stick to the coding style followed in the rest of the file; whitespace after `if`, decorate variables with `final` modifier wherever possible, etc.

adwsingh: Sure @vy . I wanted to remove `PARAM_VALUE_ESCAPE_PATTERN` but didn't because it was public.

vy:
@adwsingh, mind adding a `src/changelog/.2.x.x/1277-replace_regexp_in_Rfc5424Layout.xml` file too, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
ppkarwasz: @vy: I agree with @adwsingh: unfortunately public constants must be kept (but can be `@Deprecated`), because this is not an internal class.

ppkarwasz (approved):
LGTM

PR #1278, opened 2023-02-16
Don't use regex to detect MessagePattern

### LINKED ISSUE #1223: FormattedMessage doesn't need to use regular expressions
In order to determine which message format syntax to use in `FormattedMessage`, this uses a regular expression. This can be simplified by flipping the ordering to check for a parametrized message first before assuming it's a printf-style message.

adwsingh:
Don't use regex to detect MessagePattern. Fixes #1223

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/FormattedMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/FormattedMessage.java
@@ -33,8 +33,6 @@ public class FormattedMessage implements Message {
 
     private static final long serialVersionUID = -665975803997290697L;
     private static final int HASHVAL = 31;
-    private static final String FORMAT_SPECIFIER = "%(\\d+\\$)?([-#+ 0,(\\<]*)?(\\d+)?(\\.\\d+)?([tT])?([a-zA-Z%])";
-    private static final Pattern MSG_PATTERN = Pattern.compile(FORMAT_SPECIFIER);
 
     private String messagePattern;
     private transient Object[] argArray;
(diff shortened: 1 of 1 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/message/FormattedMessage.java
```diff
@@ -184,20 +182,20 @@ protected Message getMessage(final String msgPattern, final Object[] args, final
...
         try {
-            if (MSG_PATTERN.matcher(msgPattern).find()) {
-                return new StringFormattedMessage(locale, msgPattern, args);
+            if (ParameterizedMessage.countArgumentPlaceholders(msgPattern) > 0) {
+                return new ParameterizedMessage(msgPattern, args, aThrowable);
             }
         } catch (final Exception ignored) {
-            // Also not properly formatted.
+            // Not a valid parameterized message.
         }
```
ppkarwasz, line 197: I don't believe we need a `try...catch`.

adwsingh: I felt so but erred on the side of caution. Removed now.

vy:
@adwsingh, we have renamed `release-2.x` and `master` branches to `2.x` and `main`, respectively. As an unexpected outcome, all PRs targeting those branches are closed. Would you mind re-creating your PR against these new branches, please?

rgoers:
@vy This is really nasty. There has to have been a better way to do this than to cause all these to be closed. There are a bunch of good PRs that got hit.

PR #1290, opened 2023-02-27
Set codeql workflow top level permission

### LINKED ISSUE #1289: Set permissions to codeql github workflow
**Warning!**
It is highly recommended to discuss feature requests in [the mailing lists](https://logging.apache.org/log4j/2.x/support.html) first.

I'm talking on behalf of Google and the OpenSSF.

There is a known issue of github workflow that it grants write permission to all workflows unless defined otherwise, thus, it is both a recommendation from [OpenSSF Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and the [Github](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions) to always use credentials that are minimally scoped.

I've seen that almost all of logging-log4h2 workflows already has the permissions minimally scoped, except for codeql that, although it has the permissions set at the job, it has no top level permission defined. Just to guarantee that no job eventually added to the workflow will have undesirable write permissions, I'll send a suggestion setting the top level permission as none. 

Feel free to reach me out in case of any doubts or concerns.

joycebrum:
## Changes
Closes #1289 
- Set codeql workflo top level permission as read-all, according to other workflows' top level permissions.

## Checklist

- [ ] Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise
    >I've based on `release-2.x` since basing on 2.x was causing some conflict on CONTRIBUTING.md. Let me know if I should base on `2.x`anyway
- [ ] `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `spotless:apply` and retry)
- [ ] Changes contain an entry file in the `src/changelog/.2.x.x` directory
    > Is it also applicable to this type of changes (on the workflows)?
- [ ] Tests for the changes are provided
- [x] [Commits are signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (optional, but highly recommended)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -24,6 +24,8 @@ on:
   schedule:
     - cron: '32 12 * * 5'
 
+permissions: read-all
+
 jobs:
 
   analyze:


ppkarwasz:
@joycebrum,

This was closed automatically by Github because we renamed the `release-2.x` branch to `2.x`. Feel free to resubmit it.

PR #1312, opened 2023-02-28
Add environment variable arbiter

MouettE-SC:
This is a copy of old PR #682 which was closed following branch renames.

This is a simple arbiter using environment variables functionnally based on the existing system property arbiter (on which I corrected some typos). Tests are included using system-lambda and review remarks from #682 have been included also.


ppkarwasz:
This was squashed into e0fd5cc245e3fd8325fcbd73cca83022aec84add and merged.

Thank you very much @MouettE-SC.

PR #790, opened 2022-03-11
[LOG4J2-3040] Avoid ClassCastException in JeroMqManager with custom LoggerContextFactory

grgrzybek:
Signed-off-by: Grzegorz Grzybek <gr.grzybek@gmail.com>

I tried using `org.apache.logging.log4j.test.junit.LoggerContextFactoryExtension`, but somehow I couldn't make it work (maybe incompatible with `org.apache.logging.log4j.core.test.junit.LoggerContextRule`?

PR #1313, opened 2023-03-01
[LOG4J2-3040] Avoid ClassCastException in JeroMqManager with custom LoggerContextFactory

grgrzybek:
Created after #790 was closed because of the branch rename.


ppkarwasz:
Apparently this was already included into 2.18.0 (cf. [changelog](https://logging.apache.org/log4j/2.x/release-notes/2.18.0.html#fixed)), but the attribution might have been copy-pasted. We are really sorry about that, it will be fixed when we republish the changelog.

grgrzybek:
no problem - thanks ;)

Issue #1319, opened 2023-03-02
Spotless failures should generate artifacts


jsoref:
When spotless fails, all of the files it wants to change and the diffs it produces should be uploaded as artifacts using `actions/upload-artifact` so that contributors (especially drive-by) can see the things that it doesn't like and apply them directly without having to go through 50 iterations to get to a satisfactory state.

ppkarwasz:
Since `./mvnw spotless:apply` is a simple fix to formatting problems and it is mentioned in the error message, I don't see the need for a sophisticated formatting mechanism.

jsoref:
A quick skim shows that this repository refuses to build with anything other than java 11:
https://github.com/apache/logging-log4j2/blob/59275a93429ff5a2073aee507c3c3c07da3825cd/pom.xml#L315

Java 11 reached EOL in September and my employer just replaced Java 11 with Java 21.

My computer's default version of java is 21. It's true that I still have a couple of jdk11s installed, but I should garbage collect them.

When playing in the GitHub ecosystem, it's really nice to drive-by contributors if your CI cycles provide outputs for things you've already spent gobs of cycles on so that others don't have to waste their time regenerating them. This is one of those things. And trust me, I spent *gobs* of time trying to get the spelling PR ready for this repository banging my head against spotless.

Not to point fingers, but your responses to my requests have been incredibly counterproductive. I asked a simple question based on the fact that I had seriously tried and failed to get things to work and you gave a response that didn't work:
https://github.com/apache/logging-log4j2/pull/613#issuecomment-1449476254

I'm asking for something simple for which I'm very willing (and able!) to provide an implementation.

Anyway, I've wasted quite a bit of time on this repository, so when someone is open to assisting me in contributing, I'm available to contribute, until then, I'm writing this repository off.

ppkarwasz:
@jsoref,

Yes, we are aware of the Java 11 limitation (cf. #1851). The only reason it is in place is because some tests are failing with JDK 17 (and even more with JDK 21).

Remark that due to your comments:

- we have fixed Spotless so that it works with every module (including `log4j-bom`, which didn't have a Spotless configuration),
- we have formatted the entire source code using it, so that contributors are not asked to reformat code they did not modify,
- recently we reformatted everything again with a deterministic formatter ([palantir-java-format](https://github.com/palantir/palantir-java-format)).

So I wouldn't say you wasted your time writing those comments.

Issue #1216, opened 2023-01-18
Refresh CONTRIBUTING.md


ppkarwasz:
The [`CONTRIBUTING.md`](https://github.com/apache/logging-log4j2/blob/release-2.x/CONTRIBUTING.md) file needs some updates. E.g. the requirement for "minimal updates" needs to be more precise: Spotless might require reformatting entire files.

PR #1320, opened 2023-03-02
Resolve spotless complaints

jsoref:
`spotless` is a real pain for drive-by-contributors (c.f. https://github.com/apache/logging-log4j2/issues/1216#issue-1538041705, there was an attempt in #1196, https://github.com/apache/logging-log4j2/issues/1319)

My understanding is that the advice to everyone is that they'll need to run spotless on any files they change as part of any effort to make a pr and to tolerate any merge conflicts. That should mean that running spotless on the entire repo is fair game as it's equivalent to anyone making a PR and losing a race against any other PR.

There's one notable problem with this PR, namely: https://github.com/apache/logging-log4j2/issues/1317, but someone can fix spotless for that and then fix the entire codebase. (That someone might be me if someone can explain what spotless is doing/why)

## Checklist

- [x] use `main` otherwise
- [ ] `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `spotless:apply` and retry)
- [ ] Changes contain an entry file in the `src/changelog/.2.x.x` directory
- [x] Tests for the changes are provided -- spotless passes
- [ ] [Commits are signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (optional, but highly recommended) -- happy to do this if requested, but atm it's just overhead, and this took ~50 attempts.


ppkarwasz:
@jsoref,

When I introduced Spotless last September, I was looking for a fast way to remove most useless differences (whitespace and import ordering) between `2.x` and `main`, **without** one big commit. So I used the "ratchet" feature provided by Spotless. After 6 months of usage I must agree with you: we need to fix **all** files at once: "ratchet" does not work outside of a Git working tree and requires the CI to clone multiple commits instead of just the last one.

If we are going to do that, we need to:

 * discuss **all** aspects of code style on the `dev@logging.apache.org` mailing list: the current Spotless configuration has a minimal set of rules and I would want to merge a big commit every time we add a code style rule. A complete [Java style guide](https://logging.apache.org/log4j/2.x/javastyle.html) is on our web site. This could be a starting point for a discussion, but I am sure that some rules are no longer supported.
   Line length, modifier order, `final` keyword, chained invocations wrapping: these are only some examples of rules that are not currently enforced by Spotless, but need to be agreed on.
 * propose a Spotless configuration in one PR,
 * apply the configuration to all files in a second PR (so that committers can verify it more easily).

blacelle:
Hello. Would you consider https://github.com/marketplace/cleanthat/ to apply spotless automatically to incoming PRs ?

(I'm the author of Cleanthat, and a Spotless contributor).

blacelle:
> Line length, modifier order, final keyword, chained invocations wrapping: these are only some examples of rules that are not currently enforced by Spotless, but need to be agreed on.

Line-length can be handlded in spotless with one of the formatting options (Eclipse, GoogleJavaFormat, etc)
Modifier Order is supported through Cleanthat. Unnecessary final keywords also.
I would need more details about expectations around `chained invocations wrapping`.

ppkarwasz:
This has been solved in #1446.

PR #1322, opened 2023-03-03
Auto-merge Dependabot PRs

ppkarwasz:
If the build succeeds, we want to automatically merge Dependabot PRs. Exceptions to the rules should be explicitly written into `dependabot.yml` (i.e. do not close Dependabot PRs, add an exception).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,10 +19,68 @@ updates:
 - package-ecosystem: maven
   directory: "/"
   schedule:
-    interval: monthly
-  open-pull-requests-limit: 50
+    interval: "daily"
+  target-branch: "2.x"
+  ignore:
+    # Jetty 10.x does not have an internal logging API
+    - dependency-name: "org.eclipse.jetty:*"
+      update-types: ["version-update:semver-major"]
+    # EclipseLink 3.x is Jakarta EE 9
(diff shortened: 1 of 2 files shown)


vy:
Great job @ppkarwasz! :100: AFAIK, this setup would run the build workflow twice: the 1st one while verifying the PR, and the 2nd one after merging it. You can avoid this by adding the following block to `build.yml`:

```
on:
  push:
    # Avoid workflow run for _merged_ `dependabot` PRs.
    # They were (hopefully!) already tested in PR-triggered workflow.
    branches-ignore: "dependabot/**"
```

See [my rotating-fos configuration](/rotating-fos/blob/master/.github/workflows/build.yml#L21) for an example.

You can this PR after this change, IMO.

ppkarwasz:
I am not sure if `branches-ignore` is what prevents the double run: I think that the "branch" setting checks the branch that was pushed (`2.x` in our case).

The `GITHUB_TOKEN` used by Github Actions does not trigger workflows according to https://docs.github.com/en/actions/security-guides/automatic-token-authentication, so no `push` event should be generated.

ppkarwasz:
This CI should prove that merging the PR does not cause a second CI run: https://github.com/ppkarwasz/logging-log4j2/pull/310

Issue #1391, opened 2023-03-26
NPE in PluginElementVisitor findNamedNode


azakordonets:
## Description

In PluginElementVisitor -> findNamedNode there's no check whether childType is null or not. Therefore it can throw nullpointer sometimes. 

## Configuration

**Version:** 2.20.0

**Operating system:** Mac OS  13.1

**JDK:** amazon-corretto-11.jdk

```java
if (childType == null) {
     //System.out.println();
}
if (name.equalsIgnoreCase(childType.getElementName()) ||
 this.conversionType.isAssignableFrom(childType.getPluginClass())) {
 // FIXME: check child.getObject() for null?
```
if `childType` is null, then the check won't do anything and nullpointer will be thrown on the next if statement

I have already create [PR](https://github.com/apache/logging-log4j2/pull/1392) to fix this issue

PR #1392, opened 2023-03-26
fix: fix NPE in PluginElementVisitor findNamedNode

azakordonets:
If "childType" is null, then the "if" block just get's skipped and on line 104 nullpointer will be called for `childType.getElement` command. 

This PR fixes the #1391 issue.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginElementVisitor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/visitors/PluginElementVisitor.java
@@ -98,13 +98,12 @@ public class PluginElementVisitor extends AbstractPluginVisitor<PluginElement> {
     private Node findNamedNode(final String name, final Iterable<Node> children) {
         for (final Node child : children) {
             final PluginType<?> childType = child.getType();
-            if (childType == null) {
-                //System.out.println();
-            }
-            if (name.equalsIgnoreCase(childType.getElementName()) ||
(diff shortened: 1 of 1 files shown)


jvz:
What do you think about the adjacent "FIXME" comment? Do you think this would fix that, too?

vy:
@azakordonets, thanks so much for the report and the fix! I would appreciate it if you can next time also attach a `src/changelog/.2.x.x/<ticketId>_<description>.xml` entry along your PR.

@jvz, I don't think that FIXME is relevant any more. `visit()` contains `namedNode = findNamedNode(...); ...; return namedNode.getObject()`and its return value is checked against `null` at every call site.

vy:
@azakordonets, if you have a test, I would be happy to merge that too. :heart_eyes:

azakordonets:
Hey @vy , sorry for not including changes into the `changes` file and the test. I saw it in the template, but I have not done that in the past, so I was going to ask for help to whoever would review this. Two questions: 
1. How and in which format should I include changes to `src/changelog/.2.x.x/<ticketId>_<description>.xml` ? Any examples that I could use ? 
2. I wanted to write a test, but I could not find any tests related in that module and I was not sure whether it's on purpose or not. If not, then I will add those somewhere this week

vy:
1. For changelog entry, just clone one of the XML files there. I already added one for you in bb0f38a50ffa12a86cef3db3096efa5351746357.
2. I had the impression you get this NPE from a particular configuration. If so, you can write a test that feeds this configuration and verify the successful bootstrap of the Log4j. (There are plenty of such tests in `log4j-core` for inspiration.)

PR #1404, opened 2023-03-31
Update URL for JDK bug

mernst:
This pull request updates a broken URL.


andrei-ivanov:
Isn't https://bugs.openjdk.org/browse/JDK-4228335 better?

mernst:
> Isn't https://bugs.openjdk.org/browse/JDK-4228335 better?

Yes, that works too.  I have changed the pull request.

vy (changes requested):
Thanks so much for the fix! Mind removing the changelog entry to avoid cluttering the release notes with trivial changes.

ppkarwasz:
@mernst,

This code comes from [Apache Commons Lang](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/time/FastDateParser.java). You should consider posting the correction also there.

mernst:
> Mind removing the changelog entry to avoid cluttering the release notes with trivial changes.

I will do so.

I was following the instructions at https://github.com/apache/logging-log4j2/blob/2.x/.github/pull_request_template.md .  I have opened #1421 to correct those, since they are apparently wrong.

mernst:
> This code comes from [Apache Commons Lang](https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/time/FastDateParser.java). You should consider posting the correction also there.

[Done.](https://github.com/apache/commons-lang/pull/1050)

PR #1419, opened 2023-04-08
Fix `Log4jFixedFormatter` buffer length

### LINKED ISSUE #1418: ArrayOutOfBoundsException in Log4jFixedFormatter
The current buffer used by `Log4jFixedFormatter` is too small for formats that contain a timezone specifier, since the size of the buffer is equal to the patterns length and a `Z` pattern produces 5 characters.

ppkarwasz:
This PR fixes #1418 .

The buffer length in `Log4jFixedFormatter` was computed base on the length of the pattern. Now it is the double of the real length of a formatted date (in the English locale).

At the same time this migrates the `FastDateFormatter` tests to JUnit 5.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
@@ -163,6 +163,7 @@ public class FixedDateFormat {
         private final int millisSeparatorLength;
         private final int secondFractionDigits;
         private final FixedTimeZoneFormat fixedTimeZoneFormat;
+        private final int extraTimeZoneFormatLength;
 
         FixedFormat(final String pattern, final String datePattern, final int escapeCount, final char timeSeparator,
                     final int timeSepLength, final char millisSeparator, final int millisSepLength,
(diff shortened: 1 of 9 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
```diff
@@ -255,7 +261,7 @@ private static int[] nanoRange(final String pattern) {
          * @return the length of the resulting formatted date and time strings
          */
         public int getLength() {
-            return pattern.length() - escapeCount;
+            return pattern.length() - escapeCount + extraTimeZoneFormatLength;
```
vy, line 264: Given this is fixed, do we still need the `char[]`-size fix in `InstantFormatter`?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FixedDateFormat.java
```diff
@@ -503,6 +510,15 @@ public String getFormat() {
         return fixedFormat.getPattern();
     }
 
+    /**
+     * Returns the length of the resulting formatted date and time strings.
+     *
+     * @return the length of the resulting formatted date and time strings
```
vy, line 516: This description is not correct, otherwise we won't need to double the size of the buffer in `InstantFormatter`.

ppkarwasz: Right, I should add "in the ROOT locale" (which is basically English, but less anglocentric).

vy: I will appreciate that. Maybe also a note on _"if you want to determine the max. length, you better multiply this number by 2"_.

PR #1471, opened 2023-05-08
Add JUnit 5 test-method-scoped `LoggerContext`s

vy:
Adds `Log4jEventRecorder` et al. which provides JUnit 5 test-method-scoped `LoggerContext`s which can be used as follows:

```
@Log4jEventRecorderEnabled
class FooTest {

    @Test
    void exceptions_should_get_logged(Log4jEventRecorder log4jEventRecorder) {

        // Execute the logic
        Logger logger = log4jEventRecorder.getLoggerContext().getLogger(Foo.class);
        new Foo(logger).runBusinessLogic(...);

        // Verify the logging
        List<LogEvent> events = log4jEventRecorder.getEvents();
        // ...

    }

}
```

### DIFF
--- a/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/Log4jEventRecorder.java
+++ b/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/Log4jEventRecorder.java
@@ -16,6 +16,11 @@
  */
 package org.apache.logging.log4j.core.test.junit;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.LoggerContext;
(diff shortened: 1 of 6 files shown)


rgoers:
I really don't care for this solution. It presupposes that all one wants to do is log events to a list. The vast majority of our tests that need to run in parallel do not do this.

You are on the right track with storing the LoggerContext's in the environment but instead what should be implemented is a LoggerContextSelector that retrieve the correct logger context from the environment. The obvious choice would be to use the test class name as the key for the LoggerContext.

vy:
> I really don't care for this solution. It presupposes that all one wants to do is log events to a list. The vast majority of our tests that need to run in parallel do not do this.

Indeed it doesn't solve the problem in _our [Log4j] tests_, yet it solves dozens in my company. Hence, there are other use cases out there. That said, indeed solving it for all of them is better, if possible.

> You are on the right track with storing the LoggerContext's in the environment but instead what should be implemented is a LoggerContextSelector that retrieve the correct logger context from the environment. The obvious choice would be to use the test class name as the key for the LoggerContext.

Imagine I have two `should_jump()` and `should_crawl()` tests in a single test class and they verify the behaviour of logging component `Foo`. Each test method ideally should have its own `LoggerContext`. Though once `static final Logger LOGGER` field of `Foo` is initialized, it will trigger a single `LoggerContext` creation. Second, will this of `should_jump()` or `should_crawl()`? (Assume I want to run these tests in parallel.) In a nutshell, I am not able to see how to detach `LoggerContext`s of two different test methods. Any ideas?

rgoers:
I agree with @ppkarwasz's comment that creating new LoggerContexts for each test is wasteful. However, there are times when it is required. In many of those cases we handle it by shutting down the LoggerContext after each test method. But that doesn't allow for multi-threading. Does Junit not provide an annotation to indicate tests shouldn't be forked? Creating a ListAppender that uses a ThreadLocal seems incredibly easy to implement.

Issue #1476, opened 2023-05-11
`ParameterFormatter.appendDate` causes an `UnsupportedOperationException` with a `java.sql.Date`


gtoison:
## Description

`ParameterFormatter.appendDate` causes an `UnsupportedOperationException` when trying to call `java.sql.Date.toInstant()`

## Configuration

**Version:** 2.20.0

**Operating system:** W10

**JDK:** JDK temurin 20

## Logs

```
ERROR StatusConsoleListener An exception occurred processing Appender Console
 java.lang.UnsupportedOperationException
	at java.sql/java.sql.Date.toInstant(Date.java:316)
	at org.apache.logging.log4j.message.ParameterFormatter.appendDate(ParameterFormatter.java:492)
	at org.apache.logging.log4j.message.ParameterFormatter.appendSpecialTypes(ParameterFormatter.java:485)
	at org.apache.logging.log4j.message.ParameterFormatter.recursiveDeepToString(ParameterFormatter.java:474)
	at org.apache.logging.log4j.message.ParameterFormatter.appendMap(ParameterFormatter.java:589)
	at org.apache.logging.log4j.message.ParameterFormatter.appendPotentiallyRecursiveValue(ParameterFormatter.java:511)
	at org.apache.logging.log4j.message.ParameterFormatter.recursiveDeepToString(ParameterFormatter.java:478)
	at org.apache.logging.log4j.message.ParameterFormatter.appendArray(ParameterFormatter.java:557)
	at org.apache.logging.log4j.message.ParameterFormatter.appendPotentiallyRecursiveValue(ParameterFormatter.java:509)
	at org.apache.logging.log4j.message.ParameterFormatter.recursiveDeepToString(ParameterFormatter.java:478)
	at org.apache.logging.log4j.message.ParameterFormatter.recursiveDeepToString(ParameterFormatter.java:449)
	at org.apache.logging.log4j.message.ParameterFormatter.formatMessage2(ParameterFormatter.java:192)
	at org.apache.logging.log4j.message.ParameterizedMessage.formatTo(ParameterizedMessage.java:227)
	at org.apache.logging.log4j.core.pattern.MessagePatternConverter$SimpleMessagePatternConverter.format(MessagePatternConverter.java:121)
	at org.apache.logging.log4j.core.pattern.PatternFormatter.format(PatternFormatter.java:44)
[...]

PR #1477, opened 2023-05-11
Unit test reproducing issue #1476

gtoison:
`java.sql.Date.toInstant` throws an `UnsupportedOperationException`, we should not call that method when formatting an java.sql.Date instance

This should reproduce issue #1476


gtoison:
Does not reproduce the issue, sorry for the noise...

Issue #1211, opened 2023-01-16
Hybrid example uses wrong class


Deceptio:
It seems that this example is not quite correct on line 28? 
[Hybrid Example](https://logging.apache.org/log4j/2.x/manual/customconfig.html#Hybrid)

There is no XMLConfiguration class but only an XmlConfiguration class. 
Additionally the method createLogger is deprecated. 
And the variable "refs" is also not defined anywhere

Maybe you should write a new example for the hybrid application.

But maybe I understand the API wrong.

aawad6:
Fixed in https://github.com/apache/logging-log4j2/commit/0269ce259b6875c6082207970e81b14d56cb4444.

vy:
Fixed in #1478.

PR #1478, opened 2023-05-15
doc: Use the right classes in customconfig's Hybrid Example

AmmarYasser97:
Propose a new HypridExample.

As mentioned in [Issue#1211,](https://github.com/apache/logging-log4j2/issues/1211) the Hyprid Example in the customconfig documentation is broken. 

1. Uses a bunch of deprecated APIs, `FileAppender.createAppender` and `LoggerConfig.createLogger`
2. Using the wrong `XMLConfigration `class with all caps XML, instead of using the `XmlConfiguration `class.
3. Fix the constructor call for both `MyXmlConfigurationFactory `and `MyXmlConfiguration `calsses.
4. Missing the `AppenderRef[]` from the `createLogger `invokation.

### DIFF
--- a/src/site/xdoc/manual/customconfig.xml
+++ b/src/site/xdoc/manual/customconfig.xml
@@ -222,34 +222,36 @@ LoggerContext ctx = Configurator.initialize(builder.build());
             </p>
             <p>
               The easiest way to achieve this is to extend one of the standard Configuration classes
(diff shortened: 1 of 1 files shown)


vy:
@aawad6, thanks! :pray:
Keep these PRs coming! :rocket:

Issue #1479, opened 2023-05-16
JsonTemplateLayout Docs with wrong keyname appender.console.json.


ghost:
I think here in the examples for JsonTemplateLayout
https://github.com/apache/logging-log4j2/blob/5cc765d77daa68260dbbf79b6b2c0799d32f12bc/src/site/asciidoc/manual/json-template-layout.adoc.vm#L121-L122 the key name should be `.layout.` instead of `.json.`

```properties
appender.console.layout.type = JsonTemplateLayout
appender.console.layout.eventTemplateUri = classpath:EcsLayout.json
```

ppkarwasz:
@robert-gdv,

Thank you for your report. Actually the key used for a layout can be **arbitrary** (e.g. `!@#$%^&*()`) as long as it does not contain a dot `.` character.

When binding configuration values to JavaBean properties, Log4j:

 * binds simple properties (`int`, `String`, etc.) **by name**. Therefore you **need** to use the `target` key to specify the target of a console appender,
 * binds complex properties **by type**. That is why `appender.console.!@#$%^&*().type = JsonTemplateLayout` will work without any problems.

robert-gdv:
Hmm... reviewing the docs:
In https://logging.apache.org/log4j/2.x/manual/appenders.html#consoleappender a parameter layout is described. How does the property parser know, that `appender.console.!@#$%^&*().type` describes a layout?

PR #1480, opened 2023-05-16
.layout. instead of .json.

robert-gdv:
This request fixes https://github.com/apache/logging-log4j2/issues/1479


robert-gdv:
As discussed in https://github.com/apache/logging-log4j2/issues/1479 , the name is arbitrary, but must be used consistently in den configuration.
Since further down "layout" is used. It would be consequent, to use it here too.
Alternative: use everywhere the same, obviously arbitrary name. e.g. 
```properties
appender.myappender1.mylayoutA.type = JsonTemplateLayout 
appender.myappender1.mylayoutA.eventTemplateUri = classpath:EcsLayout.json
```

PR #1513, opened 2023-06-14
LOG4J2-3663 - fix file descriptor leak on Tomcat (config in JAR file)

lenoch7:
Fix problem with unclosed JAR file containing Log4j configuration (Tomcat - CachedResourceURLStreamHandler is configured), which causes that web application cannot be undeployed, because JAR file cannot be deleted (windows).

See detail description in [LOG4J2-3663](https://issues.apache.org/jira/projects/LOG4J2/issues/LOG4J2-3663).


ppkarwasz (approved):
@lenoch7,

Looks good to me. Thank you for your contribution.

Issue #1508, opened 2023-06-09
Zstd compression support


anuragagarwal561994:
At present log4j2 provides certain compression algorithms like

Gzip
Lz4
Etc.

Zstd in recent times have become popular and we have tried compressing our logs in production with zstd by running a fork of log4j2 and it does provide close to 20-30% saving with respect to gzip.

It depends on the file size actually. So not applicable for all use cases.

But log4j2 library today does not provide an extensible way to add a compression algorithm.

Do you think, addition of zstd to the list of compressions will be a good addition?

vy:
I really liked *both* ideas!

1. Making compression extensible (I actually have created [LOG4J2-3023](https://issues.apache.org/jira/browse/LOG4J2-3023) for something similar in the past)
2. Adding zstd

Are you interested in submitting a PR? If so, I can assist you with the design.

garydgregory:
FYI, Apache Commons Compress supports zstd and many other formats.

anuragagarwal561994:
@vy I am not sure if I will be able to contribute to in near term.

I can share the code we have been using for this purpose in our organisation. May be one doesn't have to do anything and just integrate that code itself, but I think when going open source some standards needs to be maintained.

Have been a bit occupied recently with work, still let me check if I get time over the weekend to look into this.

vy:
@anuragagarwal561994, could you at least share the outline of what you have at work? Is it a custom appender? A layout?

anuragagarwal561994:
So it is mostly built like Gzip only, just that instead of using GzipOurstream we are using ZstdOutputStream

We are using zstd-jni as a dependency here.

anuragagarwal561994:
@vy the MR is actually ready I have integrated it using commons compress itself and the tests were passing in the first go.

vy:
Thanks for the contribution @anuragagarwal561994. As I indicated in my response earlier, I was in the hope of a contribution to *first* make the compression extensible *and then* implement zstd by means of a simple plugin. (Hence, was my offer for assistance...) I am not much fond of these kinds of specialized solutions similar to the one you implemented in #1514, adding zstd only for the rolling appender. (For the records, I have also [raised this issue in the mailing list](https://lists.apache.org/thread/2kcktd7r29z1ojg6k02lr9zz3yvty9kq).)

If you would like to *collaborate* on a more general solution, please let me know.

anuragagarwal561994:
[...]

PR #1514, opened 2023-06-14
Support for ZSTD Compression

anuragagarwal561994:
This PR adds support for zstd compression using CommonsCompressAction.

Closes https://github.com/apache/logging-log4j2/issues/1508

### DIFF
--- a/log4j-core-test/pom.xml
+++ b/log4j-core-test/pom.xml
@@ -328,6 +328,11 @@
       <artifactId>xz</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.luben</groupId>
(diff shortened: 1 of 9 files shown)


anuragagarwal561994:
@vy you can check the MR here, just a suggestion `commons-compress` provides `CompressorStreamFactory` which provides constants for all the compressions available. Do you think instead of writing and changing comments everywhere we can use this constants instead and remove comments to be more verbose.

vy:
@anuragagarwal561994, I have merged this. Thanks for the contribution. This will be available in the `2.21.0` release that should come out sometime this week.

anuragagarwal561994:
Thanks @vy can finally remove a tech debt from our projects :D

PR #1518, opened 2023-06-16
Fix 3.0.0-alpha1 release notes

ppkarwasz:
The release notes for version 3.0.0-alpha1 contain features that were already shipped 2.20.0.

This PR tries to remove the additional entries and improve the overall formatting of the entries.


ppkarwasz:
@jvz,

Can you check your entries in the release notes? I have the impression that there are entries for features that were later removed from the codebase.

jvz:
The ones with my name as the author?

ppkarwasz:
Yes, they mostly deal with DI and I have the impression some may be included in others.
Maybe a single DI changelog entry with a longer description would be better?

PR #1537, opened 2023-07-04
(doc) fix the disableAnsi default value description

pbodnar:
The default value of the disableAnsi parameter (attribute) was changed for Windows by https://issues.apache.org/jira/browse/LOG4J2-2087. But the user documentation was not updated appropriately.


vy:
@pbodnar thanks for the report and the fix, much appreciated! :bow: Merging.

PR #1538, opened 2023-07-04
Avoid NPE from ContextSelector.getContext() when "entry=null" is passed

pbodnar:
This fixes an apparent bug which was reported already in https://issues.apache.org/jira/browse/LOG4J2-3217.

Note that only `ClassLoaderContextSelector` provided a custom implementation of this method, where it also checks for null-ness of `entry`. So all other selector classes like `BasicContextSelector ` would throw NullPointerException when e.g.
`Configurator.initialize(String, ClassLoader, URI)` is called.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ContextSelector.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ContextSelector.java
@@ -108,7 +108,7 @@ public interface ContextSelector {
     default LoggerContext getContext(String fqcn, ClassLoader loader, Map.Entry<String, Object> entry,
             boolean currentContext, URI configLocation) {
         final LoggerContext lc = getContext(fqcn, loader, currentContext, configLocation);
-        if (lc != null) {
+        if (lc != null && entry != null) {
             lc.putObject(entry.getKey(), entry.getValue());
         }
         return lc;


pbodnar:
Note: I've across this issue when trying to deactivate asynchronous logging as described in [CAS documentation](https://apereo.github.io/cas/6.6.x/logging/Logging.html#log4j2-properties):

> To turn off asynchronous logging, include the following in log4j2.component.properites or as a system property:
> 
>     Log4jContextSelector=org.apache.logging.log4j.core.selector.BasicContextSelector

Even after doing this and fixing the NPE, CAS still logs asynchronously, but that is possibly for another bug, possibly in another repository. 😅

---

Regarding unit tests and changelog - I'm not sure. Let me know if it is needed.

vy:
@pbodnar, thanks so much! I will take care of the changelog.

> Even after doing this and fixing the NPE, CAS still logs asynchronously

That indeed warrants another ticket.

log4j-core/src/main/java/org/apache/logging/log4j/core/selector/ContextSelector.java
```diff
@@ -108,7 +108,7 @@ default LoggerContext getContext(String fqcn, ClassLoader loader, Map.Entry<Stri
     default LoggerContext getContext(String fqcn, ClassLoader loader, Map.Entry<String, Object> entry,
             boolean currentContext, URI configLocation) {
         final LoggerContext lc = getContext(fqcn, loader, currentContext, configLocation);
-        if (lc != null) {
+        if (lc != null && entry != null) {
```
ppkarwasz, line 111: `lc` should be not `null` here, we just need to check if `entry` is not `null`.

vy: Since `getContext()` is to be implemented by the subclass, I thought we cannot have such an assumption. Nevertheless, if you think otherwise, be my guest to update it.

PR #1591, opened 2023-07-21
Document dependency upgrades for 2.20.1

ppkarwasz:
This PR adds the dependency bumps performed by Dependabot to the changelog of the next 2.x release.

### DIFF
--- a/src/changelog/.2.x.x/update-1runtime-angus-activation.xml
+++ b/src/changelog/.2.x.x/update-1runtime-angus-activation.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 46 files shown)


vy:
Could you place each update to a separate changelog entry file, please? If you stick to a certain naming convention in file names and descriptions, it would get formatted nicely.

Also please keep in mind that the whole idea of using separate changelog entry files is to avoid merge conflicts. One day :crossed_fingers: we can even make `dependabot` to create those files. Hence, squashing updates to a single entry is counter to the design principles of `log4j-changelog`.

ppkarwasz:
@vy,

I split all the updates into separate changelog entries. Looking at the result I see a problem with the way `log4j-changelog` classifies changes. I would say that in terms of importance we should sort sections as:
 - "Removed" (oops, something will break),
 - "Added" (yuppie, a new feature),
 - "Changed",
 - "Fixed",
 - dependencies should probably be a separate and last section, since they are the less important.

Issue #1545, opened 2023-07-08
HighlightConverter does not honor {STYLE_NAME} configuration


zartc:
## Description

When using the HighlightConverter like this: `%highlight{%5level}{LOGBACK}` the HighlightConverter always uses the colors of the DEFAULT preset instead of the color of the LOGBACK as asked.

## Configuration

**Version:** 2.20.0

**Operating system:** Windows 10

**JDK:** temirun-11

## Logs

none

## Reproduction

Configure a converter with: `%highlight{%5level}{LOGBACK}`
You will discover that the HighlightConverter always uses the colors of the DEFAULT preset.

## Fixe

`HighlightConverter.createLevelStyleMap()` method is missing a check at line 151 to see if the passed `options[1]` contains a map of LEVEL=color or a preset style name.

```java
if (!string.contains("=")) {
    return STYLES.getOrDefault(string, DEFAULT_STYLES);
}
```

ppkarwasz:
Hi @zartc,

As [documented](https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout) the correct syntax is:
```
%highlight{%5level}{STYLE=LOGBACK}
```

I wouldn't complexify the syntax to accept just "LOGBACK". There should be probably a warning if a user employs an incorrect syntax (without `=`). Could you submit a PR with the warning?

aawad6:
If @zartc doesn't mind, I can add the necessary warning message.

ppkarwasz:
@aawad6, go ahead!

vy:
@zartc, thanks so much for the report! :bow: The issue is fixed. It will be available with the next release.

PR #1637, opened 2023-07-30
Add a warning in case of incorrect syntax of highlighting style.

aawad6:
### Add a warning in case of incorrect syntax of highlighting style (without `=`).
This PR is related to [#issue1545](https://github.com/apache/logging-log4j2/issues/1545),  it meant to ensure that the user is providing a correct syntax for highlighting style.

e.g. `%highlight{%5level}{LOGBACK}` should be `%highlight{%5level}{STYLE=LOGBACK}`. For more information see the documentation [here](https://logging.apache.org/log4j/2.x/manual/layouts.html#PatternLayout).

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/HighlightConverter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/HighlightConverter.java
@@ -27,7 +27,6 @@ import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.layout.PatternLayout;
 import org.apache.logging.log4j.util.PerformanceSensitive;
-import org.apache.logging.log4j.util.Strings;
 
 import static org.apache.logging.log4j.util.Strings.toRootUpperCase;
 
@@ -86,6 +85,10 @@ public final class HighlightConverter extends LogEventPatternConverter implement
     private static final Map<String, String> LOGBACK_STYLES = new HashMap<>();
 
     private static final String STYLE_KEY = "STYLE";
+
+    private static final String DISABLE_ANSI_KEY = "DISABLEANSI";
+
+    private static final String NO_CONSOLE_NO_ANSI_KEY = "NOCONSOLENOANSI";
 
     private static final String STYLE_KEY_DEFAULT = "DEFAULT";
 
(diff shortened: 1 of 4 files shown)


aawad6:
@ppkarwasz I wonder If there's a way to check on a specific logged message in one of the unit tests?

garydgregory (changes requested):
Better message IMO: 
`Syntax error, missing '=': Expected "{KEY1=VALUE, KEY2=VALUE, ...}"`

aawad6:
Noted, but I have a question regarding the following code block:
```
final String string = options[1]
              .replaceAll(PatternParser.DISABLE_ANSI + "=(true|false)", Strings.EMPTY)
              .replaceAll(PatternParser.NO_CONSOLE_NO_ANSI + "=(true|false)", Strings.EMPTY);
```
To avoid false positive warning messages, aren't we need to replace also the commas with Strings.empty()?
for example: if  `Options[1] = "noConsoleNoAnsi=false, disableAnsi=false, LOGBACK"`, `string` will be `", , LOGBACK"`.

So I'd suggest to tweak the match pattern to looks like the following:
```
final String string = options[1]
        .replaceAll(",?" + PatternParser.DISABLE_ANSI + "=(true|false),?", Strings.EMPTY)
        .replaceAll(",?" + PatternParser.NO_CONSOLE_NO_ANSI + "=(true|false),?", Strings.EMPTY)
        .trim();
```

ppkarwasz:
@aawad6,

A while back I added a `@UsingStatusListener` JUnit 5 extensions, which allows you to capture status logger messages into a `ListStatusListener`:
```java
    @Test
    @UsingStatusListener
    public void testSomething(final ListStatusListener listener) {
        ...
        assertThat(listener.findStatusData(Level.WARN)).hasSize(1);
        ...
    }
```

Regarding the "disableAnsi" and "noConsoleNoAnsi" keys, we could probably deal with them better. E.g. instead of removing them through a regex, we can add them to the `dontEscapeKeys` and deal with them in the `HighlightConverter` constructor.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/HighlightConverter.java
```diff
@@ -87,6 +87,10 @@ public final class HighlightConverter extends LogEventPatternConverter implement
 
     private static final String STYLE_KEY = "STYLE";
 
+    private static final String DISABLE_ANSI_KEY = "DISABLEANSI";
+
+    private static final String NO_CONSOLE_NO_ANSI_KEY = "NOCONSOLENOANSI";
+
```
ppkarwasz, line 93: These are already defined as `PatternParser.DISABLE_ANSI` and `PatternParser.NO_CONSOLE_NO_ANSI`.
They differ in casing, but keys are matched case-insensitive anyway.

ppkarwasz (approved):
Apart from the comment below, it looks good to me.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/HighlightConverter.java
aawad6: I think we still need the UPPERCASE keys for both `disableAnsi `and `noConsoleNoAnsi`. 

The issue arises when the `key` is obtained and transformed to UPPERCASE characters. Next, search for that key in the `dontEscapeKeys` array. If the `key` was previously supplied as lowercase, the `escape` variable will be <0, which means the `value` was overridden.
```
final String key = toRootUpperCase(keyValue[0]);
final String value = keyValue[1];
final boolean escape = Arrays.binarySearch(sortedIgnoreKeys, key) < 0;
map.put(key, escape ? createSequence(value.split("\\s")) : value);
```

ppkarwasz: I missed that, sorry.

vy:
Thanks for the improvement @aawad6, much appreciated! :bow: Merging it...

Issue #1569, opened 2023-07-19
Update the Javadoc for Logger to reflect the throwable last parameter feature


bailan:
I like using {}-style parameters in logging messages. I am so used to adding the throwable as the last parameter, so it will be handled by the logging API automatically (adding the exception and stack traces to the final log).

I realized recently that not everyone knows about this feature, and I got into an argument of if the use of the API was correct or not. Specifically, there was a suspect if the exception would be print out if we don't specific that in the log message as follows
```
log.error("Failed!", exception) -> log.error("Failed! Here is the exception {}", exception)
```

I couldn't find any official documents (Javadoc, Wiki) talking about this feature other than this [stackoverflow answer]( https://stackoverflow.com/questions/35096746/substituting-parameters-in-log-message-and-add-a-throwable-in-log4j-2). It seems to me that it makes sense to make this feature explicit in the Javadoc, for example [here](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/Logger.java#L748), so it is clearer to the user how the logger would/should handle the throwable as the last parameter.

rgoers:
This really should be documented in the Log4j API section. I am quite surprised it is not.

PR #1638, opened 2023-07-30
Update Log4j API doc to reflect throwable as last parameter feature.

aawad6:
As discussed in Issue1569 Log4j API doc doesn't mention the feature of passing a `Throwable` as the last parameter of any of the `log` methods without adding an extra substituting parameter. 

This PR updates Log4j API documentation to reflect the throwable as the last parameter feature. 
`logger.debug("Logging in user {} with birthday {}", user.getName(), user.getBirthdayCalendar(), exception);`


grobmeier:
Fixes #1569. Thanks @aawad6

PR #1647, opened 2023-08-01
Fix `MemoryMappedFileAppender` buffer unmapping on JRE 9+

### LINKED ISSUE #1646: MappedByteBuffer unmapping on JDK 9+
## Description

The `MemoryMappedFileAppender` fails to unmap its buffers on JRE 9+: the unmapping code (cf. [`MemoryMappedFileAppender#unsafeUnmap()`](https://github.com/apache/logging-log4j2/blob/92538f9752651f73e7bb66e490efc233959fa143/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/MemoryMappedFileManager.java#L225) requires:
```
--add-opens java.base/java.nio=ALL-UNNAMED
--add-exports java.base/jdk.internal.ref=ALL-UNNAMED
```
These two permissions are not given by default.

Since Java 9 the cleaner is accessible through `sun.misc.Unsafe#invokeCleaner` instead.

ppkarwasz:
This moves the direct buffer unmapping logic into a new class `UnsafeUtil` for which a Java 9 version is also provided.

Closes #1646.

### DIFF
--- a/log4j-core-java9/pom.xml
+++ b/log4j-core-java9/pom.xml
@@ -30,6 +30,7 @@
     <log4jParentDir>${basedir}/..</log4jParentDir>
     <docLabel>Log4j Implementation Documentation</docLabel>
     <projectDir>/core</projectDir>
+    <maven.compiler.release>9</maven.compiler.release>
     <maven.deploy.skip>true</maven.deploy.skip>
   </properties>
   <dependencies>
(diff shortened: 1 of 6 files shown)


vy:
I still need to wrap my mind around how this _do-this-for-pre-9-and-that-for-post-9_ magic works, but from a quick glance, it all looks good to me.

ppkarwasz:
I found out that `main` solves this problem in the same exact way.
I am merging this PR.

PR #1652, opened 2023-08-01
(doc) Update log4j-spring-boot spring-boot min version

SebHeuze:
Since log4j-spring-boot 2.18.0 the required Spring Boot version is now 2.4.0 since it use `LoggingSystemFactory` which is a spring boot 2.4.0 class 
[https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/logging/LoggingSystemFactory.html](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/logging/LoggingSystemFactory.html)

### DIFF
--- a/src/site/markdown/log4j-spring-boot.md
+++ b/src/site/markdown/log4j-spring-boot.md
@@ -83,5 +83,5 @@ this purpose. Below is an example:
 ## Requirements
 
 The Log4j 2 Spring Cloud Configuration integration has a dependency on Log4j 2 API, Log4j 2 Core, and 
-Spring Boot versions 2.0.3.RELEASE or 2.1.1.RELEASE or later versions it either release series.
+Spring Boot version 2.4.0 or later versions it either release series.
 For more information, see [Runtime Dependencies](runtime-dependencies.html).


ppkarwasz:
Hi @SebHeuze,

The story is a little bit more complicated than that.

Before Spring Boot 2.4.0 the only way to select a Spring Boot [`LoggingSystem`](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/logging/LoggingSystem.html) was to set a system property:
```
org.springframework.boot.logging.LoggingSystem=org.apache.logging.log4j.spring.boot.Log4j2SpringBootLoggingSystem
```
This method is **still valid** with all the Spring Boot versions mentioned in the docs.

Log4j2 2.17.2 or earlier used a `log4j2.system.properties` file (cf. [System Properties](https://logging.apache.org/log4j/2.x/manual/configuration.html#SystemProperties)) embedded in the `log4j-spring-boot` JAR file to set this property. This method **effectively** prevents the usage of any other `LoggingSystem` (including the original Spring Boot's `Log4j2LoggingSystem`) and was removed in #777.

Since Spring Boot 2.4.0 a new mechanism to register a `LoggingSystem` is available: `spring.factories` (cf. [`SpringFactoriesLoader`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/io/support/SpringFactoriesLoader.html). This is the mechanism adopted by Log4j 2.18.0 and later versions.

SebHeuze:
Yes you are right, my mistake was because the project that had the issue forced the log4j version   so I had the compatibility issue with 2.3.9

PR #1667, opened 2023-08-06
Improve `Log4j-config.xsd` schema

ppkarwasz:
This PR backports the schema available on `main` and adapts it to validate all test configurations in `log4j-core-test`.


ppkarwasz:
Closes #1441.

Issue #1675, opened 2023-08-07
Documentation mismatch regarding Lookups


yakovsh:
## Description
There is a mismatch in documentation between the two pages on lookups:
https://logging.apache.org/log4j/2.x/manual/lookups.html
https://logging.apache.org/log4j/2.x/manual/configuration.html#PropertySubstitution

Additionally, the list of available lookups in both places doesn't match up with code. For example, the "base64" lookup is in code for the 2.xx version, and the "bundle" lookup doesn't appear on the lookups doc page, but only in the configuration doc page. Third, the table in the configuration page doesn't list all available lookups that appear in code and on the lookups page.

I can put together a pull request with changes for the documentation if desired.

P.S. Previous bug report regarding similar issue here:
https://issues.apache.org/jira/browse/LOG4J2-3504

rgoers:
PRs are very welcome!

vy:
#1681 is merged.

PR #1681, opened 2023-08-08
Update lookup documentation (#1675 LOG4J2-3504)

yakovsh:
- This pull requests updates [the Property Substitution section](https://logging.apache.org/log4j/2.x/manual/configuration.html#PropertySubstitution) of the configurations document page to list all currently valid lookups in Log4J.
- Adds information about the resource bundle lookup to the lookups page
- It also removes the information about the base64 lookup as per [LOG4J2-3504](https://issues.apache.org/jira/browse/LOG4J2-3504)
- Fixes link to the Log4J docker page


vy:
Looks great @yakovsh! Thanks so much! I am merging this and I will create a changelog entry too.

Issue #1365, opened 2023-03-13
Request to expose certain protected APIs as public


Karthi1711:
Hello,

I would like to request that certain APIs which are currently protected be exposed as public for broader access. Specifically, I am interested in accessing `createCompressAction()`, `getExtension()`, `length()` in `FileExtension` enum for customized rollover strategy which we have written at our end. 

Could you please consider this request and let me know if it is possible to make these APIs public? I would be happy to provide more information or answer any questions you may have.

Thank you for your time and consideration.

Best regards,
Karthi

ppkarwasz:
Hello @Karthi1711,

Can You submit a PR with the changes?

garydgregory:
I think we want to be careful here @ppkarwasz: Do we really want to lock down an API as public for such low level classes and be bound to binary compatibility? Or are we going to say this class is not part of BC?

ppkarwasz:
@garydgregory,

I would just prefer to see this request in the form of a PR, that we can discuss on the code.

Karthi1711:
PR created: https://github.com/apache/logging-log4j2/pull/1683.

PR #1683, opened 2023-08-09
Open `FileExtension` accessors (#1365)

Karthi1711:
#1365 

Access modifier changed from `default` to `public` of `FileExtension` class in below api's

1. `createCompressAction()`
2. `getExtension()`
3. `length()`


vy:
@Karthi1711, I am fine with these changes; except the `length()`. Let's keep the opened up methods to minimum: `createCompressAction()` and `getExtension()` are fine, but we don't need `length()`, which is accessible by `getExtension().length()` anyway.

I will
1. merge the changes
2. make `length()` package-private again
3. add a changelog entry

Issue #1650, opened 2023-08-01
Inverted rollover strategy in `log4j-1.2-api`


ppkarwasz:
## Description

The default rollover strategy in Log4j 1.x archived the current log file `File` **always** to `File.1`:

> If `MaxBackupIndex` is positive, then files `{File.1, ..., File.MaxBackupIndex -1}` are renamed to `{File.2, ..., File.MaxBackupIndex}`. Moreover, `File` is renamed `File.1` and closed. A new `File` is created to receive further log output. [source [Javadoc](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/RollingFileAppender.html#rollOver())]

This corresponds to the `fileIndex="min"` rollover strategy of the version 2.x `RollingFileAppender` (cf. [documentation](https://logging.apache.org/log4j/2.x/manual/appenders.html#default-rollover-strategy)).

However the [`RollingFileAppenderBuilder`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/RollingFileAppenderBuilder.java) from `log4j-1.2-api` creates a `fileIndex="max"` rollover strategy (the default strategy in Log4j 2.x).

This problem was reported in the SO question [Log4j 1.2 Bridge update to 2.17.2 RollingFileAppender MaxBackupIndex Problem](https://stackoverflow.com/q/76787809/11748454).

## Affected version

Versions `2.20.0` and `3.0.0-alpha1` are both affected.

krzyk:
Hi, could I work on  fix for this issue?

ppkarwasz:
@krzyk,

Sure, go ahead. Just remember to add a unity test, so we prevent future regressions.

krzyk:
@ppkarwasz So, I assume this can be closed?

PR #1721, opened 2023-08-21
Updated default fileIndex for rollover strategy in log4j1

krzyk:
For #1650 

Added default fileIndex of `min` when parsing Log4j1 configuration and when building from scratch using the provided builders (changes are only in the log4j1 packages) - modified test to check for the rollover strategy for v1 to check that it is not using `max`.


ppkarwasz (approved):
@krzyk,

Looks good to me. Don't worry about the tests, they've been failing randomly on the CI for some time now.

PR #1752, opened 2023-09-01
Migrate to ConfigurableInstanceFactory API

jvz:
It's time to rewrite the DI subsystem _one more time!_ To simplify the provenance of this long-time effort, all related changes here have been squashed back together. This addresses numerous items:

- Rename `Injector` to `InstanceFactory` with a subinterface `ConfigurableInstanceFactory` and embrace some lessons from Spring
- Teach `Condition`/`Conditional` API to support multiple conditions
- Fix `ConditionalOnProperty`
- Add `ConditionalOnMissingBinding`
- Rename `ReflectionAccessor` to `ReflectionAgent`
- Add several more specific exception classes
- Add `org.apache.logging.log4j.plugins.di.spi` package for SPI-related classes in DI
- Replace named `Function<String, String>` instance with `StringValueResolver` interface
- Replace `InjectorCallback` concept with `ConfigurableInstanceFactoryPostProcessor` service class vaguely comparable to Spring's `BeanFactoryPostProcessor` family of classes
- Add `FactoryResolver` strategy interface for abstracting resolution of "special" factories to simplify integration with configurable plugin annotation handling
- Add `InstancePostProcessor` interface similar to Spring's `BeanPostProcessor` API for processing instances created through DI
- Add and update API docs for plugin and DI things
- Remove unused `Configuration::getPluginPackages` method and backing field
- Add built-in bindings for `PropertyEnvironment` and `ClassLoader`
- Split up `DefaultBundle`
- Add `@ConfigurationFactoryType` JUnit 5 extension for defining a `ConfigurationFactory` class to use in a test
- Add `LoggerContextSource::selector` option for specifying a `ContextSelector` class to use in a test

[A clear and concise description of what the pull request is for along with a reference to the associated issue IDs, if they exist.]

## Checklist


log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/di/spi/FactoryResolver.java
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Gets a factory for the given resolvable key using existing bindings from the given instance factory.
+     * A resolvable key in this context is a {@link Key} combined with alias names and the annotated element
+     * this factory is for.
+     *
+     * @param resolvableKey   the resolvable key to create a binding for
+     * @param instanceFactory the existing instance factory to use for composing bindings
+     * @return a factory for instances described by the provided key
+     */
+    Supplier<?> getFactory(final ResolvableKey<?> resolvableKey, final InstanceFactory instanceFactory);
```
ppkarwasz, line 48: Shouldn't this method be parameterized?

ppkarwasz (approved):
With such a large PR it is virtually impossible to review the actual code, but:
  * the new naming convention makes more sense to me than the old one, so it will be easier for other contributors to use the dependency injection in a non-trivial way.
  * I think we should merge this and how it behaves.
  
  The only problem I have with the injection system as a whole is its verbosity in the status logger: it takes something like 80% of all `DEBUG` level messages, which makes it harder to debug problems in the remaining part of Log4j. I think that:
  * most of the internal processes inside the DI could log at `TRACE` level,
  * some stats and maybe non-default bindings could be printed at an `INFO` level (which for me means "visible to a power user"),
  * the resulting binding could be printed once at a `DEBUG` level.

ppkarwasz:
> Not exactly sure why the Windows build keeps failing, but seems unrelated.

The tests in `log4j-api-test` are apparently still running in parallel in the same JVM. #1800 solves the your last Windows failure.

log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/Configurable.java
```diff
@@ -22,34 +22,40 @@
...
 import org.apache.logging.log4j.util.Strings;
 
 /**
- * Annotates a plugin as being a configurable plugin. Configurable plugins are instantiated from a tree of
- * {@link Node} objects.
+ * Annotates a plugin as being a configurable plugin. A configurable plugin corresponds to a {@code Node} element
+ * of a configuration tree. Each configuration element may have zero or more {@linkplain PluginAttribute attributes}
+ * where attribute values are converted from strings into other types via {@link TypeConverter}, an optional
+ * {@linkplain PluginValue value} (another type of plugin attribute which may have dedicated syntax in some configuration
+ * formats such as XML), and zero or more child elements.
```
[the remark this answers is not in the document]
jvz: A pattern converter has its own syntax, parser, etc., compared to a configuration file. That's a good point, though, that different things can be referenced in config files.

log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/di/spi/FactoryResolver.java
jvz: I think I tried that and found it to be awkward, but I could explore that idea to see if it makes sense.

jvz: It seems like it would make more sense to parameterize the class itself. Let me try this out.

jvz: Seems as though one of the issues was that various resolvers for plugins handle both their `T` case and their `Supplier<T>` case. Splitting those up could help.

jvz: Alright, updated this. It resulted in several additional resolvers, but now they're a bit less complex.

log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/Configurable.java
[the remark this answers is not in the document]
jvz: Docs updated.

Issue #1703, opened 2023-08-16
[MS1] Repository Improvements #1


grobmeier:
- [ ] Prepare Repos
- [ ] Prepare Nexus
- [ ] Prepare Auth
- [ ] Update Documentation

grobmeier:
Duplicate of https://github.com/apache/logging-log4j2/issues/1777

PR #1754, opened 2023-09-04
Fix JeroMqAppenderTest race conditions

ppkarwasz:
This PR should eliminate the failures of `JeroMqAppenderTest` without modifying `JeroMqAppender` from `PUB` to `XPUB` (cf. commit f1b928cba5e6aec6ef3fe6e0410c2b12235595c9).

This is a prerequisite for #1703, due to the frequent CI failures due to this test.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/mom/jeromq/JeroMqAppenderTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/mom/jeromq/JeroMqAppenderTest.java
@@ -16,7 +16,9 @@
  */
 package org.apache.logging.log4j.core.appender.mom.jeromq;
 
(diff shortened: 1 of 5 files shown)


vy:
@ppkarwasz, the entire reflection logic seems brittle to me. The accessed fields can simply change in the next ZeroMQ release. Can't we simply `Awaitility.await()` on the checks with a timeout?

ppkarwasz:
@vy,

Yes, the reflection logic **is** brittle. I submitted a feature request JeroMQ (zeromq/jeromq#960) that would allow us to:
 * to retrieve the ephemeral port number currently used by the appender,
 * to detect when the test client connects to the appender.

In the meantime I replaced the custom waiting logic with `Awaitility` in 3ecf90e5e466f9d7f90dd2ceb25b46f0be9803bf.

PR #1759, opened 2023-09-05
Add JUnit 5 extension for per test-class log directories

ppkarwasz:
This PR adds a `@TempLoggingDir` extension, which is a reduced version of the standard `@TempDir` annotation.

It can be used on **static** fields and:
 * creates a directory specific for the test-class (right now in `target/logs/<fully_qualified_class_name>`,
 * injects the field with the value of the directory,
 * creates an entry in the `TestProperties` with key "logging.path",
 * deletes the directory if the test succeeds.
 
 The `TestProperties` entry is available as both a Log4j2 system property and using the `${test:logging.path}` lookup.

### DIFF
--- a/log4j-api-test/pom.xml
+++ b/log4j-api-test/pom.xml
@@ -54,6 +54,11 @@
       <artifactId>junit-jupiter-api</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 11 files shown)


jvz (approved):
Nice improvement!

ppkarwasz:
This testing infrastructure improvement is part of task #1776.

PR #1760, opened 2023-09-06
Do not warn if sun.reflect.Reflection.getCallerClass() is unavailable

Flowdalic:
Modern Java releases disallow the access to internal sun.* API. There is not much the user can do about it, hence this warning is not actionable and therefore unneeded.

It also does not tag its origin, causing confusion among users about its whereabouts.


log4j-api/src/main/java/org/apache/logging/log4j/util/StackLocator.java
```diff
@@ -77,7 +77,6 @@ public final class StackLocator {
                 }
             }
         } catch (final Exception | LinkageError e) {
-            System.out.println("WARNING: sun.reflect.Reflection.getCallerClass is not supported. This will impact performance.");
```
ppkarwasz, line 80: I would rather keep the warning, but:
 - use `LowLevelLogUtil#log` instead of `System.out.println`,
 - differentiate the message according to `Constants#JAVA_MAJOR_VERSION`:
    - if the version is <= 8 warn about "sun.reflect.Reflection.getCallerClass is not supported. This will impact location-based features.",
    - if the version is > 8 warn about "Runtime environment does not support multi-release JARs. This will impact location-base features."

ppkarwasz (changes requested):
@Flowdalic,

The `log4j-api` artifact is a multi-release JAR, so the code in your PR should only run if you are running on JRE 8. On JRE 9+ [this class](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api-java9/src/main/java/org/apache/logging/log4j/util/StackLocator.java) should run.

The typical reason why the Java 8 `StackLocator` runs on JRE 9+  is _shading_: a user copies the classes from `log4j-api`, but is not aware of other setting of the JAR (like the `Multi-Release: true` entry in the manifest).

Flowdalic:
Thanks for your reply, much appreciated. :)

> The typical reason why the Java 8 `StackLocator` runs on JRE 9+ is _shading_

That it probably what happened for me. I am using the [mill build tool for Scala](https://github.com/com-lihaoyi/mill), which is using [Jar Jar Abrams](https://github.com/eed3si9n/jarjar-abrams) for shading. Probably the shading does not support multi release jars.

BTW, see also https://github.com/apache/logging-log4j2/pull/1761.

Note that I like you suggestions, but I will probably not have the resources to implement them.

ppkarwasz (approved):

vy:
@Flowdalic, thank so much for the report and the PR, much appreciated! :bow: Merging it...

Flowdalic:
> @Flowdalic, thank so much for the report and the PR, much appreciated! 🙇 Merging it...

Thanks for working on an improved solution and merging this. Unfortunately, I only discovered now that the commit message is incomplete and may blames the wrong entity for missing multi-release JAR support. See my follow up PR at https://github.com/apache/logging-log4j2/pull/1833

PR #1784, opened 2023-09-11
Allow multiple Log4jServletContextListener registrations

### LINKED ISSUE #1782: Shutdown of Log4j2 initialized via Log4jServletContainerInitializer happens first not last
## Description

Shutdown of Log4j2 system initialized via org.apache.logging.log4j.web.Log4jServletContainerInitializer happens first before all ServletContextListener contextDestroyed methods are called in a servlet container (tomcat).

I would expect Log4j2 to shutdown itself at the very end, so other context listeners have the change to log something during their shutdown procedure.

If I intialize/shutdown log4j2 via org.apache.logging.log4j.web.Log4jServletContextListener and put the context listener at the very top in `web.xml` then log4j2 properly shuts down after every other listener.

If I debug 

            if (!"true".equalsIgnoreCase(servletContext.getInitParameter(
                    Log4jWebSupport.IS_LOG4J_AUTO_SHUTDOWN_DISABLED))) {
                servletContext.addListener(new Log4jServletContextListener());
            }

in Log4jServletContainerInitializer, all other listener already exists in the servletContext and the Log4jServletContextListener is added to the very end, thus shutdown happens as the very first.

## Configuration

**Version:** log4j2-web 2.20.0

**Operating system:** MacOS

**JDK:** JDK 11 Temurin

**Servlet Container:**: Tomcat 9.0.73

## Logs

N/A

ppkarwasz:
This PR allows the user to declare a `Log4jServletContextListener` explicitly in a `web.xml` descriptor (additionally to the one installed by `Log4jServletContainerListener`).

Due to the way the Servlet API works, the only way for `Log4jServletContextListener` to shutdown the context **after** all other listeners is to declare it explicitly in a `web.xml` descriptor.

Closes #1782.

### DIFF
--- a/log4j-jakarta-web/src/main/java/org/apache/logging/log4j/web/Log4jServletContextListener.java
+++ b/log4j-jakarta-web/src/main/java/org/apache/logging/log4j/web/Log4jServletContextListener.java
@@ -36,6 +36,8 @@ import static org.apache.logging.log4j.util.Strings.toRootUpperCase;
  */
 public class Log4jServletContextListener implements ServletContextListener {
 
+    static final String START_COUNT_ATTR = Log4jServletContextListener.class.getName() + ".START_COUNT";
+
     private static final int DEFAULT_STOP_TIMEOUT = 30;
     private static final TimeUnit DEFAULT_STOP_TIMEOUT_TIMEUNIT = TimeUnit.SECONDS;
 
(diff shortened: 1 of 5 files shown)


ppkarwasz:
This PR was tested on a simple Servlet API project.

PR #1787, opened 2023-09-13
Fix thread context loss if `<AsyncLogger>` in an all async context

### LINKED ISSUE #1786: Context data is lost if AsyncLogger and AsyncLoggerConfig are both used
## Description

If the user uses both an `AsyncLoggerContext` (which makes all loggers to be of type `AsyncLogger`) and `AsyncLoggerConfig`, a race condition can occur, which causes context data to be lost.

## Details

If both `AsyncLogger` and `AsyncLoggerConfig` are used together (which is no advised, but possible and there is a test for it):
 * `AsyncLoggerConfig#log` is called from the main disruptor thread with a `RingBufferLogEvent` argument,
 * `AsyncLoggerConfig` calls `RingBufferLogEvent#createMemento()` before sending the event to its own disruptor thread. However this method does not make a copy of the context data, nor does it mark the data as "frozen",
 * A `RingBufferLogEvent#clear()` on the main disruptor thread clears the context data. If the disruptor thread of `AsyncLoggerConfig` didn't log the message yet, it will log it without context data.

ppkarwasz:
This fixes #1786.

The `AbstractAsyncThreadContextBase` and derivative tests were modernized, so that they can run at the same time in different JVMs.

`RingBufferLogEvent#createMemento()` will now freeze the context data, which will cause a new `StringMap` instance to be created, when `clean()` is called.


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
```diff
@@ -183,11 +183,13 @@ public StringMap injectContextData(final List<Property> props, final StringMap r
             // When the ThreadContext is garbage-free, we must copy its key-value pairs into the specified reusable
             // StringMap. We cannot return the ThreadContext's internal data structure because it may be modified later
             // and such modifications should not be reflected in the log event.
-            copyProperties(props, reusable);
+            final StringMap actualReusable = reusable != null && !reusable.isFrozen() ? reusable
+                    : ContextDataFactory.createContextData();
```
vy, line 187: Doesn't this imply that, unless certain conditions are met, `ForGarbageFreeThreadContextMap.injectContextData()` is not garbage-free anymore.

ppkarwasz: Yes, you are right. Since `AsyncLoggerConfig` is the one making an immutable copy of the `RingBufferLogEvent`, it should bear the responsibility for the allocation.

BTW: the logic behind copying mutable/immutable objects should be reviewed in `3.x`. E.g. under certain conditions `AsyncLoggerConfig` uses mutable events in its ring buffer. In such a case there is no need make a copy of the thread context: a mutable destination copy already exists.

ppkarwasz:
Branch `main` was fixed in:
 - commit 3c40ae20ddd82bf7259717964818668cf57d12d7,
 - commit e9629cdc1b93abbdf7149b607359ea8e918fad70.

Issue #1779, opened 2023-09-08
Use release process on the primary repository


ppkarwasz:
Following the completion of #1774 we can update the release process of the most complex repository: https://github.com/apache/logging-log4j2

PR #1815, opened 2023-09-25
Use BND for both OSGI and JPMS

ppkarwasz:
This PR replaces `maven-bundle-plugin` with `bnd-maven-plugin`, which:
 * is maintained by the same project that releases the common backend BND, so our build system depends on one less project (Apache Felix),
 * is released at each upgrade of the BND backend.

The introduction of JPMS support in BND allows us:
 * to have real JPMS module descriptors that are **verified** to be complete at a bytecode level (even `Class.forName` calls are supported by BND).

This is part of #1779.


ppkarwasz:
Closing this since it must be split per-module.

PR #1818, opened 2023-09-25
Fix compilation with ErrorProne

ppkarwasz:
Fixes the compilation errors reported by Error Prone.

This is a prerequisite for #1779.


log4j-flume-ng/src/test/java/org/apache/logging/log4j/flume/appender/FlumeEmbeddedAppenderTest.java
```diff
@@ -208,7 +208,7 @@ public void testHeaderAddedByInterceptor() {
     }
 
     /* @Test */
-    public void testPerformance() throws Exception {
+    private void testPerformance() throws Exception {
```
vy, line 211: I'd rather keep these tests in with `@Test` annotation, though add a `@Disabled` with explanation. (There was also one in `PerformanceComparison`.)

vy (approved):

Issue #1741, opened 2023-08-29
Environment Service Missing


MLausberg:
## Description

We created an application based on Eclipse 2023-06 (4.28).
The very first time we start the application the log4j2 logic is initialized correctly and we can see log messages.
But after closing the application the following start commands will not initialize log4j2 correctly and fail.
No exyception is shown and no log messages are created anymore.

From second time on, log4j is using a default implementation for all logger variables.

it seems that the services which provide Environment variable information are not registeed correctly when initializing/calling
org.apache.logging.log4j.util.PropertiesUtil$Environment
```
sources.add(propertySource);
// We don't log anything on the status logger.
ServiceLoaderUtil.loadServices(PropertySource.class, MethodHandles.lookup(), false, false).forEach(sources::add);
```

While initializing log4j2, the System properties are not available and thats why our *.xml is not fetched
_-Dlog4j2.debug=true
-DLog4jContextSelector=org.apache.logging.log4j.core.selector.BasicContextSelector
-Dlog4j2.configurationFile=file:log4j2.xml_

## Configuration

**Version:** [Log4j version]
org.apache.logging.log4j.1.2-api_2.20.0
org.apache.logging.log4j.api_2.20.0
org.apache.logging.log4j.core_2.20.0
org.apache.logging.log4j.jcl_2.20.0
org.apache.logging.log4j.slf4j-impl_2.20.0

**Operating system:** [OS and version]
org.osgi.framework.os.name = Windows10
osgi.ws = win32
sun.arch.data.model = 64

**JDK:** [JDK distribution and version]
Eclipse Adoptium Temurin-17.0.8+7 

## Logs

Can not provide any logs, because they ar emissing.

We assume a plugin loading order issue or a service registration problem.

ppkarwasz:
@MLausberg,

Can you check with our `2.20.1-SNAPSHOT`? There has been at least one improvement for OSGI applications (#1367), since the release of `2.20.0`, which concerned `ServiceLoaderUtil` (this component loads all property sources including the one for Java system properties).

MLausberg:
I have downloaded and replaced the jars in the application
log4j-1.2-api-2.20.1-20230825.123921-76
log4j-api-2.20.1-20230825.123921-76
log4j-core-2.20.1-20230825.123921-76
log4j-jcl-2.20.1-20230825.123921-76

The error is still the same.

[...]

PR #1819, opened 2023-09-26
Use BND for both OSGI and JPMS

ppkarwasz:
This PR is a work in progress and replaces #1815.

It replaces maven-bundle-plugin with bnd-maven-plugin, which:
 *  is maintained by the same project that releases the common backend BND, so our build system depends on one less project (Apache Felix),
 * is released at each upgrade of the BND backend.

The introduction of JPMS support in BND allows us:
 * to have real JPMS module descriptors that are verified to be complete at a bytecode level (even `Class.forName` calls are supported by BND).

To help in the review process, BND configurations for various modules will have their own PRs and this one will be complete when modifications to all modules will be accepted:

- [x] `log4j-1.2-api` (#1825),
- [x] `log4j-api` (#1820),
- [x] `log4j-api-test` (#1828),
- [x] `log4j-appserver` (#1827),
- [x] `log4j-cassandra` (#1828),
- [x] `log4j-core` (#1821),
- [x] `log4j-core-test` (#1828),
- [x] `log4j-couchdb` (#1828),
- [x] `log4j-docker` (#1828),
- [x] `log4j-flume-ng` (#1828),
- [x] `log4j-iostreams` (#1827),
- [x] `log4j-jcl` (#1827),
- [x] `log4j-jdbc-dbcp2` (#1828),
- [x] `log4j-jpa` (#1828),
- [x] `log4j-jpl` (#1827),
- [x] `log4j-jul` (#1827),
- [x] `log4j-kubernetes` (#1828),
- [x] `log4j-layout-template-json` (#1828),
- [x] `logj4-layout-template-json-test` (#1828),
- [x] `log4j-mongodb3` (#1828),
- [x] `log4j-mongodb4` (#1828),
- [x] `log4j-slf4j2-impl` (#1827),
- [x] `logj4-slf4j-impl` (#1827),
- [x] `logj4-spring-boot` (#1828),
- [x] `logj4-spring-cloud-config-client` (#1828),
- [x] `log4j-taglib` (#1828),
- [x] `logj4-to-jul` (#1827),
- [x] `logj4-to-slf4j` (#1827),
- [x] `logj4-web` (#1828).


ppkarwasz:
Using BND we still need to manually overwrite the detected (or undetected) module name of **multi-release** dependency JARs. Theoretically this was fixed in bndtools/bnd#5327, but I can't get it to work. The module name of MRJs with a module descriptor in their `META-INF/versions/9` folder isn't

@HannesWell, do you know what am I missing?

ppkarwasz:
Apparently the BND (partial) fix is only in the `7.1.0-SNAPSHOT` branch.

ppkarwasz:
Closes #1830.

HannesWell:
Sorry for the late and after the fact reply.  I just checked the following slf4j bundles in more detail and everything looks good and worked well in my testing in our two OSGi applications, from which one uses `log4j-core` as back-end and redirects every other API to it and the other one uses `logback-classic` and redirects all APIs to that:
- `log4j-api`
- `log4j-core`
- `log4j-slf4j2-impl`
- `log4j-slf4j-impl`
- `log4j-to-slf4j` (Especially the extended version range in `Import-Package: org.slf4j;version="[1.7,3)",org.slf4j.spi;version="[1. 7,3)", ...` is important)

I noted that for some bundles the symbolic name changed e.g. from `o.a.l.log4j.slf4j2-impl` to `o.a.l.log4j.slf4j2.impl`. In some cases this requires adjustments, e.g. if one requires the bundle instead of importing the contained packages. But since the former is discouraged anyways it shouldn't be a blocker and is IMHO fine.

If you want to keep the MANIFEST.MF a little bit cleaner you could remove the `Private-Package`, which is only for internal use of BND and could be removed with the instruction `-removeheaders: Private-Package`.

Thank you @ppkarwasz for this work!

ppkarwasz:
@HannesWell,

That is great news, so I believe we have a green light to release 2.21.0, unless you have an idea what is going on in #1741.

Regarding the bundle names, we chose to prefer consistency (use bundle names that can be module names) over backward compatibility. If you can think of a place where bundle names are used (build scripts?), we can revert them to the original form.

HannesWell:
> That is great news, so I believe we have a green light to release 2.21.0, unless you have an idea what is going on in #1741.

From my point of view yes 🚀🙂

> Regarding the bundle names, we chose to prefer consistency (use bundle names that can be module names) over backward compatibility. If you can think of a place where bundle names are used (build scripts?), we can revert them to the original form.

That's a reasonable choice. I'm know cases that depend on the bundle-name, for example if another Bundle requires a log4j bundle using OSGi's Require-Bundle (but that's discouraged/deprecated and as far as I know mainly used in the Eclipse world only. And since you provide good OSGi metadata, i.e. versioned packages, there is no reason to not used Import-Package instead, which only depends on the package name). Or if the bundles are part of an Eclipse Feature (which is basically a group of Bundles that form a 'feature' of an Eclipse Product like the IDE), but there names can be changed too and usually a feature once build is pinned to a specific version of the Bundle so there is no compatibility problem.
So IMHO the name change is fine, personally I find bundle-names with a dash a bit odd, but it should be mentioned in the release notes so that there is a change to simplify migration.

PR #1833, opened 2023-10-05
Improve warning message if multi-release JARs are not supported

Flowdalic:
This is a follow up from my pervious PR [1]. Unfortunatly, it only occured to me after it was merged that the warning message is incomplete. In my case it was not the runtime environment which did not support multi-release JARs, but the build system. This adjusts the message accordingly.

1: https://github.com/apache/logging-log4j2/pull/1760


ppkarwasz (approved):
@Flowdalic,

Thanks, there is nothing more annoying than error messages that point you in the wrong direction.

PR #1848, opened 2023-10-09
LOG4J2-3672 Avoid invoking `DateFormatSymbols.getZoneStrings()` in `FastDateParser`

tristantarrant:
https://issues.apache.org/jira/browse//LOG4J2-3672

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FastDateParser.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/datetime/FastDateParser.java
@@ -19,7 +19,6 @@ package org.apache.logging.log4j.core.util.datetime;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
-import java.text.DateFormatSymbols;
 import java.text.ParseException;
 import java.text.ParsePosition;
 import java.util.ArrayList;
@@ -31,7 +30,6 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
-import java.util.Set;
 import java.util.TimeZone;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
@@ -39,7 +37,9 @@ import java.util.concurrent.ConcurrentMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.util.Integers;
+import org.apache.logging.log4j.status.StatusLogger;
 
 import static org.apache.logging.log4j.util.Strings.toRootUpperCase;
 
@@ -79,7 +79,6 @@ import static org.apache.logging.log4j.util.Strings.toRootUpperCase;
  * @see FastDatePrinter
  */
 public class FastDateParser implements DateParser, Serializable {
-
     /**
      * Required for serialization support.
      *
@@ -102,7 +101,7 @@ public class FastDateParser implements DateParser, Serializable {
     // comparator used to sort regex alternatives
     // alternatives should be ordered longer first, and shorter last. ('february' before 'feb')
     // all entries must be lowercase by locale.
-    private static final Comparator<String> LONGER_FIRST_LOWERCASE = (left, right) -> right.compareTo(left);
+    private static final Comparator<String> LONGER_FIRST_LOWERCASE = Comparator.reverseOrder();
 
     /**
      * <p>Constructs a new FastDateParser.</p>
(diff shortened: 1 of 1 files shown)


vy:
@tristantarrant, thanks so much for taking time to not only report the problem, but also provide a fix. :bow: I have some concerns regarding this change. For one, it is backward incompatible, hence we cannot have your toggle enabled by default. Nevertheless, I think we don't need a toggle: `FastDateParser` can do better than matching against an enormous regex to figure out the time zone.

Since `FastDateParser` is borrowed from Apache Commons Lang in 2015 (ouch!), I'd first check if the upstream already has a fix for this. Otherwise, I'd try to figure out a better way. All in all, I have an inkling that we can do better a regex.

tristantarrant:
Hi @vy, I agree we need a better solution.

- avoid constructing that ridiculous regex with all the timezone names anyway. A simple `[A-Za-z0-9\+\-/\s]+` should be enough
- if the user uses a `GMT-xx` or RFC822-style timezone, use the quick path
- if the user uses a _primary_ timezone name (e.g. `Europe/Rome`), use `TimeZone.getTimeZone()`
- otherwise fallback to obtaining the full list of zone names and aliases. Unfortunately there is no other way to look up timezones by alias... By using the approach used in my PR we can save ~700K of heap over the use of `DateFormatSymbols.getInstance(locale).getZoneStrings();` (as you can see from the table I've put in the Jira)

Unfortunately this does not solve the bigger issue: if someone uses `SimpleDateFormat` in their code, they are going to trigger the full JDK caching of timezones, rendering our fix useless anyway.

A possibility, aside from fixing this in the JDK in some way, is to install an alternate `ZoneRulesProvider` using the `java.time.zone.DefaultZoneRulesProvider` system property. Such a provider would avoid caching the full TZ dataset.

vy:
@tristantarrant, I haven't forgotten about this issue and your PR, but swamped with other priorities right now. Please allow me some time.

tristantarrant:
> @tristantarrant, I haven't forgotten about this issue and your PR, but swamped with other priorities right now. Please allow me some time.

I wasn't expecting you to act on it urgently. Take your time :)

vy:
Figured the most recent `FastDateParser` from Apache Commons Lang also still uses `getZoneStrings()`: https://github.com/apache/commons-lang/blob/5b5656a8b403fd91284b0d2acc893dac7ebd5c29/src/main/java/org/apache/commons/lang3/time/FastDateParser.java#L932

vy:
@tristantarrant, the more I digged this the more I felt like you are pulling my leg. :sweat_smile: Log4j uses date parsing nowhere in the code. I can mark all data parsing related stuff as deprecated and remove them at a certain point. Though all this deserves the following question: How did you end up having the problem described in LOG4J2-2672? :thinking: Are you trying to parse date using Log4j libraries? If so, why?

tristantarrant:
Log4j does not parse dates: it parses a date formatter which is then used to print timestamps in the logs. We use the following layout: 
```xml
<PatternLayout pattern="%highlight{%d{yyyy-MM-dd HH:mm:ss,SSS} %-5p (%t) [%c] %m%throwable}{INFO=normal, DEBUG=normal, TRACE=normal}%n"/>
```
and that `%d{....} causes all the timezone initialization.

tristantarrant:
I'll create a small verifier and attach it to the issue.

vy:
`getZoneStrings()` is only used in `FastDateParser`, and when I delete that class (and some more other code), the code still compiles fine. Anyway, apparently I am missing something. I will continue tomorrow.

tristantarrant:
I apologize: I included the wrong layout. This one triggers the call:
```xml
 <PatternLayout pattern="%d{dd/MMM/yyyy:HH:mm:ss Z} %-5p %m%throwable%n"/>
 ```
Removing the `Z` will "fix" it, but then the timestamp will not have the Common Log Format (as described here https://httpd.apache.org/docs/2.4/logs.html) which is what we need.

tristantarrant:
I've created the reproducer: https://github.com/tristantarrant/log4j-tzdb-issue

tristantarrant:
It will generate a `tz.hprof` heap dump which you can open with VisualVM or Eclipse MAT.

vy:
Okay, apparently `FastDateFormat` eagerly creates a `FastDateParser` instance. I have checked this with other PMC members and decided to remove the `FastDateParser` from the code base. I will implement the change and inform you.

vy:
f9766bb3b7cf7a914c3ae36358aa053ee088b097 should have fixed the issue. @tristantarrant, would you mind confirming the fix, please? (You can either try building the sources yourself or use `2.22.0-SNAPSHOT` after [the associated build](https://github.com/apache/logging-log4j2/actions/runs/6736244242) succeeds.)

tristantarrant:
Even better: less code is my favourite code :)
We can close this PR.

vy:
@tristantarrant, there were forgotten tests. Pushed one more commit. The new CI run: https://github.com/apache/logging-log4j2/actions/runs/6736362709

tristantarrant:
@vy just to let you know that our application initial heap footprint went down from ~31MB to 25MB (we also disabled initialization of JMX) thanks to this. Thanks again.

PR #1863, opened 2023-10-15
Remove dependency on `java.desktop`

jvz:
This adds `ConfigurationChangeListener` and `ConfigurationChangeEvent` to replace the use of `java.beans.PropertyChangeListener` and `java.beans.PropertyChangeEvent` respectively. This removes the need for the `java.desktop` module in both log4j-core and log4j-jul.

## Checklist


log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -711,7 +707,8 @@ public Configuration setConfiguration(final Configuration config) {
                 prev.stop();
             }
 
-            firePropertyChangeEvent(new PropertyChangeEvent(this, PROPERTY_CONFIG, prev, config));
+            final var event = new ConfigurationChangeEvent(this, prev, config);
+            configurationChangeListeners.forEach(listener -> listener.onChange(event));
```
ppkarwasz, line 711: Probably we should add a `try ... catch` block to prevent bugs in the listeners from breaking Log4j.

ppkarwasz (approved):
Looks good to me.

If we are introducing a breaking change in `LoggerContext`, I would also reevaluate the role of [`ConfigurationListener`](https://github.com/apache/logging-log4j2/blob/main/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationListener.java). I have the impression that the two kinds of listeners have a very similar role and can be merged (of `ConfigurationListener` can be dropped).

The role of `ConfigurationListener` is just to inform `LoggerContext` that the configuration file changed and trigger a reconfiguration. I don't think any other component really uses this interface: most components should wait for a `ConfigurationChangedEvent`.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
jvz: I suppose that makes sense as an improvement.

jvz:
> Looks good to me.
> 
> If we are introducing a breaking change in `LoggerContext`, I would also reevaluate the role of [`ConfigurationListener`](https://github.com/apache/logging-log4j2/blob/main/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationListener.java). I have the impression that the two kinds of listeners have a very similar role and can be merged (of `ConfigurationListener` can be dropped).
> 
> The role of `ConfigurationListener` is just to inform `LoggerContext` that the configuration file changed and trigger a reconfiguration. I don't think any other component really uses this interface: most components should wait for a `ConfigurationChangedEvent`.

I agree that these two should probably be combined.

ppkarwasz:
@jvz,

Thanks, `java.desktop` is one of the largest modules. The uncompressed sizes of our `log4j-samples-jlink` image are:
```
173668	.
40736	./java.base
35736	./java.desktop
16816	./java.xml
7192	./org.apache.logging.log4j.core
2924	./java.management
1256	./org.apache.logging.log4j
928	./java.rmi
632	./java.compiler
392	./java.logging
200	./java.prefs
160	./java.datatransfer
124	./java.scripting
32	./org.apache.logging.log4j.samples.jlink
```
(sizes in KiB). This means we have reduced that image by one third.

PR #1882, opened 2023-10-20
Fix detection of Apache Commons Logging bridge

### LINKED ISSUE #1865: Broken Apache Commons Logging bridge in version `2.21.0`
## Description

As part of the HtmlUnit (https://github.com/HtmlUnit/htmlunit) project i have a lot of test cases doing detailed and sometimes strange things to test some functionality.
In this special case i temporary add an appender to check if something happen on the wire.

The code looks like this:

        final Logger logger = (Logger) LogManager.getLogger("org.apache.http.headers");
        final Level oldLevel = logger.getLevel();
        Configurator.setLevel(logger.getName(), Level.DEBUG);

        final StringWriter stringWriter = new StringWriter();
        final PatternLayout layout = PatternLayout.newBuilder().withPattern("%msg%n").build();

        final WriterAppender writerAppender = WriterAppender.newBuilder().setName("writeLogger").setTarget(stringWriter)
                .setLayout(layout).build();
        writerAppender.start();

        logger.addAppender(writerAppender);
        try {
            ((DefaultCredentialsProvider) getWebClient().getCredentialsProvider())
                                            .addCredentials("jetty", "jetty".toCharArray());

            loadPage("Hi There");
            int unauthorizedCount = StringUtils.countMatches(stringWriter.toString(), "HTTP/1.1 401");
            assertEquals(1, unauthorizedCount);
        }
        finally {
            logger.removeAppender(writerAppender);
            Configurator.setLevel(logger.getName(), oldLevel);
        }

This works fine in 2.20.0 and the years before but now with 2.21.0 i fails because there is not output at all - stringWriter.toString() returns an empty string.

Any idea?
Will try to help as much as possible, doing debugging or what else needed to get this working again with the new versions.

Thanks for the great libaray
    
  Ronald

## Configuration

**Version:** 2.21.0

**Operating system:** Win

**JDK:** JDK 11

ppkarwasz:
The `log4j-jcl` artifact version 2.21.0 is not detected by `commons-logging`, because it contains a comment in the first line of the service file.

Unit tests didn't detect it because `spring-jcl` was in the classpath and it "fixed" the issue.

This PR:

-  fixes the detection of `LogFactory`,
- bans third-party logging dependencies from modules that don't expect those dependencies,
- adds a test that directly verifies that `log4j-jcl` is used.

Fixes #1865.


log4j-perf-test/pom.xml
```diff
@@ -144,6 +142,28 @@
   <build>
     <plugins>
 
+      <!--
+        ~ Unban Logback.
```
jvz, line 146: Unban or ban? The id says ban?

ppkarwasz: Unban, the `<bannedDependencies>` rule is counter-intuitive.
The banned deps are listed as `excluded`, the deps that are allowed as `included`.

ppkarwasz: The `id` of the rule must be the same as in `log4j-parent/pom.xml`.

jvz (approved):
Thanks for clarifying!

Issue #1849, opened 2023-10-10
Fix main Spotbugs issues


ppkarwasz:
After the introduction of Spotbugs in the `2.x` build, some issues were discovered that need addressing.

Spotbugs ranks the issues from 1 to 20 (cf. [documentation](https://spotbugs.readthedocs.io/en/latest/filter.html#rank)), with ranks 1-4 classified as scariest, 5-9 as scary, 10-14 as troubling and 15-20 as simple bugs.

We need to address ranks 1-14 by either correcting the code or marking them as false positives.

ppkarwasz:
Since support for a fine-grained selection of Spotbugs tests was introduced in https://github.com/apache/logging-parent/commit/871e1933845d9cf28d1cddb5c70996060da56734, the main repo build fails in the presence of:

- any bug in the `SECURITY` category,
- any bug of rank 9 or lower (the "scary" category).

PR #1885, opened 2023-10-23
[1223] Change FormattedMessage pattern heuristic

### LINKED ISSUE #1223: FormattedMessage doesn't need to use regular expressions

ppkarwasz:
We change the order in which `FormattedMessage` checks the format of the provided pattern: we first check for the presence of `{}` placeholders and only then for `java.util.Format` specifiers.

This eliminates the need for a potentially exponential regular expression evalutation, which was reported by Spotbugs (#1849).

The Javadoc and documentation were improved to clarify the heuristic used by `FormattedMessage`.

Closes #1223.

Remark: since `FormattedMessage` used the **same** regular expression as `java.util.Format`, if a message uses `java.util.Format` specifiers, it is still vulnerable to a ReDOS.

### DIFF
--- a/src/changelog/.2.x.x/1223_change_formatted_message_heuristic.xml
+++ b/src/changelog/.2.x.x/1223_change_formatted_message_heuristic.xml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="http://logging.apache.org/log4j/changelog"
+       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.2.xsd"
+       type="changed">
+  <issue id="1223" link="https://github.com/apache/logging-log4j2/issues/1223"/>
+  <description format="asciidoc">
+    Change the order of evaluation of `FormattedMessage` formatters.
+    Messages are evaluated using `java.util.Format` only if they don't comply to the `java.text.MessageFormat` or `ParameterizedMessage` format.
+  </description>
+</entry>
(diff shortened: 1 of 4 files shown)


src/changelog/.2.x.x/1223_change_formatted_message_heuristic.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="http://logging.apache.org/log4j/changelog"
+       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.2.xsd"
+       type="changed">
+  <issue id="1223" link="https://github.com/apache/logging-log4j2/issues/1223"/>
+  <description format="asciidoc">
+    Change the order of evaluation of `FormattedMessage` formatters.
+    Messages are evaluated using `java.util.Format` only if they don't comply to the `java.text.MessageFormat` or `ParameterizedMessage` format.
```
vy, line 9:
```suggestion
    Messages are evaluated using `java.util.Formatter` only if they don't comply to the `java.text.MessageFormat` or `ParameterizedMessage` format.
```

ppkarwasz:
> Remark: since `FormattedMessage` used the **same** regular expression as `java.util.Format`, if a message uses `java.util.Format` specifiers, it is still vulnerable to a ReDOS.

Let me rephrase that: the regular expression we were using was **identical** to `java.util.Formatter#formatSpecifier`. If Oracle does not consider that regex a vulnerability, I don't consider it a vulnerability either.

This PR just removes a "potential ReDOS" warning found by our static analysis tools.

PR #1920, opened 2023-10-30
Bump SLF4j to version 2.0.9

ppkarwasz:
Most of our modules with a transitive dependency on `slf4j-api` used version 1.7.36, while `log4j-slf4j-impl` used version 1.7.25 (due to a breaking change in `slf4j-ext`) and `log4j-slf4j2-impl` used version 2.0.x.

This PR simplifies this configuration by:
 * switching `log4j-slf4j-impl` to use version 1.7.36 of `slf4j-api` and version 1.7.25 of `slf4j-ext`,
 * removing the need of `slf4j-ext` where applicable,
 * switching all the other modules to SLF4J 2.x,
 * modernizing the `log4j-to-slf4j` tests to use JUnit 5 instead of JUnit 4.


rgoers:
This was a really good idea.

Issue #1707, opened 2023-08-16
[MS5] SBOM Support & Security


grobmeier:
This milestone aims to harden Log4j and its satellite projects ([`logging-parent`](/apache/logging-parent), [`log4j-tools`](/apache/logging-log4j-tools), [`log4j-scala`](/apache/logging-log4j-scala), etc.) against software supply chain attacks. Implementation tasks are detailed below.

## Consult to security experts

We had meeting with [Kamal Kothyari](https://www.linkedin.com/in/kamko) on 2023-10-13. Discussion topics and action points are as follows:
* The current distribution model (Nexus, Maven Central, ASF archive, etc.)
  * ✔ Matches the industry standards, no improvement points
  * ✔ We implemented [reproducible builds](https://reproducible-builds.org/) in #1777 and #1780
* The current security support and vulnerability disclosure model of Log4j
  * ✔ Overhauling [the Log4j security page](https://logging.apache.org/log4j/2.x/) to provide better guidance and up to date information (#1948)
  * ✔ [Private communication channels for security-related discussions](https://logging.apache.org/log4j/2.x/support.html#maintainer_discussions)
  * ✔ [CVE lists](https://logging.apache.org/log4j/2.x/security.html)
  * Project activity status (active, inactive, etc.)
    * It has been stated that it is crucial to be transparent about the project status. Is it actively maintained? Is it dormant? Is it unmaintained? etc.
    * We have been discussing to move more or less inactive projects to a dormant state; this discussion is on-going.
    * ✔ We set up [a page to monitor the activity status of all Apache Logging Services projects](https://logging-activity-monitor.staged.apache.org/). This increases our transparency and helps users to give well-informed decisions on unmaintained projects.

## Software Bill of Materials (SBOM)

SBOMs are crucial in determining the impact surface of software against vulnerabilities. They capture an inventory of software components and services and their dependency relationships. This is essential for risk identification in the face of software supply chain attacks.

Our research yielded that there are two industry standards for SBOM specification in the market:
* [CycloneDX SBOM](https://cyclonedx.org/capabilities/sbom/)
* [SPDX](https://spdx.dev/)

[...]

PR #1921, opened 2023-10-30
Fix FindSecBugs alerts

ppkarwasz:
[FindSecBugs](https://find-sec-bugs.github.io/), gives several alerts concerning alleged security problems in our code.

While these are almost certainly false positives, we need to check each one of them before suppressing the related warning.

Related to #1707.


vy (approved):
Great job @ppkarwasz! :100:

ppkarwasz:
This is also related to #1849.

Issue #1773, opened 2023-09-08
Move Log4j JMX GUI to its own repository


ppkarwasz:
Log4j JMX GUI is essentially a separate product like Chainsaw and it shouldn't reside in the main Log4j repo.

It is a Java 8 application that relies on the `jconsole.jar` file that is not available in JDK 9+, hence it blocks the rest of the Maven modules from using a more recent JDK for compilation.

vy:
`log4j-jmx-gui` has been moved to its own repository ([logging-log4j-jmx-gui](/apache/logging-log4j-jmx-gui)) with a fully functional build and CI pipeline.

PR #1925, opened 2023-10-30
Remove jmx-gui module which is no longer in this repository

jonasrutishauser:
Completes #1773


ppkarwasz:
@jonasrutishauser,

Thank you for you PR, this also closes #1926.

PR #1927, opened 2023-10-30
Expose disruptor discard count in AsyncLoggerContext

adamthom-amzn:
I would like to monitor the disruptor's discard count on a periodic basis so that I can tell if events are being discarded due to the queue being full, and how often it is happening if so. As it stands, this count is not accessible outside of the AsyncLoggerDisruptor, and the discard count is only published in a trace log when the logger is stopped.

I considered adding this to RingBufferAdminMBean but since it currently monitors the backing ring buffer, and not the disruptor itself, it did not seem to fit.

This is my first contribution so I'm unsure if this meets the bar for a changelog entry, and whether or not this needs an issue to be cut beforehand, and if I incremented the log4j-core version correctly (it was required for `mvnw verify` to work, otherwise the baseline plugin failed the build).

## Checklist

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
@@ -802,4 +802,15 @@ public class LoggerContext extends AbstractLifeCycle
     protected Logger newInstance(final LoggerContext ctx, final String name, final MessageFactory messageFactory) {
         return new Logger(ctx, name, messageFactory);
     }
+
+    /**
+     * Returns the number of logging events discarded by the disruptor's queue full policy, or 0 if the disruptor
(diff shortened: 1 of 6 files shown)


adamthom-amzn:
Is this unfeasible to add for monitoring in the interim, and I should just track the metrics API PR?

vy:
@adamthom-amzn, my _personal_ opinion (with my PMC hat on) is as follows: Your changes touch parts of the code that
1. a majority of the active maintainers don't have expertise on
2. have quite some abstraction leakage already in place, and hence maintainers are hesitant to punch another hole

You can try pitching your idea in the `dev@logging.apache.org` mailing list.

ppkarwasz:
I am closing this PR, since #2469 seems to have a larger consensus.

The latter unfortunately is stack at the end of a long TODO list.

PR #1856, opened 2023-10-11
Support for disruptor-4.0.0

Dreeam-qwq:
Add support for disruptor 4.0.0 to close [LOG4J2-3595](https://issues.apache.org/jira/browse/LOG4J2-3595) (#1829). Tested and looks good.

Issue #1829, opened 2023-09-30
Support for com.lmax:disruptor:4


ashr123:
## Description

When I run with async logger and `com.lmax:disruptor:4`, I get the following error:

> ```
> Exception in thread "main" java.lang.NoClassDefFoundError: com/lmax/disruptor/SequenceReportingEventHandler
> 	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
> 	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
> 	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
> 	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
> 	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
> 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
> 	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
> 	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
> 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
> 	at org.apache.logging.log4j.core.async.AsyncLoggerConfigDisruptor.start(AsyncLoggerConfigDisruptor.java:226)
> 	at org.apache.logging.log4j.core.config.AbstractConfiguration.start(AbstractConfiguration.java:312)
> 	at org.apache.logging.log4j.core.LoggerContext.setConfiguration(LoggerContext.java:621)
> 	at org.apache.logging.log4j.core.LoggerContext.reconfigure(LoggerContext.java:694)
> 	at org.apache.logging.log4j.core.LoggerContext.reconfigure(LoggerContext.java:711)
> 	at org.apache.logging.log4j.core.LoggerContext.start(LoggerContext.java:253)
> 	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:155)
> 	at org.apache.logging.log4j.core.impl.Log4jContextFactory.getContext(Log4jContextFactory.java:47)
> 	at org.apache.logging.log4j.LogManager.getContext(LogManager.java:196)
[...]

PR #1937, opened 2023-11-01
Upgrade LMAX Disruptor to version 4.0.0

tomerarazy:


vy:
@tomerarazy, I don't think this will work. See #1829 and [LOG4J2-3595](https://issues.apache.org/jira/browse/LOG4J2-3595). We would appreciate it if you can submit a PR fixing the issue via the way described in LOG4J2-3595.

tomerarazy:
> @tomerarazy, I don't think this will work. See #1829 and [LOG4J2-3595](https://issues.apache.org/jira/browse/LOG4J2-3595). We would appreciate it if you can submit a PR fixing the issue via the way described in [LOG4J2-3595](https://issues.apache.org/jira/browse/LOG4J2-3595).

If I understand the issue is backward compatibility with older Java versions, I don't mind taking a shot at this but I'm not familiar with the log4j (other than as a user) so I don't even know where to start looking on how different Java versions affect which code we're loading, can you please provide some pointers to where I can start looking?

ppkarwasz:
I would prefer to use the trick I described in #1856.

In Disruptor 4.x two things changed:

- `EventHandler` has new default methods that come from its previous sub-interfaces,
- some sub-interfaces of `EventHandler` disappeared.

We can easily work with this, so that:

- our code compiles with Disruptor 3.x,
- users can use both Disruptor 3.x and Disruptor 4.x at runtime.

The details are in #1856.

jvz:
I'm going to go ahead and apply this idea to main since that has a baseline requirement of Java 11 already.

jvz:
Straightforward approach in main: https://github.com/apache/logging-log4j2/commit/9dcb10bf300432b353cb793a81d1202bce36ef77

tcmot:
Refer to this ticket.

https://github.com/LMAX-Exchange/disruptor/issues/467#issuecomment-1855643967

PR #1948, opened 2023-11-06
Refactor security page

vy:
This implements the following changes in [the security page](https://logging.apache.org/log4j/2.x/security.html):

* Switched to using [the Maven version range syntax](https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html) to denote affected versions
* Switched from [MITRE](https://cve.mitre.org) to [NVD](https://nvd.nist.gov/) – Both are in essence funded by the same organization (i.e., the U.S. Government) and contain identical data. Though MITRE seems dead and NVD looks alive with more bells and whistles.
* Switched to CVSS 3.x severity and metrics
* Improved navigation with more user-centric sections and expanded menu
* Added _"Components affected"_ (e.g., `log4j-core`) field
* Added _"Versions fixed"_ field
* Removed duplicate content
* Simplified textual content – kept basic description and mitigation techniques
* Removed mitigation techniques worded as _"alternatively you can do X, but it won't really help"_

### Review aid

- Old [security.md](https://github.com/apache/logging-log4j2/blob/2.x/src/site/markdown/security.md)
- New [security.adoc](https://github.com/apache/logging-log4j2/blob/security-page/src/site/asciidoc/security.adoc)

### References

- Structuring and wording are inspired by [the PostgreSQL Security page](https://www.postgresql.org/support/security/).
- This work is a part of the #1707 bundle.


src/site/asciidoc/security.adoc
```diff
@@ -0,0 +1,313 @@
...
+====
+
+[#reporting]
+== Reporting vulnerabilities
+
+If you have encountered an unlisted security vulnerability or other unexpected behaviour that has a security impact, or if the descriptions here are incomplete, please report them **privately** to mailto:security@logging.apache.org[the Log4j Security Team].
+
+[WARNING]
+====
+Reports assuming attacker's access to the Log4j configuration will not qualify as a vulnerability.
```
jvz, line 46: We can probably expand on this a little bit. Kind of like:

"The threat model that Log4j uses considers configuration files as safe input controlled by the programmer; potential vulnerabilities that require the ability to modify a configuration are not considered vulnerabilities as the required access to do so implies the attacker can execute arbitrary code."

Or something along those lines. Essentially, our general threat model revolves around log messages or other user input that can cause itself to be processed in such a way that it breaks security expectations. For example, logging a string that looks like a template string shouldn't expand the template since that's not an intentional feature.

src/site/asciidoc/security.adoc
```diff
@@ -0,0 +1,313 @@
...
+====
+
+[#policy]
+== Vulnerability handling policy
+
+The Apache Log4j Security Team follows the https://www.apache.org/security/committers.html[ASF Project Security] guide for handling security vulnerabilities.
+
+Found security vulnerabilities are subject to voting (by means of https://logging.apache.org/guidelines.html[_lazy approval_], preferably) in the private mailto:security@logging.apache.org[security mailing list] before creating a CVE and populating its associated content.
```
jvz, line 54: I think it would make a little more sense to say "reported" instead of "found" in this sentence, but either way, it makes sense.

jvz (approved):
Reads well to me. Thanks for organizing this!

src/site/asciidoc/security.adoc
vy: Implemented in 2f3662881185b85337dbc73bd22cf38f674a5189.

src/site/asciidoc/security.adoc
vy: Implemented in 2f3662881185b85337dbc73bd22cf38f674a5189.

PR #1961, opened 2023-11-10
Fix Error Prone warnings in `log4j-api`

### LINKED ISSUE #1850: Fix main ErrorProne issues
Since #1780, [Error Prone](https://errorprone.info/) is running on both development branches.

Error Prone classifies [bug patterns](https://errorprone.info/bugpatterns) into two categories: `ERROR` and `WARNING`, while a third category `SUGGESTION` is experimental.

While all `ERROR`s have been recently fixed, we need to scroll through the detected warnings to see, which one should be corrected.

ppkarwasz:
This suppresses or fixes most Error Prone warning in `log4j-api`.

Only two dozens warnings are left, mainly in the
[TypeParameterUnusedInFormals](https://errorprone.info/bugpattern/TypeParameterUnusedInFormals), which can not be removed for source compatibility reasons.

Closes #1850.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
@@ -65,7 +65,7 @@ public class LogManager {
 
     private static volatile LoggerContextFactory factory;
 
-    /**
+    /*
      * Scans the classpath to find all logging implementation. Currently, only one will be used but this could be
      * extended to allow multiple implementations to be used.
      */
@@ -92,7 +92,7 @@ public class LogManager {
                     final Class<? extends LoggerContextFactory> factoryClass = provider.loadLoggerContextFactory();
                     if (factoryClass != null) {
                         try {
-                            factories.put(provider.getPriority(), factoryClass.newInstance());
+                            factories.put(provider.getPriority(), factoryClass.getDeclaredConstructor().newInstance());
                         } catch (final Exception e) {
                             LOGGER.error("Unable to create class {} specified in provider URL {}",
                                     factoryClass.getName(), provider.getUrl(), e);
--- a/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
@@ -92,7 +92,7 @@ public class LogManager {
                     final Class<? extends LoggerContextFactory> factoryClass = provider.loadLoggerContextFactory();
                     if (factoryClass != null) {
                         try {
-                            factories.put(provider.getPriority(), factoryClass.getDeclaredConstructor().newInstance());
+                            factories.put(provider.getPriority(), LoaderUtil.newInstanceOf(factoryClass));
                         } catch (final Exception e) {
                             LOGGER.error("Unable to create class {} specified in provider URL {}",
                                     factoryClass.getName(), provider.getUrl(), e);
--- a/log4j-api/src/main/java/org/apache/logging/log4j/MarkerManager.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/MarkerManager.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+import com.google.errorprone.annotations.InlineMe;
 import org.apache.logging.log4j.util.PerformanceSensitive;
 import org.apache.logging.log4j.util.StringBuilderFormattable;
 
@@ -83,7 +84,7 @@ public final class MarkerManager {
         if (parentMarker == null) {
             throw new IllegalArgumentException("Parent Marker " + parent + " has not been defined");
         }
-        return getMarker(name, parentMarker);
+        return getMarker(name).addParents(parentMarker);
     }
 
     /**
@@ -95,6 +96,9 @@ public final class MarkerManager {
      * @throws IllegalArgumentException if any argument is {@code null}
      * @deprecated Use the Marker add or set methods to add parent Markers. Will be removed by final GA release.
      */
+    @InlineMe(
+            replacement = "MarkerManager.getMarker(name).addParents(parent)",
+            imports = "org.apache.logging.log4j.MarkerManager")
     @Deprecated
     public static Marker getMarker(final String name, final Marker parent) {
         return getMarker(name).addParents(parent);
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
@@ -58,8 +58,8 @@ public class DefaultFlowMessageFactory implements FlowMessageFactory, Serializab
 
     private static MessageFactory createDefaultMessageFactory() {
         try {
-            return AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.newInstance();
-        } catch (final InstantiationException | IllegalAccessException e) {
+            return AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.getDeclaredConstructor().newInstance();
+        } catch (final ReflectiveOperationException e) {
             throw new IllegalStateException(e);
         }
     }
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
@@ -19,6 +19,7 @@ package org.apache.logging.log4j.message;
 import java.io.Serializable;
 
 import org.apache.logging.log4j.spi.AbstractLogger;
+import org.apache.logging.log4j.util.LoaderUtil;
 import org.apache.logging.log4j.util.StringBuilderFormattable;
 import org.apache.logging.log4j.util.StringBuilders;
 import org.apache.logging.log4j.util.Strings;
@@ -58,7 +59,7 @@ public class DefaultFlowMessageFactory implements FlowMessageFactory, Serializab
 
     private static MessageFactory createDefaultMessageFactory() {
         try {
-            return AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.getDeclaredConstructor().newInstance();
+            return LoaderUtil.newInstanceOf(AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS);
         } catch (final ReflectiveOperationException e) {
             throw new IllegalStateException(e);
         }
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/MessageFormatMessage.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/MessageFormatMessage.java
@@ -29,14 +29,15 @@ import org.apache.logging.log4j.status.StatusLogger;
 
 /**
  * Handles messages that consist of a format string conforming to java.text.MessageFormat.
- *
- * @serial In version 2.1, due to a bug in the serialization format, the serialization format was changed along with
- * its {@code serialVersionUID} value.
  */
 public class MessageFormatMessage implements Message {
 
     private static final Logger LOGGER = StatusLogger.getLogger();
 
+    /**
+     * @serial In version 2.1, due to a bug in the serialization format, the serialization format was changed along with
+     * its {@code serialVersionUID} value.
+     */
     private static final long serialVersionUID = 1L;
 
     private static final int HASHVAL = 31;
@@ -124,7 +125,7 @@ public class MessageFormatMessage implements Message {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof MessageFormatMessage)) {
             return false;
         }
 
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
@@ -30,6 +30,10 @@ import org.apache.logging.log4j.util.PerformanceSensitive;
  * @since 2.6
  */
 @PerformanceSensitive("allocation")
+/*
+ * https://errorprone.info/bugpattern/ThreadLocalUsage
+ * Instance thread locals are not a problem here, since this class is almost a singleton.
+ */
 @SuppressWarnings("ThreadLocalUsage")
 public final class ReusableMessageFactory implements MessageFactory2, Serializable {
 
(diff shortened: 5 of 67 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
```diff
@@ -92,7 +92,7 @@ public class LogManager {
                     final Class<? extends LoggerContextFactory> factoryClass = provider.loadLoggerContextFactory();
                     if (factoryClass != null) {
                         try {
-                            factories.put(provider.getPriority(), factoryClass.newInstance());
+                            factories.put(provider.getPriority(), factoryClass.getDeclaredConstructor().newInstance());
```
jvz, line 95: You can also use `LoaderUtil` here. I think that's what we're doing in main anyways.

log4j-api/src/main/java/org/apache/logging/log4j/MarkerManager.java
```diff
@@ -95,6 +96,9 @@ public static Marker getMarker(final String name, final String parent) {
      * @throws IllegalArgumentException if any argument is {@code null}
      * @deprecated Use the Marker add or set methods to add parent Markers. Will be removed by final GA release.
      */
+    @InlineMe(
```
jvz, line 99: Oh this looks cool!

log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
```diff
@@ -58,8 +58,8 @@ public DefaultFlowMessageFactory(final String entryText, final String exitText)
 
     private static MessageFactory createDefaultMessageFactory() {
         try {
-            return AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.newInstance();
-        } catch (final InstantiationException | IllegalAccessException e) {
+            return AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.getDeclaredConstructor().newInstance();
```
jvz, line 61: Similarly, this could use `LoaderUtil::newInstanceOfUnchecked` or similar.

log4j-api/src/main/java/org/apache/logging/log4j/message/MessageFormatMessage.java
```diff
@@ -124,7 +125,7 @@ public boolean equals(final Object o) {
         if (this == o) {
             return true;
         }
-        if (o == null || getClass() != o.getClass()) {
+        if (!(o instanceof MessageFormatMessage)) {
```
jvz, line 128: You got an IntelliJ template for this? I'm pretty sure this is generated from the default equals and hashCode templates there.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
```diff
@@ -30,6 +30,7 @@
  * @since 2.6
  */
 @PerformanceSensitive("allocation")
+@SuppressWarnings("ThreadLocalUsage")
```
jvz, line 33: Heh, this should end up getting removed once we finalize the recycler API!

log4j-api/src/main/java/org/apache/logging/log4j/util/LowLevelLogUtil.java
```diff
@@ -32,6 +32,7 @@
  */
 final class LowLevelLogUtil {
 
+    @SuppressWarnings("DefaultCharset")
```
jvz, line 35: Might be useful to leave a comment, too, that this is intentional functionality? Though I suppose the warning suppression sort of implies that.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
```diff
@@ -137,9 +137,12 @@ final class Util {
...
          * @return the property broken into lower case tokens
          */
+        // https://errorprone.info/bugpattern/CollectionUndefinedEquality
+        @SuppressWarnings("CollectionUndefinedEquality")
         public static List<CharSequence> tokenize(final CharSequence value) {
-            if (CACHE.containsKey(value)) {
-                return CACHE.get(value);
+            // `value` should be a `String`
+            if (CACHE.containsKey(value.toString())) {
+                return CACHE.get(value.toString());
```
jvz, line 145: Oops, I think this was my bad.

jvz (approved):
LGTM!

log4j-api/src/main/java/org/apache/logging/log4j/message/MessageFormatMessage.java
ppkarwasz: No, I just applied the patches Error Prone suggested (cf. [patching](http://errorprone.info/docs/patching)).

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
ppkarwasz: Sorry I didn't add the meaning of all these warnings in comments (cf. [Error Prone documentation](http://errorprone.info/bugpatterns))

[ThreadLocalUsage](http://errorprone.info/bugpattern/ThreadLocalUsage) actually means: only use `static` thread locals.

I believe we should use instance thread locals for singleton's like this. We only need to ensure that this is a singleton (except in tests). E.g. if a user asks for the message factory `org.apache.logging.log4j.message.ReusableMessageFactory`, we can do like `ServiceLoader` in Java 9+: look for a static `#provider` method and only if this is not found instantiate the class.

log4j-api/src/main/java/org/apache/logging/log4j/util/LowLevelLogUtil.java
ppkarwasz: Sure, I'll add some comments and links to the error descriptions.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
ppkarwasz: I noticed this before, but the method is only called with `String` parameters, so it is not actually a problem.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableMessageFactory.java
jvz: Ugh, that rule is more of a warning than an error as it's easy to accidentally use up far more memory than expected depending on the usage pattern. Using a static thread local field can also leak memory.

ppkarwasz: I agree, all of the changes in this PR are warnings and some code changes are there just to help Error Prone understanding that we understand the problem.

log4j-api/src/main/java/org/apache/logging/log4j/message/SimpleMessage.java
```diff
@@ -89,17 +90,24 @@ public Object[] getParameters() {
...
         final SimpleMessage that = (SimpleMessage) o;
 
-        return !(charSequence != null ? !charSequence.equals(that.charSequence) : that.charSequence != null);
+        /*
+         * https://errorprone.info/bugpattern/UndefinedEquals
+         *
+         * If the char sequences are different, we fall back on string comparison.
+         */
+        return Objects.equals(this.charSequence, that.charSequence)
+                || Objects.equals(this.getFormattedMessage(), that.getFormattedMessage());
```
ppkarwasz, line 110: This change makes the `equals` method more useful: two objects are equal if they have the same message.

However it has the side-effect of evaluating `charSequence.toString()`. @jvz, what do you think?

jvz: How often are messages checked for equality? If it's not a hot path, then I say go for it. `CharSequence::compare` appears to be a Java 11 method, so there's no standard way to compare two char sequences apparently.

ppkarwasz:
I have refactored all `Class#newInstance()` calls to use `LoaderUtil#newInstanceOf` and some `clazz.getDeclaredConstructor().newInstance()` calls.

Since `Class#newInstance` is caller sensitive, this might break the instantiation of private/package-access classes. However I believe that the callers always use `public` and exported classes.

log4j-api/src/main/java/org/apache/logging/log4j/spi/ThreadContextMapFactory.java
```diff
@@ -88,7 +89,7 @@ public static ThreadContextMap createThreadContextMap() {
             try {
                 final Class<?> clazz = cl.loadClass(ThreadContextMapName);
                 if (ThreadContextMap.class.isAssignableFrom(clazz)) {
-                    result = (ThreadContextMap) clazz.newInstance();
+                    result = (ThreadContextMap) clazz.getDeclaredConstructor().newInstance();
```
vy, line 92: Why is this not a `LoaderUtil.newInstanceOf(clazz)`?

log4j-api/src/main/java/org/apache/logging/log4j/util/Base64Util.java
```diff
@@ -53,7 +54,7 @@ public static String encode(final String str) {
         if (str == null) {
             return null;
         }
-        final byte [] data = str.getBytes();
+        final byte [] data = str.getBytes(Charset.defaultCharset());
```
vy, line 57: Shouldn't rather we getting the charset from the caller? _Default charset_ is an assumption and can very well be wrong.

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
```diff
@@ -107,7 +107,7 @@ private static int getMajorVersion() {
     }
 
     static int getMajorVersion(final String version) {
-        final String[] parts = version.split("-|\\.");
+        final String[] parts = version.split("-|\\.", -1);
```
vy, line 110: We are only interested in first two parts, shouldn't this be `2` instead?

log4j-api/src/main/java/org/apache/logging/log4j/util/ProcessIdUtil.java
```diff
@@ -38,7 +40,7 @@ public static String getProcessId() {
             final Object runtimeMXBean = getRuntimeMXBean.invoke(null);
             final String name = (String) getName.invoke(runtimeMXBean);
             //String name = ManagementFactory.getRuntimeMXBean().getName(); //JMX not allowed on Android
-            return name.split("@")[0]; // likely works on most platforms
+            return name.split("@", -1)[0]; // likely works on most platforms
```
vy, line 43: Shouldn't this be `1` instead?

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -82,35 +84,31 @@ static int calculateRingBufferSize(final String propertyName) {
...
+        ExceptionHandler<RingBufferLogEvent> handler = null;
         try {
-            @SuppressWarnings("unchecked")
-            final Class<? extends ExceptionHandler<RingBufferLogEvent>> klass =
-                (Class<? extends ExceptionHandler<RingBufferLogEvent>>) Loader.loadClass(cls);
-            return klass.newInstance();
-        } catch (final Exception ignored) {
-            LOGGER.debug("Invalid AsyncLogger.ExceptionHandler value: error creating {}: ", cls, ignored);
-            return new AsyncLoggerDefaultExceptionHandler();
+            handler =
```
vy, line 89: You don't need a `handler` variable'; you can simply place a `return` statement here and at the `catch` return the default.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -82,35 +84,31 @@ static int calculateRingBufferSize(final String propertyName) {
...
+        ExceptionHandler<AsyncLoggerConfigDisruptor.Log4jEventWrapper> handler = null;
         try {
-            @SuppressWarnings("unchecked")
-            final Class<? extends ExceptionHandler<AsyncLoggerConfigDisruptor.Log4jEventWrapper>> klass =
-                    (Class<? extends ExceptionHandler<AsyncLoggerConfigDisruptor.Log4jEventWrapper>>) Loader.loadClass(cls);
-            return klass.newInstance();
-        } catch (final Exception ignored) {
-            LOGGER.debug("Invalid AsyncLoggerConfig.ExceptionHandler value: error creating {}: ", cls, ignored);
-            return new AsyncLoggerConfigDefaultExceptionHandler();
+            handler = LoaderUtil.newCheckedInstanceOfProperty(
```
vy, line 103: See my comment above. `handler` variable is not needed.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ContextDataInjectorFactory.java
```diff
@@ -66,21 +67,17 @@ public class ContextDataInjectorFactory {
...
-            return result;
+            injector =
+                    LoaderUtil.newCheckedInstanceOfProperty(CONTEXT_DATA_INJECTOR_PROPERTY, ContextDataInjector.class);
+        } catch (final ReflectiveOperationException e) {
+            StatusLogger.getLogger().warn("Could not create ContextDataInjector: {}", e.getMessage(), e);
+        }
+        if (injector != null) {
+            return injector;
         }
+        return createDefaultInjector();
```
vy, line 80: `injector` variable is not needed. See my earlier comments.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
ppkarwasz: [`LoaderUtil.newCheckedInstanceOfProperty`](https://github.com/apache/logging-log4j2/blob/ce3a6de7a53a879764dadf01aae35298a039e6be/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java#L330) can return `null` if the property is not defined. Since instantiation can also fail, leaving the variable at its default value of `null`, I do a single null-check for both cases.

log4j-api/src/main/java/org/apache/logging/log4j/spi/ThreadContextMapFactory.java
ppkarwasz: `LoaderUtil.newInstanceOf` caused problem with internal (package-private) implementations. I added a switch-case for those.

log4j-api/src/main/java/org/apache/logging/log4j/util/Base64Util.java
ppkarwasz: I marked the method as `@Deprecated` and removed its only user. The entire `Base64Util` class can be replaced with `Base64.getEncoder()` since Java 8.

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
ppkarwasz: Thanks, I corrected it.
(actually we need 3 parts: `major.minor.rest_of_string`).

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
ppkarwasz: I added a version of `newCheckedInstanceOfProperty` that takes a `Supplier<T>` as additional parameter to provide a default value.

However in `3.x` we should probably consider returning `Optional<T>`: this way we can always call `Optional#orElseGet(Supplier<T>)` without the need of two versions of each method.

ppkarwasz:
Besides refactoring some bugs indicated by Error Prone, this PR also replaces BND Baseline package exclusions with more limited [`@BaselineIgnore`](https://javadoc.io/doc/biz.aQute.bnd/biz.aQute.bnd.annotation/latest/aQute/bnd/annotation/baseline/BaselineIgnore.html) annotations.

Issue #1851, opened 2023-10-10
Increase build environment requirement to Java 17


ppkarwasz:
While Maven plugins have started requiring Java 11 a while ago (e.g. Spotless 2.30+), some new plugins like `bnd-maven-plugin` have bumped the requirement to Java 17.

Most of our code is Java 17-ready, but we must update some of our tests that modify `final private` fields.

This is a blocker for apache/logging-parent#41.

ppkarwasz:
Fixed by #2021

PR #1964, opened 2023-11-12
Work towards supporting Java 17 builds

jvz:
If you add a dependency on GraalJS to `log4j-core-test`, that brings down the test failure count to 14 when run with Java 21. There are some tests that have to be disabled for later versions due to Java changes such as the removal of custom security managers (throws an exception right now in Java 21) and the removal of Nashorn (we need an explicit JavaScript `ScriptEngine` dependency in Java 17+). It's possible that Nashorn is available as a standalone dependency somewhere, but I haven't looked yet. When run without a JavaScript engine, we get 36 test failures and 2 test errors, all of which are script-related tests. Thus, this gets everything working again except for JavaScript `ScriptEngine` use.

### DIFF
--- a/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
+++ b/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
@@ -18,6 +18,8 @@ package org.apache.logging.log4j.test.junit;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
@@ -146,9 +148,15 @@ class StatusLoggerExtension extends TypeBasedParameterResolver<ListStatusListene
                         .withThrowable(data.getThrowable())
                         .withLocation(data.getStackTraceElement())
                         .log("{} {}",
-                                DateTimeFormatter.ISO_LOCAL_TIME.format(Instant.ofEpochMilli(data.getTimestamp())),
+                                formatLocalTime(data.getTimestamp()),
                                 data.getMessage());
             });
+        }
+
+        private static String formatLocalTime(final long epochMilli) {
+            final Instant instant = Instant.ofEpochMilli(epochMilli);
+            final LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+            return localDateTime.format(DateTimeFormatter.ISO_LOCAL_TIME);
         }
     }
 
--- a/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
+++ b/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
@@ -18,7 +18,6 @@ package org.apache.logging.log4j.test.junit;
 
 import java.io.IOException;
 import java.time.Instant;
-import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
@@ -154,9 +153,9 @@ class StatusLoggerExtension extends TypeBasedParameterResolver<ListStatusListene
         }
 
         private static String formatLocalTime(final long epochMilli) {
-            final Instant instant = Instant.ofEpochMilli(epochMilli);
-            final LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
-            return localDateTime.format(DateTimeFormatter.ISO_LOCAL_TIME);
+            return DateTimeFormatter.ISO_LOCAL_TIME
+                    .withZone(ZoneId.systemDefault())
+                    .format(Instant.ofEpochMilli(epochMilli));
         }
     }
 
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
@@ -22,24 +22,20 @@ package org.apache.logging.log4j.util;
  * @since 2.6.2
  */
 public final class Constants {
-    /**
-     * {@code true} if we think we are running in a web container, based on the boolean value of system property
-     * "log4j2.is.webapp", or (if this system property is not set) whether the  {@code javax.servlet.Servlet} class
-     * is present in the classpath.
-     */
-    public static final boolean IS_WEB_APP = PropertiesUtil.getProperties().getBooleanProperty(
-            "log4j2.is.webapp", isClassAvailable("javax.servlet.Servlet")
-                    || isClassAvailable("jakarta.servlet.Servlet"));
+    private static final LazyBoolean WEB_APP = new LazyBoolean(() -> isWebApp(PropertiesUtil.getProperties()));
+    private static final LazyBoolean USE_THREAD_LOCALS = new LazyBoolean(() -> isUseThreadLocals(PropertiesUtil.getProperties()));
 
     /**
-     * Kill switch for object pooling in ThreadLocals that enables much of the LOG4J2-1270 no-GC behaviour.
-     * <p>
-     * {@code True} for non-{@link #IS_WEB_APP web apps}, disable by setting system property
-     * "log4j2.enable.threadlocals" to "false".
-     * </p>
+     * @deprecated use {@link #isWebApp()}
      */
-    public static final boolean ENABLE_THREADLOCALS = !IS_WEB_APP && PropertiesUtil.getProperties().getBooleanProperty(
-            "log4j2.enable.threadlocals", true);
+    @Deprecated
+    public static final boolean IS_WEB_APP = WEB_APP.getAsBoolean();
+
+    /**
+     * @deprecated use {@link #isUseThreadLocals()}
+     */
+    @Deprecated
+    public static final boolean ENABLE_THREADLOCALS = USE_THREAD_LOCALS.getAsBoolean();
 
     /**
      * Java major version.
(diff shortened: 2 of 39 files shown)


log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
```diff
@@ -146,10 +148,16 @@ public void handleException(final ExtensionContext context, final Throwable thro
...
+                                formatLocalTime(data.getTimestamp()),
                                 data.getMessage());
             });
         }
+
+        private static String formatLocalTime(final long epochMilli) {
+            final Instant instant = Instant.ofEpochMilli(epochMilli);
+            final LocalDateTime localDateTime = LocalDateTime.ofInstant(instant, ZoneId.systemDefault());
+            return localDateTime.format(DateTimeFormatter.ISO_LOCAL_TIME);
+        }
```
jvz, line 160: Turns out this change didn't work, so I fixed it and forgot to push to 2.x directly, so here you go.

log4j-core-test/pom.xml
```diff
@@ -379,6 +379,22 @@
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>
+            <!-- for environment variable access -->
+            --add-opens java.base/java.util=ALL-UNNAMED
+            --add-opens java.base/java.lang=ALL-UNNAMED
+            <!-- for URLStreamHandlerFactory access -->
+            --add-opens java.base/java.net=ALL-UNNAMED
```
jvz, line 394: We have some tests that try to modify internals of `URLStreamHandlerFactory` to test something related to handling `URL`s. Yet another sign we could use our own `Resource`-style API. :)

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/ClockFactoryTest.java
```diff
@@ -16,32 +16,15 @@
...
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledOnJre;
-import org.junit.jupiter.api.condition.JRE;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertSame;
 
-// as of Java 12, final fields can no longer be overwritten via reflection
-@EnabledOnJre({ JRE.JAVA_8, JRE.JAVA_9, JRE.JAVA_10, JRE.JAVA_11 })
```
jvz, line 32: No longer requiring reflection here, so we get more tests to run on Java 17, too! Yay.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
ppkarwasz: Somehow I always get the time API wrong. It should be:
```
DateTimeFormatter.ISO_LOCAL_TIME.withZone(ZoneId.systemDefault()).format(Instant.ofEpochMilli(data.getTimestamp())),
```

jvz: Oh that makes sense; something needs to be converted to a time zone.

ppkarwasz:
Fixes #1851.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Constants.java
```diff
@@ -76,8 +76,10 @@ public final class Constants {
      * {@code true} if we think we are running in a web container, based on the boolean value of system property
      * "log4j2.is.webapp", or (if this system property is not set) whether the  {@code javax.servlet.Servlet} class
      * is present in the classpath.
+     * @deprecated use {@link org.apache.logging.log4j.util.Constants#isWebApp()}
      */
-    public static final boolean IS_WEB_APP = org.apache.logging.log4j.util.Constants.IS_WEB_APP;
+    @Deprecated
```
vy, line 81: Why are we adding deprecated fields?

vy (changes requested):
@jvz, I miss a clear scope for this PR. To me it feels like (could very well be due to my lack of understanding) changes are at arbitrary places and doesn't much contribute to the PR goal: Java 17 support.

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
```diff
@@ -22,24 +22,20 @@
  * @since 2.6.2
  */
 public final class Constants {
+    private static final LazyBoolean WEB_APP = new LazyBoolean(() -> isWebApp(PropertiesUtil.getProperties()));
```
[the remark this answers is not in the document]
jvz: It's to make testing this stuff not require breaking encapsulation or other hacks which aren't allowed in newer versions of Java. This is largely the same basic refactoring as was done in main for these two properties.

log4j-core-its/src/test/java/org/apache/logging/log4j/core/async/perftest/ResponseTimeTest.java
```diff
@@ -141,7 +141,7 @@ public static void main(final String[] args) throws Exception {
         runLatencyTest(logger, WARMUP_DURATION_MILLIS, WARMUP_COUNT, loadMessagesPerSec, idleStrategy,
                 warmupServiceTmHistograms, warmupResponseTmHistograms, threadCount);
         System.out.println("-----------------Warmup done. load=" + loadMessagesPerSec);
-        if (!Constants.ENABLE_DIRECT_ENCODERS || !Constants.ENABLE_THREADLOCALS) {
+        if (!Constants.ENABLE_DIRECT_ENCODERS || !org.apache.logging.log4j.util.Constants.isUseThreadLocals()) {
```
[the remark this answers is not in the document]
jvz: I didn't want to refactor all the use of constants in 2.x as that's something related to the context-scoped properties feature introduced in 3.x. However, if we have tests that would otherwise need to, say, modify a static final field, then yes, I'd refactor this.

log4j-core-test/pom.xml
jvz: Also seems like the tests using this are related to `ResolverUtil` which is going away in 3.x as we no longer do class path scanning.

jvz:
> @jvz, I miss a clear scope for this PR. To me it feels like (could very well be due to my lack of understanding) changes are at arbitrary places and doesn't much contribute to the PR goal: Java 17 support.

The scope of this PR is to fix a ton of tests that fail in Java 17 or Java 21. Most of those tests were failing due to breaking encapsulation rules of Java such as trying to redefine a static final field somewhere. Believe me, I wouldn't be making changes to 2.x that aren't in scope of anything important! I'm all-in on the main branch.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/StatusLoggerExtension.java
ppkarwasz: I pushed this change to both `2.x` and `main` (it was also breaking #1961), so you can rebase.

log4j-core-its/src/test/java/org/apache/logging/log4j/core/async/perftest/ResponseTimeTest.java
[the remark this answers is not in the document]
vy: So we are converting constants that are subject to modification in tests to `Lazy*` instances and leaving out the rest intact?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/Constants.java
vy: I think we are allowed to break binary backward compatibility in non-API[1] classes. Otherwise, doesn't this conflict with the rationale of a major release? Maybe this is indeed a gray area and we should discuss further. Though my stance is, as long as it is improving the architecture and not breaking the API[1], go ahead.

[1] For me, API translates to the `log4j-api` and the file-based configuration API.

vy:
> > @jvz, I miss a clear scope for this PR. To me it feels like (could very well be due to my lack of understanding) changes are at arbitrary places and doesn't much contribute to the PR goal: Java 17 support.
> 
> The scope of this PR is to fix a ton of tests that fail in Java 17 or Java 21. Most of those tests were failing due to breaking encapsulation rules of Java such as trying to redefine a static final field somewhere. Believe me, I wouldn't be making changes to 2.x that aren't in scope of anything important! I'm all-in on the main branch.

Okay. I would really appreciate it if you can update the description of the ticket accordingly. What is the goal? What is failing? The list of failures addressed, etc. One last remark: I would only address 17 and, if needed, worry about 21 in a follow-up ticket.

jvz:
I ran 2.x branch in Java 17 and Java 21 by adjusting the enforcer version range in the pom. Based on the numerous failing and erring tests, I iterated through them by fixing the errors and came to this. I didn't refactor all the static properties because that's something we're working on more so in main, and I only wanted to update the minimum required to get the tests updated.

At this point, just reuse whatever you'd like to incorporate to 2.x; I'll continue working on main.

log4j-core-its/src/test/java/org/apache/logging/log4j/core/async/perftest/ResponseTimeTest.java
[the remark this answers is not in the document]
jvz: Yes, that was the idea here. In main, that's something we're trying to clean up more for the properties enhancement, but it requires a lot of refactoring, hence why it's limited to main.

ppkarwasz:
@jvz,

I'll apply the Palantir formatter to `main`. As Carter suggested, it should be pretty easy to merge this PR if you do the same here.

ppkarwasz:
@jvz,

When you fix the tests under Java 17, can you close #1851 and fix apache/logging-parent#62 ?

ppkarwasz (commented):
@jvz,

Could you split the `Constants` refactoring into another PR? As far as I can tell, this is related to running tests in the same JVM, but has nothing to do with test failures under JRE 17.

I am a big fan of minimal solutions, so that we can solve one problem at a time. I don't know about the others, but I have difficulties reviewing more than 10 files of changes.

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
```diff
@@ -120,4 +115,44 @@ static int getMajorVersion(final String version) {
...
+
+    @InternalApi
+    public static void setWebApp(final boolean webApp) {
+        WEB_APP.setAsBoolean(webApp);
+    }
+
+    @InternalApi
+    public static void resetWebApp() {
+        WEB_APP.reset();
+    }
```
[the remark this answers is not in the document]
jvz: We can use `@RestrictedApi` from error prone as that's something the compiler can enforce (at least when using error prone). But right, this whole concept of these two values shouldn't really be public as it's an internal detail used by some tests. Should think of another way to do this.

jvz:
I'm working on the main branch; this PR is a draft of ideas needed to get Java 17 working for tests. I won't be making more changes to this branch as I don't have the time.

ppkarwasz:
@jvz,

I'll cherry-pick some changes from here to provide a minimal fix to tests on JRE 17.

PR #1970, opened 2023-11-14
Switch Basic Authentication encoding to UTF-8

ppkarwasz:
[RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617) introduces a new `charset` parameter for the Basic authentication scheme with a single allowed value "UTF-8".

Clients that comply to this RFC should encode the user name and password using UTF-8 if the parameter is present. Clients that **always** use UTF-8 obviously comply to this rule.

Up until now Log4j used the system encoding for Basic authentication. This PR:

- switches the default encoding to UTF-8,
- adds a `log4j2.configurationAuthorizationEncoding` property to overwrite the default value.

This problem was detected by Error Prone in #1961.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/BasicAuthorizationProvider.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/BasicAuthorizationProvider.java
@@ -17,12 +17,15 @@
 package org.apache.logging.log4j.core.util;
 
 import java.net.URLConnection;
+import java.nio.charset.Charset;
+import java.util.Base64;
 
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.status.StatusLogger;
-import org.apache.logging.log4j.util.Base64Util;
 import org.apache.logging.log4j.util.LoaderUtil;
 import org.apache.logging.log4j.util.PropertiesUtil;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 /**
  * Provides the Basic Authorization header to a request.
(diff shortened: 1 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/util/BasicAuthorizationProvider.java
```diff
@@ -35,6 +38,11 @@ public class BasicAuthorizationProvider implements AuthorizationProvider {
     public static final String CONFIG_USER_NAME = "log4j2.configurationUserName";
     public static final String CONFIG_PASSWORD = "log4j2.configurationPassword";
     public static final String PASSWORD_DECRYPTOR = "log4j2.passwordDecryptor";
+    /*
+     * Properties used to specify the encoding in HTTP Basic Authentication
+     */
+    private static final String BASIC_AUTH_ENCODING = "log4j2.configurationAuthorizationEncoding";
+    private static final String SPRING_BASIC_AUTH_ENCODING = "logging.auth.encoding";
```
vy, line 45: I am not happy with
1. having `configuration` in a property name (isn't everything has something to do with some _configuration_?)
2. adding Spring-specific functionality to a Spring-agnostic product

But... The rest of the class has already embraced these wrong conventions. Hence, you successfully have mirrored it.

vy (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/util/BasicAuthorizationProvider.java
ppkarwasz: I think these property are transitional, because:

- most password are ASCII anyway, since you need sometimes to type them on a keyboard of an OS without advanced input methods,
- most modern OSes use `UTF-8` as system encoding anyway (Windows is the exception),
- using system encodings is obsolete (besides the system encoding of the client and server might differ).

I will **drop** these configuration properties in `3.x` and leave `UTF-8` as only encoding.

PR #2071, opened 2023-12-06
Replaces `spring-test` with custom JNDI mock

ppkarwasz:
Since Spring 6.x the JNDI mock is no longer available.

We replace it with a simplified custom version based on Tomcat's JNDI implementation.


ppkarwasz:
Closing this in favor of #2072.

Issue #2018, opened 2023-11-29
Port `log4j-spring-cloud-config-client` changes from `2.x`


vy:
This issue is more intricate than simply diff'ing the sources and porting necessary changes. Log4j 3 is incompatible with Spring Boot 2 and it needs to be upgraded to 3. See the disabled `Log4j2EventListenerTest` for details.

PR #2072, opened 2023-12-06
Upgrade to Spring Boot 3

vy:
This PR addresses ports `log4j-spring-cloud-config-client` changes from `2.x` (#2018) and upgrades to Spring Boot 3.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
@@ -121,6 +121,11 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
         return true;
     }
 
+    @Deprecated(since = "3.0.0", forRemoval = true)
+    public static ConfigurationFactory getInstance() {
+        return LoggerContext.getContext(false).getInstanceFactory().getInstance(KEY);
+    }
+
     /**
      * Required for Spring Boot.
      * @param props PropertiesUtil.
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/net/UrlConnectionFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/net/UrlConnectionFactory.java
@@ -56,6 +56,16 @@ public class UrlConnectionFactory {
     private static final String NO_PROTOCOLS = "_none";
     public static final String ALLOWED_PROTOCOLS = "log4j2.Configuration.allowedProtocols";
 
+    @Deprecated(since = "3.0.0", forRemoval = true)
+    public static <T extends URLConnection> T createConnection(
+            final URL url,
+            final long lastModifiedMillis,
+            final SslConfiguration sslConfiguration,
+            final AuthorizationProvider authorizationProvider)
+            throws IOException {
+        return createConnection(url, lastModifiedMillis, sslConfiguration, authorizationProvider, PropertiesUtil.getProperties());
+    }
+
     @SuppressFBWarnings(
             value = "URLCONNECTION_SSRF_FD",
             justification = "The URL parameter originates only from secure sources.")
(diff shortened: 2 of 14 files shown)


rgoers:
This looks good to me.

rgoers (approved):

jvz (approved):

log4j-jndi/src/main/java/org/apache/logging/log4j/jndi/JndiManager.java
```diff
@@ -194,6 +195,14 @@ protected boolean releaseSub(final long timeout, final TimeUnit timeUnit) {
         return JndiCloser.closeSilently(this.context);
     }
 
+    /**
+     * @return the active context
+     */
+    @Nullable
+    public Context getContext() {
+        return context;
+    }
```
ppkarwasz, line 204: I have mixed feeling about exposing `Context` (which is usually `InitialContext`, the source of all evil) from the public API.

log4j-jndi-test/src/main/java/org/apache/logging/log4j/jndi/test/junit/JndiRule.java
```diff
@@ -16,42 +16,113 @@
...
  *
  * @since 2.8
  */
+@SuppressWarnings("BanJNDI")
 public class JndiRule implements TestRule {
 
-    private final Map<String, Object> initialBindings;
+    static {
+        final InitialContextFactoryBuilder factoryBuilder =
+                factoryBuilderEnv -> factoryEnv -> new MemoryContext(new Hashtable<>()) {};
```
ppkarwasz, line 51: Why not storing this `MemoryContext` somewhere?

The `NamingManager.setInitialContextFactoryBuilder` method is irreversible (can be called only once), so the context used by `JndiManager.getDefaultManager()` will always be the one defined in here.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -121,6 +121,11 @@ protected boolean isActive() {
         return true;
     }
 
+    @Deprecated(since = "3.0.0", forRemoval = true)
+    public static ConfigurationFactory getInstance() {
+        return LoggerContext.getContext(false).getInstanceFactory().getInstance(KEY);
+    }
+
```
ppkarwasz, line 128: Not sure why this is required. I can not find any usage in this PR.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/UrlConnectionFactory.java
```diff
@@ -56,6 +56,17 @@ public class UrlConnectionFactory {
...
+    public static <T extends URLConnection> T createConnection(
+            final URL url,
+            final long lastModifiedMillis,
+            final SslConfiguration sslConfiguration,
+            final AuthorizationProvider authorizationProvider)
+            throws IOException {
+        return createConnection(
+                url, lastModifiedMillis, sslConfiguration, authorizationProvider, PropertiesUtil.getProperties());
+    }
+
```
ppkarwasz, line 69: Same as above.

ppkarwasz (commented):
I am not entirely sure about the usage of [Simple-JNDI](https://github.com/h-thurow/Simple-JNDI):

- the author still answered some issue reports last March,
- but the project didn't have any release since 2020 and its dependencies are from 2016 (`commons-io` has even a CVE).

`MemoryContext` could be easily replaced with [`NamingContext`](https://github.com/h-thurow/Simple-JNDI) for Apache Tomcat. The only difference with the latter is that it requires an explicit `createSubcontext` to create missing subcontexts.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: `Log4J2LoggingSystem` of Spring Boot uses this.

ppkarwasz:
> @ppkarwasz, I am not strongly opinionated about `simple-jndi`. It is what Spring Framework recommends as a successor and uses itself too.

If Spring recommends it, it is good enough for me.

PR #2076, opened 2023-12-08
Removes support for `EventData`

ppkarwasz:
The `org.slf4j.ext.EventData` class was a primitive equivalent of Log4j's `StructuredDataMessage`. QOS.CH published CVE-2018-8088 against this class and removed it from `slf4j-ext` version 1.7.26.

For a long time we kept support code to convert `EventData` to its Log4j equivalent for compatibility's sake: users that relied on this class could still use Log4j Core, but the functionality was **automatically** disabled once they upgraded `slf4j-ext`.

Since 5 years passed since `slfj4-ext` 1.7.26 was published, we can safely assume that all the users upgraded the library and the support code can be removed.


ppkarwasz:
@rgoers,

I have two doubts about the support for `slfj4-ext`:

- should I backport this change to `2.x`?
- once I remove the support for `EventData`, `log4j-slf4j-impl` has **no** code specifically written for `slf4j-ext`. Should I remove the tests that use `XLogger` too? The way I see it these tests can only discover errors in `slf4j-ext` itself, not our own code.

vy:
@ppkarwasz, LGTM.

I support the idea of removing `XLogger` tests too and backporting all this to `2.x`.

Issue #2075, opened 2023-12-07
Merge queue factory interfaces


ppkarwasz:
Right now there are multiple interfaces with a similar purpose in `main`:

- [`QueueFactory`](https://github.com/apache/logging-log4j2/blob/main/log4j-api/src/main/java/org/apache/logging/log4j/util/QueueFactory.java),
- [`QueueFactories.BoundedQueueFactory`](https://github.com/apache/logging-log4j2/blob/main/log4j-api/src/main/java/org/apache/logging/log4j/util/QueueFactories.java),
- [`BlockingQueueFactory`](https://github.com/apache/logging-log4j2/blob/main/log4j-core/src/main/java/org/apache/logging/log4j/core/async/BlockingQueueFactory.java).

I believe we should merge them into a single interface.

We should also move the two implementations of `BlockingQueueFactory` that depend on JCTools and Conversant Disruptor into their own modules. Since these implementations can be used in both `log4j-api` and `log4j-core` they should:

- be accessible via `ServiceLoader` and configurable via Log4j properties,
- be annotated as Log4j Plugins and configurable as Log4j plugins.

vy:
This issue is (mostly) fixed by #2104:

* `QueueFactories` and `QueueFactory` are removed
* `BlockingQueueFactory` is left in `log4j-core` with a JCTools-based implementation in the new `log4j-jctools` module

Issue #2016, opened 2023-11-28
[MS7] Branch `main` overhaul #2


vy:
This story aims to build on the work delivered in #1969 and continue with following tasks:

- Improve queue factory interface (#2075)
- Address leftover recycler reviews (d029b335b024ca79d2b039ddd51af88c56350594, 47ff3fd4db475a0aec3ba761e3ebd72254f46b36, 716eca0128303cbbc1aac26a75d69590efc8bf02, 156b34e848253db7510048bf9c57a36e707bd9b0)
- Overhaul `main` (targeting Log4j version `3.0.0`) branch
  - Upgrade compiler baseline to Java 17 (77245ddae8e9ffc2b355def5d87ce4d8b61ac819)
  - Auxiliary support files (9651905cbefbc2a92c6b98ff41bc60e4959b1aff)
  - #2142
  - #2154
  - #2076 
  - Port changes from `2.x`:
    - [Port `log4j-jpl` changes from `2.x`](https://github.com/apache/logging-log4j2/commit/6eed880f20dcd6ff1e2e2198d8de5053c67462d1)
    - #2160
    - [Port `log4j-kubernetes` changes from `2.x`](https://github.com/apache/logging-log4j2/commit/64226fef0b6dae54d4e11d461f0805a4ace271d7)
    - #2164
    - #2159
    - #2158
    - #2156
    - #2018
    - #2150
    - #2015

ppkarwasz:
I would also propose to bind #2076 to this task as a cleanup job.

The whole `log4j-slf4j-impl` should IMHO be removed from `3.x` or at least reimplemented in terms of `log4j-slf4j2-impl`.

ppkarwasz:
I am currently working on:

- `log4j-script`,
- `log4j-slf4j-impl`,
- `log4j-slf4j2-impl`,
- `log4j-to-jul`

so I marked them as done to prevent collisions.

ppkarwasz:
Related to this are PRs that remove additional dependencies from `log4j-core`:

- [ ] #2142 
- [ ] #2154

grobmeier:
completed

PR #2104, opened 2023-12-18
Move JCTools-based queues to a new `log4j-jctools` module

vy:
This PR addresses #2075 by means of following changes:
- Recycler factories in the API are converted to `ServiceProvider`s
- JCTools dependencies in `log4j-api` and `log4j-core` are removed in favor of a new module: `log4j-jctools`

### Why does `log4j-jctools` depend on `log4j-core`?

There are two places where JCTools is used as an optional dependency:
1. In `log4j-api` as a `Recycler[Factory]` (used almost everywhere) implementation
2. In `log4j-core` as a `BlockingQueueFactory` (used by `AsyncAppender` in `log4j-core`) implementation

There are two approaches I can think of to remove the JCTools optional dependencies:
1. Hoist `BlockingQueueFactory` from `log4j-core` to `log4j-api`
   - :smiley: `log4j-jctools` will only need to depend on `log4j-api`
   - :frowning_face: `log4j-api` will be introduced a component that only matters for some `log4j-core` internal
   - :frowning_face: `log4j-core` has access to plugins and hence, users were expected to provide custom `@Configurable @Plugin BlockingQueueFactory` implementations that they can refer to in `log4j2.xml`. Though moving `BlockingQueueFactory` to `log4j-api` will break this contract completely, since `log4j-api` has no notion of plugins.
2. Make `log4j-jctools` depend on `log4j-core` and provide both `Recycler[Factory]` and `BlockingQueueFactory` implementations
   - :frowning_face: Users who wants a JCTools-based recycler, but no `log4j-core` et al. (if there is such a use case at all) needs to manually exclude `log4j-core`

### Review kit

1. You can safely ignore files changed due to package relocations.
1. Start with changes in `o.a.l.log4j.internal.recycler` and `o.a.l.log4j.spi` packages of `log4j-api` module
1. Proceed to `log4j-core`
1. Now it is safe to check `log4j-jctools`


log4j-api/src/main/java/org/apache/logging/log4j/internal/recycler/ThreadLocalRecyclerFactoryProvider.java
```diff
@@ -0,0 +1,125 @@
...
+
+    @Nullable
+    @Override
+    public RecyclerFactory createForEnvironment(PropertyEnvironment environment) {
+        requireNonNull(environment, "environment");
+        final boolean threadLocalEnabled =
+                environment.getBooleanProperty(LoggingSystemProperty.THREAD_LOCALS_ENABLE, true);
+        if (!threadLocalEnabled) {
+            return null;
+        }
```
ppkarwasz, line 64: I would just drop (and document) the support for the `ThreadLocals.enable`.

Unless I am mistaken it was only used in the recycling process, setting `ThreadLocals.enable=false` never disabled **all** thread locals. For example `ThreadContext` still used them. The `ThreadLocals.enable` property seems misleading and deprecated, since `Recycler.factory` replaces it.

ppkarwasz (approved):
Nice job, looks good to me.

We might reevaluate the need for alternative recycler implementations in `log4j-api` in the future, but for now I agree that `log4j-core` is all we need.

I start to wonder how useful is it to fill `log4j-api` with partial implementations. For example a recycled `LogBuilder` is useful for `log4j-core`, but not so much for `log4j-to-slf4j`. If `log4j-to-slf4j` is bound to Logback, the Log4j API builder will be recycled, but the underlying SLF4J builder won't. We might as well remove `Recycler` from `AbstractLogger` and move it into `core.Logger`.

log4j-api/src/main/java/org/apache/logging/log4j/internal/recycler/ThreadLocalRecyclerFactoryProvider.java
vy: Created #2105 and linked to it in #2016.

vy:
> I start to wonder how useful is it to fill `log4j-api` with partial implementations. For example a recycled `LogBuilder` is useful for `log4j-core`, but not so much for `log4j-to-slf4j`. If `log4j-to-slf4j` is bound to Logback, the Log4j API builder will be recycled, but the underlying SLF4J builder won't. We might as well remove `Recycler` from `AbstractLogger` and move it into `core.Logger`.

Created #2106 and linked to it in #2016.

PR #2115, opened 2023-12-20
Fix NPE caused by MDCContextMap

### LINKED ISSUE #1426: MDCContextMap::getCopy can return null causing an NPE in CloseableThreadContext
## Description

`MDCContextMap::getCopy` delegates to `MDC::getCopyOfContextMap` which is documented as potentially returning `null`. This fact is respected within `MDCContextMap`, but `ThreadContextMap::getCopy` is documented as returning a non-null value. In `CloseableThreadContext.Instance`, this causes `NullPointerException`s in the `putAll` method when `ThreadContext::getContext` is invoked and later used.

## Configuration

**Version:** 2.20.0

**Operating system:** [OS and version]

**JDK:** [JDK distribution and version]

## Logs

```
java.lang.NullPointerException: null
        at org.apache.logging.log4j.CloseableThreadContext$Instance.putAll(CloseableThreadContext.java:170)
        at org.apache.logging.log4j.CloseableThreadContext.putAll(CloseableThreadContext.java:100)
```

## Reproduction

ppkarwasz:
Fixes an NPE caused by `MDCContextMap`: the [`MDC#getCopyOfContextMap`](https://www.slf4j.org/api/org/slf4j/MDC.html#getCopyOfContextMap--) method can return `null`, but [`ThreadContextMap#getCopy`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/ThreadContextMap.html) can not return `null`.

Closes #1426

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingFileManagerTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingFileManagerTest.java
@@ -16,7 +16,17 @@
  */
 package org.apache.logging.log4j.core.appender.rolling;
 
-import java.io.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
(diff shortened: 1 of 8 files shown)


ppkarwasz:
Closes #1426

PR #2120, opened 2023-12-21
Simplify configuration after BND 7.x upgrade

ppkarwasz:
BND 7.x supports multi-release JARs, so we can remove many manual overrides of module names.

We also **add** an override from `disruptor` to
`com.conversantmedia.disruptor`, since new (Java 11 based) versions of `com.conversantmetida:disruptor` have a JPMS descriptor.

The OSGi descriptor is identical to the one before this change.


log4j-core/pom.xml
```diff
@@ -66,15 +66,16 @@
...
-      com.fasterxml.jackson.dataformat.yaml;substitute="jackson-dataformat-yaml";transitive=false,
+      com.fasterxml.jackson.core;transitive=false,
+      com.fasterxml.jackson.databind;transitive=false,
+      com.fasterxml.jackson.dataformat.xml;transitive=false,
+      com.fasterxml.jackson.dataformat.yaml;transitive=false,
       java.naming;transitive=false,
       org.apache.commons.csv;transitive=false,
       org.fusesource.jansi;transitive=false,
+      <!-- Newer versions have a module descriptor -->
+      com.conversantmedia.disruptor;substitute="disruptor";transitive=false;static=true,
```
vy, line 77: Why `substitute` is not necessary for Jackson et al., whereas it is for Disruptor?

src/changelog/.2.x.x/simplify_configuration_after_bnd_7_update.xml
```diff
@@ -0,0 +1,25 @@
...
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="http://logging.apache.org/log4j/changelog"
+       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.3.xsd"
+       type="updated">
+  <description format="asciidoc">
+    Simplify BND configuration after upgrade from version 6.4.1 to 7.0.0.
```
vy, line 23: Please note that this is of type `changed`, not `updated`. I guess you play this trick to get a patch release out. Let's not trick anyone, but be transparent. (I don't have energy to argue about patch-vs-minor anymore.) Please use `updated` only for, as its name implies and we agreed on, dependency updates.

```suggestion
       type="changed">
  <description format="asciidoc">
    Simplify BND configuration after upgrade from version `6.4.1` to `7.0.0`
```

vy (changes requested):

log4j-core/pom.xml
ppkarwasz: Good question!

In `2.x` we are stuck with `com.conversant:disruptor:1.2.15`, which is the last version that runs on JRE 8. In the 1.2.16 **patch** release, Conversant:

 - bumped the required JRE version to 9,
 - added a JPMS module descriptor.

Therefore I assume that all the users that use Conversant Media Disruptor and JPMS, will use version 1.2.16+.

src/changelog/.2.x.x/simplify_configuration_after_bnd_7_update.xml
ppkarwasz: Technically it **is** a dependency update: it should have **no** effect on the user. The only reason I add it to the changelog is that these kind of things _might_ contain errors and I want to be plenty transparent about the "no-op" changes that occurred.

Maybe it should be a "fixed" change, since it actually fixes the usage of Conversant Media Disruptor.

log4j-core/pom.xml
vy: Got it. Jackson is still at version `2.16.0`, right? That is, no JPMS descriptor. Shouldn't we be keeping `substitute` there?

ppkarwasz: Jackson has a JPMS descriptor in `META-INF/versions/9`. BND 6.x could not find it, but BND 7.x can.

ppkarwasz:
Let's retarget this for `2.23.0`, since it is of "changed" type.

vy:
@ppkarwasz, thanks for adapting changelog types accordingly. You can cut either a patch or minor release.

Issue #2129, opened 2023-12-26
Regression in Log4j 2.22.0 on Android because of changes to getThreadContextClassLoader() in LoaderUtil


centic9:
## Description

When upgrading an Android Application which has a transitive dependency on Log4j from log4j-core 2.21.1 to 2.22.0, it fails with an exception because it seems the method AccessController.doPrivileged() with the required parameters is not available on Android, but calls to it are introduced in LoaderUtil now via commit https://github.com/apache/logging-log4j2/commit/80de81652e0d3eaf195402c2bdc28316a8c71dde

This prevents us from upgrading log4j when including any Java libraries which depend on log4j-core in an Android application.

Is it possible to not introduce these changes in the 2.x series? The commit-message sounds like this was sort of "let's backport some stuff from 3.x branches".

## Configuration

**Version:** 2.22.0

**Operating system:** Android application with min SDK Level 26

**JDK:** Android

## Logs

```
java.lang.NoSuchMethodError: No static method doPrivileged(Ljava/security/PrivilegedAction;Ljava/security/AccessControlContext;[Ljava/security/Permission;)Ljava/lang/Object; in class Ljava/security/AccessController; or its super classes (declaration of 'java.security.AccessController' appears in /system/framework/core-oj.jar)
at org.apache.logging.log4j.util.LoaderUtil.getThreadContextClassLoader(LoaderUtil.java:155)
at org.apache.logging.log4j.util.LoaderUtil.findUrlResources(LoaderUtil.java:473)
at org.apache.logging.log4j.util.LoaderUtil.findResources(LoaderUtil.java:462)
at org.apache.logging.log4j.util.PropertyFilePropertySource.loadPropertiesFile(PropertyFilePropertySource.java:45)
at org.apache.logging.log4j.util.PropertyFilePropertySource.<init>(PropertyFilePropertySource.java:37)
at org.apache.logging.log4j.util.PropertiesUtil.<init>(PropertiesUtil.java:84)
at org.apache.logging.log4j.util.PropertiesUtil.<init>(PropertiesUtil.java:80)
at org.apache.logging.log4j.status.StatusLogger.<clinit>(StatusLogger.java:76)
at org.apache.logging.log4j.status.StatusLogger.getLogger(StatusLogger.java:149)
at org.apache.logging.log4j.LogManager.<clinit>(LogManager.java:60)
at org.apache.logging.log4j.LogManager.getLogger(LogManager.java:610)
[...]

PR #2138, opened 2023-12-29
Fix use of AccessController in LoaderUtil

jvz:
This fixes issue #2129 where `AccessController::doPrivileged` is needlessly invoked when no `SecurityManager` is installed.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
@@ -50,36 +50,31 @@ public final class LoaderUtil {
     // wants to use PropertiesUtil, but then PropertiesUtil wants to use LoaderUtil.
     private static Boolean ignoreTCCL;
 
-    private static final RuntimePermission GET_CLASS_LOADER = new RuntimePermission("getClassLoader");
-    private static final boolean GET_CLASS_LOADER_DISABLED;
+    static final RuntimePermission GET_CLASS_LOADER = new RuntimePermission("getClassLoader");
+    static final LazyBoolean GET_CLASS_LOADER_DISABLED = new LazyBoolean(() -> {
+        if (System.getSecurityManager() == null) {
+            return false;
+        }
+        try {
+            AccessController.checkPermission(GET_CLASS_LOADER);
+            // seems like we'll be ok
+            return false;
+        } catch (final SecurityException ignored) {
+            try {
+                // let's see if we can obtain that permission
+                AccessController.doPrivileged((PrivilegedAction<Void>) () -> {
+                    AccessController.checkPermission(GET_CLASS_LOADER);
+                    return null;
+                });
+                return false;
+            } catch (final SecurityException ignore) {
+                // no chance
+                return true;
+            }
+        }
+    });
 
     private static final PrivilegedAction<ClassLoader> TCCL_GETTER = new ThreadContextClassLoaderGetter();
-
-    static {
-        if (System.getSecurityManager() != null) {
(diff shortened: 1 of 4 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
```diff
@@ -143,22 +140,29 @@ private static boolean isChild(final ClassLoader loader1, final ClassLoader load
...
 
+    private static <T> T runActionInvolvingGetClassLoaderPermission(final PrivilegedAction<T> action) {
+        return System.getSecurityManager() != null
+                ? runPrivilegedActionWithGetClassLoaderPermission(action)
+                : action.run();
+    }
+
+    private static <T> T runPrivilegedActionWithGetClassLoaderPermission(final PrivilegedAction<T> action) {
+        return AccessController.doPrivileged(action, null, GET_CLASS_LOADER);
+    }
```
ppkarwasz, line 164:
```suggestion
    private static <T> T runPrivilegedActionWithGetClassLoaderPermission(final PrivilegedAction<T> action) {
        return AccessController.doPrivileged(action);
    }
```
Now that we know that Android does not have the `AccessController.doPrivileged(PrivilegedAction<T>, AccessControlContext, Permission...)` method (cf. [Android javadoc](https://developer.android.com/reference/java/security/AccessController)), why don't we call `AccessController.doPrivileged(PrivilegedAction<T>)` method?

Sure, you took care to protect every call site with a check to `System.getSecurityManager() != null`, but I don't understand what is the advantage of calling the 3-param method? Even if the `action` lambda only requires the "getClassLoader" permission, I don't see the problem of giving to it **all** the permissions of the caller.

By eliminating the 3-param method from our codebase, we prevent some future errors.

ppkarwasz (changes requested):
@jvz,

Thank You for the fast fix!

Personally I think we should use as little `SecurityManager` code as possible, see my remark below.

vy:
@jvz, would you also mind adding tests too, please?

log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
jvz: Updated.

log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
```diff
@@ -50,39 +50,31 @@ public final class LoaderUtil {
...
 
-    private static final RuntimePermission GET_CLASS_LOADER = new RuntimePermission("getClassLoader");
-    private static final boolean GET_CLASS_LOADER_DISABLED;
-
-    private static final PrivilegedAction<ClassLoader> TCCL_GETTER = new ThreadContextClassLoaderGetter();
-
-    static {
-        if (System.getSecurityManager() != null) {
-            boolean getClassLoaderDisabled;
+    static final RuntimePermission GET_CLASS_LOADER = new RuntimePermission("getClassLoader");
```
ppkarwasz, line 53: Unless I am mistaken, this is no longer necessary.

ppkarwasz (approved):
LGTM, although it would be nice to have an Android test.

log4j-api/src/main/java/org/apache/logging/log4j/util/LoaderUtil.java
jvz: It's used for checking if we have the ability to use the permission in full-permission mode to see if we should just fall back to the class loader of this class.

PR #2142, opened 2023-12-30
Split off YAML configuration into its own module

ppkarwasz:
This PR splits `YamlConfiguration` into its own module and removes Jackson from `log4j-core`'s optional dependencies.

## Changes

- `YamlConfigurationFactory` is moved to a new `log4j-config-yaml` module together with YAML specific tests,
- Converts the configuration of 3 `log4j-core-test` from YAML to XML, since the tests are not YAML specific,
- Removes `log4j2-2134.yaml`, which was referenced in `JiraLog4j2_2134Test`, but in practice the `DefaultConfiguration` was used instead.


ppkarwasz:
I rebase the PR on top of #2155, so that we can immediately remove Jackson.

log4j-config-yaml/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+
+    <dependency>
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
```
vy, line 39: `jackson-dataformat-yaml` already depends on `jackson-databind`, hence this is not needed.

src/changelog/.3.x.x/create_config_yaml.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="http://logging.apache.org/log4j/changelog"
+       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.2.xsd"
```
vy, line 4: Now you know what I mean by people blindly copy paste changelog entries. :stuck_out_tongue_winking_eye: 

```suggestion
       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.3.xsd"
```

vy (approved):

log4j-config-yaml/pom.xml
ppkarwasz: I often follow `./mvnw dependency:analyze` suggestions, which uses the public types in our code. Since we use types from `core`, `databind` and `databind-yaml`, I would put all three deps, but I am not strict about it.

ppkarwasz:
This has actually been successfully merged. I just deleted the branch too soon.

PR #2150, opened 2024-01-02
Port `log4j-to-jul` changes from `2.x`

ppkarwasz:
This PR adds to the `log4j-to-jul` module the missing changes from the 2.x branch.

Regarding the `log4j-to-jul` module, the only differences between the 2.x and 3.x version is the absence of an OSGi activator in the 3.x version: we rely on Service Loader Mediator in 3.x.


vy (approved):
LGTM. :+1:

PR #2155, opened 2024-01-03
Remove Jackson dependency from `MutableThreadContextMapFilter`

ppkarwasz:
This PR rewrites `MutableThreadContextMapFilter` to use the embedded `JsonReader` instead of Jackson. Except `YamlConfiguration`, this is the only part of `log4j-core` that requires Jackson.

Alternatively, we could move the filter to a separate artifact.


vy (approved):
@ppkarwasz, great job! LGTM. :+1:

ppkarwasz:
Inadvertedly I merged it, when merging the YAML config: 8803ecbd1d5d556c0e94cb2cdef561c5e7ea4606

Feel free to revert the commit if you disagree with the change.

PR #2156, opened 2024-01-03
Port `log4j-slf4j-impl` changes from `2.x`

ppkarwasz:
This PR adds to the `log4j-slf4j-impl` module the missing changes from the `2.x` branch.

After this change `git diff 2.x main -- log4j-slf4j-impl` should only contain changes related to serialization.


log4j-slf4j-impl/src/main/java/org/apache/logging/slf4j/package-info.java
```diff
@@ -21,7 +21,7 @@
  */
 @Export
 @Header(name = Constants.BUNDLE_ACTIVATIONPOLICY, value = Constants.ACTIVATION_LAZY)
-@Version("2.21.0")
+@Version("2.23.0")
```
PascalSchumacher, line 24: I guess this should have a different value on `3.x` branch?

ppkarwasz: @PascalSchumacher,

thanks for the remark.

We don't bump package versions automatically (if there are no minor/major changes we leave them the same, but in this case BND remarks that we have hidden (removed `public`) from two classes:

```
[ERROR] * org.apache.logging.slf4j                           PACKAGE    MAJOR      2.23.0     2.21.0     3.0.0      -
[ERROR]   MAJOR                PACKAGE    org.apache.logging.slf4j
...
[ERROR]     MAJOR              CLASS      org.apache.logging.slf4j.Log4jLogger
[ERROR]       REMOVED          IMPLEMENTS java.io.Serializable
[ERROR]     REMOVED            CLASS      org.apache.logging.slf4j.Log4jMarker
```

vy (approved):
LGTM. :+1:

Issue #1956, opened 2023-11-08
Extract configuration metadata from source code


ppkarwasz:
Following the completion of #1955, as a second step towards #1954, we should:

- [ ] develop a tool that extracts the configuration options of all Log4j plugins, together with Javadoc associated to it and convert it into the chosen metadata format,
- [ ] discuss a mechanism to bundle the plugins metadata with the Maven artifacts (e.g. we might include it in `META-INF`).

PR #2172, opened 2024-01-08
Move validation constraints to `log4j-plugins-api`

ppkarwasz:
**Remark**: the target of this PR is a feature branch.

Plugin annotations are employed by `log4j-docgen` to generate documentation (cf. #1956)

Since the tool is compiled using Java 8 bytecode, we need to split `log4j-plugins` into:

 * an API part that contains mostly annotations and interfaces,
 * an implementation part that can use all the improvements of Java 17.

In this PR:

 * we move the constraint annotation into a new artifact,
 * we make the `value` property of the `@Constraint` annotation optional,
 * we introduce a `ConstraintValidatorFactory` that produces the appropriate `ConstraintValidator` for each annotation. This way we don't need to reference `ConstraintValidator` implementations in the constraint annotations.
 * we add the `2.x` constraint annotations that were lost in `main`.

This PR leaves the `o.a.l.l.plugins` package split between two artifacts, so it can not be merged to `main`.


log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/di/DefaultInstanceFactory.java
```diff
@@ -99,6 +104,9 @@ private DefaultInstanceFactory(
         this.bindings.put(Key.forClass(InstanceFactory.class), () -> this);
         this.bindings.put(Key.forClass(PropertyEnvironment.class), PropertiesUtil::getProperties);
         this.bindings.put(Key.forClass(ClassLoader.class), LoaderUtil::getClassLoader);
+        this.bindings.put(
+                Key.forClass(ConstraintValidatorFactory.class),
+                () -> new DefaultConstraintValidatorFactory(Integer::valueOf));
```
jvz, line 109: I suppose this works for now, though we may be able to move most (or all) of these default bindings to a `ConfigurableInstanceFactoryPostProcessor` as I noted in a todo comment.

log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/di/DefaultInstanceFactory.java
```diff
@@ -392,6 +400,29 @@ protected <T> void injectField(final Field field, final Object instance) {
         }
     }
 
+    public <A extends Annotation> ConstraintValidator<A> createValidator(
```
jvz, line 403: Nit: missing `@Override`

jvz (approved):
LGTM. I see some type erasure issues to clean up later, but no big deal.

ppkarwasz:
From my perspective it is more interesting to check if the API part (interfaces and annotations) do not depend on their implementations. And as it turns out they do.

A use can for an independent `log4j-plugins-api` would be a `RecyclerFactory` that only depends on `log4j-api` and `log4j-plugins-api` and can still provide a component usable in `log4j-core`.

**Remark**: the Java 17 bytecode currently used by `log4j-plugins` is not a blocker for `log4j-docgen`. The latter can still be used as an annotation processor to compile Java 8 bytecode if a JDK 17 compiler is used. However many projects still compile Java 8 bytecode using ... JDK 8. ;-)

vy:
>  the Java 17 bytecode currently used by `log4j-plugins` is not a blocker for `log4j-docgen`

Then this PR simply complicates `log4j-plugins` more than it already is without any benefit, hence -1 from me.

PR #2174, opened 2024-01-09
Remove legacy `2.x` properties configuration

ppkarwasz:
The old properties configuration factory from `2.x` started with simple XML to properties mapping rules, but ended up with a lot of exceptions to those rules:

 * some main components must use a specific property name, which is different than the plugin name. E.g. the list of appenders is called `appender` and does not require an `appender.type = Appenders` property. The plugin is called `Appenders`.
 * some subcomponents are treated differently, e.g. a subcomponent of an appender with key "layout" or "filter" uses a different visitor the all the other subcomponents,
 * the `Properties` sub-tree uses different mapping rules than the rest of the configuration,
 * there are stil some bugs in this configuration factory. E.g. the "filter" property of an appender accepts an array of filter, but only the first one is used,
 * there are some issues with the order of elements: the order in which filters are applied is random.

For these reasons I would remove this configuration factory and see how users like `log4j-config-properties`.


vy (approved):
I support this step. Let me be more elaborate: I disagree with two competing solutions: `log4j-config-properties-legacy` and `log4j-config-properties`. This is confusing for users and a burden on the code base.

PR #2188, opened 2024-01-12
Port `log4j-1.2-api` changes from `2.x`

ppkarwasz:
This PR ports several changes to the `log4j-1.2-api` module that are present in `2.x`, but were never ported to `main`.


vy (approved):
It is difficult to review this PR. Given you haven't removed any tests, on the contrary, added more, and build still passes... What can go wrong? Right? :sweat_smile:

PR #2198, opened 2024-01-16
Remove `log4j-layout-jackson` and `log4j-layout-jackson-xml`

vy:
Removes `XmlLayout` (i.e., `log4j-layout-jackson-xml`) and `log4j-layout-jackson`, since it is not used anymore.

### Review kit

Ignore the changes in `log4j-layout-jackson` and `log4j-layout-jackson-xml` folders, since these folders are simply deleted. Focus on the rest instead.


ppkarwasz (approved):
Since @garydgregory approved, it is fine with me.

Issue #2163, opened 2024-01-04
[MS8] Branch `main` overhaul #3


ppkarwasz:
Before we release 3.0.0 we need to:

- port changes from `2.x` to these modules:
    - [ ] `log4j-1.2-api`
    - [ ] `log4j-api`
    - [ ] `log4j-api-test`
    - [ ] `log4j-appserver`
    - [ ] `log4j-core`
    - [ ] `log4j-core-its`
    - [ ] `log4j-core-test`
    - [ ] `log4j-csv`
    - [ ] `log4j-docker`
    - [ ] `log4j-flume-ng`
    - [ ] `log4j-gctests`
    - [ ] `log4j-iostreams`
    - [ ] `log4j-jcl`
    - [ ] `log4j-jdbc`
    - [ ] `log4j-jdbc-dbcp2`
    - [ ] `log4j-jndi`
    - [ ] `log4j-jndi-test`
    - [ ] `log4j-layout-jackson`
    - [ ] `log4j-layout-jackson-xml`
    - [ ] `log4j-layout-template-json`
    - [ ] `log4j-layout-template-json-test`
    - [ ] `log4j-osgi-test`
    - [ ] `log4j-parent`
    - [ ] `log4j-perf-test`

ppkarwasz:
While working on #2227 I checked that no regressions are present in the `o.a.l.l.core.async` package in the `main` branch.

grobmeier:
Checked and considered complete.

Issue #1977, opened 2023-11-17
Classes should no longer be specified via system properties


jvz:
Over time, we collected many places in the codebase that accept an override by specifying a class name through a system property. These typically worked by guessing at an appropriate `ClassLoader` to use to load the class and then using reflection to instantiate it. This pattern is problematic in module systems like the Java module system and OSGi due to stronger encapsulation guarantees. As we've already implemented for plugins, anything else remaining that customizes class selection based on system properties should be updated to use `java.util.ServiceLoader` (when customizing in log4j-api) or specified in bundle classes installed via a service loader (which is also where some of the old system properties customizations are currently supported and can be simplified).

In general, the goal of this is to eliminate the use of `LoaderUtil::load` and any other implicit `ClassLoader` usage.

PR #2203, opened 2024-01-16
Restore compatibility of Log4j API

ppkarwasz:
This restores backward compatibility of `log4j-api` from a user's perspective and adds comments to classes that should not have breaking changes.

I have checked all the types in `org.apache.logging.log4j` and those reachable through inheritance or signature of public methods in other packages:

```
org.apache.logging.log4j.message.EntryMessage
org.apache.logging.log4j.message.ExitMessage
org.apache.logging.log4j.message.FlowMessage
org.apache.logging.log4j.message.FlowMessageFactory
org.apache.logging.log4j.message.Message
org.apache.logging.log4j.message.MessageFactory
org.apache.logging.log4j.spi.LoggerContext
org.apache.logging.log4j.spi.ExtendedLogger
org.apache.logging.log4j.spi.LoggerContextFactory
org.apache.logging.log4j.spi.LoggerRegistry
org.apache.logging.log4j.spi.StandardLevel
org.apache.logging.log4j.spi.ReadOnlyThreadContextMap
org.apache.logging.log4j.util.BiConsumer
org.apache.logging.log4j.util.ReadOnlyStringMap
org.apache.logging.log4j.util.StringMap
org.apache.logging.log4j.util.StringBuilderFormattable
org.apache.logging.log4j.util.Supplier
org.apache.logging.log4j.util.TriConsumer
```

The only breaking changes I found are `Logger#entry/exit` <s>and in `MarkerFactory`</s>. I believe we should restore them and have a 100% binary compatibility with `2.x`, rather then breaking BC for almost nothing.

This PR also adds `forRemoval` to all the classes that where deprecated in `2.x`, but not removed in `3.x`.

Part of #2163


rgoers (approved):

log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
```diff
@@ -40,7 +40,13 @@
 public class LogManager {
 
     /**
-     * Constant holding the name of the root logger (the empty string).
+     * Kept for binary compatibility.
+     */
+    @Deprecated(forRemoval = true)
+    public static final String FACTORY_PROPERTY_NAME = "log4j2.loggerContextFactory";
```
jvz, line 46: I suppose we can keep these strings, but by 3.0.0, they'll be functionally useless due to https://github.com/apache/logging-log4j2/issues/1977

jvz (approved):
Generally seems ok. Might be useful to define an annotation for the comments about public APIs (something similar to `@InternalApi`, perhaps `@PublicApi`). Would be nice to include the `since` attribute of the deprecation annotations, too.

log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
ppkarwasz: There is actually a compatibility mechanism developed by Ralph (cf. `log4j2.propertyMapping.json`) that allows `log4j2.loggerContextFactory` to still work.

I had doubts about reintroducing this: it is a **compile-time** constant, so it will be inlined in all binary artifacts. As far as source compatibility is concerned, I think we could remove it. What do you think?

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
```diff
@@ -122,6 +122,8 @@ public interface Logger {
      * @param messageSupplier A function, which when called, produces the desired log message.
      * @since 2.4
      */
+    @Deprecated(forRemoval = true)
```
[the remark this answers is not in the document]
ppkarwasz: I use `git blame` and annotated all deprecations with the minor version they were introduced.

ppkarwasz:
I have removed the two deprecated `MarkerFactory#getMarker` methods, since they were deprecated in April 2014, i.e. **before** the release of `2.0`. IMHO we should treat them as if they never existed in the 2.x branch.

log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java
jvz: What I mean is that specifying custom classes to use for things will likely use some sort of code-based way of doing so rather than a system property.

PR #2209, opened 2024-01-17
Remove ineffective tests

vy:
This PR either fixes or removes ineffective tests.


rgoers:
What criteria was used to determine that the tests were "ineffective"?  Anything with "perf" in the name was just deemed ineffective?

vy:
> What criteria was used to determine that the tests were "ineffective"? Anything with "perf" in the name was just deemed ineffective?

@rgoers, removed tests either were disabled long ago (causing compilation time waste) or contained no assertions (causing compilation + testing time waste). I tried saving whatever I can and removed the rest of such _ineffective_ tests.

rgoers:
@vy There were a couple of "tests" you removed that did not have any assertions that I used for occaisional sanity checks. Namely FilterPerformanceComparison, SimplePerfTest, and ThreadedPerfTest.  I never found a way to assert what I wanted to check - namely that someone hadn't inserted code that seriously impacted the performance of what those are doing- SimplePerfTest primarily validates that the cost of logging is close to zero when disabled.  FilterPerformanceComparison ensures we haven't added something that causes filtering to slow down - comparing it to Logback is a decent way to do that.  One of the main issues is the performance varies from machine to machine so it is very tough to make assertions. Instead, I just eyeball the results based on what I expect on my machine.

vy:
@rgoers, thanks for sharing the context, now I understand. These use cases weren't apparent to me from the code. I ~am~ was actively in the search of finding the right way to establish a continuous (and reproducible!) performance test bed. The checks you mentioned indeed serve an important purpose an should ideally be a part of the build. I will check if we can flesh out an attack plan. I will first consult to INFRA if we can get our hands on a machine where we can run performance tests without external noise and, ideally, manage it in GitHub Actions so that we can make it a part of our CI pipeline.

PR #2215, opened 2024-01-18
Adds lighter v3 `Logger` interface (PoC)

ppkarwasz:
This is a proof-of-concept to provide a lighter `Logger` interface without breaking compatibility with libraries compiled against v2 `Logger`.

In the `v3.Logger` interface:

 - we replace `o.a.l.l.util.Supplier` with `java.util.function.Supplier`,
 - we provide the methods available in SLF4J,
 - we replace methods that log `Object` with three methods that accept a `CharSequence` (as better "lazy" string), `Message` or `Throwable`,
 - we unroll the `(String, Supplier<?>..)` methods up to two arguments,
 - we replace `logMessage` with `log` and remove the FQCN argument (normal users don't need it).

## How to evaluate this draft?

Look at the `o.a.l.l.v3` package and the modifications to `o.a.l.l.Logger`. A user of `v3.LogManager` will have a lighter interface and his code will use `java.util.function.Supplier`. A user of `LogManager` will have access to the old v2 methods and his code will use `o.a.l.l.util.Supplier` (without the user noticing it).

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Category.java
@@ -334,7 +334,7 @@ public class Category implements AppenderAttachable {
         final org.apache.logging.log4j.Level lvl = level.getVersion2Level();
         final Message msg = createMessage(message);
         if (logger instanceof ExtendedLogger) {
-            ((ExtendedLogger) logger).logMessage(fqcn, lvl, null, msg, t);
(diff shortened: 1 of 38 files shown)


vy:
@ppkarwasz, given [we decided to use `log4j-api-2.x` in `main`](https://lists.apache.org/thread/gvk0to32c13xsfz0ytczfs506svx3hxm), is this PR still relevant for me to review?

ppkarwasz:
No, it might be relevant in the future, but right now I am closing this PR.

Issue #2100, opened 2023-12-15
Document Log4j security model


ppkarwasz:
We should document our security model on our [security page](https://logging.apache.org/log4j/2.x/security.html).
An example of such a description is the [Commons Security Model](https://commons.apache.org/security.html)

ppkarwasz:
@rm5248, @swebb2066, @fluffynuts and @grobmeier: maybe this could be a document shared by all the projects?
The security risks seems pretty much similar to me.

ppkarwasz:
From the top of my head, our security model should relieve us from any responsibility for:

- CRLF injections if the user uses a line-oriented tool to parse logs,
- other kinds of data corruption if the user uses `PatternLayout` to generate JSON, XML or other structured formats,

Attacks that require access to a configuration file should not be considered vulnerabilities.

grobmeier:
I agree, one security model should match for all logging projects - maybe some extensions are necessary for Flume and Chainsaw

ppkarwasz:
This [Commons discussion](https://lists.apache.org/thread/3z4p76oq8g1npvdxpv0zs0j7dqv45c1q) might be interesting.

When the document is ready, we might ask @raboof to proofread it.

jvz:
Some initial thoughts about this:

* Configuration files or configuration code have the ability to execute arbitrary code by using custom plugins; as such, configurations must be controlled by the programmer, not arbitrary users.
* Log message parameters are not evaluated for any sort of template strings like `{}`; template strings are limited to the non-parameter message argument. As such, log message parameters are not treated specially (such as with a prepared statement in SQL). When structure-preserving encoding is desired for things besides the log message as a whole, the `Message` API should be used for logging structured data rather than strings.
* Unstructured layouts shouldn't be used if any user-controlled data can be logged through them; this largely includes logging exceptions if said exceptions can contain user input (which is a common use case for exceptions).
* Code signing and verification is only performed at the artifact level (i.e., our jars, zips, and tarballs); we do not rely on jar signing to validate plugins or extensions (unlike, say, the Java cryptography APIs which require signed jars typically).
* Given the equivalence to the halting problem, we cannot generically support any form of log masking for sensitive data. While external plugins may support limited forms of masking (e.g., for credit card numbers or similar easily identifiable byte patterns), it is up to the programmer to explicitly mark sensitive data for masking.
[...]

PR #2219, opened 2024-01-19
Add security threat model to docs

jvz:
This updates the security page with notes about the threat model used for Log4j. Some related documentation has been fixed during review of this threat model.

Related to and adapted from commentary in https://github.com/apache/logging-log4j2/issues/2100.

### DIFF
--- a/src/site/asciidoc/security.adoc
+++ b/src/site/asciidoc/security.adoc
@@ -50,7 +50,7 @@ The threat model that Log4j uses considers configuration files as safe input con
 == Threat Model
 
 Log4j is a low level library where configuration inputs and the classpath are expected to be controlled by the programmer.
-Configurations have the ability to execute arbitrary code through custom plugins.
+Configurations have the ability to execute arbitrary code.
 While specific Log4j plugins (such as a JNDI lookup) may use constraint validators or conditionals to require additional settings to opt in to functionality, this is not universally required by custom plugins.
 Specific security considerations involved in our threat model are detailed below.
 
--- a/src/site/asciidoc/security.adoc
+++ b/src/site/asciidoc/security.adoc
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
 The threat model that Log4j uses considers configuration files as safe input controlled by the programmer; **potential vulnerabilities that require the ability to modify a configuration are not considered vulnerabilities** as the required access to do so implies the attacker can execute arbitrary code.
 ====
 
+[#model]
+== Threat Model
+
+Log4j is a low level library where configuration inputs and the classpath are expected to be controlled by the programmer.
+Configurations have the ability to execute arbitrary code through custom plugins.
+While specific Log4j plugins (such as a JNDI lookup) may use constraint validators or conditionals to require additional settings to opt in to functionality, this is not universally required by custom plugins.
+Specific security considerations involved in our threat model are detailed below.
+
+=== Parameterized Logging
+
+When using a log message containing template parameters like `{}`, only the format string is evaluated for parameters to be substituted.
+The message parameters themselves are not evaluated for parameters; they are only included in the format string corresponding to their template position.
+The conversion of message parameters into a string is done on-demand depending on the layout being used.
+When structure-preserving transformations of log message data are required, the `Message` API should be used for logging structured data.
+
+=== Unstructured Logging
+
+When using an unstructured layout such as `PatternLayout`, no guarantees can be made about the output format.
+This layout is mainly useful for development purposes and should not be relied on in production applications.
+For example, if a log message contains new lines, these are not escaped or encoded specially unless the configured pattern uses the `%encode{pattern}{CRLF}` wrapper pattern converter (which will encode a carriage return as the string `\r` and a line feed as the string `\n`) or some other `%encode` option.
+Similarly, other encoding options are available for other formats, but pattern layouts cannot make assumptions about the entire output.
+As such, when using unstructured layouts, no user-controlled input should be included in logs.
+It is strongly recommended that a structured layout is used instead for these situations.
+Note that `StrLookup` plugins (those referenced by `${...}` templates in configuration files) that contain user-provided input should not be referenced by a pattern layout.
+
+=== Structured Logging
+
+When using a structured layout (most layouts besides pattern layout), log messages are encoded according to various output formats.
+These safely encode the various fields included in a log message.
+For example, the JSON template layout can be configured to output log messages in various JSON structures where all log data is properly encoded into safely parseable JSON.
+This is the recommended mode of operation for use with log parsing and log collection tools that rely on log files or arbitrary output streams.
+
+=== Code Signing
+
+Log4j artifacts are all signed using PGP using a key from the Logging Services PMC https://downloads.apache.org/logging/KEYS[KEYS file].
+Information on how to verify releases signed with PGP is https://httpd.apache.org/dev/verification.html[documented here].
+Individual jar files are not signed using `jarsigner` (only PGP), and the Log4j plugin system does not rely on signed jars for validation (unlike the Java cryptography APIs for example).
+Thus, PGP signatures should be validated in your build process.
+
+=== Java Security Manager
+
+Log4j no longer supports running in or using a custom `SecurityManager`.
+This Java feature has been deprecated for removal in Java 21.
+Previous versions of Log4j 2.x include partial support for running with a security manager.
+
+=== Log Masking
+
+Log4j, like any other generic logging library, cannot generically support log masking of sensitive data.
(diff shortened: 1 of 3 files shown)


src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+Configurations have the ability to execute arbitrary code through custom plugins.
+While specific Log4j plugins (such as a JNDI lookup) may use constraint validators or conditionals to require additional settings to opt in to functionality, this is not universally required by custom plugins.
+Specific security considerations involved in our threat model are detailed below.
+
+=== Parameterized Logging
+
+When using a log message containing template parameters like `{}`, only the format string is evaluated for parameters to be substituted.
+The message parameters themselves are not evaluated for parameters; they are only included in the format string corresponding to their template position.
+The conversion of message parameters into a string is done on-demand depending on the layout being used.
+When structure-preserving transformations of log message data are required, the `Message` API should be used for logging structured data.
```
ppkarwasz, line 62: I would advise users to only use compile-time constants as format strings and under no circumstance build format strings from user data.

ppkarwasz (approved):
LGTM

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
 The threat model that Log4j uses considers configuration files as safe input controlled by the programmer; **potential vulnerabilities that require the ability to modify a configuration are not considered vulnerabilities** as the required access to do so implies the attacker can execute arbitrary code.
 ====
 
+[#model]
+== Threat Model
+
+Log4j is a low level library where configuration inputs and the classpath are expected to be controlled by the programmer.
+Configurations have the ability to execute arbitrary code through custom plugins.
```
vy, line 53: Arbitrary code can very well be executed in a mere configuration, say, using `ScriptFilter`; using plugins is not a necessity.
```suggestion
Configurations have the ability to execute arbitrary code.
```

src/site/asciidoc/security.adoc
vy, line 62: Shouldn't we rather have said the following instead?
```suggestion
When structure-preserving transformations of log message data are required, layouts supporting this functionality (e.g., `JsonTemplateLayout`) should be used.
```

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+When structure-preserving transformations of log message data are required, the `Message` API should be used for logging structured data.
+
+=== Unstructured Logging
+
+When using an unstructured layout such as `PatternLayout`, no guarantees can be made about the output format.
+This layout is mainly useful for development purposes and should not be relied on in production applications.
+For example, if a log message contains new lines, these are not escaped or encoded specially unless the configured pattern uses the `%encode{pattern}{CRLF}` wrapper pattern converter (which will encode a carriage return as the string `\r` and a line feed as the string `\n`) or some other `%encode` option.
+Similarly, other encoding options are available for other formats, but pattern layouts cannot make assumptions about the entire output.
+As such, when using unstructured layouts, no user-controlled input should be included in logs.
+It is strongly recommended that a structured layout is used instead for these situations.
```
vy, line 71:
```suggestion
It is strongly recommended that a structured layout (e.g., `JsonTemplateLayout`) is used instead for these situations.
```

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+Note that `StrLookup` plugins (those referenced by `${...}` templates in configuration files) that contain user-provided input should not be referenced by a pattern layout.
```
vy, line 72: JTL also uses lookups too. Hence, they should not be used anywhere anyway.

```suggestion
Note that `StrLookup` plugins (those referenced by `${...}` templates in configuration files) that contain user-provided input should not be referenced by layouts.
```

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+
+=== Structured Logging
+
+When using a structured layout (most layouts besides pattern layout), log messages are encoded according to various output formats.
+These safely encode the various fields included in a log message.
+For example, the JSON template layout can be configured to output log messages in various JSON structures where all log data is properly encoded into safely parseable JSON.
```
vy, line 78: To comply with the naming convention used in previously shared layout names...
```suggestion
For example, the `JsonTemplateLayout` can be configured to output log messages in various JSON structures where all log data is properly encoded into safely parseable JSON.
```

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+This is the recommended mode of operation for use with log parsing and log collection tools that rely on log files or arbitrary output streams.
+
+=== Code Signing
+
+Log4j artifacts are all signed using PGP using a key from the Logging Services PMC https://downloads.apache.org/logging/KEYS[KEYS file].
+Information on how to verify releases signed with PGP is https://httpd.apache.org/dev/verification.html[documented here].
+Individual jar files are not signed using `jarsigner` (only PGP), and the Log4j plugin system does not rely on signed jars for validation (unlike the Java cryptography APIs for example).
```
vy, line 85: We can do without mentioning every single option we do not do.
```suggestion
```

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+When configurations are provided through JNDI, these should only use the `java` scheme for sharing configurations in a JavaEE or JakartaEE application service.
+JNDI-sourced configurations should not use other JNDI providers such as LDAP, DNS, or RMI, as all these providers are difficult to properly secure.
+
+=== Cryptographic Logs
+
+When using custom plugins to calculate message authentication codes, signatures, or encrypted logs, proper key management must be considered.
+Cryptography is a tricky subject, and key management lies at the heart of most design problems to solve when using it.
+If a custom plugin relies on a pre-shared key or password of some sort, it is likely broken beyond repair if being used for cryptographic purposes.
+Proper key management typically uses public key infrastructure (PKI) such as X.509 certificates, though this is not the only option.
+It is beyond the scope of Log4j to implement key management as this would depend on other infrastructure in use for validating or decrypting logs combined with the deployment and runtime environment of the logging application.
```
vy, line 125: Cryptographic logs are neither a feature we slightly offer, nor something we even briefly mention. This is not a place we educate users on cryptographic logs either. Hence... I think we can do without this section.

vy (changes requested):
@jvz, left some minor remarks. In overall, it LGTM. Thanks for your kind effort! :pray:

It also is not slipped my attention that you adhered to one-sentence-per-line convention. :heart_eyes:

src/site/asciidoc/security.adoc
jvz: It's both. Since the APIs for specifying non-string values into the thread context is kind of awkward, and it can be wasteful to store things there which are only relevant for a single log message, it's useful to use something like a `MapMessage`, and yes, you need to combine that with something like JTL to preserve the structure of the message.

src/site/asciidoc/security.adoc
jvz: True, it's not a current offering, but:

* The Flume plugin has some sort of half-assed support for encryption that violates these guidelines.
* We have a few feature requests related to this in Jira (mostly around encrypting logs, but there was another issue for signed logs).

I did research about this a few years ago, but my original implementation idea was fairly complex. Anyways, this section could perhaps be deferred until we have such a feature.

src/site/asciidoc/security.adoc
```diff
@@ -46,6 +46,89 @@ If you have encountered an unlisted security vulnerability or other unexpected b
...
+When using a log message containing template parameters like `{}`, only the format string is evaluated for parameters to be substituted.
+The message parameters themselves are not evaluated for parameters; they are only included in the format string corresponding to their template position.
+The conversion of message parameters into a string is done on-demand depending on the layout being used.
+When structure-preserving transformations of log message data are required, the `Message` API should be used for logging structured data.
+
+=== Unstructured Logging
+
+When using an unstructured layout such as `PatternLayout`, no guarantees can be made about the output format.
+This layout is mainly useful for development purposes and should not be relied on in production applications.
+For example, if a log message contains new lines, these are not escaped or encoded specially unless the configured pattern uses the `%encode{pattern}{CRLF}` wrapper pattern converter (which will encode a carriage return as the string `\r` and a line feed as the string `\n`) or some other `%encode` option.
```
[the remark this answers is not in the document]
jvz: Added

vy (approved):

ppkarwasz:
@raboof,

Could you take a look at this? It looks fine from our perspective.

PR #2227, opened 2024-01-22
Simplify async logger in `main`

ppkarwasz:
Due to the introduction of `Recycler` in `main` and related changes, many async loggers can be simplified:

 * no thread name caching is required,
 * in many methods we can use `ReusableLogEvent` instead of concrete classes,
 * many casts from `ReadOnlyStringMap` to `StringMap` are not necessary.


log4j-core-test/src/test/java/org/apache/logging/log4j/core/async/QueueFullAbstractTest.java
```diff
@@ -302,9 +302,11 @@ protected static void assertAsyncLoggerConfig(final LoggerContext ctx, final int
         final Configuration config = ctx.getConfiguration();
         assertThat(config).isNotNull();
         assertThat(config.getRootLogger()).isInstanceOf(AsyncLoggerConfig.class);
-        final AsyncLoggerConfigDisruptor disruptor = (AsyncLoggerConfigDisruptor) config.getAsyncLoggerConfigDelegate();
-        final Field sizeField = field(AsyncLoggerConfigDisruptor.class, "ringBufferSize");
-        assertThat(sizeField.get(disruptor)).isEqualTo(expectedBufferSize);
+        final AsyncLoggerConfigDisruptor disruptorWrapper =
+                (AsyncLoggerConfigDisruptor) config.getAsyncLoggerConfigDelegate();
+        final Disruptor<?> disruptor = (Disruptor<?>)
+                field(AsyncLoggerConfigDisruptor.class, "disruptor").get(disruptorWrapper);
```
vy, line 308: Just a reminder that `AsyncLoggerConfigDisruptor` could have had a package-local `getDisruptor` to ease testing, just like `getWaitStrategy()`. I am fine with the currently implemented reflection approach too.

We have a mixture of both all over the code base. [Excluding the reflection usage when package-local has accessibility problems due to tests being located in a different package,] I wish we would have had a convention for this.

vy (approved):
LGTM, can be merged.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/async/QueueFullAbstractTest.java
ppkarwasz: I corrected this part of code in #2228.

PR #2228, opened 2024-01-22
Remove JMX support

ppkarwasz:
As partial solution to #1344, this PR drop JMX support and replaces the usage of `RingBufferAdmin` in tests with a direct access to the `RingBuffer`.

## Motivation

JMX is an old monitoring technology, which less and less used today. Given its inherent insecurity, we should evaluate if JMX support should even be part of Log4j Core.

Our JMX support code is the main obstacle to the separation of `AsyncLogger` and related code into its own artifact.

IMHO we should drop JMX support in 3.0.x and evaluate if such a technology should be added again in 3.1.x.


log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -436,13 +435,7 @@ public boolean stop(final long timeout, final TimeUnit timeUnit) {
...
 
             this.setStopping();
-            String name = getName();
-            try {
-                Server.unregisterLoggerContext(name); // LOG4J2-406, LOG4J2-500
-            } catch (final LinkageError | Exception e) {
-                // LOG4J2-1506 Hello Android, GAE
-                LOGGER.error("Unable to unregister MBeans", e);
-            }
+
```
vy, line 438: Why did we remove this and another below Android-specific guards?

vy (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
ppkarwasz: I am not sure I understand: I removed the `Server` class and method call, so the `try ... catch` becomes empty.

vy: Oh! :facepalm: Got it.

vy (approved):

ppkarwasz:
Part of  #2163

Issue #2234, opened 2024-01-24
`RingBufferLogEvent` does not `clear()` the `threadName`


vy:
The issue was reported by @piecebook in https://github.com/apache/logging-log4j2/discussions/2233 discussion.

PR #2239, opened 2024-01-24
Fix forgotten `threadName` field in `RingBufferLogEvent#clear()` (#2234)

vy:


ppkarwasz (approved):
LGTM,

Can you port it to `main` too, so that we don't have regressions?

vy:
@ppkarwasz, will do.

PR #2242, opened 2024-01-24
Remove instance of checks for async components

ppkarwasz:
This PR removes all occurences of `instanceof Async*` that might prevent the separation of the async support.

The removal requires a couple of new API methods and binary incompatible changes.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Marker;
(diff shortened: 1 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -614,15 +611,18 @@ public ReliabilityStrategy getReliabilityStrategy() {
...
-        if (predicate == null || predicate.allow(this)) {
+        if (hasDispatchType(dispatchType)) {
             callAppenders(event);
         }
-        logParent(event, predicate);
+        logParent(event, dispatchType);
+    }
+
+    protected boolean hasDispatchType(final DispatchType dispatchType) {
+        return dispatchType != DispatchType.ASYNCHRONOUS;
```
vy, line 625: I am not able to follow the logic here. `hasDispatchType()` sounds like we will check if there are any dispatch types. Though we certainly don't do that. Plus, we only check if it is not async; again something I couldn't understand. Mind elaborating on what are you trying to do here? I was expecting something like `boolean dispatchTypeAllowed(DispatchType)`.

Also, I am puzzled to see _async_ in `LoggerConfig`. `DispatchType` will still litter the `LoggerConfig` with `async` even if we split it into a separate module. Am I wrong? To avoid this, shouldn't `DispatchType` be a protected interface that integrators extend from instead?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -649,7 +649,7 @@ public boolean requiresLocation() {
         return false;
     }
 
-    private void logParent(final LogEvent event, final LoggerConfigPredicate predicate) {
+    private void logParent(final LogEvent event, final DispatchType predicate) {
```
vy, line 652: Could you rename `predicate` to `dispatchType`, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
ppkarwasz: I replaced `LoggerConfigPredicate` with `Predicate<LoggerConfig>`.
All references to async in `log4j-core` should be gone.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
ppkarwasz: In the corrected version I use `Predicate<LoggerConfig>`.

vy (approved):

Issue #2204, opened 2024-01-16
Replace `LowLevelLogUtil` with a more sophisticated bootstrap logger


jvz:
This was something I experimented with in the spi-updates branch. Essentially, the strategy used by the low-level logging util should be an interface that begins with a simple implementation that enqueues status data elements until this reference is replaced with a `StatusLogger`-backed version once `StatusLogger` has been initialized. This would drain the accumulated status data to the logger.

Such an API would look something like this (abbreviated):

```java
@NullMarked
public interface BootstrapLogger {
    void log(
            final Level level,
            @Nullable final Marker marker,
            final String message,
            @Nullable final Throwable exception);
}
```

This could potentially be changed to log a `StatusData` object or a newly defined object for bootstrap messages. In any case, this API should be internal to `log4j-api` (i.e., not exported in the module) as it's only relevant for classes involved in bootstrapping.

vy:
@jvz, do we really need this sophistication? Both `LowLevelLogUtil` and `StatusLogger` are supposed to help Log4j components to log. That is, they are the logging framework of the logging framework – the Log4j. It is not expected to be used by anything but Log4j. Right? (Please correct me if I am wrong.) Why can't we simply have a `StatusLogger` such that

* It doesn't extend the logger API of Log4j
* It provides a `public static void log(Level level, String message)` method
* It provides a `public static void log(Level level, String format, Object... args)` method supporting `Throwable` in the last argument
* It provides a `public static void isLevelEnabled(Level level)` method
* It is only activated by a system property `log4j.statusLogger.level=(debug|info|warn|error)`
* It only buffers logged messages iff instructed so by a system property `log4j.statusLogger.bufferLength=<positiveInteger>`
* Buffering is exposed via `public static Collection<StatusMessage> buffer()`, where `StatusMessage` is composed of
  * `Instant instant`
  * `Level level`
  * `String message`
  * `Throwable error`

Note the similarity to [Spring's `DeferredLog`](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/logging/DeferredLog.html).

I have a feeling this single-class-file design will not only greatly simplify the code, but will also serve all existing use cases. This will render the `status` attribute in the configuration files redundant.

@jvz, WDYT?

jvz:
Hmm, that's an interesting proposal! I can try your idea out. The main purpose of my proposal is to make it easier to configure logging there as `LowLevelLogUtil` is option-less.

[...]

PR #2249, opened 2024-01-25
Draft: Make `StatusLogger` self-contained and testable

vy:
Motivated by #2204 and making `log4j-api-2.x` adaptable by Log4j 3, this work revamps the `StatusLogger` to make it self-contained and testable. Noteworthy changes are as follows:

### Simplifications and improvements for self-contained implementation

* Dependencies on `PropertiesUtil` and `LoaderUtil` are eliminated. To do so, simple system property and resource from classpath readers are implemented. See the new `StatusLogger.Config` class.
* Dependency on `SimpleLogger` is eliminated. The default listener is fixed to a `StatusConsoleListener` which dumps `StatusData#getFormattedStatus()` output to a `PrintStream`.
* **[behavioural change]** `StatusConfiguration` does not create a new `StatusListener` each time anymore. That is, there will not be a dedicated `StatusConsoleListener` for every each `LoggerContext`. Existing behaviour was already problematic:
   * Each `StatusConfiguration` created was resetting the `StatusLogger` listener level (hence, last one always wins)
   * Created `StatusConsoleListener`s weren't cleaned up
   * Even though there was a dedicated listener for each `LoggerContext`, there is always a single `StatusLogger`, defeating the purpose of per-`LoggerContext` listeners
* Hence, we embraced a simpler model: There is one default `StatusConsoleListener` for the `StatusLogger` singleton. If user overrides the `StatusLogger` configuration in a `log4j2.xml`, last read configuration will be the effective one.
* Several `StatusLogger` javadocs are rewritten.
* `LowLevelLogUtil` is removed.
* `AbstractLogger` doesn't act on `log4j2.messageFactory` and `log4j2.flowMessageFactory` properties anymore; its message and flow message factories are fixed to `ParameterizedMessageFactory` and `DefaultFlowMessageFactory`, respectively. This was necessary to avoid cyclic instantiation: `PropertiesUtil` is using `StatusLogger`, which extends `AbstractLogger`, which indirectly reaches back to `PropertiesUtil` for these two functionalities removed.

### Testing-related changes

* `StatusLogger` is made extendable (to enable mocking/spying)
* `StatusLogger` static instance is made replaceable (to enable substituting the implementation for tests)
* Leveraging these, `@UsingStatusLoggerMock` introduced

### TODO

* Introduce `log4j2.messageFactory` and `log4j2.flowMessageFactory` properties back
* Update docs
* Fix last remaining failing tests (already fixed some)
* Check if there are any refactoring opportunities for `@UsingStatusListener`

### Review kit

1. Start with reviewing changes in `o.a.l.l.status` package: `StatusLogger`, `StatusData`, `StatusConsoleListener`, etc.
2. Review `StatusConfiguration`
3. Review `o.a.l.l.test.junit.StatusLoggerMockExtension` to see testing conveniences
4. Review the rest


rgoers:
I will go through this as I can. It might take a couple of days depending on my available time. However, the description of the solution above seems to aim right on target.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,567 @@
...
             }
         }
         return null;
     }
 
     @Override
-    public boolean isEnabled(final Level level, final Marker marker, final String message, final Throwable t) {
+    public boolean isEnabled(final Level level, final Marker marker, final String message, final Throwable throwable) {
         return isEnabled(level, marker);
     }
```
ppkarwasz, line 581: It is probably outside of the scope of this PR, but we implement these `ExtendedLogger` methods in the same way multiple times in our codebase.

This should be a `default` method on the `ExtendedLogger`, since it works for all implementations except `log4j-core`. We could even add:

```java
default boolean isEnabled(final Level level, final Marker marker) {
    return isEnabled(level);
}
```
since this works for all implementations except `log4j-to-slf4j` (and obviously `log4j-core`).

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -462,53 +718,30 @@ public boolean isEnabled(
...
-            super.add(object);
-            while (messages.size() > size) {
-                messages.poll();
-            }
-            return size > 0;
+        requireNonNull(level, "level");
+        listenerReadLock.lock();
+        try {
+            return listeners.stream()
+                    .anyMatch(listener -> listener.getStatusLevel().isLessSpecificThan(level));
```
ppkarwasz, line 742: The maximum level of all listeners, should be cached.
Status messages are not so rare (e.g. you get one each time a new SLF4J logger is created) and the stream API will introduce a lot of garbage.

ppkarwasz (commented):
This looks almost ready.

Remark the `PropertiesUtil.Environment` does not log anything about broken property sources to prevent a recursive initialization. Now we can start logging again.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,567 @@
...
-        super(name, messageFactory);
-        final Level loggerLevel = DEBUG_ENABLED ? Level.TRACE : Level.ERROR;
-        this.logger = loggerFactory.createSimpleLogger("StatusLogger", loggerLevel, messageFactory, System.err);
-        this.listenersLevel = Level.toLevel(DEFAULT_STATUS_LEVEL, Level.WARN).intLevel();
+    public StatusLogger() {
+        this(
+                StatusLogger.class.getSimpleName(),
+                ParameterizedNoReferenceMessageFactory.INSTANCE,
+                Config.getInstance(),
+                new StatusConsoleListener(Config.getInstance().defaultListenerLevel));
```
ppkarwasz, line 322: > Users should not create new instances, but use {@link #getLogger()} instead!

I would remove this constructor instead and inline it in the only caller.

In our code we will rather use the other constructor.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,567 @@
...
+     * @return a thread-safe read-only collection of buffered events
      */
     public List<StatusData> getStatusData() {
-        msgLock.lock();
-        try {
-            return new ArrayList<>(messages);
-        } finally {
-            msgLock.unlock();
-        }
+        return Collections.unmodifiableList(new ArrayList<>(buffer));
```
ppkarwasz, line 493: Do we need to prevent modification of the returned data, since it is a copy?

A typical application for us would be in tests where we can:

- get a snapshot of the data,
- assert the presence of some `WARN` messages,
- assert that there are no other `WARN` messages.

vy: I like the idea of making the returned collection read-only. This makes it clear that mutations on the returned buffer will not be propagated.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: No-arg ctor allows restore once the state is corrupted. Consider the following:
```
StatusLogger.setLogger(mockStatusLogger);
// Okay, I am done with my tests.
// How can I restore the `StatusLogger`?
StatusLogger.setLogger(new StatusLogger());
```
One can argue that `StatusLogger.getLogger()` should be safely stored to a reference before being replaced.

Anyway... This was sort of my reasoning. I am not strongly opinionated. @ppkarwasz, what do you prefer?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Fixed in 21a47acf352b790fbbe6e1db6c815f05f6f8a331. Could you check that and, if it is okay, resolve the conversation, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: We can indeed enhance `ExtendedLogger` with `default`s. But I prefer to do it outside this PR.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,567 @@
...
-     * entries will be removed as new entries are added.
+     * The name of the system property that can be configured with the maximum number of events buffered.
+     * <p>
+     * Once the limit is reached, older entries will be removed as new entries are added.
+     * </p>
+     *
+     * @deprecated Use {@link #BUFFER_CAPACITY_PROPERTY_NAME} instead.
+     */
+    @Deprecated
+    public static final String MAX_STATUS_ENTRIES = BUFFER_CAPACITY_PROPERTY_NAME;
```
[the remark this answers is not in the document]
vy: Trying to adhere to `*_PROPERTY_NAME` and `*_DEFAULT_VALUE` convention.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusConsoleListener.java
```diff
@@ -56,50 +54,59 @@ public StatusConsoleListener(final Level level) {
...
-        logger
-                // Logging using _only_ the following 4 fields set by `StatusLogger#logMessage()`:
-                .atLevel(data.getLevel())
-                .withThrowable(data.getThrowable())
-                .withLocation(data.getStackTraceElement())
-                .log(data.getMessage());
+        requireNonNull(data, "data");
+        if (data.getLevel().isMoreSpecificThan(level)) {
+            final String formattedStatus = data.getFormattedStatus();
+            stream.println(formattedStatus);
```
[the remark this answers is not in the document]
vy: Call me opinionated, but I (strongly?) prefer `var a = a(); var b = b(a); c(b)` over `c(b(a))`, since
1. Variables give you documentation opportunity (e.g., `boolean validUserName = userName.matches(userNamePattern)`)
2. One-statement-per-line is easier to read and understand
3. Variables allow you to track intermediate values while stepping through code in a debugger. AFAIK, this is not possible in `c(b(a))` form.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/package-info.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the license.
  */
 @Export
-@Version("2.21.1")
+@Version("2.22.0")
```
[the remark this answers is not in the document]
vy: I thought we were sticking to `bnd-baseline-maven-plugin` suggestions?

> keep the OSGi version somehow related to the artifact version

I thought these two are almost unrelated. Am I missing something?

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/UsingStatusLoggerMock.java
```diff
@@ -0,0 +1,35 @@
...
+import java.lang.annotation.Target;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+/**
+ * Shortcut to {@link StatusLoggerMockExtension}.
+ */
+@Retention(RUNTIME)
+@Target({TYPE, METHOD})
+@Documented
+@ExtendWith(StatusLoggerMockExtension.class)
```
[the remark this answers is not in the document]
vy: Fixed in fd67e38cdab2e7562809bc192a89f1bf1d4c6493.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/package-info.java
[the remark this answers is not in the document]
ppkarwasz: You might have missed my proposal: https://lists.apache.org/thread/tv3t11pwq54ch56qrj6odz9hngys1761

Rationale: it might be confusing to users to match the package version with the artifact version. If they see "2.22.0", they might think `log4j-api` 2.22.0 is enough. So I would propose:

- for a micro bump, just increase the micro number (e.g. `2.21.1` -> `2.21.2`). If users compiled their code against package number `2.21.7`, it should still work with `log4j-api` 2.21.0 (module bug fixes),
- for a minor bump, align the version number with the version of `log4j-api`. If the package is `2.21.1` we bump it to `2.23.0` in 2.x and `3.0.0` in 3.x (unless we can guarantee that the changes in 2.x and 3.x are synchronized).

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
ppkarwasz: In my experience every `public` method **will** be used by users. So my strategy is to expose only those methods that are necessary or that we are willing to support.

Mocking the status logger is a very nice use case, but it will probably used in a single test class, because it is not parallelisable. All the other test classes should register a separate listener (like `JUnitListStatusListener`) or even better: inject a `Logger/StatusLogger` into the components they are testing.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
ppkarwasz: Here we might need a `public` method like `updateListenerLevels` (yes, I said `public` ;-)).

Imagine if a user updates the level of its listener, he must be able to also update the cached value (cf. Martin Fowler's [two hard things](https://martinfowler.com/bliki/TwoHardThings.html)).

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/package-info.java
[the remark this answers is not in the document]
jvz: If we use the module version as the package version, then we should also update the various packages marked as version `1.0.0`. And if `Foo.class.getPackage().getSpecificationVersion()` (and `getImplementationVersion()`) return the module version rather than the OSGi package version, that's just further confusion!

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusData.java
```diff
@@ -34,58 +36,93 @@ public class StatusData implements Serializable {
 
     private static final long serialVersionUID = -4341916115118014017L;
 
-    private final long timestamp;
+    private final Instant instant;
+
+    @Nullable
+    private final DateTimeFormatter instantFormatter;
```
jvz, line 42: Thank you for this! I was thinking along these same lines myself.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,590 @@
...
 
-    private static final String DEFAULT_STATUS_LEVEL = PROPS.getStringProperty(DEFAULT_STATUS_LISTENER_LEVEL);
+        // We need to roll out our own `.properties` reader.
+        // We could have used `PropertiesUtil`, `PropertyFilePropertySource`, etc.
+        // Consequently, they would delegate to `LoaderUtil`, etc.
+        // All these mechanisms expect a working `StatusLogger`.
+        // Hence, in order to be self-sufficient, we cannot rely on them.
+        private static Properties readPropertiesFile() {
+            final Properties properties = new Properties();
+            final URL url = StatusLogger.class.getResource(PROPERTIES_FILE_NAME);
```
jvz, line 280: This is an area I struggled with when I attempted a variation of this PR. The original `LoaderUtil` code attempts to find this file using multiple `ClassLoader` instances. At the very least, it could be useful to include a fallback here to try to load the file from the thread context class loader, though on the other hand, this is loaded so early in code that a custom TCCL is unlikely to have been set yet.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,590 @@
...
+ * {@code StatusLogger} is designed as a singleton class accessed statically.
+ * If you are running an application containing multiple Log4j configurations (e.g., in a servlet environment with multiple containers) and you happen to have differing {@code StatusLogger} configurations (e.g, one {@code log4j2.xml} containing {@code <Configuration status="ERROR">} while the other {@code <Configuration status="INFO">}), the last loaded configuration will be effective one.
+ * </p>
+ * <h3>Debug mode</h3>
+ * <p>
+ * When the {@value Constants#LOG4J2_DEBUG} system property is present, any level-related filtering will be skipped and all events will be notified to listeners.
+ * </p>
  */
-public final class StatusLogger extends AbstractLogger {
+public class StatusLogger extends AbstractLogger {
```
jvz, line 88: Since this class is still serializable in 2.x, I suggest using a serialization proxy for this logger rather than relying on any particular `transient` or non-transient fields. That way, you can serialize the configuration (which is currently not serializable), transform the status data entries into a basic array (or `ArrayList`), and exclude any fields that can be constructed from the proxy state.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -16,311 +16,590 @@
...
+                // There is nothing we can do but simply dumping the failure.
+                error.printStackTrace(System.err);
+            }
+            return properties;
+        }
+    }
 
-    static final boolean DEBUG_ENABLED =
-            PropertiesUtil.getProperties().getBooleanProperty(Constants.LOG4J2_DEBUG, false, true);
+    private static volatile StatusLogger INSTANCE = new StatusLogger();
```
jvz, line 295: Note that by putting this in a static field, this means the static init of this class will cause the construction of the field. You could wrap it in a `Lazy` to defer that until something calls `getLogger` if you'd like.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
jvz: Usually, an unmodifiable list view is made on the modifiable list. If you're returning an immutable copy, at least in newer Java, you'd just use `List::copyOf`. If that's the intention here, then go for it; otherwise, returning an `ArrayList` copy is sufficient since the caller can't change the `StatusLogger` collection.

jvz (approved):
I like this approach so far!

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
ppkarwasz: The `log4j2.StatusLogger.properties` file provides options that apply to all logger contexts.
Looking it up in the TCCL does not make sense to me.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Can't they better re-register? That is, `logger.removeListener(this); logger.addListener(this)`? One less public method?

vy:
@ppkarwasz, I have introduced the fallback listener concept back in cd3fadb3bc65e34e6fa7f53628bfd4f77f5824c2. Would you mind reviewing it, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Removed the `StatusLogger()` ctor in f1e3e2f3238fbdd196b6bafe22451e10353690b2.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Documented the rationale in c4c941e9125cf259ad36e24facd91c046bf0ad67.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: I am taking the initiative to mark this conversation as resolved. @jvz, if you have objections, please feel free to re-open the conversation.

log4j-core/src/main/java/org/apache/logging/log4j/core/jmx/StatusLoggerAdmin.java
```diff
@@ -92,6 +80,50 @@ private static MBeanNotificationInfo createNotificationInfo() {
...
+        super.addNotificationListener(listener, filter, handback);
+        registeredNotificationListeners.add(listener);
+        updateStatusListenerRegistration();
+    }
+
+    @Override
+    public void removeNotificationListener(final NotificationListener listener) throws ListenerNotFoundException {
+        super.removeNotificationListener(listener);
+        registeredNotificationListeners.remove(listener);
+        updateStatusListenerRegistration();
```
ppkarwasz, line 104: Looking at the `NotificationBroadcasterSupport` class, the logic to add and remove listeners is more complex.

If I understand correctly:

 * NBS remembers a list of triplets `(listener, filter, handback`,
 * the `removeNotificationListener(listener)` method removes all the triplets with the give `listener` value,
 * the `removeNotificationListener(listener, filter, handback)` method remove **only** the triplet with the given values of `listener`, `filter` and `handback`.

I am not convinced we need to implement this logic, but we should either:
 * implement the complete logic,
 * or limit ourselves to registering the status listener in `addNotificationListener` and removing the status listener only when `MBeanRegistration#postDeregister` is called. This way we ignore all the intricacies of notifications.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
ppkarwasz: My initial remark was mostly about `isEnabled` generating garbage (a new `Stream`).

Maybe the ideal solution is not caching as I proposed, but replacing `stream()` with a for loop:

```java
for (int i = 0; i < listeners.size(); i++) {
    if (listeners.get(i).getStatusLevel().isLessSpecificThan(level)) {
        return true;
    }
}
```

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: Good catch. Fixed in 3684a2e2de6132f2d367e3887f164d0626878a27.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/package-info.java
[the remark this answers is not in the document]
vy: Fixed in 2c01bcd2d94a4f5a20daa2b04181b939a25b83db.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: I see your point. I did not improve the serializability of `StatusLogger`, nor worsened it. Hence, I am inclined to keep the things the way they are for the time being. If it deems necessary, I would prefer to work on the `StatusLogger` serializability in a separate PR.

ppkarwasz: Since we don't assign `StatusLogger` to instance fields, I don't think serialization is a problem.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: @ppkarwasz, could you check 40cc7218ce2e1d3a56e25244ed0975fcb37ba2de, please?

ppkarwasz: LGTM

ppkarwasz (approved):
LGTM

log4j-core/src/main/java/org/apache/logging/log4j/core/jmx/StatusLoggerAdmin.java
vy: @ppkarwasz, thanks for the due diligence during review. I implemented the `postDeregister()` approach in 88eb8fcb415bbe7abf544467d632cde12675d375. Would you mind checking it out, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusConsoleListener.java
```diff
@@ -56,50 +54,59 @@ public StatusConsoleListener(final Level level) {
...
     }
 
     /**
-     * Return the Log Level for which the Listener should receive events.
-     * @return the Log Level.
+     * Sets the output stream to a new value.
+     *
+     * @param stream the new output stream
+     * @throws NullPointerException on null {@code stream}
+     * @since 2.23.0
```
ppkarwasz, line 76:
```suggestion
     * @since 2.23
```
Nit: we don't need to specify the patch version where this was introduced, since it **always** must be `0`.

vy: If we can, why not? Further, I don't think `2.23` is valid. That is, `@since` is supposed to refer to a version, `2.23` is not a version, but a version prefix. I find this pretty confusing, in particular, given Log4j had releases of `2.1`, `2.3`, etc. versioning scheme.

ppkarwasz: I don't know if the parameter of `@since` must be a valid version number. In practice JDK methods have annotations like `@since 1.8` and `@since 9`.

I prefer `2.23`, but I don't mind too much `2.23.0`. I am mostly happy you annotated this method, since most methods are not annotated.

lauredogit:
Hi,
This changes caused an issue when running on Java 8 and having configured a DateFormat.

It will cause a StackOverflowError such as:

`Caused by: java.lang.StackOverflowError
	at org.apache.logging.log4j.message.ParameterFormatter.analyzePattern(ParameterFormatter.java:93)
	at org.apache.logging.log4j.message.ParameterizedMessage.<init>(ParameterizedMessage.java:142)
	at org.apache.logging.log4j.message.ParameterizedMessage.<init>(ParameterizedMessage.java:176)
	at org.apache.logging.log4j.message.ParameterizedNoReferenceMessageFactory.newMessage(ParameterizedNoReferenceMessageFactory.java:105)
	at org.apache.logging.log4j.message.AbstractMessageFactory.newMessage(AbstractMessageFactory.java:103)
	at org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:2686)
	at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:2460)
	at org.apache.logging.log4j.spi.AbstractLogger.warn(AbstractLogger.java:3674)
	at org.apache.logging.log4j.spi.AbstractLogger.handleLogMessageException(AbstractLogger.java:2927)
	at org.apache.logging.log4j.spi.AbstractLogger.tryLogMessage(AbstractLogger.java:2909)
	at org.apache.logging.log4j.spi.AbstractLogger.logMessageTrackRecursion(AbstractLogger.java:2859)
	at org.apache.logging.log4j.spi.AbstractLogger.logMessageSafely(AbstractLogger.java:2841)
	at org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:2687)
	at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:2460)
	at org.apache.logging.log4j.spi.AbstractLogger.warn(AbstractLogger.java:3674)
	at org.apache.logging.log4j.spi.AbstractLogger.handleLogMessageException(AbstractLogger.java:2927)`

which is caused by

`java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: YearOfEra
        at java.time.Instant.getLong(Instant.java:603)
        at java.time.format.DateTimePrintContext.getValue(DateTimePrintContext.java:298)
        at java.time.format.DateTimeFormatterBuilder$NumberPrinterParser.format(DateTimeFormatterBuilder.java:2551)
        at java.time.format.DateTimeFormatterBuilder$CompositePrinterParser.format(DateTimeFormatterBuilder.java:2190)
        at java.time.format.DateTimeFormatter.formatTo(DateTimeFormatter.java:1746)
        at java.time.format.DateTimeFormatter.format(DateTimeFormatter.java:1720)
        at org.apache.logging.log4j.status.StatusData.getFormattedStatus(StatusData.java:170)
        at org.apache.logging.log4j.status.StatusConsoleListener.log(StatusConsoleListener.java:138)
        at org.apache.logging.log4j.status.StatusLogger.notifyListener(StatusLogger.java:555)
        at org.apache.logging.log4j.status.StatusLogger.notifyListeners(StatusLogger.java:549)
        at org.apache.logging.log4j.status.StatusLogger.logMessage(StatusLogger.java:526)
        at org.apache.logging.log4j.spi.AbstractLogger.log(AbstractLogger.java:2812)
        at org.apache.logging.log4j.spi.AbstractLogger.tryLogMessage(AbstractLogger.java:2906)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessageTrackRecursion(AbstractLogger.java:2859)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessageSafely(AbstractLogger.java:2841)
        at org.apache.logging.log4j.spi.AbstractLogger.logMessage(AbstractLogger.java:2620)
        at org.apache.logging.log4j.spi.AbstractLogger.logIfEnabled(AbstractLogger.java:2567)
        at org.apache.logging.log4j.spi.AbstractLogger.warn(AbstractLogger.java:3469)`

because in org.apache.logging.log4j.status.StatusLogger.Config#readInstantFormatter, the code is merely

`        private static DateTimeFormatter readInstantFormatter(final Properties fileProvidedProperties) {
            final String format = readProperty(fileProvidedProperties, STATUS_DATE_FORMAT);
            return format != null ? DateTimeFormatter.ofPattern(format) : null;
        }
`

with only **DateTimeFormatter.ofPattern(format)**

This works on Java 11, but on Java 8, it breaks as in https://stackoverflow.com/questions/25229124/unsupportedtemporaltypeexception-when-formatting-instant-to-string

A very simple fix is to use instead

`        private static DateTimeFormatter readInstantFormatter(final Properties fileProvidedProperties) {
            final String format = readProperty(fileProvidedProperties, STATUS_DATE_FORMAT);
            return format != null ? DateTimeFormatter.ofPattern(format).withZone(ZoneId.systemDefault()) : null;
        }`

vy:
@lauredogit, thanks so much for the report. I will submit a fix. We are considering a `2.23.1` release sometime soon, maybe next week?

lauredogit:
Perfect, thanks!

I think there are 2 distinct issues at play.

The easy one to fix is the actual _UnsupportedTemporalTypeException_ on Java 8, which will be fixed easily by adding the ".withZone(ZoneId.systemDefault())".

However, the other issue is that having such an exception thrown by the StatusLogger was causing a _StackOverflowError_... 
and this was not mitigated by AbstractLogger._logMessageTrackRecursion()_.

This can be investigated later on, though.

ppkarwasz:
Well, _logMessage_**Track**_Recursion_ only tracks recursion, it doesn't do anything about it. Other parts of the system like async logger read the recursion level (and log synchronously for example).

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
carterkozak: I believe this line introduced a bug. Unlike [`ClassLoader.getResource`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/ClassLoader.html#getResource(java.lang.String)), which takes absolute resource paths, [`Class.getResource`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Class.html#getResource(java.lang.String)) supports relative resource paths, so the `log4j2.StatusLogger.properties` resource must be placed into jar resources as `org/apache/logging/log4j/status/log4j2.StatusLogger.properties`, and existing configuration resources which worked with previous log4j2 releases are ignored.
```suggestion
            final URL url = StatusLogger.class.getResource('/' + PROPERTIES_FILE_NAME);
```

vy: @carterkozak, doh! :facepalm: Thanks for the report. Could you review #2354, please?

PR #2101, opened 2023-12-15
Mark `JdkMapAdapterStringMap` as frozen if map is immutable

ppkarwasz:
This PR marks `StringMap`s generated from Java `Map`s as frozen if the wrapped map is immutable.

Fixes #2098

Issue #2098, opened 2023-12-15
Using `Map.of` in a `ContextDataProvider` implementation throws UOE 


slinkydeveloper:
## Description

I'm using logj2 with a custom `ContextDataProvider`, and then using the [aws lambda appender](https://docs.aws.amazon.com/lambda/latest/dg/java-logging.html#java-logging-log4j2). My `ContextDataProvider` implementation uses `Map.of` methods from JDK to create cheap immutable maps, but apparently that doesn't work and throws UOE.

In the stacktrace it's shown that a method down the stack invokes `ImmutableCollections.AbstractImmutableMap::clear`, which throws UOE. Interesting enough, in many online examples they use `Collections.emptyMap()` for the empty map, which doesn't throw UOE on `clear`.

What puzzles me though is that I wasn't able to reproduce this issue with the log4j2-core console appender. Is the amazon appender doing something strange, or is this a bug of log4j2-core?

## Configuration

**Version:** 2.22.0

**Operating system:** Linux

**JDK:** 21 (amazon)

## Logs

```
2023-12-15T11:26:15.337000+00:00  WARN StatusConsoleListener org.apache.logging.log4j.spi.AbstractLogger caught java.lang.UnsupportedOperationException logging ReusableSimpleMessage: Incremented counter!
2023-12-15T11:26:15.338000+00:00  java.lang.UnsupportedOperationException
2023-12-15T11:26:15.338000+00:00  at java.base/java.util.ImmutableCollections.uoe(Unknown Source)
2023-12-15T11:26:15.338000+00:00  at java.base/java.util.ImmutableCollections$AbstractImmutableMap.clear(Unknown Source)
2023-12-15T11:26:15.338000+00:00  at org.apache.logging.log4j.core.impl.JdkMapAdapterStringMap.clear(JdkMapAdapterStringMap.java:122)
2023-12-15T11:26:15.338000+00:00  at org.apache.logging.log4j.core.impl.MutableLogEvent.clear(MutableLogEvent.java:143)
2023-12-15T11:26:15.338000+00:00  at org.apache.logging.log4j.core.impl.ReusableLogEventFactory.release(ReusableLogEventFactory.java:141)
2023-12-15T11:26:15.338000+00:00  at org.apache.logging.log4j.core.config.LoggerConfig.log(LoggerConfig.java:578)
[...]

PR #2256, opened 2024-01-29
Fix regression in `JdkMapAdapterStringMap` performance

### LINKED ISSUE #2238: Avoid a slow exception catch in JdkMapAdapterStringMap constructor
### Discussed in https://github.com/apache/logging-log4j2/discussions/2235

<div type='discussions-op-text'>

<sup>Originally posted by **jdgenerix** January 24, 2024</sup>
The `org.apache.logging.log4j.core.impl.JdkMapAdapterStringMap` constructor is called from `org.apache.logging.log4j.core.impl.ThreadContextDataInjector` with an UnmodiableMap (`copy`) and the constructor is trying to modify and catch an exception to detect if this is an immutable Map.
This code is triggered a lot of time and could be more efficient.
Adding a package protected constructor which take the immutable boolean flag as an argument and using it from `ThreadContextDataInjector` would avoid this slow try catch.

```java
        @Override
        public ReadOnlyStringMap rawContextData() {
            final ReadOnlyThreadContextMap map = ThreadContext.getThreadContextMap();
            if (map instanceof ReadOnlyStringMap) {
                return (ReadOnlyStringMap) map;
            }
            // note: default ThreadContextMap is null
            final Map<String, String> copy = ThreadContext.getImmutableContext();
            return copy.isEmpty() ? ContextDataFactory.emptyFrozenContextData() : new JdkMapAdapterStringMap(copy);
        }


        public JdkMapAdapterStringMap(final Map<String, String> map) {
		this.map = Objects.requireNonNull(map, "map");
		try {
			map.replace(Strings.EMPTY, Strings.EMPTY, Strings.EMPTY);
		} catch (final UnsupportedOperationException ignored) {
			immutable = true;
		}
	}
```</div>

### PR DESCRIPTION
**Pull request by ppkarwasz**
We introduce a new constructor for `JdkMapAdapterStringMap` to prevent the performance penalty of an exception if the map is immutable. The regression was introduced in #2101.

Closes #2238.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
@@ -44,6 +44,7 @@ public class JdkMapAdapterStringMapTest {
     @Test
     public void testConstructorDisallowsNull() {
         assertThrows(NullPointerException.class, () -> new JdkMapAdapterStringMap(null));
+        assertThrows(NullPointerException.class, () -> new JdkMapAdapterStringMap(null, false));
     }
 
     @Test
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
@@ -43,7 +43,6 @@ public class JdkMapAdapterStringMapTest {
 
     @Test
     public void testConstructorDisallowsNull() {
-        assertThrows(NullPointerException.class, () -> new JdkMapAdapterStringMap(null));
         assertThrows(NullPointerException.class, () -> new JdkMapAdapterStringMap(null, false));
     }
 
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMap.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMap.java
@@ -21,6 +21,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.WeakHashMap;
 import org.apache.logging.log4j.util.BiConsumer;
 import org.apache.logging.log4j.util.ReadOnlyStringMap;
 import org.apache.logging.log4j.util.StringMap;
@@ -41,13 +42,16 @@ public class JdkMapAdapterStringMap implements StringMap {
         }
         return left.compareTo(right);
     };
+    // Cache of known unmodifiable map implementations.
+    // It is a cache, no need to synchronise it between threads.
+    private static Map<Class<?>, Void> UNMODIFIABLE_MAPS_CACHE = new WeakHashMap<>();
 
     private final Map<String, String> map;
     private boolean immutable = false;
     private transient String[] sortedKeys;
 
     public JdkMapAdapterStringMap() {
-        this(new HashMap<String, String>(), false);
+        this(new HashMap<>(), false);
     }
 
     /**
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java [hunk 1]
```diff
@@ -44,6 +44,7 @@ public class JdkMapAdapterStringMapTest {
     @Test
     public void testConstructorDisallowsNull() {
         assertThrows(NullPointerException.class, () -> new JdkMapAdapterStringMap(null));
```
vy, line 46: [Optional] I would remove this line using the deprecated ctor.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
vy, line 1: [Optional] I would also replace the use of the deprecated ctor in the `testImmutability()` method too.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMap.java
```diff
@@ -47,9 +47,14 @@ public class JdkMapAdapterStringMap implements StringMap {
...
     }
 
+    /**
+     * @deprecated for performance reasons since 2.23.
+     *             Use {@link #JdkMapAdapterStringMap(Map, boolean)} instead.
+     */
+    @Deprecated
     public JdkMapAdapterStringMap(final Map<String, String> map) {
         this.map = Objects.requireNonNull(map, "map");
         try {
```
vy, line 60: I agree that the new ctor explicitly requiring an immutability flag is the way to go forward, though I feel uncomfortable leaving this deprecated ctor as is. This try-catch is the problem, yet we work around it rather than fixing it. Can't we do something else here? For instance,

```
if (knownImmutableMapClasses.contains(map.getClass())) {
    immutable = true;
} else {
    // try-catch
}
```

vy (changes requested):
Thanks so much for promptly handling this. I guess this warrants a quick `2.23.0` release.

Changes LGTM. But I am not feeling confident enough to say we addressed the problem. I don't see a test triggering the code path users reported. Can you add such a test, please? Maybe we can craft/mock a custom map throwing an [unexpected] exception on `replace()` and verify that `replace()` is not called?

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMap.java
ppkarwasz: What about:
```java
this(new HashMap<>(map), false);
```
you proposed in #2101?

Now that I removed all references to this constructor in **our** code, I don't have objections about it generating garbage. We could even use:
```java
this(map, true);
```

vy: @ppkarwasz, given we are trying to mitigate a performance regression, doesn't `HashMap::new` (or `this(map, true)`) create another by replacing the one from `try-catch`? I mean, are these significantly faster than `try-catch`?

ppkarwasz: Anything is faster than catching an exception. ;-)

I like the `HashMap::new` version better, because it enforces semantics: IMHO a **frozen** `StringMap` should be immutable, while a `StringMap` created from a `Collections.unmodifiableMap` can be modified by changing the underlying map.

As I remarked, I removed all references to the deprecated constructor in our code. On Github I could only find two other projects that use `JdkMapAdapterStringMap`:
- [zenmoto/opentelemetry_logging_prototypes](https://github.com/zenmoto/opentelemetry_logging_prototypes/blob/main/log4j/src/main/java/io/opentelemetry/sdk/proto/LayoutWrapper.java),
- [apple/servicetalk](https://github.com/apple/servicetalk/blob/main/servicetalk-opentracing-log4j2/src/main/java/io/servicetalk/opentracing/log4j2/ServiceTalkTracingThreadContextMap.java).

log4j-core-test/src/test/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMapTest.java
ppkarwasz: I would prefer to keep it to test that if someone uses it, it works correctly.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/JdkMapAdapterStringMap.java
```diff
@@ -41,27 +42,60 @@ public class JdkMapAdapterStringMap implements StringMap {
...
+     */
+    public JdkMapAdapterStringMap(final Map<String, String> map, final boolean frozen) {
+        this.map = Objects.requireNonNull(map, "map");
+        this.immutable = frozen;
+    }
+
     @Override
     public Map<String, String> toMap() {
-        return map;
+        return new HashMap<>(map);
```
ppkarwasz, line 98: This change is required by the `toMap` contract, which requires a "mutable snapshot of the underlying data".

rgoers:
This seems strange to me. Why wouldn't you create an ImmutableJdkMapAdapterStringMap that extends JdkMapAdapterStringMap (or vice versa) instead of using a boolean parameter? It seems like it would be cleaner.

ppkarwasz:
@rgoers,

> This seems strange to me. Why wouldn't you create an ImmutableJdkMapAdapterStringMap that extends JdkMapAdapterStringMap (or vice versa) instead of using a boolean parameter? It seems like it would be cleaner.

The original problem in #2098 was that a user implemented [`ContextDataProvider#supplyContextData`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/util/ContextDataProvider.html) using `Map#of`. This caused the generation of a `StringMap` that wasn't frozen, but wasn't modifiable either, since the underlying map was immutable.

I tried auto-detecting the problem with the performance issue we are faced today. By adding a new constructor I want to give users clear directions that:

- if the provided map is not immutable, Log4j might reuse it to prevent the allocation of new instances,
- if the provided map is immutable, it must be marked so.

I can create an `ImmutableJdkMapAdapterStringMap`, but then the contract of the original `JdkMapAdapterStringMap` constructor must be changed. What do you think?

ppkarwasz:
@rgoers,

I think that an entire `ImmutableJdkMapAdapterStringMap` class is a little bit overkill for this, but maybe a pair of factory methods would be more appropriate (IDEA complains that `boolean` in a constructor is bad practice).
Any idea how to call them? `createMutable` and `createFrozen`?

PR #2259, opened 2024-02-02
[LOG4J2-3680] Allow deserialization of arrays

ppkarwasz:
The `DefaultObjectInputFilter` is too restrictive and it does not allow the deserialization of **arrays** of whitelisted classes.

Most notably `RuntimeException` can not be deserialized, since it contains an array of `StackTraceElement`s. Both classes are allowed, but deserialization of the `StackTraceElement[]` array fails.

We also replace all manual serializations/deserializations in test classes to calls to `SerialUtil`. This guarantees a consistent behavior of all tests under Java 8 and Java 9+.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/internal/SerializationUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/internal/SerializationUtil.java
@@ -163,8 +163,7 @@ public final class SerializationUtil {
      * @see Class#getName()
      */
     public static String stripArray(final String name) {
-        int offset;
-        for (offset = 0; offset < name.length() && name.charAt(offset) == '['; offset++) {}
+        final int offset = name.lastIndexOf('[') + 1;
         if (offset == 0) {
             return name;
         }
@@ -191,8 +190,7 @@ public final class SerializationUtil {
             case "S":
                 return "short";
             default:
-                // Should never happen
-                return name;
+                throw new IllegalArgumentException("Unsupported array class signature '" + name + "'");
         }
     }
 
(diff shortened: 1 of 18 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/internal/SerializationUtil.java
```diff
@@ -133,5 +139,62 @@ public static void assertFiltered(final java.io.ObjectInputStream stream) {
...
+     * Gets the class name of an array component recursively.
+     * <p>
+     *     If {@code name} is not the name of an array class it is returned unchanged.
+     * </p>
+     * @param name the name of a class.
+     * @see Class#getName()
+     */
+    public static String stripArray(final String name) {
+        int offset;
+        for (offset = 0; offset < name.length() && name.charAt(offset) == '['; offset++) {}
```
vy, line 167: Why not simply using `String#indexOf(char)` instead?

log4j-api/src/main/java/org/apache/logging/log4j/util/internal/SerializationUtil.java
```diff
@@ -133,5 +139,62 @@ public static void assertFiltered(final java.io.ObjectInputStream stream) {
...
+                return "double";
+            case "F":
+                return "float";
+            case "I":
+                return "int";
+            case "J":
+                return "long";
+            case "S":
+                return "short";
+            default:
```
vy, line 193: Shouldn't we be throwing an exception? In particular, we are in pretty dangerous waters of serialization...

vy (approved):
LGTM. But I would have expected `stripArray()` to thrown an exception on unknown array type.

PR #2275, opened 2024-02-09
Fix `StringBuilder` cache corruption on recursive access

vy:


log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
```diff
@@ -32,10 +33,18 @@
...
+
+    private static final StringBuilderRecycler STRING_BUILDER_RECYCLER = StringBuilderRecycler.of(
+            Constants.MAX_REUSABLE_MESSAGE_SIZE,
+            // This value indicates the maximum recursion depth before the recycler starts creating new instances.
+            // Consider a `ParameterizedMessage` containing an argument such that its `toString()` causes another (i.e.,
+            // recursive) `ParameterizedMessage` formatting. This value indicates the depth we support garbage-free
+            // formatting in such nested formatting situations. When this depth is exceeded, code still works, but
+            // starts generating garbage due to new `StringBuilder` allocations.
+            3,
+            Constants.ENABLE_THREADLOCALS);
```
ppkarwasz, line 47: In the case of `ReusableParameterizedMessage` is recursion even possible? The `buffer` field is an instance field and `getFormattedMessage` does not call itself recursively.

I would simplify this with:
```java
private transient StringBuilder buffer;
```

No thread local is needed: if `ReusableParameterizedMessage` is used concurrently on two threads, the problem lies elsewhere.

ppkarwasz (commented):
This certainly looks great and builds up on your work on `main`, but do we really need to backport `StringBuilderRecycler` to `2.x`?

This is an edge case that is only triggered, if we two nested log calls. I would prefer a fix with the minimal amount of LOCs required. In this case I think that `ParameterizedMessage` needs a:

```java
private static final ThreadLocal<AtomicInteger> recursionCount;
```

field.

ppkarwasz (approved):
LGTM

vy:
> I would prefer a fix with the minimal amount of LOCs required.

You are right. I have changed the implementation as you suggested and it indeed looks better. :100: Thanks for the review. :heart_eyes:

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableParameterizedMessage.java
vy: You're right. Simplified it with a `private final StringBuilder` instead.

Issue #2274, opened 2024-02-08
Custom Level and Properties does not work together


paulonicolau-rj:
## Description

I´m implementing a log file contains trace of request for a funcionality. When i use absolute path in filename attribute of RoliingFileappender, it works fine. When I use property to declare filename, this is ignored and his key is used. 

## Configuration

**Version:** [Log4j 2.22.1]

**Operating system:** [Windows 10/Amazon Linux 2]

**JDK:** [Oracle 17]

## Logs

```
[Stacktraces, errors, etc. relevant applications logs.]
```

## Reproduction

[An isolated test reproducing the test.
JUnit tests similar to the ones in the code base are extremely appreciated.]

![image](https://github.com/apache/logging-log4j2/assets/78874860/495e3de7-bf28-46ff-99ab-762835498fc8)
![image](https://github.com/apache/logging-log4j2/assets/78874860/ce912784-3481-4ec7-8e98-f934818e7288)

RapidTransit:
Version: 2.21.1
I had a similar problem, I don't recall what fixed it but here's a snippet of my config, I want to say maybe the property was coming up null?:
AFAIK I haven't tweaked my production config.
`    <Properties>
        <Property name="log4j2.disableJmx">true</Property>
        <Property name="log4j2.disable.jmx">true</Property>
        <Property name="log4j2.enableJndiLookup">false</Property>
        <Property name="filename">${env:PROXY_LOG_LOCATION:-/var/log/proxy}</Property>
    </Properties>

    <Select>
    <SystemPropertyArbiter propertyName="log.environment" propertyValue="production">

        <Appenders>

            <RollingFile name="GenericRollingFile" fileName="${filename}/application.log" filePattern="logs/application-%d{MM-dd-yyyy}-%i.log.gz">
`

paulonicolau-rj:
I can´t see, but in my case, the problem is when i use a customlevels. If I remove the tag CustomLevels, all works fine.

ppkarwasz:
@paulonicolau-rj,

Can you add your full configuration as text? Since all other nodes are interpolated using the properties that you provide, the `<Properties>` element must be the **first** child of `<Configuration>`.

paulonicolau-rj:
> @paulonicolau-rj,
> 
> Can you add your full configuration as text? Since all other nodes are interpolated using the properties that you provide, the `<Properties>` element must be the **first** child of `<Configuration>`.

Yes, follow:
[...]

PR #2278, opened 2024-02-09
Allow arbitrary position of `<Properties>` element

ppkarwasz:
Until now the `<Properties>` element had to be the first child of `<Configuration>`. In the current architecture this restriction is no longer necessary and can be lifted.

This is related to #2274.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationPropertiesOrderTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationPropertiesOrderTest.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.config;
+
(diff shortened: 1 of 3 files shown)


log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationPropertiesOrderTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.apache.logging.log4j.core.Appender;
+import org.apache.logging.log4j.core.appender.ConsoleAppender;
+import org.junit.jupiter.api.Test;
+
+public class ConfigurationPropertiesOrderTest {
+
+    @Test
+    void propertiesCanComeLast() {
+        final Configuration config = new AbstractConfiguration(null, ConfigurationSource.NULL_SOURCE) {
```
vy, line 29: Should we eat our own dog food and use `ConfigurationBuilderFactory.newConfigurationBuilder()` instead?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -670,7 +673,7 @@ protected void doConfigure() {
         boolean setLoggers = false;
         boolean setRoot = false;
         for (final Node child : rootNode.getChildren()) {
-            if (child.getName().equalsIgnoreCase("Properties")) {
+            if ("Properties".equalsIgnoreCase(child.getName())) {
                 if (tempLookup == runtimeStrSubstitutor.getVariableResolver()) {
                     LOGGER.error("Properties declaration must be the first element in the configuration");
                 }
```
vy, line 679: Does this block serve any purpose still?

vy (approved):
Sweet! :candy:

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
ppkarwasz: Done

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationPropertiesOrderTest.java
ppkarwasz: I prefer to test `AbstractConfiguration` directly. Besides `BuiltConfiguration` reorders the elements so that properties come first, so it wouldn't be much of a test.

Issue #2281, opened 2024-02-12
The `core.Logger#setLevel` method should work like `Configurator#setLevel`


ppkarwasz:
The main difference between the [`core.Logger#setLevel`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/Logger.html#setLevel(org.apache.logging.log4j.Level)) method and [`Configurator#setLevel`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/config/Configurator.html#setLevel(org.apache.logging.log4j.Logger,org.apache.logging.log4j.Level)) is that the former changes only the level of a specific logger, while the latter also changes the level of child loggers.

As documented in its Javadoc, the `core.Logger#setLevel` method is not part of the public API (so it can do whatever it wants). However it is used in `log4j-1.2-api` and `log4j-jul` instead of the more correct `Configurator#setLevel` method and it causes those modules to misbehave.

We can therefore:

- either change the way `Logger#setLevel` works to match the behavior of `Configurator#setLevel` (IMHO it is the best option, since it prevents mistakes from committers and users),
- or simply replace `Logger#setLevel` with `Configurator#setLevel` in `log4j-1.2-api` and `log4j-jul`.

garydgregory:
I bitten by this issue in Apache Commons Logging

vy:
> The main difference between the [`core.Logger#setLevel`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/Logger.html#setLevel(org.apache.logging.log4j.Level)) method and [`Configurator#setLevel`](https://logging.apache.org/log4j/2.x/javadoc/log4j-
> We can therefore:
> 
> 1. either change the way `Logger#setLevel` works to match the behavior of `Configurator#setLevel` (IMHO it is the best option, since it prevents mistakes from committers and users),
> 2. or simply replace `Logger#setLevel` with `Configurator#setLevel` in `log4j-1.2-api` and `log4j-jul`.

Thanks for investigation, @ppkarwasz. I would like to add a 3rd option to the list above:

3. do both (would be of my preference)

garydgregory:
TY @ppkarwasz for helping out with https://github.com/apache/logging-log4j2/pull/2282, merged! 👍

ppkarwasz:
Let's leave this issue open, until we check `log4j-jul`.

[...]

PR #2282, opened 2024-02-13
The core.Logger#setLevel method should work like

garydgregory:
The core.Logger#setLevel method should work like Configurator#setLevel #2281

@ppkarwasz 

I'm not sure how to really test this from our 1.2 module but I did install a snapshot locally and tested it with my Apache Commons Logging branch here https://github.com/garydgregory/commons-logging/tree/log4j1-log42-api

The POM change to apache-rat-plugin was required for `mvn clean verify` to pass locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -542,6 +542,7 @@
       <plugin>
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
+        <version>0.16.1</version>
         <configuration>
           <consoleOutput>true</consoleOutput>
           <excludes combine.children="append">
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -542,6 +542,7 @@
       <plugin>
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
+        <version>0.16.1</version>
```
ppkarwasz, line 545: We should bump this in `logging-parent` if necessary or even better in the ASF parent: apache/maven-apache-parent#193

ppkarwasz (commented):
Can we simply test that setting a new level on a "foo" logger changes the effective level on a "foo.bar" logger?

ppkarwasz:
BTW: you should also add a file in `src/changelog/.2.x.x`.

garydgregory:
I added tests but they fail. I can't tell if the issue is in main or the test itself...

PR #2288, opened 2024-02-14
Register a single reference counted StatusConsoleListener

ppkarwasz:
This adds a single `StatusConsoleListener` that is registered by the first Log4j Core configuration active and removed by the last one.

Compare to the current status this:

1. Prevents users from messing with the stream of the fallback `StatusConsoleListener`. Right now if they set it to something different from `System.out/System.err`, they need to remember to close it at shutdown,
2. Log4j Core configurations are still modifying a **shared** `StatusConsoleListener`, but they don't touch the fallback listener. The fallback listener is disabled when the first configuration starts and enabled again, when the last one stops.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/BuiltConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/BuiltConfiguration.java
@@ -38,7 +38,6 @@ import org.apache.logging.log4j.core.util.Patterns;
  * @since 2.4
  */
 public class BuiltConfiguration extends AbstractConfiguration {
-    private final StatusConfiguration statusConfig;
     protected Component rootComponent;
     private Component loggersComponent;
     private Component appendersComponent;
@@ -51,7 +50,6 @@ public class BuiltConfiguration extends AbstractConfiguration {
     public BuiltConfiguration(
             final LoggerContext loggerContext, final ConfigurationSource source, final Component rootComponent) {
         super(loggerContext, source);
-        statusConfig = new StatusConfiguration().withStatus(getDefaultStatus());
         for (final Component component : rootComponent.getComponents()) {
             switch (component.getPluginType()) {
                 case "Scripts": {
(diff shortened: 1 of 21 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/BuiltConfiguration.java
```diff
@@ -131,8 +129,9 @@ public void createAdvertiser(final String advertiserString, final ConfigurationS
         super.createAdvertiser(advertiserString, configSource, buffer, contentType);
     }
 
+    @Override
     public StatusConfiguration getStatusConfiguration() {
-        return statusConfig;
+        return super.getStatusConfiguration();
     }
```
vy, line 135: Can't we simply remove this method definition here? Dynamic dispatching will call `super` anyway.

vy (commented):
@ppkarwasz, I am -0 on this one, that is, I don't approve it, though I am not standing in the way either. My reasoning is as follows:

1. [`StatusConfiguration` is planned to be removed in `3.x`.](https://lists.apache.org/thread/3sbk3rxgp522w97l81snlz3n4tmd17xt) Any effort put into this is not only a waste of time, but a maintenance burden; the changes increase the complexity.
2. Since its incarnation, nobody communicated a single complaint about the issue of non-system `StatusConsoleListener` streams not getting closed. Given diff stats indicate `+141 -63`, I don't see the pressure to increase the complexity for an unsolicited fix.

ppkarwasz:
I think we can close this.

In 3.x we can find a proper way to allow different status loggers in different logger contexts.

vy:
@ppkarwasz, I am not feeling good about being the reason to drop your PR. Hence, sorry for that and thanks for your understanding.

> In 3.x we can find a proper way to allow different status loggers in different logger contexts.

Please, don't, Piotr. [*Even Ralph* finds this an overkill.](https://lists.apache.org/thread/t13f6ljl07ozwzvy5lxvh65wdx6jvgdb) `StatusLogger` is `printf`-on-steroids. Anything more than this is a logging framework and we already have one: Log4j. I think `StatusLogger` should not be subject to any enhancement discussions, e.g., what else we can improve it with. Rather, it should be subject to discussions debating what else we can remove from it.

PR #2289, opened 2024-02-15
[Issue2281]  The core.Logger#setLevel method should work like Configurator#setLevel()

garydgregory:
Do for this module what we did for `log4j-1.2-api`.


garydgregory:
Note that the new test fails... @ppkarwasz

ppkarwasz:
Are we testing the right thing? Unless I am mistaken JUL's [`Logger#getLevel`](https://docs.oracle.com/en/java/javase/17/docs/api/java.logging/java/util/logging/Logger.html#getLevel()) returns the **explicitly** configured level of the logger. A call to `Logger#setLevel` should not change the **explicit** level of any other logger, but it should change the **effective** level of all its children.

Unfortunately the effective level can be only be guessed by calling [`Logger#isLoggable`](https://docs.oracle.com/en/java/javase/17/docs/api/java.logging/java/util/logging/Logger.html#isLoggable(java.util.logging.Level)) multiple times.

I think that our `CoreLogger` implementation has an additional bug: it does not overwrite `Logger#getLevel`. So if a user changes the level of a logger through Log4j Core directly, `CoreLogger` will not be able to detect it.

garydgregory:
Ah, good point on the tests, I update the tests to use `isLoggable()` and now they pass.

log4j-jul/src/test/java/org/apache/logging/log4j/jul/test/CoreLoggerTest.java
```diff
@@ -100,6 +103,35 @@ public void testSetLevel() throws Exception {
...
+            Level.INFO,
+            Level.OFF,
+            Level.SEVERE,
+            Level.WARNING
+        }) {
+            a.setLevel(level);
+            assertEquals(level, a.getLevel());
+            assertTrue(a.isLoggable(level));
+            assertTrue(a_b.isLoggable(level));
+            assertTrue(a_b_c.isLoggable(level));
```
ppkarwasz, line 131: The fact that `level` is enabled does not mean that the effective level is `level`. What about a helper functions that return the first of `ALL`,  `FINEST`, `FINER` up to `OFF` that is loggable?

I am also a little bit concerned about the initial conditions: if the only explicitly configured logger is the root logger, shouldn't all `getLevel()` calls return `null` instead of `INFO`?

garydgregory: > The fact that level is enabled does not mean that the effective level is level. 

According to the Javadoc, JUL returns null when the level is not set and the effective level is the parent's level, and on up the chain presumably.

> What about a helper functions that return the first of ALL, FINEST, FINER up to OFF that is loggable?

Not useful IMO here since this test goes over all values, and passes, not that this means it's correct ;-) So we should figure this out first before adding more stuff ;-)

> I am also a little bit concerned about the initial conditions: if the only explicitly configured logger is the root logger, shouldn't all getLevel() calls return null instead of INFO?

The new test does not use the root logger though.

garydgregory: Well, at least the build is green ;-)

ppkarwasz (approved):
LGTM

garydgregory:
> LGTM

@ppkarwasz 
TY for your help; merged!

Issue #2290, opened 2024-02-15
[MS11] Log4j API generalization #2


ppkarwasz:
Following the conclusion of #2273, we continue our task of removing version 3.x of `log4j-api` and replace it with version 2.x.

This task requires changes in both the `2.x` and `main` branches.

## Changes to Log4j API 2.x

Since `log4j-api` is not a pure API artifact and contains a lot of utility classes, we need to create extension points that will allow Log4j Core 3.x to replace the utility classes from `log4j-api` that require breaking changes.

All the functionality accessible from the `org.apache.logging.log4j` package (the public API) is provided by five services:

* [`LoggerContextFactory`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/LoggerContextFactory.html) provides the functionality of the `LogManager` static methods,
* [`ThreadContextMap`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/ThreadContextMap.html) and [`ThreadContextStack`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/ThreadContextStack.html) provide the functionality of the `ThreadContext` static methods,
* [`MarkerManager`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/MarkerManager.html) is a non-customizable service that handles the creation of `Marker`s,
* [`Level`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/Level.html) is a non-customizable service that provides support for custom log levels.

Of these five services, only `LoggerContextFactory` and `ThreadContextMap` **need** to be provided by a logging implementation.

While in Log4j Core 2.x these services can be created independently and access each other through the `LogManager/ThreadContext` static methods, in Log4j Core 3.x they are linked together by a common dependency injection context and need to be created by a common factory.

Therefore we need to:
- [x] #2374, which is the common factory that creates the `LoggerContextFactory` and `ThreadContextMap`,
- [x] #2392

## Changes to Log4j Core 3.x

The lion's share of this task concentrates on Log4j Core 3.x, which must incorporate all the enhancements that are currently part of Log4j API 3.x. Currently the main differences between Log4j API 2.x and 3.x are:

* the `Recycler` API introduced in #1969,
[...]

PR #2291, opened 2024-02-15
Revamp `AbstractLogger` in `log4j-sdk`

ppkarwasz:
In order for Log4j Core 3.x not to depend on the utility classes in Log4j API 2.x, we provide a new `AbstractLogger` in `log4j-sdk`.

The main improvements to this `AbstractLogger` are:

- The methods are grouped by functionality: public API for users, location-aware method for system integrators, "create `Message` and log" methods and filter methods. The most important methods are at the **beginning**.
- All the dependencies of the class (message factory, flow message factory, recycler factory) must be explicitly provided. Only the log builder implementation is hardcoded.
- The `logMessageSafely` message has an additional `StackTraceElement` parameter and **all** the logging methods are routed through it,
- All the methods are tested for their bytecode length so that most of them are below 35 bytes. The exceptions are listed in `AbstractLoggerTest`. It is basically impossible for **unrolled** vararg methods with more than 2 parameters and methods that use supplier to follow a 100% inline-able path.
- The class is annotated with JSpecify.
- The class has **only 2** unimplemented methods: `doLog` (formerly `log`) and `isEnabled(Level, Marker)`.

Part of #2290


garydgregory:
"SDK" implies to me something bigger, not an API, while "SPI" implies something like a template to be implemented or classes to be extended.

ppkarwasz:
> "SDK" implies to me something big, not an API, while "SPI" implies something like a template to be implemented or classes to be extended.

I don't like calling things SDK either, but I like a simple rule to determine the main package name from the artifact name. `o.a.l.l.spi` is already taken, what about `log4j-spi3` and `o.a.l.l.spi3`?

vy:
Please no enumerated naming, e.g., `log4j-spi3`.

I am fine with `log4j-sdk`. Other alternatives I can think of:
* `log4j-runtime`
* `log4j-backend`
* `log4j-kit`
* `log4j-base`

log4j-sdk/src/main/java/org/apache/logging/log4j/sdk/logger/AbstractLogger.java
```diff
@@ -0,0 +1,3935 @@
...
+            final @Nullable Marker marker,
+            final @Nullable Message message,
+            final @Nullable Throwable throwable) {
+        try {
+            doLog(fqcn, getLocation(fqcn, location), level, marker, message, throwable);
+        } catch (final Throwable t) {
+            // LOG4J2-1990 Log4j2 suppresses all exceptions that occur once application called the logger
+            handleLogMessageException(t, fqcn, message);
+        }
+    }
```
ppkarwasz, line 266: This call to `getLocation` is IMHO precocious:

- if `requiresLocation()` returns true, it computes the location of each log message,
- however `requiresLocation()` does not evaluate the filters that a user configured down the line (in `LoggerConfig` or `Appender` for instance), so we could compute location that is not logged.

A better solution would be to move the call further down the logging pipeline before a thread jump or after an appender's filter has been evaluated.

The bonus of such a change would be that if I remove the `getLocation(fqcn, location` call I can introduce a:

```java
private @Nullable Throwable getThrowable(message, throwable) {
    return throwable != null ? throwable : message != null : message.getThrowable() : null;
}
```
method call and still maintain the 35 bytes bytecode limit. This will simplify the logic of many methods below.

ppkarwasz:
@garydgregory,

What name would you propose?

Regardless of the naming of the artifact  can you review the [AbstractLogger](https://github.com/apache/logging-log4j2/blob/2dc91155688fbb49bdbc2ef08a957cf8a28cbd89/log4j-sdk/src/main/java/org/apache/logging/log4j/sdk/logger/AbstractLogger.java) class? There are some new methods like `logEnterMessage`, `logCatchingMessage` and so on. The most important ones should be at the top.

Since this PR targets the feature branch `feature/log4j-sdk` and not `main`, we can solve the naming problem just before merging the feature branch into `main`.

log4j-sdk/src/main/java/org/apache/logging/log4j/sdk/logger/AbstractLogger.java
vy: 1. Does moving `getLocation()` further down in the call stack cause any behavioural change? As a start: would all tests still pass? (Tests do not always represent the precise behaviour, but they generally are pretty close to that.)
2. Can't we store the location capture in a continuation and make it available only when actually used?

log4j-sdk/src/test/java/org/apache/logging/log4j/sdk/logger/AbstractLoggerTest.java
```diff
@@ -0,0 +1,71 @@
...
+            // logging methods with Supplier, MessageSupplier or more than 2 parameters
+            "logMessage(Ljava/lang/String;Lorg/apache/logging/log4j/Level;Lorg/apache/logging/log4j/Marker;Lorg/apache/logging/log4j/util/MessageSupplier;Ljava/lang/Throwable;)V",
+            "logMessage(Ljava/lang/String;Lorg/apache/logging/log4j/Level;Lorg/apache/logging/log4j/Marker;Lorg/apache/logging/log4j/util/Supplier;Ljava/lang/Throwable;)V",
+            "logMessage(Ljava/lang/String;Lorg/apache/logging/log4j/Level;Lorg/apache/logging/log4j/Marker;Ljava/lang/String;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;",
+            "logMessage(Ljava/lang/String;Lorg/apache/logging/log4j/Level;Lorg/apache/logging/log4j/Marker;Ljava/lang/String;[Lorg/apache/logging/log4j/util/Supplier;)V",
+            // logging methods with more than 3 parameters
+            "logIfEnabled(Ljava/lang/String;Lorg/apache/logging/log4j/Level;Lorg/apache/logging/log4j/Marker;Ljava/lang/String;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;");
+
+    @Test
+    void does_not_exceed_MaxInlineSize() throws Exception {
```
vy, line 48: This is a really sweet test @ppkarwasz! :100: Can't we introduce an _internal_ `@Inlined` annotation to `log4j-api`, create a test utility for it in `log4j-api-test` and use it for every method that is supposed to be inlined?

vy (approved):
This change is massive; maybe not in size, but its implications. I am not able to see anything fishy. But I guess we will iron out most wrinkles along the way.

log4j-sdk/src/main/java/org/apache/logging/log4j/sdk/logger/AbstractLogger.java
ppkarwasz: 1. We can delegate the burden of computing the location to `LogEvent.getSource()` (actually most implementation already do it) and **document** it in the Javadoc. This way we can treat location the same way we treat `Message.getFormattedMessage()`: we call it just before a thread jump, otherwise we let the layouts do it. I prefer to do these changes in a separate PR.
2. The way it currently works the `location` parameter is passed all the way down to the `LogEvent`. Whether the location appears in the log or not follows the following logic:

- if `location` is not null (e.g. it was provided by the caller or Log4j Transform) it will be available,
- if `!isIncludeLocation() && requiresLocation()` it is computed,
- otherwise it will be unavailable.

log4j-sdk/src/test/java/org/apache/logging/log4j/sdk/logger/AbstractLoggerTest.java
ppkarwasz: There is already a `@PerformanceSensitive` annotation. We might have a single test that scans the classes for it and checks all the methods where the value is "inline" or `"maxInlineSize".

I can add it to my TODO list.

ppkarwasz:
@vy and @garydgregory,

I renamed the module to `log4j-kit`. Is it OK with you?

log4j-sdk/src/main/java/org/apache/logging/log4j/sdk/logger/AbstractLogger.java
ppkarwasz: The actual implementation was a little bit different, see #2329.

Issue #2292, opened 2024-02-15
Performance optimization in CloseableThreadContext.closeMap()


jengebr:
## Description

Excess work is performed in CloseableThreadContext.closeMap() that shows up as a production hotspot in our client-facing, latency-sensitive application.  Specifically, the use of multiple calls to ThreadContext.remove() and ThreadContext.put() causes multiple copy-and-modify operations in the underlying ThreadContextMap.  The attached java files compare the performance of the original code vs. a modification that groups keys by remove vs. replace, and makes zero or one calls to ThreadContext.remove() and ThreadContext.put().

On my system, tests with thread locals disabled (`Constants.ENABLE_THREADLOCALS == false`) show a 40% improvement:

```
Done with old in 521
Done with old in 535
Done with old in 537
Done with old in 516
Done with old in 516
Done with new in 284
Done with new in 285
Done with new in 284
Done with new in 301
Done with new in 292
```

With thread locals enabled (`Constants.ENABLE_THREADLOCALS == true`) show a 20% improvement:

```
Done with old in 386
Done with old in 388
Done with old in 386
Done with old in 389
Done with old in 408
Done with new in 308
Done with new in 309
Done with new in 307
Done with new in 307
Done with new in 313
```

The impact of the thread locals flag is because of the different implementations of ThreadContextMap.  However, both cases demonstrate substantial improvement.

## Configuration

**Version:** 2.20

**Operating system:** Affects all - tested on Linux and Windows

**JDK:** Tested on OpenJDK 8 and OpenJDK 17

## Logs
[CloseableThreadContextSpeedTest.java.txt](https://github.com/apache/logging-log4j2/files/14299669/CloseableThreadContextSpeedTest.java.txt)

## Reproduction

See attached file.

ppkarwasz:
@jengebr, can you submit a PR?

jengebr:
As requested: https://github.com/apache/logging-log4j2/pull/2296 

Hope I got the details right....

vy:
Fixed in #2296.

PR #2296, opened 2024-02-16
Performance optimization in CloseableThreadContext.closeMap() (#2292)

jengebr:
This substantially improves the performance of  CloseableThreadContext.closeMap(), reducing the aggregate create/close cost by 40% or 20%, depending on the ThreadContext implementation in use.  Further performance explanation is on the related issue, 
 https://github.com/apache/logging-log4j2/issues/2292

## Checklist


vy (approved):
@jengebr, thanks so much! Would you mind adding a changelog entry (i.e., `src/changelog/.2.x.x/improve_CloseableThreadContext_closeMap.xml`) too, please?

ppkarwasz:
@jengebr,

Personally I think there is an inherent problem in the way `CloseableThreadContext.Instance` works: it stores the **difference** between the context data available when it was created and the current context data.

Since you pointed out that `Map` or [`StringMap`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/util/StringMap.html) (a replacement of Java `Map` with garbage-free iteration) allocation is a performance bottleneck, maybe we could change the logic of `CloseableThreadContext.Instance` to:

- take a snapshot of the current context data in its constructor,
- restore the snapshot in `Instance#close()`.

This might require API changes, that I am more than willing to accept, but it should give better results: closing `Instance` would be simple and garbage-free.

Unfortunately there are **3** implementations of [`ThreadContextMap`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/ThreadContextMap.html):

- [`DefaultThreadContextMap`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/DefaultThreadContextMap.html) keeps the data in an **immutable** `Map`,
- [`CopyOnWriteSortedArrayThreadContextMap`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/CopyOnWriteSortedArrayThreadContextMap.java) keeps the data in an **immutable** `StringMap`,
- [`GarbageFreeSortedArrayThreadContextMap`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/GarbageFreeSortedArrayThreadContextMap.java) keeps the data in a **mutable** `StringMap`.

In order to work with all 3, `Instance` would at least need to store the appropriate native format of the context data, to prevent conversions from `Map` to `StringMap` and viceversa.

jengebr:
@vy thanks for adding the finals!  I've added the changelog entry under 2.24.0, which matches the milestone, hopefully that's correct.

@ppkarwasz I like the garbage-free proposal but I don't think I'm up for that - not enough familiarity with the log4j codebase or designs.  I am however able to performance test and profile such changes in a real-life, heavy-load environment, and am happy to help in that way.

vy:
@jengebr, could you move the changelog file to `src/changelog/.2.x.x/improve_CloseableThreadContext_closeMap.xml`, please? (`.2.x.x` is literally the folder name, not a placeholder. It denotes the changes that will be published with the next release. `2.24.0` folder will be created and populated during a release.)

jengebr:
@vy done, I think - and sorry, my IDE was hiding that folder and I misunderstood the intent.  I kept the change you made (thanks for that!).

jengebr:
@ppkarwasz - fyi, our conversation above sparked an idea for improving DefaultThreadContextMap.  I opened #2319 to discuss further, including an implementation and benchmark.

PR #2307, opened 2024-02-21
Add `PropertyEnvironment` to `log4j-sdk`

ppkarwasz:
This PR creates a simplified version of `PropertyEnvironment` in `log4j-sdk` that supports Spring-like property classes.

It provides aggregation features to allow the conversion of multiple properties at once in a typesafe way.

This is based on #2291 and part of #2290.

The main functionality of `PropertyEnvironment` is to replace a set of properties with keys "Configuration.file", "Configuration.factory", "Configuration.timeout" into a Java record:

```java
@Log4jProperty
public record Configuration(String file, Class<? extends ConfigurationFactory> factory, long timeout)
```


log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.kit.env.ConfigurablePropertyEnvironment;
+import org.apache.logging.log4j.kit.env.PropertyEnvironment;
+import org.apache.logging.log4j.kit.env.PropertySource;
+import org.apache.logging.log4j.kit.env.internal.CopyOnWriteNavigableSet;
+import org.jspecify.annotations.Nullable;
+
+public class PropertySourcePropertyEnvironment extends ClassloaderPropertyEnvironment
```
vy, line 32: Piotr, `PropertySourcePropertyEnvironment`... just reading the class name is making my mind whirl. I get totally confused of what is a `PropertySource`, what was a `PropertyEnvironment`, etc. In essence, this is a `ClassLoaderPropertyEnvironment` with some sources. Can't we say `SourcedPropertyEnvironment` instead, please?

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
```diff
@@ -0,0 +1,94 @@
...
+        implements ConfigurablePropertyEnvironment {
+
+    private final Collection<PropertySource> sources =
+            new CopyOnWriteNavigableSet<>(Comparator.comparing(PropertySource::getPriority));
+    private final Map<String, Optional<String>> stringCache = new ConcurrentHashMap<>();
+    private final Map<Class<?>, Object> classCache = new ConcurrentHashMap<>();
```
vy, line 38: Don't we +1 the problems by introducing caching at this early stage. Is it really necessary?

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
vy, line 32: Does this class need to be thread-safe?

log4j-kit/src/test/java/org/apache/logging/log4j/kit/env/support/BasicPropertyEnvironmentTest.java
```diff
@@ -0,0 +1,226 @@
...
+import org.jspecify.annotations.Nullable;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+/**
+ * Tests the property values that are used as properties.
+ */
+class BasicPropertyEnvironmentTest {
```
vy, line 45: This class is plain beauty!

log4j-kit/src/test/java/org/apache/logging/log4j/kit/env/support/BasicPropertyEnvironmentTest.java
```diff
@@ -0,0 +1,226 @@
...
+
+    @Log4jProperty
+    record BasicValues(boolean boolAttr, int intAttr, long longAttr) {}
```
vy, line 48: Spring's `@ConfigurationProperties` support nesting. I don't see any such examples here. What is our stance on nesting?

log4j-kit/src/test/java/org/apache/logging/log4j/kit/logger/TestListLogger.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.logging.log4j.message.Message;
+import org.apache.logging.log4j.message.MessageFactory;
+import org.apache.logging.log4j.message.ParameterizedNoReferenceMessageFactory;
+import org.apache.logging.log4j.spi.recycler.RecyclerFactory;
+import org.apache.logging.log4j.status.StatusLogger;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+@NullMarked
+public class TestListLogger extends AbstractLogger {
```
vy, line 36: Even though, I liked the idea, I am not really sure if this is needed. You pass `TestListLogger` to `BasicPropertyEnvironment(Logger statusLogger)` ctor. Why don't you use a `new StatusLogger()` and leverage its message buffer instead?

vy (commented):
I think this is a really good start! There are certain things I don't agree with, though I really liked the direction where this is going. :100:

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/PropertySource.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Provides the priority of the property source.
+     * <p>
+     *     Property sources are ordered according to the natural ordering of their priority. Sources with lower
+     *     numerical value take precedence over those with higher numerical value.
+     * </p>
+     *
+     * @return priority value
+     */
+    int getPriority();
```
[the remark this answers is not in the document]
ppkarwasz: There is no coherence here in Log4j or other Java projects:

- Spring [`@Order`](https://www.baeldung.com/spring-order), Log4j's old [`PropertySource`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/util/PropertySource.html#getPriority()) and [`Provider`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/spi/Provider.html#getPriority()) use a **lowest-first** ordering,
- Log4j's [`@Order`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/config/Order.html) annotation uses a **highest-first** ordering.

log4j-kit/src/test/java/org/apache/logging/log4j/kit/env/support/BasicPropertyEnvironmentTest.java
ppkarwasz: There should be a test case later on.

In my preliminary rewrite of the current properties I do use nesting, although I am not sure if others agree (it breaks the original `<component>.<property>` assumption).

log4j-kit/src/test/java/org/apache/logging/log4j/kit/logger/TestListLogger.java
ppkarwasz: Good point.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/PropertySource.java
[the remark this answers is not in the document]
vy: I agree.

Note that `@Order` does not use the word `priority`. Hence, I don't think it is a valid example to compare with.

Regarding `PropertySource` and `Provider`, IMHO, they got it wrong. :stuck_out_tongue_closed_eyes: I will consult to native speakers in Slack...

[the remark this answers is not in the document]
rgoers: Well, a lot of this is my fault and I am a native speaker.  The problem starts with DEBUG, INFO, WARN, and ERROR.  Many people want to discuss them as one Level being "higher" or "lower" than another. However, the whether a Level is "higher" or "lower" differs based on whether you are viewing it for the develoepr writing code or from the configuration.  In the code, debug would typically be thought of low level logging. But when you set the log level to DEBUG in the configuration you are capturing MORE logging data, which could be perceived as being higher.  That is why I avoided that terminology completely when discussing levels and used less specific and more specific. 

Ordering has the same problem. When something has the highest value does that mean it is processed first or last? You could argue it either way. It also has to do with how defaults get handled. For example, if the item with the highest value is processed first what happens when you get a duplicate entry on an item with a lower value. It could either replace the previous item or be ignored by the previous item, depending on which item should take precedence.

In short, I don't think there is any kind of standard for this.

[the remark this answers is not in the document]
vy: @ppkarwasz, I don't want to this conversation to block your PR. You can proceed as you wish. Please share your conclusion here and resolve this conversation at your convenience.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/internal/CopyOnWriteNavigableSet.java
```diff
@@ -0,0 +1,269 @@
...
+import java.util.NavigableSet;
+import java.util.SortedSet;
+import java.util.Spliterator;
+import java.util.TreeSet;
+import java.util.function.Consumer;
+import java.util.function.IntFunction;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+public class CopyOnWriteNavigableSet<E> extends TreeSet<E> {
```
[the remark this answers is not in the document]
ppkarwasz: Removed.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/BasicPropertyEnvironment.java
```diff
@@ -0,0 +1,322 @@
...
+import org.apache.logging.log4j.kit.env.PropertyEnvironment;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * An implementation of {@link PropertyEnvironment} that only uses basic Java functions.
+ * <p>
+ * Conversion problems are logged using a status logger.
+ * </p>
+ */
+public abstract class BasicPropertyEnvironment implements PropertyEnvironment {
```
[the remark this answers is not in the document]
ppkarwasz: That is probably something we can decide later.

My initial idea was that frameworks like Spring Boot don't need to implement our `PropertySource` abstraction, but they can directly implement `PropertyEnvironment`. This class might be useful, other implementations not necessarily.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
ppkarwasz: I renamed it to `CompositePropertyEnvironment`.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/ClassloaderPropertyEnvironment.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.kit.env.support;
+
+import org.apache.logging.log4j.Logger;
+
+/**
+ * An environment implementation that uses a specific classloader to load classes.
+ */
+public abstract class ClassloaderPropertyEnvironment extends BasicPropertyEnvironment {
```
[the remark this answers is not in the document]
ppkarwasz: Done

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
ppkarwasz: I removed caching and the runtime modification of property sources: these sources should be available when the context is created.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/support/PropertySourcePropertyEnvironment.java
ppkarwasz: Thread-safety is only a concern if the list of sources is modified. I removed the feature.

log4j-kit/src/main/java/org/apache/logging/log4j/kit/env/internal/ContextEnvironmentPropertySource.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.logging.log4j.util.PropertySource;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * PropertySource backed by the current environment variables.
+ * <p>
+ *     Should haves a slightly lower priority than global environment variables.
+ * </p>
+ */
+public class ContextEnvironmentPropertySource implements PropertySource {
```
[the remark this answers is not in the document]
ppkarwasz: Done

log4j-kit/src/test/java/org/apache/logging/log4j/kit/logger/TestListLogger.java
ppkarwasz: I added a not to #2325.

Currently I can not use `StatusLogger`, because I am working against the `main` branch, where you modifications are not available.

Issue #2319, opened 2024-02-23
Performance improvements to DefaultThreadContextMap


jengebr:
`DefaultThreadContextMap` maintains non-blocking thread safety by guaranteeing immutability, and performing copy-and-modify operations when adding or removing.  The referenced pull request  improves performance while maintaining functionality by creating a custom data structure (`UnmodifiableArrayBackedMap`).  According to the attached JMH benchmark, it outperforms the current implementation in every way - except that search operations (get/containsKey) become linear-time rather than constant. This appears to be a good tradeoff, especially in the common cases when the map is small (< 15 entries).  I welcome discussion on this point.

Details: `DefaultThreadContextMap` currently implements the copy-and-modify by allocating a new `HashMap` (with a padded array), iterating through the existing `HashMap` (reading null buckets, and chasing nested `Node`s), then inserting each element into the new `HashMap` (`Nod`e allocation, hash/equals calculations). The new implementation reduces copying to a single array allocation and a single call to `System.arrayCopy(`.
 
The JMH benchmarks attached to this issue show that copying and adding one entry to the new map is 1.5x faster when the existing map is empty, and the delta increases steadily up to 5x faster when 75 items are already present.  Similarly, adding 5 entries at a time is 2x faster when the map is empty, and the delta climbs to 5x faster when the map contains 75 items.

Separate profiling suggests a reduction in memory allocation MB/s around 20%, mostly by eliminating the `Node` instances.

Other JMH highlights:
1. Iteration across the `entrySet()` improves by 8x to 20x.
2. `get()` times are faster with 0 or 1 items in the map, but slower after 5+ items are present.

JMH benchmark class: 
[MapSpeedBenchmark.java.txt](https://github.com/apache/logging-log4j2/files/14387210/MapSpeedBenchmark.java.txt)

JMH benchmark results:  
[MapSpeedBenchmark_results.txt](https://github.com/apache/logging-log4j2/files/14387226/MapSpeedBenchmark_results.txt)

jengebr:
Pull request: https://github.com/apache/logging-log4j2/pull/2321

jengebr:
Closing based on discussion on pull request.  No changes made.

PR #2321, opened 2024-02-23
Performance improvements to DefaultThreadContextMap

jengebr:
This implements a performance optimization described on issue #2319 and detailed discussion of the performance characteristics is present there, including a JMH benchmark w/ results.

Note that building change causes complaints about api versioning, and I doubt I got those flags correct.  I was able to build locally by updating the package info and pom to 2.24.0.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/spi/UnmodifiableArrayBackedMapTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/spi/UnmodifiableArrayBackedMapTest.java
@@ -1,0 +1,255 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.spi;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+
+public class UnmodifiableArrayBackedMapTest {
+    private static final int TEST_DATA_SIZE = 5;
+    private HashMap<String, String> getTestParameters() {
+        HashMap<String, String> params = new HashMap<>();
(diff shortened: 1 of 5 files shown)


ppkarwasz:
@jengebr,

Yes, bumping the package version to `2.24.0` is the right thing to do. We are using OSGi semantics for package versioning, so every change that **adds** some public API (like a new class) requires at least a minor version bump.

Unfortunately your `UnmodifiableArrayBackedMap` can not replace the usage of `HashMap` in `DefaultThreadContextMap`, because it will cause memory leaks in servlet environments. The `DefaultThreadContextMap` uses a thread local field to store the map. If this field contains an instance of a class from a web application classloader (like `UnmodifieableArrayBackedMap`) instead of a class from the JRE, the entire web application classloader will not be garbage collected upon shutdown.

This potential memory leak problem is the only reason we even keep `DefaultThreadContextMap`.

The performance improvements in your benchmark are however very promising. Can you check the performance of your `DefaultThreadContextMap` implementation against [`CopyOnWriteSortedArrayThreadContextMap`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/CopyOnWriteSortedArrayThreadContextMap.java)? You can force Log4j API to use the latter by adding:
```
-Dlog4j2.threadContextMap=org.apache.logging.log4j.spi.CopyOnWriteSortedArrayThreadContextMap
```
to your JVM parameters.

jengebr:
Ugh, the memory leak problem makes sense, and using a WeakReference to track the instance would potentially lose the state.  :(  Thank you for explaining.

I'll do the perf comparison you suggest.

jengebr:
Your point is valid, this is comparable to [CopyOnWriteSortedArrayThreadContextMap](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/CopyOnWriteSortedArrayThreadContextMap.java).  Closing.

jengebr:
@ppkarwasz  I had an idea: my map implementation was based on a single object (Object[]), which is a JVM class.  If we only put the Object[] into the ThreadLocal, we'd avoid the memory leak and could create custom Maps for not-quite-free, since all they do is wrap the array.  Could even turn the copy-and-modify operations into static methods and avoid allocating the map for internal operations.

A similar approach could be applied to the existing SortedArrayStringMap, although the extra internal flags (immutable, iterating, keys, size, threshold, values) would need to combined into a single Array, probably with reserved slots for the flags.

Thoughts?

ppkarwasz:
@jengebr,

Great idea, it should work. @jvz, do you confirm?

Can you make another implementation of `DefaultThreadContextMap` (called e.g. `WebappThreadContextMap` or `StringArrayThreadContextMap`) and submit it as PR? No need to modify `SortedArrayStringMap`: this is conceived for systems that have a single classloader or do not reload classloaders (like Spring Boot).

In 3.x we got rid of most `ThreadLocal`s thanks to the [`Recycler`](https://github.com/apache/logging-log4j2/blob/main/log4j-api/src/main/java/org/apache/logging/log4j/spi/recycler/Recycler.java) abstraction. For obvious reasons the `ThreadContextMap` is the only place we use `ThreadLocal`s directly.

**Remark**: the `StringMap`-based thread context maps can also be safely used in a Servlet environment. You just need to add Log4j API in the system classloader. Many servlet containers allow such a configuration. For Apache Tomcat I have developed some small extensions in [copernik-eu/log4j-plugins](https://github.com/copernik-eu/log4j-plugins) to ease such a configuration.

jvz:
Yeah I think that would work.

jengebr:
PR #2330 adds `StringArrayThreadContextMap` wrapping `UnmodifiableArrayBackedMap`.  The new version of `UnmodifiableArrayBackedMap` differs slightly but the JMH test shows that performance is comparable.

PR #2322, opened 2024-02-23
Fix `StatusLogger` time-zone issues and stack overflow

vy:
This PR improves `StatusLogger` such that
1. A new `log4j2.StatusLogger.DateFormatZone` configuration property is introduced. This sets the time-zone `StatusLogger` uses to format `java.time.Instant`. Without this formatting patterns accessing to time-zone-specific fields (e.g., year-of-era) cause failures. The value defaults to `ZoneId.systemDefault()`.
2. `StatusLogger` is improved to avoid stack overflow caused by `StatusLogger#logMessage()` failures.

These issues are originally reported by @lauredogit here: https://github.com/apache/logging-log4j2/pull/2249#issuecomment-1960973519

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
@@ -33,30 +33,16 @@ import org.junit.jupiter.params.provider.MethodSource;
 
 class StatusLoggerPropertiesUtilDoubleTest {
 
-    private static final String[] NOT_PREFIXED_MATCHING_PROPERTY_NAMES = new String[] {
-        // For system properties:
-        "StatusLogger.DateFormat",
-        "StatusLogger.dateFormat",
-        "StatusLogger.dateformat",
-        "StatusLogger.Dateformat",
-        "Status.Logger.Dateformat",
-        "Status.Logger.Date.Format",
-        "statusLoggerDateFormat",
-        "Status.Logger.Date.Format",
-        "status.logger.date.format",
-        // For environment variables:
-        "STATUSLOGGER_DATEFORMAT",
-        "STATUS_LOGGER_DATE_FORMAT"
+    private static final String[] MATCHING_PROPERTY_NAMES = new String[] {
+        // System properties for version range `[, 2.10)`
+        "log4j2.StatusLogger.DateFormat",
+        // System properties for version range `[2.10, 3)`
+        "log4j2.statusLoggerDateFormat",
+        // System properties for version range `[3,)`
+        "log4j2.StatusLogger.dateFormat",
+        // Environment variables
+        "LOG4J_STATUS_LOGGER_DATE_FORMAT"
     };
-
-    private static final String[] MATCHING_PROPERTY_NAMES = Stream.of(
-                    // For system properties:
-                    "log4j.", "log4j2.",
-                    // For environment variables:
-                    "LOG4J_", "LOG4J2_")
-            .flatMap(prefix -> Arrays.stream(NOT_PREFIXED_MATCHING_PROPERTY_NAMES)
-                    .map(notPrefixedPropertyName -> prefix + notPrefixedPropertyName))
-            .toArray(String[]::new);
 
     private static final String[] NOT_MATCHING_PROPERTY_NAMES =
             new String[] {"log4j2.StatusLogger$DateFormat", "log4j2.StàtusLögger.DateFormat"};
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
@@ -1,0 +1,126 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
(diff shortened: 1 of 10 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -158,11 +162,26 @@ public class StatusLogger extends AbstractLogger {
...
+    /**
+     * The name of the system property that can be configured with a time-zone ID (e.g., {@code Europe/Amsterdam}, {@code UTC+01:00}) that will be used while formatting the created {@link StatusData}.
+     * <p>
+     * When not provided, {@link ZoneId#systemDefault()} will be used.
+     * </p>
+     *
+     * @see #STATUS_DATE_FORMAT
+     * @since 2.23.1
+     */
+    static final String STATUS_DATE_FORMAT_ZONE = "log4j2.StatusLogger.DateFormatZone";
```
ppkarwasz, line 183: In `3.x` the format of all properties will be `Component.propertyInCamelCase`, so I would already use `StatusLogger.dateFormatZone` here.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -251,7 +276,13 @@ private static Level readFallbackListenerLevel(final Properties fileProvidedProp
 
         private static DateTimeFormatter readInstantFormatter(final Properties fileProvidedProperties) {
             final String format = readProperty(fileProvidedProperties, STATUS_DATE_FORMAT);
-            return format != null ? DateTimeFormatter.ofPattern(format) : null;
+            if (format == null) {
+                return null;
+            }
+            final DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format);
+            final String zoneId = readProperty(fileProvidedProperties, STATUS_DATE_FORMAT_ZONE);
+            final ZoneId effectiveZoneId = zoneId != null ? ZoneId.of(zoneId) : ZoneId.systemDefault();
```
ppkarwasz, line 284: What happens if `ZoneId.of` throws an exception?

ppkarwasz (approved):
LGTM

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
```diff
@@ -158,11 +162,26 @@ public class StatusLogger extends AbstractLogger {
...
     /**
      * The name of the system property that can be configured with a {@link java.time.format.DateTimeFormatter} pattern that will be used while formatting the created {@link StatusData}.
+     * <p>
+     * When not provided, {@link Instant#toString()} will be used.
+     * </p>
      *
+     * @see #STATUS_DATE_FORMAT_ZONE
      * @since 2.11.0
      */
     public static final String STATUS_DATE_FORMAT = "log4j2.StatusLogger.DateFormat";
```
rgoers, line 173: Piotr's comment below also applies to this property name.

rgoers (approved):
The property name change is pretty minor. Everything else looks good.

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: I've pushed a fix for this. Could you check it again, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: I've extensively improved such checks along with tests. Could you check it again, please?

log4j-api/src/main/java/org/apache/logging/log4j/status/StatusLogger.java
vy: I've pushed a fix for this. Could you check it again, please?

log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
```diff
@@ -0,0 +1,126 @@
...
+        "StatusLogger.Dateformat",
+        "Status.Logger.Dateformat",
+        "Status.Logger.Date.Format",
+        "statusLoggerDateFormat",
+        "Status.Logger.Date.Format",
+        "status.logger.date.format",
+        // For environment variables:
+        "STATUSLOGGER_DATEFORMAT",
+        "STATUS_LOGGER_DATE_FORMAT"
+    };
```
ppkarwasz, line 50: I would relax the test to only check these cases:

- the legacy (pre-2.10) values `log4j2.StatusLogger.level` and `log4j2.StatusLogger.DateFormat`,
- the 2.10 system properties values `log4j2.statusLoggerLevel` and `log4j2.statusLoggerDateFormat`,
- the 2.10 environment variables values `LOG4J_STATUS_LOGGER_LEVEL` and `LOG4J_STATUS_LOGGER_DATE_FORMAT`,
- future (3.0) values `log4j.StatusLogger.level` and `log4j.StatusLogger.dateFormat`.

I believe we should test the externally mandated behavior, in such a way that the test does not require any changes if the internal implementation changes.

Since we don't officially support `log4j2.status.logger.date.format`, we don't accept issue reports if it stops working and we should probably not test for it.

(Reminder to myself: I overspecified a lot of tests too)

ppkarwasz (approved):
LGTM

log4j-api-test/src/test/java/org/apache/logging/log4j/status/StatusLoggerPropertiesUtilDoubleTest.java
vy: Fixed in 9c52bfebd8e2292c1eb549f058ae16394de009f8.

Issue #1458, opened 2023-05-03
AsyncLogger does not compute location for some methods


ppkarwasz:
## Description

A logging call that is not filtered out will eventually call one of these methods:

 * [`AbstractLogger#log(Level, Marker, String, StackTraceElement, Message, Throwable)`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/spi/AbstractLogger.html#log-org.apache.logging.log4j.Level-org.apache.logging.log4j.Marker-java.lang.String-java.lang.StackTraceElement-org.apache.logging.log4j.message.Message-java.lang.Throwable-),
 * [`Logger#logMessage(Level, Marker, String, StackTraceElement, Message, Throwable)`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/Logger.html#logMessage-org.apache.logging.log4j.Level-org.apache.logging.log4j.Marker-java.lang.String-java.lang.StackTraceElement-org.apache.logging.log4j.message.Message-java.lang.Throwable-): effectively a `final` method, a recursion-safe version of the previous one,
 * [`ExtendedLogger#logMessage(String, Level, Marker, Message, Throwable)`](https://logging.apache.org/log4j/2.x/log4j-api/apidocs/org/apache/logging/log4j/spi/ExtendedLogger.html#logMessage-java.lang.String-org.apache.logging.log4j.Level-org.apache.logging.log4j.Marker-org.apache.logging.log4j.message.Message-java.lang.Throwable-).

`AsyncLogger` correctly computes the location, if the third method is called. If the first method is called with `location == null` no attempt to compute location occurs.

## Configuration

**Version:** 2.20.0

PR #2329, opened 2024-02-29
Make location computation and message formatting more lazy

ppkarwasz:
We move the computation of location and message formatting to the last method call executed before a message leaves the current thread.

This has several consequences:

 - the refactoring of location computations should solve #1458,
 - since `AbstractLogger` no longer needs to compute the location of the caller, many messages were refactored to fit in the 35 bytes of bytecode limit. Only `logIfEnabled` methods with 4 or more `Object` parameters are over the limit and `logMessage` methods with 8 or more `Object` parameters, `handleMessageException` and the constructors,
 - messages that are filtered out by e.g. the filter of `AsyncAppender` do not include location information.

Part of #2290.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
@@ -19,7 +19,7 @@ package org.apache.logging.log4j.core.impl;
 import java.util.Objects;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Marker;
-import org.apache.logging.log4j.ThreadContext;
+import org.apache.logging.log4j.ThreadContext.ContextStack;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.config.LoggerConfig;
 import org.apache.logging.log4j.core.time.Instant;
@@ -32,6 +32,7 @@ import org.apache.logging.log4j.util.InternalApi;
 import org.apache.logging.log4j.util.ReadOnlyStringMap;
 import org.apache.logging.log4j.util.StringMap;
 import org.apache.logging.log4j.util.Strings;
+import org.jspecify.annotations.Nullable;
 
 /**
  * Immutable copy of a LogEvent.
@@ -50,8 +51,8 @@ public class MementoLogEvent implements LogEvent {
     private boolean endOfBatch;
     private final Message message;
     private final ReadOnlyStringMap contextData;
-    private final ThreadContext.ContextStack contextStack;
-    private final StackTraceElement source;
+    private final ContextStack contextStack;
+    private final @Nullable StackTraceElement source;
     private final String threadName;
     private final long threadId;
     private final int threadPriority;
@@ -72,9 +73,9 @@ public class MementoLogEvent implements LogEvent {
         if (instant.getEpochMillisecond() == 0 && message instanceof TimestampMessage) {
             instant.initFromEpochMilli(((TimestampMessage) message).getTimestamp(), 0);
         }
-        contextData = memento(event.getContextData());
+        contextData = mementoOfContextData(event.getContextData());
         contextStack = event.getContextStack();
-        source = includeLocation ? event.getSource() : null;
+        source = includeLocation ? event.getSource() : event.peekSource();
         threadName = event.getThreadName();
         threadId = event.getThreadId();
         threadPriority = event.getThreadPriority();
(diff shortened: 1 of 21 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
```diff
@@ -72,50 +73,37 @@ public MementoLogEvent(final LogEvent event) {
...
 
-    @Override
-    public LogEvent toImmutable() {
-        return this;
+    private static Message mementoOfMessage(final LogEvent event) {
+        final Message message = event.getMessage();
+        if (message instanceof LoggerNameAwareMessage) {
+            ((LoggerNameAwareMessage) message).setLoggerName(event.getLoggerName());
+        }
+        return message instanceof final ReusableMessage reusable ? reusable.memento() : message;
```
vy, line 101: _[I know this code block is not actually changed.]_ When the message is not a reusable one, we are assuming it to be immutable. Is this a valid assumption? I would have actually expected a `memento()` (or `freeze()`?) method on the `Message` contract... (Maybe I am not able to see the implications of this.)

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
```diff
@@ -72,50 +73,37 @@ public MementoLogEvent(final LogEvent event) {
...
     }
 
-    @Override
-    public LogEvent toImmutable() {
-        return this;
+    private static Message mementoOfMessage(final LogEvent event) {
+        final Message message = event.getMessage();
+        if (message instanceof LoggerNameAwareMessage) {
+            ((LoggerNameAwareMessage) message).setLoggerName(event.getLoggerName());
+        }
```
vy, line 100: _[I know this code block is not actually changed.]_ I am not able to follow why this is the right place to do this. Shouldn't this already be done while the message gets populated?

vy (approved):
Great simplification @ppkarwasz! :100:

1. Since the `Message` is already populated with a `loggerFqcn` and that is enough to resolve the `location` in the `LogEvent`, we don't need to resolve the location in `AbstractLogger`, right?
1. Due to to `peekSource()`, when a message gets copied elsewhere, `peekSource()` can be used to initialize the `source`, and reuse the already populated value, if present. Otherwise, `getSource()` can calculate it, if requested. This is the crux of this locate-at-the-end implementation, right?

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
ppkarwasz: I started to look at these, but I got such a headache, that I decided to split the work in three parts:

- location computation issues (calling `LogEvent#getSource` should prepare a log event for a thread jump),
- thread data computation issues (I guess these should just be created, when the log event is created),
- context data computation issues (probably this should be computed just before the jump),
- message immutability issue (no message is immutable until `getFormattedMessage()` is called).

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/MementoLogEvent.java
ppkarwasz: Same as above: let's deal with it in another PR.

ppkarwasz:
>     1. Since the `Message` is already populated with a `loggerFqcn` and that is enough to resolve the `location` in the `LogEvent`, we don't need to resolve the location in `AbstractLogger`, right?

Right. Having the FQCN of the logger class allows us to retrieve the location at any time or never.

Previously the location was computed eagerly based on the result of `requiresLocation()`. This method can tell us if location information will be necessary, when the message will be logged. However it can not tell if the message will be logged at all.

>     2. Due to to `peekSource()`, when a message gets copied elsewhere, `peekSource()` can be used to initialize the `source`, and reuse the already populated value, if present. Otherwise, `getSource()` can calculate it, if requested. This is the crux of this locate-at-the-end implementation, right?

Partially, the original `getSource()` method was introduced when location was a binary choice: you could either compute it and log it or neither compute it nor log it. However the `LogBuilder#withLocation` method changed this: we can know the location without computing it.

If we just keep the `getSource()` method and the user sets `includeLocation == false` on a component (which means "do not **compute** location), we need to be extra careful while calling `getSource`:

```java
boolean oldIncludeLocation = event.isIncludeLocation();
event.setIncludeLocation(false);
event.getSource();
event.setIncludeLocation(oldIncludeLocation);
```

The `peekSource` is a way to make the process easier.

Issue #2337, opened 2024-03-01
The parameter `-Dlog4j2.debug` does not `take effect` in version `2.23.0`, and StatusLogger.debug will no longer print logs


panbingkun:
## Description
When we upgraded `log4j2` from version `2.22.1` to `2.23.0` in Spark, we found that logs that could have been printed before are no longer being printed.

## Configuration
```
# This log4j config file is for integration test SparkConfPropagateSuite.
rootLogger.level = debug
rootLogger.appenderRef.stdout.ref = console

appender.console.type = Console
appender.console.name = console
appender.console.target = SYSTEM_ERR
appender.console.layout.type = PatternLayout
appender.console.layout.pattern = %d{HH:mm:ss.SSS} %p %c: %maxLen{%m}{512}%n%ex{8}%n
```

**Version:** [Log4j version]
2.23.0

**Operating system:** [OS and version]
`Ubuntu 22.04.4 LTS` And `macos Sonama 14.3.1`.

**JDK:** [JDK distribution and version]
JDK 17

## Logs

The following are the logs printed in version `2.22.1`:
```
...

TRACE StatusLogger DefaultConfiguration cleaning Appenders from 1 LoggerConfigs.
DEBUG StatusLogger Stopped org.apache.logging.log4j.core.config.DefaultConfiguration@384ad17b OK
TRACE StatusLogger Reregistering MBeans after reconfigure. Selector=org.apache.logging.log4j.core.selector.ClassLoaderContextSelector@5852c06f
TRACE StatusLogger Reregistering context (1/1): '5ffd2b27' org.apache.logging.log4j.core.LoggerContext@31190526
TRACE StatusLogger Unregistering but no MBeans found matching 'org.apache.logging.log4j2:type=5ffd2b27'
TRACE StatusLogger Unregistering but no MBeans found matching 'org.apache.logging.log4j2:type=5ffd2b27,component=StatusLogger'
TRACE StatusLogger Unregistering but no MBeans found matching 'org.apache.logging.log4j2:type=5ffd2b27,component=ContextSelector'
TRACE StatusLogger Unregistering but no MBeans found matching 'org.apache.logging.log4j2:type=5ffd2b27,component=Loggers,name=*'
TRACE StatusLogger Unregistering but no MBeans found matching 'org.apache.logging.log4j2:type=5ffd2b27,component=Appenders,name=*'
[...]

PR #2343, opened 2024-03-04
Don't fail on insufficient parameters in `ParameterFormatter` (#2337)

vy:


log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
```diff
@@ -17,21 +17,27 @@
...
 import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.ArgumentCaptor;
 
 /**
  * Tests {@link ParameterFormatter}.
  */
-public class ParameterFormatterTest {
+@UsingStatusLoggerMock
+class ParameterFormatterTest {
```
ppkarwasz, line 40: I am not a big fan of mocking the entire `StatusLogger`: this prevents the test to run in parallel with other tests.

I prefer the [`@UsingStatusListener`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/UsingStatusListener.java) extension to register a per-test status listener.

Of course neither of us is impartial in the choice of JUnit extension to use. ;-)

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -236,7 +240,10 @@ static void formatMessage(
             final String message = String.format(
                     "found %d argument placeholders, but provided %d for pattern `%s`",
                     analysis.placeholderCount, args.length, pattern);
-            throw new IllegalArgumentException(message);
```
ppkarwasz, line 239: Personally I am pretty much in favor of throwing an exception here: [`java.util.Formatter`](https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html) does it, so we can also throw.

On the other hand we should probably catch the exception in `ParameterizedMessage#getFormattedMessage`, as it happens in other kinds of messages:

https://github.com/apache/logging-log4j2/blob/28b70d1d7fe63417e8eaa1abe8bf992363060515/log4j-api/src/main/java/org/apache/logging/log4j/message/StringFormattedMessage.java#L120-L126

https://github.com/apache/logging-log4j2/blob/28b70d1d7fe63417e8eaa1abe8bf992363060515/log4j-api/src/main/java/org/apache/logging/log4j/message/MessageFormatMessage.java#L113-L119

ppkarwasz (changes requested):
Personally I would maintain the current logic that throws exceptions, I would just catch them in the call sites.

I looked in Log4j 2.19.0 how formatting was done. Using the `StringFormatterMessageFactory` a call to `logger.info("Hello %s and %s!", "world")` gives:

```
2024-03-04 11:18:57,550 main ERROR Unable to format msg: Hello %s and %s!
java.util.MissingFormatArgumentException: Format specifier '%s'
Hello %s and %s!
```

whereas the default factories for a call to `logger.info("Hello {} and {}!", "world")` give just:
```
Hello world and {}!
```

I don't think the behavior of Log4j 2.19.0 is correct. I would expect rather:
```
2024-03-04 11:18:57,550 main ERROR Unable to format msg: Hello {} and {}!
java.lang.IllegalArgumentException: Found 2 argument placeholders, but only 1 was provided.
Hello {} and {}!
```

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
vy: I have updated it as you requested. Could you review the changes in 1084855467f554f2dabf34af4d860018fb098fd8, please?

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterizedMessage.java
```diff
@@ -274,7 +278,12 @@ public void formatTo(final StringBuilder buffer) {
             buffer.append(formattedMessage);
         } else {
             final int argCount = args != null ? args.length : 0;
-            ParameterFormatter.formatMessage(buffer, pattern, args, argCount, patternAnalysis);
+            try {
+                ParameterFormatter.formatMessage(buffer, pattern, args, argCount, patternAnalysis);
+            } catch (final Exception error) {
+                STATUS_LOGGER.error("message formatting failure", error);
```
ppkarwasz, line 284: Nit: the message is "Unable to format msg: {}" in the case of `StringFormattedMessage`. Can we use the same message everywhere?

ppkarwasz (approved):
LGTM

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterizedMessage.java
vy: Fixed in a40714e415813433df099f6d15de5b511ea3f910.

log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
vy: Piotr, if I would have been a programmer who also happens to be a serial killer, you would not want me to see what you have done in `UsingStatusListener` et al. using `ThreadLocal`s. :sunglasses: I find it a liability with several fragile assumptions just to make tests needing `StatusLogger` work in parallel. I have so much dislike towards such design choices, but I will spare that for another time.

Just to show how much I like you, I have replaced `@UsingStatusLoggerMock` with `@UsingStatusListener` in 186c580d2cda1f301e7b5eab14ded8a15e5b54f5 as you requested. :stuck_out_tongue_closed_eyes:

ppkarwasz (approved):

Issue #1954, opened 2023-11-08
[MS9] Schema and documentation generation from source code


grobmeier:
**Abstract**: This PR completes the STF milestone 12 by implementing the infrastructure to auto-generate 1) the Log4j plugin documentation (to be deployed as a part of the Log4j website) and 2) the XSD (for assisting the Log4j runtime configuration, i.e., `log4j2.xml`, in user applications).

### The problem description

Log4j runtime is a composition of 300+ plugins, which is analogous to _beans_ in Java EE and Spring applications. Developers document these plugins in the website. Users assemble them in the `log4j2.xml` file describing their Log4j runtime configurations.

Currently, several plugins are either not documented, or their documentation doesn't match the behaviour of the source code. This not only causes users to misconfigure their applications or not be aware of all options available to them, but can also pose security implications due to unintended (mis)configurations. Next to lacking documentation, the `log4j-config.xsd` file users use to get assisted while assembling their `log4j2.xml` configurations is _severely_ outdated. It doesn't cover a very big part of the available plugin ecosystem.

### Deliverable 1: Log4j Docgen

**Log4j Docgen (Documentation Generator)** project aims to attack this problem by generating 1) the documentation, 2) plugin configuration options, and 3) XSD from the Java source code of the Log4j project itself. Effectively, this enables the following improvements:

* _Avoid the need to maintain documentation in multiple places_ – For instance, Log4j JSON Template Layout is composed of several plugins with sufficient Javadoc explaining their behaviour. Next to this we also maintain a big `json-template-layout.adoc` page for the website, again, documenting each component. Last, but not least, `log4j-config.xsd` also needs to be updated (which is not!) to document all these whistles and bells too. By generating documentation from the source code, developers can precisely capture the behaviour of their components only in the very code itself that they change. As a result, there will be one and only place of the truth: the source code itself.
* _Accurately capture the component configuration_ –  It is pretty common that source code changes are not reflected to documentation and XSD. By generating the documentation from the source code, this discrepancy gets removed.

[...]

PR #2357, opened 2024-03-07
Fix regex error in Log4j-config.xsd

MalKeshar:
Trivial fix of error "The value of the facet 'pattern' is not a valid regular expression"

Before:
```bash
$ xmllint --schema Log4j-config.xsd /tmp/log4j2.xml --noout
regexp error : failed to compile: Wrong escape sequence, misuse of character '\'
regexp error : failed to compile: internal: no atom generated
regexp error : failed to compile: generate transition: atom == NULL
regexp error : failed to compile: xmlFAParseAtom: expecting ')'
regexp error : failed to compile: xmlFAParseRegExp: extra characters
Log4j-config.xsd:131: element pattern: Schemas parser error : Element '{[http://www.w3.org/2001/XMLSchema}pattern':](http://www.w3.org/2001/XMLSchema%7Dpattern':) The value '(|disable|\$\{[^{}]+\})' of the facet 'pattern' is not a valid regular expression.
...
```
And few same messages

After:
```bash
$ xmllint --schema log4j-core/src/main/resources/Log4j-config.xsd /tmp/log4j2.xml --noout
/tmp/log4j2.xml validates
```

### DIFF
--- a/log4j-core-test/src/main/resources/Log4j-config.xsd
+++ b/log4j-core-test/src/main/resources/Log4j-config.xsd
@@ -128,7 +128,7 @@
             </annotation>
             <simpleType>
                <restriction base="string">
-                  <pattern value="(|disable|\$\{[^{}]+\})" />
+                  <pattern value="(|disable|\\$\{[^{}]+\})" />
                </restriction>
             </simpleType>
          </attribute>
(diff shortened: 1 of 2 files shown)


vy:
@MalKeshar, thanks, much appreciated! The XSD misses a *huge* set of the available components and I am actively working on a project that will auto-generate this file from scratch. I expect to deliver a complete XSD within a couple of weeks. To avoid merge conflicts, and since there is a better solution in the making, I am inclined to close this PR. @ppkarwasz, thoughts?

ppkarwasz:
I am closing this, since we'll remove the manually written schema soon.

I added a task pointing to this PR in the #1954 milestone.

PR #2366, opened 2024-03-11
Generalize DI usage

ppkarwasz:
This commit ensures that:

* an `InstanceFactory` is created in the Log4j `Provider`,
* the same factory is injected as parameters to all the components of the chain `Log4jContextFactory`, `ContextSelector`, `LoggerContext` (child), `Configuration` (grand-child).
* `DI.createInstanceFactory` is called only when strictly necessary.

## Review kit

The most important changes are:

- `ConfigurationFactory` is split into an interface (`ConfigurationFactory`) and an abstract class (`AbstractConfigurationFactory`). The `ConfigurationFactory` type becomes an interface, which might be a non-acceptable change,
- The entire tree `Provider` -> `LoggerContextFactory` -> `ContextSelector` -> `LoggerContext` -> `Configuration` is created using an `InstanceFactory`,
- While it is impossible to eliminate the default constructors for `DefaultConfiguration` and `NullConfiguration` (the tests are filled with these), their usage is limited to tests,
- `LoggerConfig` is refactored to require a `Configuration` parameter to retrieve the services it need (e.g. `LogEventFactory`),
- The `log4j-async-logger` artifact has been refactored to create the necessary disruptors using the DI. Since the `AsyncLoggerConfigDisruptor` requires some configuration data only available in a `Configuration`, this forced a change in the way bundles of beans are registered with the DI: **each** child instance factory applies all available `ConfigurableInstanceFactoryPostProcessor` services to itself. Previously on the first instance factory did it.

Part of #2290.


log4j-1.2-api/src/test/java/org/apache/log4j/CategoryTest.java
```diff
@@ -334,25 +335,43 @@ private <M extends Message> void testMessageImplementation(
...
             rootLogger.info("testAddLogger");
-            assertEquals(++count, version1Appender.getEvents().size(), "adding at root works");
-            assertEquals(0, appender.getEvents().size(), "adding at child works");
+            assertThat(version1Appender.getEvents())
+                    .as("check root logger events")
+                    .extracting(LoggingEvent::getRenderedMessage)
+                    .containsExactly("testAddLogger");
+            assertThat(appender.getEvents())
+                    .as("check logger '%s' events", logger.getName())
+                    .isEmpty();
```
ppkarwasz, line 352: These improvements in `CategoryTest` are the reason for the changes in `LogEventWrapper/LogEventAdapter`.
I could push these in a separate PR.

ppkarwasz:
@jvz,

I am closing this as "partially reviewed", but I am waiting patiently to replace the 3 fully-fledged and configured `InstanceFactory`s with a nifty use of scopes.

jvz:
Sounds good to me!

Issue #2363, opened 2024-03-11
Add Logback's throwable-consuming semantics as an option


SeasonPanPan:
## Description
Hello, we used to use logback in our springboot project, recently migrated to log4j2 and we found that a lot of the exception stack is gone. 

For example, in the following example, using logback, you can clearly see the error stack, but log4j2 only has one sentence "java.lang.NullPointerException", I read the source code of log4j2 and feel that this should be a bug. I also test it with version:2.23.1, and the same error.
Additionally, I found that these classes are correct: MessageFormatMessage and StringFormattedMessage.

If you can confirm that this is a bug, I'd be pleasure to submit a PR to fix it.

## Configuration

**Version:** 2.23.1

**Operating system:** windows/Linux

**JDK:** "1.8.0_291"

## Logs

```
==> log4j2 result:

20:32:33.887 [main] [ERROR] com.pandas.springboot.demo.DemoApplicationTests - fail to process null, the error is:java.lang.NullPointerException

==> logback result:
2024-03-11 20:37:35.173 ERROR 16840 --- [           main] c.p.s.demo.DemoApplicationTests          : fail to process null, the error is:{}

java.lang.NullPointerException: null
	at com.pandas.springboot.demo.DemoApplicationTests.log(DemoApplicationTests.java:20) [test-classes/:na]
	at com.pandas.springboot.demo.DemoApplicationTests.testLog(DemoApplicationTests.java:15) [test-classes/:na]
	
```

## Reproduction

```java

@SpringBootTest
public class DemoApplicationTests {

	private static final Logger log = LoggerFactory.getLogger(DemoApplicationTests.class);

	@Test
	public void testLog() {
		log(null);
	}

	private static void log(String param) {
		try {
			int b = param.length();
		} catch (Exception e) {
			log.error("fail to process {}, the error is:{}", param, e);
		}
	}
}

```

vy:
@SeasonPanPan, could you provide us a **minimal** `log4j2.xml` and a **minimal** `pom.xml` that we can reproduce the problem using your `DemoApplicationTests` shared above?

**minimal**: containing only what is necessary to reproduce the issue and nothing more

SeasonPanPan:
@vy This is the **log4j2.xml**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Configuration status="WARN">
    <Appenders>
        <Console name="Console" target="SYSTEM_OUT">
[...]

PR #2368, opened 2024-03-12
Fix issue #2363: Extract Throwable should be irrelevant to the count of patternAnalysis

SeasonPanPan:
Fix issue #2363: Extract Throwable should be irrelevant to the count of patternAnalysis

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterizedMessageTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterizedMessageTest.java
@@ -17,6 +17,8 @@
 package org.apache.logging.log4j.message;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


SeasonPanPan:
Hello, for this issue, I have some questions about it, the first modification is to keep the original judgment of the count of parameters and placeholders, and change `args.length > analysis.placeholderCount` to `args.length >= analysis.placeholderCount`, but I also thought that if the user sets the last parameter to an exception, it is more appropriate to print out the exception stack, It has nothing to do with the count of placeholders. 

I hope that teachers can explain the intent of this code?

ppkarwasz (changes requested):
I like the idea of using the last parameter as `Throwable` (and removing it from the parameter list), whenever possible, but this constitutes a breaking change.

You can achieve the same effect implementing a separate `MessageFactory2`.

SeasonPanPan:
@ppkarwasz I got your point,  thank you.

PR #2374, opened 2024-03-13
Centralize initialization in `Provider` class

ppkarwasz:
We move the code responsible for the instantiation of `LoggerContextFactory` and `ThreadContextMap` from the static entry points to the logging system (`LogManager` and `ThreadContext`) to the `Provider` class.

The `Provider` class is instantiated using `ServiceLoader`, so `log4j-core` 2.x and 3.x can reimplement the initialization process according to their own rules. E.g. `log4j-core` 3.x can use the DI to create an instance of `LoggerContextFactory` and `ThreadContextMap`.

The following modification were performed:

* a **new** system property `log4j.provider` was introduced,
* the old `log4j2.loggerContextFactory` has been deprecated and revised: if set it selects the first provider that uses the given `LoggerContextFactory`. Therefore it selects now both the context factory and thread context map implementations,
* private static configuration values were removed from `ThreadContextMap` implementations, helping test parallelisation,
* a distinct `NoOpThreadContextStack` implementation has been introduced.

This PR only includes tests for the first two items. The remaining points **should** be covered by existing tests.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
@@ -48,7 +48,7 @@ public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyString
     }
 
     /**
-     * @deprecated Since 2.24.0 use the default constructor or {@link NoOpThreadContextMap} instead.
+     * @deprecated Since 2.24.0. See {@link Provider#getThreadContextMap()} on how to obtain a no-op map.
      */
     @Deprecated
     public DefaultThreadContextMap(final boolean useMap) {
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
@@ -43,40 +43,30 @@ public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyString
     private final boolean useMap;
     private final ThreadLocal<Map<String, String>> localMap;
 
-    private static boolean inheritableMap;
-
-    static {
-        init();
-    }
-
-    // LOG4J2-479: by default, use a plain ThreadLocal, only use InheritableThreadLocal if configured.
-    // (This method is package protected for JUnit tests.)
-    static ThreadLocal<Map<String, String>> createThreadLocalMap(final boolean isMapEnabled) {
-        if (inheritableMap) {
-            return new InheritableThreadLocal<Map<String, String>>() {
-                @Override
-                protected Map<String, String> childValue(final Map<String, String> parentValue) {
-                    return parentValue != null && isMapEnabled //
-                            ? Collections.unmodifiableMap(new HashMap<>(parentValue)) //
-                            : null;
-                }
-            };
-        }
-        // if not inheritable, return plain ThreadLocal with null as initial value
-        return new ThreadLocal<>();
-    }
-
-    static void init() {
-        inheritableMap = PropertiesUtil.getProperties().getBooleanProperty(INHERITABLE_MAP);
-    }
-
     public DefaultThreadContextMap() {
         this(true);
     }
 
+    /**
+     * @deprecated Since 2.24.0 use the default constructor or {@link NoOpThreadContextMap} instead.
+     */
+    @Deprecated
     public DefaultThreadContextMap(final boolean useMap) {
+        this(useMap, PropertiesUtil.getProperties());
+    }
+
+    DefaultThreadContextMap(final boolean useMap, final PropertiesUtil properties) {
         this.useMap = useMap;
-        this.localMap = createThreadLocalMap(useMap);
+        localMap = properties.getBooleanProperty(INHERITABLE_MAP)
+                ? new InheritableThreadLocal<Map<String, String>>() {
+                    @Override
+                    protected Map<String, String> childValue(final Map<String, String> parentValue) {
+                        return parentValue != null && useMap
+                                ? Collections.unmodifiableMap(new HashMap<>(parentValue))
+                                : null;
+                    }
+                }
+                : new ThreadLocal<Map<String, String>>();
     }
 
     @Override
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
@@ -18,6 +18,7 @@ package org.apache.logging.log4j.spi;
 
 import java.lang.ref.WeakReference;
 import java.net.URL;
+import java.util.Objects;
 import java.util.Properties;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.simple.SimpleLoggerContextFactory;
@@ -39,6 +40,9 @@ import org.apache.logging.log4j.util.PropertiesUtil;
  * </p>
  */
 public class Provider {
+    /**
+     * Constant inlined by the compiler
+     */
     protected static final String CURRENT_VERSION = "2.6.0";
 
     /**
@@ -68,9 +72,6 @@ public class Provider {
      */
     public static final String PROVIDER_PROPERTY_NAME = "log4j.provider";
 
-    // Bundled context map implementations
-    private static final String BASE = "org.apache.logging.log4j.internal.map.";
-
     /**
      * Constant used to disable the {@link ThreadContextMap}.
      * <p>
@@ -78,7 +79,7 @@ public class Provider {
      * </p>
      * @see #getThreadContextMap
      */
-    public static final String NO_OP_CONTEXT_MAP = BASE + "NoOp";
+    protected static final String NO_OP_CONTEXT_MAP = "NoOp";
 
     /**
      * Constant used to select a web application-safe implementation of {@link ThreadContextMap}.
@@ -90,7 +91,7 @@ public class Provider {
      * </p>
      * @see #getThreadContextMap
      */
-    public static final String WEB_APP_CONTEXT_MAP = BASE + "WebApp";
+    protected static final String WEB_APP_CONTEXT_MAP = "WebApp";
 
     /**
      * Constant used to select a copy-on-write implementation of {@link ThreadContextMap}.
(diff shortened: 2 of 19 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
         return null;
     }
 
     /**
-     * Gets the class name of the {@link org.apache.logging.log4j.spi.ThreadContextMap} implementation of this Provider.
-     *
+     * Extension point for providers to create a {@link LoggerContextFactory}.
+     * @since 2.24.0
+     */
+    protected LoggerContextFactory createLoggerContextFactory() {
```
vy, line 277: I have a feeling this is here only for tests. I cannot see it getting it used in a way that necessitates a `public` visibility. Am I missing something?

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
+     * </p>
+     * <ol>
+     *     <li>calls {@link #loadThreadContextMap},</li>
+     *     <li>if the previous call returns {@code null}, it calls {@link #getThreadContextMap} to instantiate one of
+     *     the internal implementations,</li>
+     *     <li>it returns a no-op map otherwise.</li>
+     * </ol>
+     * @since 2.24.0
+     */
+    protected ThreadContextMap createThreadContextMap() {
```
vy, line 388: Similar to `createLoggerContextFactory()`, I don't understand why this needs to have `public` visibility.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
+        }
+        LOGGER.warn("Falling back to {}", SimpleLoggerContextFactory.INSTANCE);
+        return SimpleLoggerContextFactory.INSTANCE;
+    }
+
+    /**
+     * @return A lazily initialized logger context factory
+     * @since 2.24.0
+     */
+    public final LoggerContextFactory getLoggerContextFactory() {
```
vy, line 295: Why do we seal this method with `final`? If I would have been a provider, I would rather only implement this method
1. don't deal with the ctor param
2. throw `UnsupportedOperationException` for `loadLoggerContextFactory()`

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
+                LOGGER.error(
+                        "Unable to create instance of class {} specified in {}", factoryClass.getName(), getUrl(), e);
+            }
+        }
+        LOGGER.warn("Falling back to {}", SimpleLoggerContextFactory.INSTANCE);
+        return SimpleLoggerContextFactory.INSTANCE;
+    }
+
+    /**
+     * @return A lazily initialized logger context factory
```
vy, line 292:
```suggestion
     * @return a new logger context factory instance
```
_"A lazily initialized"_ statement is not correct, the returned instance is already initialized.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
+        return new NoOpThreadContextMap();
+    }
+
+    // Used for testing
+    void resetThreadContextMap() {
+        threadContextMapLazy.set(null);
+    }
+
+    /**
+     * @return A lazily initialized thread context map.
```
vy, line 435:
```suggestion
     * @return a new thread context map instance
```
_"A lazily initialized"_ statement is not correct, the returned instance is already initialized.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
+     * @since 2.24.0
+     */
+    public final ThreadContextMap getThreadContextMapInstance() {
```
vy, line 438: Similar to `getLoggerContextFactory()`, I don't understand why this is `final`.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -97,127 +217,266 @@ public String getVersions() {
...
-        if (classLoader == null || (loader = classLoader.get()) == null) {
-            result.append(", classLoader=null(not reachable)");
-        } else {
-            result.append(", classLoader=").append(loader);
+        if (Provider.class.equals(getClass())) {
+            final ClassLoader loader = classLoader.get();
+            if (loader == null) {
+                result.append("\n\tclassLoader          = null or not reachable");
+            } else {
+                result.append("\n\tclassLoader          = ").append(loader);
```
vy, line 477: Please don't do this type of indentation. It is a matter of time somebody will screw up the alignment. I prefer the earlier version.

log4j-api/src/main/java/org/apache/logging/log4j/util/ProviderUtil.java
```diff
@@ -163,6 +172,88 @@ static void lazyInit() {
         }
     }
 
+    /**
+     * Used to test the public {@link #getProvider()} method.
+     */
+    static Provider selectProvider(final PropertiesUtil properties) {
```
vy, line 178: I presume you sufficiently covered this method with tests. :see_no_evil:

log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
```diff
@@ -43,40 +43,30 @@ public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyString
...
+        localMap = properties.getBooleanProperty(INHERITABLE_MAP)
+                ? new InheritableThreadLocal<Map<String, String>>() {
+                    @Override
+                    protected Map<String, String> childValue(final Map<String, String> parentValue) {
+                        return parentValue != null && useMap
+                                ? Collections.unmodifiableMap(new HashMap<>(parentValue))
+                                : null;
+                    }
+                }
+                : new ThreadLocal<Map<String, String>>();
```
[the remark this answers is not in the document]
ppkarwasz: `localMap` must be a `ThreadLocal`, ;-)

If `!useMap` I could probably set `localMap` to `null`, but our code (see `Provider`) does not use `DefaultThreadContextMap` in such a case. I leave the `useMap` parameter only for backward compatibility.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -20,39 +20,135 @@
...
+     * <p>
+     *     This implementation must ensure that common operations don't create new object instances. The drawback is
+     *     the necessity to bind custom classes to {@link ThreadLocal} variables.
+     * </p>
+     * <p>
+     *     <strong>Warning:</strong> the value of this constant does not point to a concrete class name.
+     * </p>
+     * @see #getThreadContextMap
+     */
+    public static final String GARBAGE_FREE_CONTEXT_MAP = BASE + "GarbageFree";
```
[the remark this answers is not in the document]
ppkarwasz: You are right, they could be `protected`.

The reason I introduced these constants is:

- half of these implementations are package-private. An implementation of `Provider` can not instantiate them, but `Provider` itself can. To use these context maps an implementation should:
   1. implement `#getThreadContextMap` and return one of these special constants,
   2. make a call to `super.createThreadContextMap()` in its `#createThreadContextMap` implementation.
- I wanted to separate the characteristics of the context map, from the implementation. So the constant `WebApp` can switch to @jengebr's implementation under the hood. Probably I should make these constants shorter.

BTW: IMHO the inclusion of `ThreadContextMap` implementations in the `log4j-api` unnecessarily bloats the artifact. `ThreadContext` is not a standalone service for users to store thread-dependent data (although I saw some users use it for this purpose). It must be tightly linked to the `Logger` implementation. As such, except Log4j Core, no other Log4j API implementation requires these 3 implementations.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
ppkarwasz: This is actually one of the two methods that implementations should implement.

Due to the long history of the class, there are other methods:

- `#getClassName()` gives the name of the class, but then Log4j API must guess the classloader to use to load it,
- `#loadLoggerContextFactory()` returns a class, but unless I am mistaken JPMS will not allow Log4j API to instantiate the class unless it is `public` and in an exported package,

I introduced this method so that implementations can instantiate it in any way they need. E.g. Log4j Core 3.x uses:
```java
new Log4jContextFactory(instanceFactory);
```

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
ppkarwasz: I wanted to stress out the `LoggerContextFactory` is a singleton.

log4j-api/src/main/java/org/apache/logging/log4j/util/ProviderUtil.java
ppkarwasz: I tested the externally observable behavior:

1. the `log4j.provider` system property creates a provider with the specified class name,
2. the `log4j2.loggerContextFactory` system property create a provider with the specified logger context factory class,
3. if more than one implementation is available, the implementation with highest `getPriority` wins.

What happens in the remaining cases (multiple system properties set, typos in the names of properties) are in my opinion implementation details and we can change them at will without breaking the `#getProvider()` contract.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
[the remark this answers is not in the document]
ppkarwasz: I marked the constants as `protected`.

log4j-api/src/main/java/org/apache/logging/log4j/util/ProviderUtil.java
ppkarwasz: I increased the test coverage with some smoke tests.

If users specify a faulty configuration, I just check that a warning or error appears and no exception is thrown.

rgoers:
I apologize for taking so long to look at this but while working on moving context related stuff to log4j-context-data I ran into this and it makes things much more difficult as it hard-wires a bunch of classes that should be internal/private into the Provider class.  I would request that those changes be undone.

ppkarwasz:
Those internal/private classes are hardwired in `Provider`, since implementations of `Provider` have no way to access those classes.

Once I'll promote the `StringArrayThreadContextMap` to default thread context map, I'll be able to either:

* remove those private implementations (since the default context map can be used for garbage-free logging)
* or move them to `log4j-core` (since no other implementation really profits from them).

PR #2381, opened 2024-03-14
Add Logback's throwable-consuming semantics as an option

### LINKED ISSUE #2363: Add Logback's throwable-consuming semantics as an option

ppkarwasz:
The semantics used to determine the `Throwable` attached to a log event changed in Logback 1.1.0 (in the fix to [LOGBACK-873](https://jira.qos.ch/browse/LOGBACK-873)).

When there are not enough arguments to the log call to fill both the throwable field and the placeholders contained in the message pattern, Logback 1.1.0 and later prefer to fill the throwable field and leave some placeholders empty. Log4j Core does the opposite choice: it first fills all the placeholders and only fills the throwable field if there is something left.

This change allows `log4j-slf4j-impl` and `log4j-slf4j2-impl` users to switch between the two behaviors by setting the property `log4j2.messageFactory` to:

```
org.apache.logging.slf4j.message.ThrowableConsumingMessageFactory
```

In this PR two choices have been made:

- the message factory is only included in `log4j-slf4j-impl` and `log4j-slf4j2-impl`, since it should be mostly required by Logback users during a transition period. Note that Logback's and Log4j's semantics to deal with throwables agree whenever enough arguments are provided. In a second step users can modify their code using openrewrite/rewrite-logging-frameworks#141, so that it behaves in the same way in Logback and Log4j Core out-of-the-box. However I can also create a new `log4j-message` artifact for this.
- no **garbage-free** version of this factory is provided. SLF4J creates temporary arrays whenever **more than 2** arguments are provided. I believe that an additional creation of a `ParameterizeMessage` instance does not make a difference.

Closes #2363


vy (changes requested):
LGTM, thanks for taking care of this @ppkarwasz! :100:
I have two nits:

1. Would you mind explicitly stating in **all** newly added Java source files something like the following, please?
```
This class should be identical to its clone in {@code log4j-slf4j2-impl}.
Any changes made here should be reflected to the other one, and vice versa.
```
2. Would you mind documenting this in the following files, please?
   1. `migrate-from-logback.adoc`
   2. `installation.adoc#impl-logback`

ppkarwasz:
@vy,
>     1. Would you mind explicitly stating in **all** newly added Java source files something like the following, please?
>
> ```
> This class should be identical to its clone in {@code log4j-slf4j2-impl}.
> Any changes made here should be reflected to the other one, and vice versa.
> ```

I don't think this is necessary. Every time one of `log4j-slf4j-impl` or `log4j-slf4j2-impl` is modified, the other one needs to be modified too. 

>     2. Would you mind documenting this in the following files, please?
>        
>        1. `migrate-from-logback.adoc`
>        2. `installation.adoc#impl-logback`

I added some documentation to `migrate-from-logback.adoc` in f7a63206b831043399f04320f533053300114df0.
The `ThrowableConsumingMessageFactory` is only useful for people that migrate from Logback, so I didn't add it to `installation.adoc`.

src/changelog/.2.x.x/2363_add_logback_throwable_consuming_semantics.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-       xmlns="http://logging.apache.org/log4j/changelog"
-       xsi:schemaLocation="http://logging.apache.org/log4j/changelog https://logging.apache.org/log4j/changelog-0.1.3.xsd"
-       type="added">
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="changed">
```
vy, line 5: Shouldn't this be `added` instead?

src/site/antora/modules/ROOT/examples/migrate-from-logback/MigrateFromLogback.java
```diff
@@ -0,0 +1,28 @@
...
+import org.apache.logging.log4j.Logger;
+
+class MigrateFromLogback {
+    private static Logger logger = LogManager.getLogger();
+
+    doLogWrong() {
+        try {
+            // do something
+            // tag::wrong[]
+        } catch (Exception e) {
```
vy, line 29: Could you, please, please, please use self-explanatory variable names in the docs? For instance, `error`, `failure`, etc.

src/site/antora/modules/ROOT/pages/migrate-from-logback.adoc
```diff
@@ -106,3 +106,62 @@ To assist with migrating Logback configuration components to Log4j Core, see the
...
+[source]
+----
+The foo process exited with and error.
+java.lang.RuntimeException: Message
+    at example.MigrateFromLogback.doLogWrong(MigrateFromLogback.java:10)
+...
+----
+
+[TIP]
+====
```
vy, line 163: Block is not closed.

src/changelog/.2.x.x/2363_add_logback_throwable_consuming_semantics.xml
ppkarwasz: Nice catch! It is a copy/paste error.

Fixed in 42b5ebbb41b4b23f665d395833a10c431a2a602f.

src/site/antora/modules/ROOT/examples/migrate-from-logback/MigrateFromLogback.java
ppkarwasz: Fixed in 42b5ebbb41b4b23f665d395833a10c431a2a602f.

src/site/antora/modules/ROOT/pages/migrate-from-logback.adoc
ppkarwasz: Fixed in 42b5ebbb41b4b23f665d395833a10c431a2a602f.

vy (approved):

PR #2385, opened 2024-03-17
2214 - Add support for ScopedContext

rgoers:
This adds support for a ScopedContext as requested in apache/logging-log4j-kotlin#71 and similar to #2214

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
@@ -51,8 +51,12 @@ public class ScopedContext {
      */
     public static Map<String, Renderable> getContextMap() {
         Optional<ScopedContext> context = ScopedContextAnchor.getContext();
-        return context.map(scopedContext -> Collections.unmodifiableMap(scopedContext.contextMap))
-                .orElse(Collections.emptyMap());
+        if (context.isPresent()
+                && context.get().contextMap != null
+                && !context.get().contextMap.isEmpty()) {
+            return Collections.unmodifiableMap(context.get().contextMap);
+        }
+        return Collections.emptyMap();
     }
 
     /**
@@ -80,9 +84,31 @@ public class ScopedContext {
         return ScopedContextAnchor.getContext();
     }
 
-    private final Map<String, Renderable> contextMap = new HashMap<>();
+    private final ScopedContext parent;
+    private final String key;
+    private final Renderable value;
+    private final Map<String, Renderable> contextMap;
 
-    private ScopedContext() {}
+    private ScopedContext() {
+        this.parent = null;
+        this.key = null;
+        this.value = null;
+        this.contextMap = null;
+    }
+
+    private ScopedContext(Map<String, Renderable> map) {
+        this.parent = null;
+        this.key = null;
+        this.value = null;
+        this.contextMap = map;
+    }
+
+    private ScopedContext(ScopedContext parent, String key, Renderable value) {
+        this.parent = parent;
+        this.key = key;
+        this.value = value;
+        this.contextMap = null;
+    }
 
     /**
      * Adds a key/value pair to the ScopedContext being constructed.
@@ -106,14 +132,8 @@ public class ScopedContext {
 
     private ScopedContext addObject(String key, Object obj) {
         if (obj != null) {
-            ScopedContext context = new ScopedContext();
-            context.contextMap.putAll(this.contextMap);
-            if (obj instanceof Renderable) {
-                context.contextMap.put(key, (Renderable) obj);
-            } else {
-                context.contextMap.put(key, new ObjectRenderable(obj));
-            }
-            return context;
+            Renderable renderable = obj instanceof Renderable ? (Renderable) obj : new ObjectRenderable(obj);
+            return new ScopedContext(this, key, renderable);
         }
         return this;
     }
@@ -123,12 +143,7 @@ public class ScopedContext {
      * @param op the code block to execute.
      */
     public void run(Runnable op) {
-        ScopedContextAnchor.addScopedContext(this);
-        try {
-            op.run();
-        } finally {
-            ScopedContextAnchor.removeScopedContext();
-        }
+        new ScopedContextRunner(this, op).run();
     }
 
     /**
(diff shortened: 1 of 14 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
```diff
@@ -0,0 +1,224 @@
...
+        return new ScopedContext(inherit);
+    }
+
+    /**
+     * Return the key from the current ScopedContext, if there is one and the key exists.
+     * @param key The key.
+     * @return The value of the key in the current ScopedContext.
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T get(String key) {
```
ppkarwasz, line 102: I believe that `ThreadContext` and `ScopedContext` should be **write-only**.

As far as I see it, these are logging-only services. The presence of `get()` methods allows user to use these services to store and retrieve their thread-dependent data, which is in my opinion and **abuse** of the API.

There are more proper ways to store data, e.g. in a servlet application the username should be stored in the `ServletRequest` attributes, not in `ThreadContext`.

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
```diff
@@ -0,0 +1,224 @@
...
+    /**
+     * Adds a key/value pair to the ScopedContext being constructed.
+     * @param key the key to add.
+     * @param value the value associated with the key.
+     * @return the ScopedContext being constructed.
+     */
+    public ScopedContext where(String key, Object value) {
+        addObject(key, value);
+        return this;
+    }
```
ppkarwasz, line 140: This kind of methods should return a new copy of `ScopedContext` with a new copy of the map, in my opinion.

Code like:

```java
ScopedContext ctx = ScopedContext.newInstance().where("userId", "lala");
ctx.where("spanId", "abcdef").run(...);
ctx.run(...);
```

will fail otherwise.

ppkarwasz (commented):
In this epoch of concurrent programming, I am not sure if introducing another API class based on static methods and `ThreadLocal`'s is such a good idea.

At a first glance, I would probably prefer to see a wrapper like this:

```java
public class ContextLogger implements Logger, CloseableThreadContext.Instance {
    private final Logger delegate;
    private final Map<String, String> contextData;
    ...
}
```

that would allow users to modify context data and log messages at the same time.

However this subject should be discussed on `dev@logging` and with many other Open Source projects. For example I believe that the first logging API that will be included into "advanced" `Runnable`s such as [`CoreSubscriber`](https://projectreactor.io/docs/core/release/api/reactor/core/CoreSubscriber.html) or Kotlin's coroutines will have a very bright future.

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
rgoers: I have to disagree with this but here we are just expressing our own opinions. Your example would require passing the ServletRequest through many methods that have no business knowing what a ServletRequest is or that they are even running in a servlet, but do need to know who the user is. For example, I have plenty of code that is shared between REST requests and Kafka consumers.  It is obviously very late to suggest that the ThreadContext be write only.   FWIW, the enhancement going into SLF4J does something sort of similar to this but instead of having the scoped context completely separate there you still do MDC.get to fetch attributes from the scoped context.

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
rgoers: Actually, it won't fail. It just won't behave the same way you are believing it should. You are thinking that the run method on the second line should get a different instance then the run method on the third line. As written they will both get the same context. That doesn't make it wrong. Just different.  Again, this is simply a difference of opinion over how the API should behave. I am fine having a discussion about that.

However, I do see one possible concern. Someone could co
```
final ScopedContext ctx = ScopedContext.newInstance().where("userId", "lala");
ctx.where("spanId", "abcdef").run(() -> {
    ...
    ctx.where("spanId", "12345");
)};
```

Obviously calling where from inside the run or call methods should not be allowed, but that is easily preventable.  Admittedly though, returning an immutable object makes this impossible to cause any harm. This alone might convince me it should be immutable, as much as I really dislike that.

rgoers:
@ppkarwasz For some reason I cannot reply to your original comment directly.

> At a first glance, I would probably prefer to see a wrapper like this:
> 
> public class ContextLogger implements Logger, CloseableThreadContext.Instance {
>     private final Logger delegate;
>     private final Map<String, String> contextData;
>     ...
> }

This I could not agree to. It is a completely different paradigm. The ScopedContext creates attributes to be logged for a block of code across all the Loggers that may be encountered. What you propose here is a context for a single Logger. While that may be useful it is not the same thing at all and serves a completely different purpose.

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
rgoers: It occurs to me that making the where method return an immutable object makes newInstance worthless. The boolean is no longer needed to clone the previous context since the where method would always do that.

rgoers: I have made ScopedContext immutable

ppkarwasz:
I think we should discuss it in `dev@logging`. I started a [thread dedicated to this subject](https://lists.apache.org/thread/ossk8ckz4yy59z88l8hhbb3k87hbbk0r).

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
ppkarwasz: Yes, I believe it is a matter of taste and commodity.

I am a little bit worried about users using `ThreadContext` for business-critical functionality, since a change of logging backend (e.g. from `log4j-core` to `log4j-to-jul`), might break that functionality. For example in #2374 I assumed that the `ThreadContextMap` implementation used by an implementation is strictly linked to the logger context implementation; since `java.util.logging` formatters don't support context data, `log4j-to-jul` might as well use no-op `ThreadContextMap`. Am I wrong?

rgoers:
@garydgregory It is modeled after https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ScopedValue.html. Obviously it is not the same API but it is similar enough that I would hope users would easily understand it.

vy (changes requested):
I have shared [my remarks in the mailing list](https://lists.apache.org/thread/zmcqrxkzkk0khdvb48b34hntr2w7c6x4). In short, IMHO, in its current form, `ScopedContext` addresses neither #2214, nor apache/logging-log4j-kotlin#71, and its API, and consequently its API's integration with the rest of the system still needs some work.

log4j-api/src/main/java/org/apache/logging/log4j/ScopedContext.java
```diff
@@ -0,0 +1,264 @@
...
+            } finally {
+                ScopedContextAnchor.removeScopedContext();
+            }
+        }
+    }
+
+    /**
+     * Interface for converting Objects stored in the ContextScope to Strings for logging.
+     */
+    public static interface Renderable {
```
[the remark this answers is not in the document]
rgoers: With regard to the get() method. This API supports Objects - something users have been requesting for a long time. I added them here because they can be safely used the way this is implemented. There is no point in accepting an object if user's cannot query them since logging will NEVER log an object, only its string representation.

How? There are no StringBuilders at all in ScopedContext. It is used to convert an object into its String representation so it can be included in the LogEvent's ContextMap as a single String item. While the Object's render method certainly might use a StringBuilder internally the code storing the data into the Map certainly doesn't want one. Furthermore, if you look at the default implementation the render method simply calls toString() on the object. For the 90+% of use cases where the object is already a String this is effectively a no-op as it just returns "this". Using a StringBuilder would require much more overhead.

log4j-api/src/main/java/org/apache/logging/log4j/internal/ScopedContextAnchor.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.Optional;
+import org.apache.logging.log4j.ScopedContext;
+
+/**
+ * Anchor for the ScopedContext. This class is private and not for public consumption.
+ */
+public class ScopedContextAnchor {
+    private static final ThreadLocal<Deque<ScopedContext>> scopedContext = new ThreadLocal<>();
```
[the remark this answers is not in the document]
rgoers: 1. This is NOT meant for Java 21 and above.
2. ThreadLocals are NOT risky even with Virtual Threads when they are used in this way. I specifically looked for a mention of being able to disable ThreadLocals and could not find anything as I recall you previously mentioning that.

PR #2392, opened 2024-03-19
Backport some Log4j API 3.x features

ppkarwasz:
We move some features found in the `main` branch of `log4j-api` back to `2.x`:

- It adds a `ServiceLoaderUtil#safeStream` method that takes an instance of `ServiceLoader` as a parameter. Since now the caller must call `ServiceLoader#load`, we can remove the previous methods that created **lambdas through reflection** to workaround the caller-sensitivity of `ServiceLoader#load`. Needless to say, the new approach is much more GraalVM-friendly,
- It makes all the `ReusableParameterizedMessage#set` methods public,
- It removes the `@Deprecated` annotation from `Supplier` and `BiConsumer` and let them extend the `j.u.f` interfaces of the same name,
- It directly instantiates the default `MessageFactory` and `FlowMessageFactory`,
- It directly instantiates the `LoggerContextFactory` and `ThreadContextMap` for the `log4j-to-jul` and `log4j-to-slf4j` implementations.

After these changes we reach two goals:

- Log4j Core 3.x works with Log4j API 2.x (#2290),
- The Log4j API implementations `log4j-to-jul` and `log4j-to-slf4j` work with GraalVM without any additional reachability metadata.


ppkarwasz:
As a curiosity, when working with GraalVM and either `log4j-to-jul` or `log4j-to-slf4j` (i.e. **without** Log4j Core) I get this nice warning from GraalVM:
```
Warning: The log4j library has been detected, but the version is unavailable. Due to Log4Shell, please ensure log4j is at version 2.17.1 or later.
```

log4j-api/pom.xml
```diff
@@ -36,6 +36,10 @@
     <bnd-multi-release>true</bnd-multi-release>
     <!-- Differs from OSGi bundle name -->
     <bnd-module-name>org.apache.logging.log4j</bnd-module-name>
+    <bnd-extra-package-options>
+      <!-- Not exported by most OSGi system bundles, hence we use the system classloader to load `sun.reflect.Reflection` -->
+      !sun.reflect
+    </bnd-extra-package-options>
```
vy, line 42: Has this problem ever been reported? I feel like we are solving a problem of your C64 that you tinker at nights in your attic.

log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
```diff
@@ -51,17 +52,22 @@ public DefaultFlowMessageFactory() {
      * @param exitText the text to use for trace exit, like {@code "Exit"}.
      */
     public DefaultFlowMessageFactory(final String entryText, final String exitText) {
+        this(entryText, exitText, createDefaultMessageFactory());
+    }
+
+    public DefaultFlowMessageFactory(final MessageFactory messageFactory) {
```
vy, line 58: Where do we use this?

log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
```diff
@@ -57,29 +58,53 @@ public static <T> Stream<T> loadServices(final Class<T> serviceType, final Looku
...
+                                case Bundle.STARTING:
+                                    return "STARTING";
+                                case Bundle.STOPPING:
+                                    return "STOPPING";
+                                case Bundle.ACTIVE:
+                                    return "ACTIVE";
+                                default:
+                                    return "UNKNOWN";
+                            }
+                        });
```
vy, line 90: I can't stop but wonder, why not simply use `bundle.getState()` and assume its `toString()` will do the right job.

log4j-api/src/main/java/org/apache/logging/log4j/util/Strings.java
```diff
@@ -288,6 +289,19 @@ public static String trimToNull(final String str) {
...
+    /**
+     * Removes control characters from both ends of this String returning {@code Optional.empty()} if the String is
+     * empty ("") after the trim or if it is {@code null}.
+     * @param str The String to trim.
+     * @return An Optional containing the String.
+     *
+     * @see #trimToNull(String)
+     * @since 2.24.0
+     */
+    public static Optional<String> trimToOptional(final String str) {
```
vy, line 301: Where do we use this?

vy (commented):
Really liked the new `LoaderUtil#safeStream()` usage allowing GraalVM compat! :star_struck:

log4j-api/pom.xml
ppkarwasz: A `sun.reflect;resolution:=optional` declaration has been in `log4j-api` since I joined the team.

We must have lost it, when we switched to BND, because it couldn't detect that we are using `sun.reflect.Reflect`. However after I replaced `LoaderUtil.loadClass` with `Class.forName` the entry reappeared and broke our OSGi tests.

log4j-api/src/main/java/org/apache/logging/log4j/util/OsgiServiceLocator.java
ppkarwasz: `bundle.getState()` is an integer. ;-)

Here I wanted to provide some debugging data for us in case `OsgiServiceLocator` starts misbehaving again, due to the changes in this PR.

log4j-api/src/main/java/org/apache/logging/log4j/message/DefaultFlowMessageFactory.java
ppkarwasz: I planned to use it in Log4j Core 3.x, so that the same `MessageFactory` is used by `DefaultFlowMessageFactory` and `AbstractLogger`.

Thinking about it, I might have been overzealous in #1171, when I introduced the `messageFactory` field. I am removing the field and hardcoding it to `ParameterizedMessageFactory.INSTANCE`.

log4j-api/src/main/java/org/apache/logging/log4j/util/Strings.java
ppkarwasz: In about 10 locations in `log4j-plugins`, but I can move it there.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/Resources.java
```diff
@@ -20,11 +20,36 @@
 
 /**
  * Constants to use the {@link ResourceLock} annotation.
- *
  */
-public class Resources {
+public final class Resources {
```
ppkarwasz, line 24: This class is used in `@ResourceLock` JUnit 5 annotations.

Thanks to the `DI`,  most tests in Log4j Core 3.x will not require to access service classes through static methods (yes, even for `ThreadContext`), but we want to mark those that need to be rewritten.

ppkarwasz:
@vy,

I am merging this PR in order to be able to use `2.24.0-SNAPSHOT` versions in the PRs concerning the `main` branch.

Please let me know if you find something else that needs to be corrected.

PR #2393, opened 2024-03-19
Fix issue #2380: Log messages with insufficient parameters should not throw exception.

### LINKED ISSUE #2380: Log messages with partially missing parameters
Log4j version `2.20.0` did also log messages with missing parameters. A call like:

```java
logger.info("{}, {} and {}", 1, 2);
```

resulted in a partially replaced message: `1, 2, and {}`.

The behavior changed in `2.21.0` and should be restored for backward compatibility.

SeasonPanPan:
Fix issue #2380: Log messages with insufficient parameters should not throw exception.

In this PR, I modified the class `org.apache.logging.log4j.message.ParameterFormatter`.
I deleted the codes about checking placeholderCount and argCount, if they weren't equal, it would throw an IllegalArgumentException. 

At the same time, I also modified the corresponding test cases.

Close #2380

@ppkarwasz Please review the codes when you have time.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
@@ -22,8 +22,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.message.ParameterFormatter.MessagePatternAnalysis;
-import org.apache.logging.log4j.test.junit.UsingStatusLoggerMock;
+import org.apache.logging.log4j.status.StatusData;
+import org.apache.logging.log4j.test.ListStatusListener;
+import org.apache.logging.log4j.test.junit.UsingStatusListener;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
@@ -32,8 +36,14 @@ import org.junit.jupiter.params.provider.MethodSource;
 /**
  * Tests {@link ParameterFormatter}.
  */
-@UsingStatusLoggerMock
+@UsingStatusListener
 class ParameterFormatterTest {
+
+    final ListStatusListener statusListener;
+
+    ParameterFormatterTest(ListStatusListener statusListener) {
+        this.statusListener = statusListener;
+    }
 
     @ParameterizedTest
     @CsvSource({
@@ -67,12 +77,23 @@ class ParameterFormatterTest {
         }
     }
 
-    @Test
-    void formatWithInsufficientArgs() {
-        final String pattern = "Test message {}-{} {}";
-        final Object[] args = {"a", "b"};
-        final String message = ParameterFormatter.format(pattern, args, args.length);
-        assertThat(message).isEqualTo("Test message a-b {}");
+    @ParameterizedTest
+    @CsvSource({"2,pan {} {},a,pan a {}", "3,pan {}{}{},a-b,pan ab{}", "1,pan {},a-b-c,pan a"})
+    void format_should_fail_on_insufficient_args(
+            final int placeholderCount, final String pattern, final String argsStr, final String expectedMessage) {
+        final String[] args = argsStr.split("-");
+        final int argCount = args.length;
+
+        String actualMessage = ParameterFormatter.format(pattern, args, argCount);
+        assertThat(actualMessage).isEqualTo(expectedMessage);
+        final List<StatusData> statusDataList = statusListener.getStatusData().collect(Collectors.toList());
+        assertThat(statusDataList).hasSize(1);
+        final StatusData statusData = statusDataList.get(0);
+        assertThat(statusData.getLevel()).isEqualTo(Level.WARN);
+        assertThat(statusData.getMessage().getFormattedMessage())
+                .isEqualTo(
+                        "found %d argument placeholders, but provided %d for pattern `%s`",
+                        placeholderCount, argCount, pattern);
     }
 
     @ParameterizedTest
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
@@ -17,7 +17,6 @@
 package org.apache.logging.log4j.message;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -68,15 +67,12 @@ class ParameterFormatterTest {
         }
     }
 
-    @ParameterizedTest
-    @CsvSource({"1,foo {}", "2,bar {}{}"})
-    void format_should_fail_on_insufficient_args(final int placeholderCount, final String pattern) {
-        final int argCount = placeholderCount - 1;
-        assertThatThrownBy(() -> ParameterFormatter.format(pattern, new Object[argCount], argCount))
-                .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage(
-                        "found %d argument placeholders, but provided %d for pattern `%s`",
-                        placeholderCount, argCount, pattern);
+    @Test
+    void formatWithInsufficientArgs() {
+        final String pattern = "Test message {}-{} {}";
+        final Object[] args = {"a", "b"};
+        final String message = ParameterFormatter.format(pattern, args, args.length);
+        assertThat(message).isEqualTo("Test message a-b {}");
     }
 
     @ParameterizedTest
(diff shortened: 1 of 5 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -240,14 +240,6 @@ static void formatMessage(
             return;
         }
 
-        // Fail if there are insufficient arguments
-        if (analysis.placeholderCount > args.length) {
-            final String message = String.format(
-                    "found %d argument placeholders, but provided %d for pattern `%s`",
-                    analysis.placeholderCount, args.length, pattern);
-            throw new IllegalArgumentException(message);
-        }
```
ppkarwasz, line 249: I would prefer to issue a `StatusLogger` warning whenever the number of arguments:

- is less than `analysis.placeholderCount` (some arguments are missing),
- is higher than `analysis.placeholderCount + (args[args.length - 1] instance of Throwable ? 1 : 0)` (some arguments will be ignored).

Since there will be no more exceptions throws, you can remove the `try...catch`es at the call sites.

SeasonPanPan: I got your point, I'm modifying the codes.

vy: Agreed with @ppkarwasz's remark. Please provide sufficient test coverage for all possibilities he shared. In particular, test not just the unhappy flow, but also happy flow.

vy (changes requested):
I see several tests being replaced by single-shot examples, please don't. Instead use a `@ParameterizedTest` and provide several examples providing sufficient coverage. I also would like to see examples where arguments contain `Throwable`s to match against the cases @ppkarwasz stated.

log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterFormatterTest.java
```diff
@@ -68,15 +67,12 @@ void test_pattern_analysis(
...
 
-    @ParameterizedTest
-    @CsvSource({"1,foo {}", "2,bar {}{}"})
-    void format_should_fail_on_insufficient_args(final int placeholderCount, final String pattern) {
-        final int argCount = placeholderCount - 1;
-        assertThatThrownBy(() -> ParameterFormatter.format(pattern, new Object[argCount], argCount))
-                .isInstanceOf(IllegalArgumentException.class)
-                .hasMessage(
-                        "found %d argument placeholders, but provided %d for pattern `%s`",
-                        placeholderCount, argCount, pattern);
```
ppkarwasz, line 79: While we don't provide an equivalent `assertThatLoggedBy` method to test that a certain status logger message was issued (we probably should), you can add an `org.apache.logging.log4j.Logger statusLogger` parameter to `ParameterFormatter.format` and detect status messages as in [`ProviderUtilTest`](https://github.com/apache/logging-log4j2/blob/ed4c72e679951fe78fe4121436085444bc87b388/log4j-api-test/src/test/java/org/apache/logging/log4j/util/ProviderUtilTest.java).

vy: No, this is not needed. This file was already testing logged `StatusLogger` messages – scroll down, you will see the associated @SeasonPanPan deleted. They need to simply revert those changes and adapt them accordingly.

SeasonPanPan: Yes, I find the codes to check status logger what I deleted. I am adding the test cases.

SeasonPanPan:
@vy @ppkarwasz Done, and added some test cases with the following scenarios:  
      
     * 1. The placeholders are greater than the count of arguments. 
     * 2. The placeholders are less than the count of arguments.  
     * 3. The arguments contains an exception, and the placeholder is greater than normal arguments. 
     * 4. The arguments contains an exception, and the placeholder is less than the arguments. 
     All of these should logged in status logger with WARN level.
 
But this case with no warn log:
*  The arguments contains an exception, and the placeholder count is equal to normal arguments.

log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterizedMessageTest.java
```diff
@@ -185,16 +185,73 @@ void testSerializable(final Object arg) {
...
+                new Object[] {"more {} {}", 2, new Object[] {"a"}, "more a {}"},
+                new Object[] {"more {} {} {}", 3, new Object[] {"a"}, "more a {} {}"},
+                new Object[] {"less {}", 1, new Object[] {"a", "b"}, "less a"},
+                new Object[] {"less {} {}", 2, new Object[] {"a", "b", "c"}, "less a b"},
+                new Object[] {
+                    "more throwable {} {}",
+                    2,
+                    new Object[] {"a", new RuntimeException()},
+                    "more throwable a java.lang.RuntimeException"
+                },
```
ppkarwasz, line 236: This case should issue no warning: the `RuntimeException` is simply classified as normal argument and it fills the second placeholder.

ppkarwasz (changes requested):
Nice job!

It almost looks good to me, except we need to maintain Log4j semantics (which is the same as Logback < 1.1.0 semantics) for classifying a trailing `Throwable` as part of the arguments or as "extra" argument.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -240,12 +244,14 @@ static void formatMessage(
...
+        // check if there are insufficient arguments that do not include Throwable arg
+        final int realArgCount = args.length;
+        final int noThrowableArgCount = realArgCount - ((args[realArgCount - 1] instanceof Throwable) ? 1 : 0);
+        if (analysis.placeholderCount != noThrowableArgCount) {
             final String message = String.format(
                     "found %d argument placeholders, but provided %d for pattern `%s`",
-                    analysis.placeholderCount, args.length, pattern);
-            throw new IllegalArgumentException(message);
+                    analysis.placeholderCount, realArgCount, pattern);
+            STATUS_LOGGER.warn(message);
```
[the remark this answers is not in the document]
SeasonPanPan: Yes, yours is good example.

log4j-api-test/src/test/java/org/apache/logging/log4j/message/ParameterizedMessageTest.java
SeasonPanPan: I will adjust this case.

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -240,12 +244,14 @@ static void formatMessage(
             return;
         }
 
-        // Fail if there are insufficient arguments
-        if (analysis.placeholderCount > args.length) {
+        // check if there are insufficient arguments that do not include Throwable arg
+        final int realArgCount = args.length;
+        final int noThrowableArgCount = realArgCount - ((args[realArgCount - 1] instanceof Throwable) ? 1 : 0);
+        if (analysis.placeholderCount != noThrowableArgCount) {
```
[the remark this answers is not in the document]
ppkarwasz: Ah, you mean that code like:

```java
logger.info("An error in {} occurred: {}", "foo", new RuntimeException());
```

should give a warning? I might agree that it should give a warning, since such statements prevent users from switching to Logback in the future. However for backward compatibility, such statements should be silently accepted.

@vy, what do you think?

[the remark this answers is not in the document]
SeasonPanPan: Please confirm if I need to give a warning log. I will change my codes or not. @vy @ppkarwasz

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -240,12 +244,17 @@ static void formatMessage(
...
-        // Fail if there are insufficient arguments
-        if (analysis.placeholderCount > args.length) {
-            final String message = String.format(
-                    "found %d argument placeholders, but provided %d for pattern `%s`",
-                    analysis.placeholderCount, args.length, pattern);
-            throw new IllegalArgumentException(message);
+        // #2380: check if the count of placeholder is not equal to the count of arguments
+        if (analysis.placeholderCount != argCount) {
+            final int realArgCount = args.length;
+            final int noThrowableArgCount = realArgCount - ((args[realArgCount - 1] instanceof Throwable) ? 1 : 0);
```
ppkarwasz, line 250:
```suggestion
            final int noThrowableArgCount = realArgCount - ((args[argCount - 1] instanceof Throwable) ? 1 : 0);
```

Use `argCount` instead of `args.length`: the array can have an arbitrary length (it can be reused by multiple messages), that is why the method accepts `argCount` as argument.

ppkarwasz (approved):
Apart from the usage of `args.length`, it looks good to me.

SeasonPanPan:
> Apart from the usage of `args.length`, it looks good to me.

Done. Because the argCount might be 0 in ReusableMessage, I made a non-zero judgment in advance.

ppkarwasz:
@SeasonPanPan

PR #2404, opened 2024-03-25
Add auto-generated plugin reference to the `2.x` site

vy:
This PR adds the following new steps to the project build.

1. At `compile` phase, auto-generate plugin descriptors using an annotation processor
2. At `site` phase, using the generated plugin descriptors, auto-generate
   1. `/target/log4j-config.xsd`
   2. `/target/generated-source/site/plugin-descriptors.adoc` et al.

Employed annotation processor, XSD and AsciiDoc generators are provided by the Log4j Docgen shipped with the recently released Log4j Tools `0.8.0`.

This work is a part of #1954.

PR #2407, opened 2024-03-25
Add auto-generated plugin reference to the `3.x` site

vy:
This PR ports #2404 from `2.x` to `main` for Log4j 3.


log4j-parent/pom.xml
```diff
@@ -1028,6 +1031,118 @@
...
+          <!-- Generate a random target package for `Log4jPlugins.java` -->
+          <plugin>
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>build-helper-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>generate-plugin-package-for-tests</id>
+                <goals>
+                  <goal>bsh-property</goal>
+                </goals>
```
ppkarwasz, line 1064: This is not necessary, we could hardcode the parameter to `org.apache.logging.current.test.module` or similar.

In the current setup the tests of a module `A` do not depend on the content of `B/src/test`.

ppkarwasz (approved):
LGTM

log4j-parent/pom.xml
vy: AFAIC, this is necessary, see the following block deleted from `log4j-core-test/pom.xml`:

```
              <compilerArgs combine.children="append">
                <!-- Disable automatic package discovery, which overwrites `log4j-core` plugins. -->
                <arg>-ApluginPackage=org.apache.logging.log4j.core.test2</arg>
              </compilerArgs>
```

I would prefer to ensure a unique package path rather than hardcoding it to a constant. The latter approach is prone to causing difficult to reason about compilation failures in case of a package conflict.

ppkarwasz: I don't like an idea of adding a BSH script to the POM, for a single case (`log4j-core-test`).

Issue #2218, opened 2024-01-19
Remove `log4j-kubernetes` module


ppkarwasz:
The `log4j-kubernetes` module is problematic, because:

 * the `kubernetes-model-core` dependency follows Kubernetes lifecycle. It might be (and it already occurred) that breaking changes occurred on a minor version change,
 * on the other hand it depends only on `StrLookup`, which is unlikely to ever change,
 * the Fabric8 team has a better idea, which properties could be interesting to users.

Therefore if fabric8io/kubernetes-client#5718 is accepted, I propose to remove `log4j-kubernetes` from our codebase.

ppkarwasz:
An improved `kubernetes-log4j` module is available since Kubernetes Client version 6.11.0 (cf. [announcement](https://github.com/fabric8io/kubernetes-client/releases/tag/v6.11.0)), therefore we can remove `log4j-kubernetes`.

PR #2408, opened 2024-03-25
Remove `log4j-kubernetes` in version 3.x

ppkarwasz:
Due to differences in the lifecycle of Log4j Core and Kubernetes Client, we remove `log4j-kubernetes` from the 3.x release and redirect users to Fabric8's own `kubernetes-log4j` artifact introduced in [version 6.11.0](https://github.com/fabric8io/kubernetes-client/discussions/5834).

## Motivation

The `log4j-kubernetes` lookup depends on:

* the very stable `StrLookup` interface from Log4j Core,
* the evolving set of Kubernetes metadata provided by Kubernetes Client.

Therefore it makes more sense to distribute the lookup together with Kubernetes Client.


rgoers (approved):
This looks good to me. Should it be dropped in 2.x as well?

ppkarwasz:
> This looks good to me. Should it be dropped in 2.x as well?

I guess we can also drop it from 2.x.

vy:
+1 for dropping it from `2.x`. @ppkarwasz, could you also submit a PR for that too, please?

ppkarwasz:
@vy, I submitted a 2.x equivalent in #2412

ppkarwasz:
Closes #2218

PR #2410, opened 2024-03-26
Rebase Log4j Core 3.x on Log4j API 2.x

ppkarwasz:
This PR depends on  #2400 being accepted.

We rebase Log4j Core 3.x and its submodules on top of Log4j API 2.x, which concludes #2290.

## Main changes

- Classes moved to Log4j API 3.x (e.g. `ReflectionUtil`, `JsonReader`) are moved back to their original modules or `log4j-kit`,
- Seldom used methods like those in `NameUtil` are moved to the class of their caller,
- `Log4jExtension` is adapted to modify the global `Provider` instead of `LoggerContextFactory`,
- the `log4j-to-jul` and `log4j-to-slf4j` implementations are rolled back to their 2.x version (they were almost identical),
- a simplified OSGi activator is created in the `log4j-core` module.


vy (approved):
Skimmed through the changes, noticed nothing standing out.

PR #2411, opened 2024-03-26
Remove Log4j API 3.x

ppkarwasz:
This PR is based on #2410.

Since the Log4j API 3.x modules are no longer used, they are simply deleted.


vy (approved):
_"I like the smell of napalm in the morning."_

ppkarwasz:
My Github stats will never recover from the -50k lines hit.

PR #2412, opened 2024-03-26
Remove `log4j-kubernetes` in version 2.x

ppkarwasz:
This is the 2.x equivalent of #2408

Due to differences in the lifecycle of Log4j Core and Kubernetes Client, we remove `log4j-kubernetes` from the 2.x release and redirect users to Fabric8's own `kubernetes-log4j` artifact introduced in fabric8io/kubernetes-client#5718.

The `log4j-kubernetes` lookup depends on:

* the very stable `StrLookup` interface from Log4j Core,
* the evolving set of Kubernetes metadata provided by Kubernetes Client.

Therefore it makes more sense to distribute the lookup together with Kubernetes Client.

**Note**: the Spring Cloud manual page has been synchronized with 3.x.


ppkarwasz:
Closes #2218.

Issue #2273, opened 2024-02-08
[MS10] Log4j API generalization #1


vy:
Currently, Log4j 3 contains a new API (`log4j-api-3.x`) and this is not a good thing. We want to generalize the Log4j 2 API (`log4j-api-2.x`) as _"the Log4j API"_ and use it in Log4j 3 too. (See [the maintainer resolution](https://lists.apache.org/thread/gvk0to32c13xsfz0ytczfs506svx3hxm).) This milestone contributes to that effort.

### Log4j 2 API generalization

Log4j 2 API has several parts that are specialized for Log4j 2. This specialization makes it difficult to be a drop-in replacement for Log4j 3 API, which we intend to replace, yet Log4j 3 depends on it with several assumptions. We need to refactor these specialized Log4j 2 API mechanisms such that
1. They are isolated and self-sustained (so that it doesn't matter if they are used in a Log4j 2 or Log4j 3 context)
2. They allow plugging in 3rd party business logic (so that Log4j 2 or Log4j 3 can customize their behaviour as they see fit)

- `StatusLogger` is the lowest level component of Log4j: the logger for the logging system itself. It is crucial to isolate it from the rest to avoid unexpected cyclic dependencies, e.g, `StatusLogger` depending on `SimpleLogger` depending on `PropertiesUtil` depending on `StatusLogger` again.
  - #2249
  - #2280

### Miscellaneous improvements

In accordance with [Boy Scout Rule](https://wiki.c2.com/?BoyScoutRule), things we have improved along the way:

- #2266
- #2275

grobmeier:
Confirmed

PR #2424, opened 2024-03-28
Add logging to `PropertiesUtil` and fix `Duration` parser

### LINKED ISSUE #1969: [MS6] Unified memory management #1
This umbrella story aims to consolidate all object recycling strategies into a single abstraction. The idea is already in-use by the `log4j-layout-template-json` module: _recyclers_. In the first iteration of this pursuit, we will hoist recyclers up to `log4j-core` and employ there to replace `ThreadLocal`s.

- The initial recycler draft (#1401)
- Revamp `main` branch for the upcoming `3.0.0` major release:
  - Sync Java EE modules from `2.x` branch and move them a separate repository and website (#1966)
  - Remove [deprecated modules](/apache/logging-log4j2/discussions/1822) (#1951)

ppkarwasz:
Profiting from the `StatusLogger` enhancements in #2273, we add logging statement in `PropertiesUtil`.

We also add tests for the `getDurationProperty` method and fix its behavior.

Part of #2290.

Closes #1969

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -23,6 +23,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.time.Duration;
+import java.time.format.DateTimeParseException;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalUnit;
 import java.util.ArrayList;
@@ -36,6 +37,9 @@ import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.regex.MatchResult;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.status.StatusLogger;
 
@@ -63,6 +67,8 @@ public final class PropertiesUtil {
 
     private static final Lazy<PropertiesUtil> COMPONENT_PROPERTIES =
             Lazy.lazy(() -> new PropertiesUtil(LOG4J_PROPERTIES_FILE_NAME, false));
+
+    private static final Pattern DURATION_PATTERN = Pattern.compile("([+-]?\\d+)\\s*(\\w+)?", Pattern.CASE_INSENSITIVE);
 
     private final Environment environment;
 
@@ -200,9 +206,8 @@ public final class PropertiesUtil {
      * @return The value or null if it is not found.
      * @since 2.13.0
      */
-    @SuppressWarnings("deprecation")
     public Boolean getBooleanProperty(final String[] prefixes, final String key, final Supplier<Boolean> supplier) {
-        for (String prefix : prefixes) {
+        for (final String prefix : prefixes) {
             if (hasProperty(prefix + key)) {
                 return getBooleanProperty(prefix + key);
             }
@@ -243,7 +248,7 @@ public final class PropertiesUtil {
                 return Charset.forName(mapped);
             }
         }
-        LOGGER.error(
+        LOGGER.warn(
                 "Unable to read charset `{}` from property `{}`. Falling back to the default: `{}`",
                 charsetName,
                 name,
@@ -263,8 +268,13 @@ public final class PropertiesUtil {
         if (prop != null) {
             try {
                 return Double.parseDouble(prop);
-            } catch (final Exception ignored) {
-                // returns default value
+            } catch (final NumberFormatException e) {
+                LOGGER.warn(
+                        "Unable to read double `{}` from property `{}`. Falling back to the default: `{}`",
+                        prop,
+                        name,
+                        defaultValue,
+                        e);
             }
         }
         return defaultValue;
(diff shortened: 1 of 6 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -649,45 +674,57 @@ public static Map<String, Properties> partitionOnCommonPrefixes(
...
-        static Duration getDuration(final String time) {
-            final String value = time.trim();
-            TemporalUnit temporalUnit = ChronoUnit.MILLIS;
-            long timeVal = 0;
-            for (TimeUnit timeUnit : values()) {
-                for (String suffix : timeUnit.descriptions) {
-                    if (value.endsWith(suffix)) {
-                        temporalUnit = timeUnit.timeUnit;
-                        timeVal = Long.parseLong(value.substring(0, value.length() - suffix.length()));
+        private static long parseAmount(final MatchResult matcher) {
```
vy, line 705: This method has nothing to do with `TimeUnit`. I would rather place it right after `parseDuration` and rename it to `parseDurationAmount`.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -649,45 +674,57 @@ public static Map<String, Properties> partitionOnCommonPrefixes(
...
+            final String amount = matcher.group(1);
+            try {
+                return Long.parseLong(amount);
+            } catch (final NumberFormatException e) {
+                throw new DateTimeParseException(
+                        "Invalid time amount '" + amount + "'", matcher.group(), matcher.start(1), e);
+            }
+        }
+
+        private static TemporalUnit parseUnit(final MatchResult matcher) {
```
vy, line 715: Why does `TimeUnit#parseUnit()` receive a `MatchResult` instead of a `String`?

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -649,45 +674,57 @@ public static Map<String, Properties> partitionOnCommonPrefixes(
...
-        return getStringProperty("os.name", "").startsWith("Windows");
+        return SystemPropertiesPropertySource.getSystemProperty("os.name", "").startsWith("Windows");
+    }
+
+    static Duration parseDuration(final String value) {
+        final Matcher matcher = DURATION_PATTERN.matcher(value);
+        if (matcher.matches()) {
+            return Duration.of(TimeUnit.parseAmount(matcher), TimeUnit.parseUnit(matcher));
+        }
+        throw new DateTimeParseException("Invalid duration value: should be in the format nnn[unit].", value, 0);
```
vy, line 685: I don't think someone who is not able to correctly type a duration value can understand what is expected from them by looking at the hinted `nnn[unit]` format. Can we make the example/pattern a little bit more human-friendly?

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -23,6 +23,7 @@
 import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.time.Duration;
+import java.time.format.DateTimeParseException;
```
vy, line 26: `DateTimeParseException` doesn't display any of its input arguments when printed. Could you instead use `IllegalArgumentException` with self-explanatory messages, please?

For instance, consider the following _example_:
```
final String message = String.format("provided duration `%s` doesn't match the expected `%s` pattern", input, pattern);
throw new IllegalArgumentException(message, cause);
```

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -64,6 +68,8 @@ public final class PropertiesUtil {
     private static final Lazy<PropertiesUtil> COMPONENT_PROPERTIES =
             Lazy.lazy(() -> new PropertiesUtil(LOG4J_PROPERTIES_FILE_NAME, false));
 
+    private static final Pattern DURATION_PATTERN = Pattern.compile("([+-]?\\d+)\\s*(\\w+)?", Pattern.CASE_INSENSITIVE);
```
vy, line 71: *Nit:* I would rather dynamically build this pattern using the exact list of supported values from `TimeUnit`.

log4j-api/src/main/java/org/apache/logging/log4j/util/SystemPropertiesPropertySource.java
```diff
@@ -32,19 +34,32 @@
...
+
     /**
-     * Used by bootstrap code to get system properties without loading PropertiesUtil.
+     * Method used by Java 9+ to instantiate providers
+     * @since 2.24.0
+     * @see java.util.ServiceLoader
      */
+    public static PropertySource provider() {
+        return INSTANCE;
+    }
```
vy, line 50: 1. What do these methods do?
2. What do they have to do with this PR?

vy (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/util/SystemPropertiesPropertySource.java
vy: > A provider method is a public static method named "provider" with no formal parameters and a return type that is assignable to the service's interface or class.

States [the Java 9 `ServiceLoader` javadoc](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/util/ServiceLoader.html).

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: `matcher` is used for error handling, but I can remove it.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: I don't think such an optimization is necessary: the `getDurationProperty` method is only maintained for BC and is not used by any of our artifacts.

In Log4j Core I would prefer a standard `Duration.of` converter instead of this custom format.

vy (approved):

Issue #1206, opened 2023-01-15
Replace withers with setters in `main`


vy:
Certain accessor patterns in `master` has diverged from `release-2.x` and gravitated from setters to withers. As [discussed in the mailing list](https://lists.apache.org/thread/fd2p0z697xzm5jnxfrkx2xxgq0ny0ff4), it is better use setters everywhere in the code base.

jvz:
This would probably exclude the builders in `log4j-api` as those are already published APIs. Can update the rest, though!

renepanke:
I think this issue needs an update or can be closed. I checked out all branches and there I made the following discoveries:

1. There is no `master` branch anymore, but a `main` branch.
2. In the `main` branch there are no methods using `with` to be found.
3. Also to be absolutely sure I checked the `2.x` branch, where I can also not find methods using `with`.

_To whoever reads this: Have an awesome day :>_

PR #2425, opened 2024-03-28
Deprecate `o.a.l.l.c.a.r.a.Duration` class for removal

ppkarwasz:
We deprecate the [`o.a.l.l.c.a.r.a.Duration`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/rolling/action/Duration) and replace it with `java.time.Duration`.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
@@ -117,13 +117,13 @@ public final class IfLastModified implements PathCondition {
     public static final class Builder implements org.apache.logging.log4j.core.util.Builder<IfLastModified> {
         @PluginBuilderAttribute
         @Required(message = "No age provided for IfLastModified")
-        private Duration age;
+        private org.apache.logging.log4j.core.appender.rolling.action.Duration age;
 
         @PluginElement("nestedConditions")
         private PathCondition[] nestedConditions;
 
         public Builder setAge(final Duration age) {
-            this.age = age;
+            this.age = org.apache.logging.log4j.core.appender.rolling.action.Duration.ofMillis(age.toMillis());
             return this;
         }
 
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
@@ -27,7 +27,7 @@ import java.util.Objects;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.Core;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
-import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
 import org.apache.logging.log4j.core.config.plugins.PluginElement;
 import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
@@ -63,12 +63,6 @@ public final class IfLastModified implements PathCondition {
         return Collections.unmodifiableList(Arrays.asList(nestedConditions));
     }
 
-    /*
-     * (non-Javadoc)
-     *
-     * @see org.apache.logging.log4j.core.appender.rolling.action.PathCondition#accept(java.nio.file.Path,
-     * java.nio.file.Path, java.nio.file.attribute.BasicFileAttributes)
-     */
     @Override
     public boolean accept(final Path basePath, final Path relativePath, final BasicFileAttributes attrs) {
         final FileTime fileTime = attrs.lastModifiedTime();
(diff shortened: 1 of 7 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/convert/TypeConverters.java
```diff
@@ -224,11 +224,26 @@ public Double convert(final String s) {
...
+            return org.apache.logging.log4j.core.appender.rolling.action.Duration.parse(s);
+        }
+    }
+
+    /**
+     * Converts a {@link String} into a {@link Duration}.
+     * @since 2.24.0
+     */
+    @Plugin(name = "JavaDuration", category = CATEGORY)
+    public static class JavaDurationConverter implements TypeConverter<Duration> {
```
vy, line 246: Just checking: We don't have any usage of this yet, right?

vy (approved):
I would have rather removed `action.Duration`, but anyway...

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/convert/TypeConverters.java
ppkarwasz: Nice catch!

For compatibility sake I had to revert the `age` field in the `IfLastModified` builder to "our" `Duration`. Therefore this one is not used.

norrisjeremy:
Hi @ppkarwasz,

Can you update the release notes for 2.24.0 to document the fact that this deprecation also means that the relaxed syntax of `with 'P' and 'T' optional` is also thus deprecated?
This is not clearly defined in the release notes and thus means consumers of [o.a.l.l.c.a.r.a.Duration](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/rolling/action/Duration) that naively migrate to `java.time.Duration` can thus suddenly encounter `java.time.format.DateTimeParseException`'s for strings that could previously be parsed without issue.

Thanks,
Jeremy

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
```diff
@@ -79,35 +79,62 @@ public boolean accept(final Path basePath, final Path relativePath, final BasicF
...
+     */
+    public static final class Builder implements org.apache.logging.log4j.core.util.Builder<IfLastModified> {
+        @PluginBuilderAttribute
+        @Required(message = "No age provided for IfLastModified")
+        private org.apache.logging.log4j.core.appender.rolling.action.Duration age;
+
+        @PluginElement("nestedConditions")
+        private PathCondition[] nestedConditions;
+
+        public Builder setAge(final Duration age) {
```
norrisjeremy, line 125: Shouldn't this be spelled as `withAge()` to be more consistent with other Builders syntax?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
```diff
@@ -79,35 +79,62 @@ public boolean accept(final Path basePath, final Path relativePath, final BasicF
...
+            this.age = org.apache.logging.log4j.core.appender.rolling.action.Duration.ofMillis(age.toMillis());
+            return this;
+        }
+
+        public Builder setNestedConditions(final PathCondition... nestedConditions) {
```
norrisjeremy, line 130: Shouldn't this be spelled as `withNextedConditions()` to be more consistent with other Builders syntax?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
ppkarwasz: Usually a `with*` method is provided by _immutable_ objects and creates a new instance of the object, so `set*` is more appropriate here.
In `2.x` you have a mixture of "setters" and "withers" and if both are present, the "wither" is **deprecated**.

The consensus is to replace all "withers" with "setters" in `3.x` (see #1206).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
ppkarwasz: See comment above: new methods are introduced as "setters" so they don't need to be deprecated in `3.x`.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
norrisjeremy: Ok, understood.
Was just confused because other nearby Builder's already used the `withXyz()` syntax, so it seemed inconsistent when I encountered it here.

norrisjeremy:
> Hi @ppkarwasz,
> 
> Can you update the release notes for 2.24.0 to document the fact that this deprecation also means that the relaxed syntax of `with 'P' and 'T' optional` is also thus deprecated? This is not clearly defined in the release notes and thus means consumers of [o.a.l.l.c.a.r.a.Duration](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/rolling/action/Duration) that naively migrate to `java.time.Duration` can thus suddenly encounter `java.time.format.DateTimeParseException`'s for strings that could previously be parsed without issue.
> 
> Thanks, Jeremy

I also am not sure if it would be worthwhile calling out to users in the Release Notes that configuration elements such as `<IfLastModified age="7d"/>` are also thus deprecated and should be updated to syntax such as `<IfLastModified age="P7D"/>`?

vy:
> I also am not sure if it would be worthwhile calling out to users in the Release Notes that configuration elements such as `<IfLastModified age="7d"/>` are also thus deprecated and should be updated to syntax such as `<IfLastModified age="P7D"/>`?

@norrisjeremy, to keep the backward compatibility, Log4j 2 still uses the (deprecated) Log4j `Duration` class to read `IfLastModified#age` – the new `j.t.Duration` syntax is not supported yet. We only marked the class and its usages as deprecated to warn users accessing those fields programmatically, etc. But you have a point: Log4j 3 completely switches to `j.t.Duration` and this will break incompatible `IfLastModified` configurations. That said, Log4j 3 will be a big bang anyway... :shrug:

PR #2426, opened 2024-03-28
Remove `o.a.l.l.c.a.r.a.Duration` class

ppkarwasz:
This is the 3.x counterpart of the `o.a.l.l.c.a.r.a.Duration` deprecation in #2425.

It also makes the `Clock` argument for the `IfLastModified` plugin compulsory. The argument is usually supplied by injection.


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
```diff
@@ -97,17 +91,24 @@ public String toString() {
...
+     * @since 2.24.0
+     */
+    public static final class Builder implements Supplier<IfLastModified> {
         private Duration age;
         private PathCondition[] nestedConditions;
         private Clock clock;
 
-        public Builder setAge(@PluginAttribute final Duration age) {
+        public Builder setAge(
+                @PluginAttribute @Required(message = "No age provided for IfLastModified") final Duration age) {
```
vy, line 111: Is this `message` really needed? I think the DI should be smart enough to generate that, right? Doesn't it?

vy (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/IfLastModified.java
ppkarwasz: The default message is somehow generic: `The parameter is null or empty: age`.

The message I added is the same as in `2.x`.

PR #2435, opened 2024-04-01
Remove CLI tools from `log4j-core` and `log4j-1.2-api`

ppkarwasz:
We remove the command line tools to generate logger wrappers and convert Log4j 1.x configuration.

The tools have been submitted to Log4j Transform:

 * apache/logging-log4j-transform#92
 * apache/logging-log4j-transform#93

and are not necessary at runtime.

The `Log4j1ConfigurationFactory` that has been removed from `log4j-1.2-api` is an **older** and deprecated version of the `PropertiesConfigurationFactory`.


ppkarwasz:
The build failure is an April's Fool I played on myself: the failing test was disabled until `2024-04-01`. I'll fix it in `main`.

vy (approved):
You remove Java files and that does not cause any changes in the source code. Hilarious! A solid indicator of a-feature-that-could-have-been-a-separate-project. Good job @ppkarwasz!

Could you double-check that the `src/site` doesn't need to be updated, please?

PR #2443, opened 2024-04-04
Migrate `3.x` website to Antora

vy:
This work ports #2427 to `main`, i.e., Log4j 3.


src/site/resources/.htaccess
```diff
@@ -8,13 +7,16 @@
...
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-# ----------------------------------------------------------------------------
-./mvnw spotless:apply -pl '!log4j-bom' -T2C
+#
+RewriteEngine On
+RewriteRule "^(/log4j/[23].x)/log4j-(core|api)/apidocs(.*)" "$1/javadoc/$2$3" [R=permanent]
+RewriteRule "^(/log4j/[23].x)/manual/scala-api.html" "/log4j/scala" [R=permanent]
+RewriteRule "^(/log4j/[23].x)/release-notes/index.html" "$1/release-notes.html" [R=permanent]
```
ppkarwasz, line 22:
```suggestion
RewriteEngine On
RewriteBase "/log4j/3.x/"
RewriteRule "^manual/scala-api(\.html)?$" "/log4j/scala/" [R=permanent]
RewriteRule "^release-notes/(index(\.html)?)?$" "release-notes" [R=permanent]
```

This seems to work on my local server.

Remark that according to Git history we never had a `/log4j/3.x/log4j-api/apidocs/` folder.

It is also pretty much a personal taste, but I prefer to use URLs that do not end in `.html`. Redirects to folders on the other way must terminate in `/` to spare the client an additional redirection.

src/site/resources/.htaccess
```diff
@@ -1,5 +1,4 @@
-#!/bin/sh
-# ----------------------------------------------------------------------------
+#
```
ppkarwasz, line 1: _Nit_: this hidden file in a standalone folder is hard to find.

Could we have everything under the same tree? Also is `src/site/antora` the native location for Antora? If it is not, could we move it?

ppkarwasz (approved):
It is hard to review more than a hundred of files, but I assume most of these changes are in common with #2427.

src/site/resources/.htaccess
vy: I will discuss this personally and apply @ppkarwasz's suggestion to both `2.x` and `main`.

PR #2450, opened 2024-04-08
Simplify `DefaultScopedContextProvider`

ppkarwasz:
This PR simplifies the logic of `DefaultScopedContextProvider`, by using two kinds of `ScopedContext.Instance` implementations:

- `MapInstance` is a simple container for an immutable map and is the only implementation that can be set in the `ThreadLocal`,
- `KeyValueInstance` represents a node of a linked list of `key/value` pairs. It is used to record differences from a `MapInstance` and is converted into a `MapInstance` in the `wrap` methods.


ppkarwasz:
I closed it, since a similar solution appeared in #2438

Issue #1799, opened 2023-09-18
Spring Boot shudown fails with an error


rgoers:
## Description

This issue is captured in https://github.com/spring-projects/spring-boot/issues/33450. During shutdown properties are being accessed, which causes the Spring Property Source to be invoked. However, it needs an active LoggerContext to work.

## Configuration

**Version:** 2.20.0

## Logs

```
[Stacktraces, errors, etc. relevant applications logs.]
```

## Reproduction

While building a sample Spring Boot app I ran into a similar, but related, error.

WARN StatusConsoleListener Unable to register Log4j shutdown hook because JVM is shutting down. Using SimpleLogger
Exception in thread "SpringApplicationShutdownHook" java.lang.ClassCastException: class org.apache.logging.log4j.simple.SimpleLoggerContext cannot be cast to class org.apache.logging.log4j.core.LoggerContext (org.apache.logging.log4j.simple.SimpleLoggerContext and org.apache.logging.log4j.core.LoggerContext are in unnamed module of loader org.springframework.boot.loader.LaunchedURLClassLoader @5b480cf9)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.getLoggerContext(Log4J2LoggingSystem.java:480)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.lambda$getShutdownHandler$2(Log4J2LoggingSystem.java:452)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.springframework.boot.SpringApplicationShutdownHook.run(SpringApplicationShutdownHook.java:110)
	at java.base/java.lang.Thread.run(Thread.java:833)

We have seen this in other cases but have never come up with a permanent solution.

rgoers:
Fixed in 3.x with https://github.com/apache/logging-log4j2/pull/2062

PR #2454, opened 2024-04-09
Backport part of #2062 to `2.x` branch.

### LINKED ISSUE #2453: spring boot 3.2.4 + spring-boot-starter-log4j2 + spring-boot-starter-undertow
## Stop service error

2024-04-09 11:36:26.124 traceId: INFO  c.z.h.HikariDataSource                   HikariPool-1 - Shutdown completed.
ERROR StatusConsoleListener Unable to unregister MBeans
 java.lang.IllegalStateException: UT015023: This Context has been already destroyed
	at io.undertow.servlet.spec.ServletContextImpl.getDeploymentInfo(ServletContextImpl.java:210)
	at io.undertow.servlet.spec.ServletContextImpl.getInitParameterNames(ServletContextImpl.java:434)
	at org.springframework.web.context.support.ServletContextPropertySource.getPropertyNames(ServletContextPropertySource.java:41)
	at org.springframework.core.env.EnumerablePropertySource.containsProperty(EnumerablePropertySource.java:74)
	at org.springframework.core.env.PropertySourcesPropertyResolver.containsProperty(PropertySourcesPropertyResolver.java:51)
	at org.springframework.boot.context.properties.source.ConfigurationPropertySourcesPropertyResolver.containsProperty(ConfigurationPropertySourcesPropertyResolver.java:56)
	at org.springframework.core.env.AbstractEnvironment.containsProperty(AbstractEnvironment.java:546)
	at org.springframework.boot.logging.log4j2.SpringEnvironmentPropertySource.containsProperty(SpringEnvironmentPropertySource.java:55)
	at org.apache.logging.log4j.util.PropertiesUtil$Environment.get(PropertiesUtil.java:531)
	at org.apache.logging.log4j.util.PropertiesUtil$Environment.access$200(PropertiesUtil.java:448)
	at org.apache.logging.log4j.util.PropertiesUtil.getStringProperty(PropertiesUtil.java:396)
	at org.apache.logging.log4j.util.PropertiesUtil.getBooleanProperty(PropertiesUtil.java:169)
	at org.apache.logging.log4j.util.PropertiesUtil.getBooleanProperty(PropertiesUtil.java:158)
	at org.apache.logging.log4j.core.jmx.Server.isJmxDisabled(Server.java:131)
	at org.apache.logging.log4j.core.jmx.Server.unregisterLoggerContext(Server.java:244)
	at org.apache.logging.log4j.core.LoggerContext.stop(LoggerContext.java:364)
	at org.apache.logging.log4j.core.AbstractLifeCycle.stop(AbstractLifeCycle.java:136)
	at org.springframework.boot.logging.log4j2.Log4J2LoggingSystem.lambda$getShutdownHandler$2(Log4J2LoggingSystem.java:453)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.springframework.boot.SpringApplicationShutdownHook.run(SpringApplicationShutdownHook.java:116)
	at java.base/java.lang.Thread.run(Thread.java:1583)

ppkarwasz:
This backports the part of PR #2062 that concerns spring-projects/spring-boot#33450 to the `2.x` branch.

It contains two sets of changes:

- it catches exceptions thrown by property sources and logs them as warnings,
- it adds a `PropertiesUtil#removePropertySource` to allow Spring Boot to remove its custom property source, when the Spring `Environment` becomes unavailable at shutdown.

Note that the second change will not be effective **until** Spring Boot starts using it, which can happen only after a `2.24.0` release of Log4j API.

**Remark**: the part of PR #2062 that concerns #1799 was not backported, since I believe it to be a Spring Boot problem:

1. Spring Boot should not perform an unchecked cast of `o.a.l.l.spi.LoggerContext`,
2. Spring Boot should not call `LogManager.getContext` multiple times, but cache the `LoggerContext` it received at startup.

Fixes #2453

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
@@ -36,13 +36,16 @@ import java.util.ResourceBundle;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.status.StatusLogger;
+import org.apache.logging.log4j.util.PropertySource.Comparator;
 
 /**
  * <em>Consider this class private.</em>
@@ -139,13 +142,21 @@ public final class PropertiesUtil {
     }
 
     /**
-     * Allows a PropertySource to be added after PropertiesUtil has been created.
-     * @param propertySource the PropertySource to add.
+     * Allows a {@link PropertySource} to be added after {@code PropertiesUtil} has been created.
+     * @param propertySource the {@code PropertySource} to add.
+     * @since 2.19.0
      */
     public void addPropertySource(final PropertySource propertySource) {
-        if (environment != null) {
-            environment.addPropertySource(propertySource);
-        }
+        environment.addPropertySource(Objects.requireNonNull(propertySource));
+    }
+
+    /**
+     * Removes a {@link PropertySource}.
+     * @param propertySource the {@code PropertySource} to remove.
+     * @since 2.24.0
+     */
+    public void removePropertySource(final PropertySource propertySource) {
+        environment.removePropertySource(Objects.requireNonNull(propertySource));
     }
 
     /**
@@ -488,7 +499,11 @@ public final class PropertiesUtil {
      */
     private static final class Environment {
 
-        private final Set<PropertySource> sources = new ConcurrentSkipListSet<>(new PropertySource.Comparator());
+        private final List<PropertySource> sources = new ArrayList<>();
+        private final ReadWriteLock sourceLock = new ReentrantReadWriteLock();
+        private final Lock sourceReadLock = sourceLock.readLock();
+        private final Lock sourceWriteLock = sourceLock.writeLock();
+
         /**
          * Maps a key to its value or the value of its normalization in the lowest priority source that contains it.
          */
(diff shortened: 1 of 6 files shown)


ppkarwasz:
Closes #2453

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
```diff
@@ -101,10 +101,12 @@ class Comparator implements java.util.Comparator<PropertySource>, Serializable {
...
-        public int compare(final PropertySource o1, final PropertySource o2) {
-            return Integer.compare(
-                    Objects.requireNonNull(o1).getPriority(),
-                    Objects.requireNonNull(o2).getPriority());
+        public int compare(final PropertySource left, final PropertySource right) {
+            Objects.requireNonNull(left);
+            Objects.requireNonNull(right);
+            final int result = Integer.compare(left.getPriority(), right.getPriority());
+            // Two property sources can have the same priority
+            return result != 0 || left.equals(right) ? result : left.hashCode() - right.hashCode();
```
vy, line 109: I agree with Phillip Webb's proposal in [LOG4J2-3618](https://issues.apache.org/jira/browse/LOG4J2-3618):

> use `equals/hashCode` to determine if items can be added, and only use the Comparator for sorting

I am not able to see how this achieves that. I would be in favor of the following instead:
1. Leave this comparator as is (only compare using ranks)
2. Initialize the `PropertiesUtil.Environment#sources` using `new ConcurrentSkipListSet()`, i.e., no explicit comparator
3. Find iterations over `PropertiesUtil.Environment#sources`, and employ the comparator there (i.e., `.stream().sorted(Comparator.INSTANCE).forEach(...)`) _iff_ an order is needed.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -540,18 +552,18 @@ private synchronized void reload() {
                 final List<CharSequence> tokens = PropertySource.Util.tokenize(key);
                 final boolean hasTokens = !tokens.isEmpty();
                 sources.forEach(source -> {
```
vy, line 554: Why do we only guard against the particular failure cases reported rather than the entire block? That is, I was expecting:
```
keys.stream().filter(Objects.nonNull).forEach(key -> {
    try {
       // Old logic goes here verbatim
    } catch (final Exception error) {
      // Report the failure
    }
});
for (final PropertySource source : sources) {
    try {
       // Old logic goes here verbatim
    } catch (final Exception error) {
      // Report the failure
    }
}
```

vy (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
rgoers: Reporting failures will actually generate bug reports. It is very likely that Log4j will try to load the Spring PropertySource before the Spring environment exists. As I recall that will cause an exception during startup of every Spring Boot application, which obviously is unacceptable.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertySource.java
ppkarwasz: Fixed in 2a40dc5bfa2a368b432f057194e6241f6626bb85

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: If a single property source fails, we don't want the entire lookup to fail.

Anyway I am not a big fan of this _preemptive_ caching strategy adopted by `PropertiesUtil`. In practice the caching process is quite slow, since the values of properties such a `user.home` or all environment variables are cached.
On the other hand users rarely use Log4j configuration properties, so most of the dozens of calls to `PropertiesUtil` result in cache misses.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
vy, line 1: [As recommended by Java itself](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/Lock.html), shouldn't all `lock()` calls be placed outside the `try` block?

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
```diff
@@ -517,45 +532,64 @@ private Environment(final PropertySource propertySource) {
...
-         * @param propertySource The PropertySource to add.
-         */
-        public void addPropertySource(final PropertySource propertySource) {
-            sources.add(propertySource);
+        private void addPropertySource(final PropertySource propertySource) {
+            try {
+                sourceWriteLock.lock();
+                if (!sources.contains(propertySource)) {
+                    sources.add(propertySource);
+                    sources.sort(Comparator.INSTANCE);
```
vy, line 545: I thought we agreed to only sort while iterating, but, AFAICT, we are doing something else. Any particular reason for this choice?

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: Fixed in 1e437b777abc0062518a64463e5e306896435e78.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: `ConcurrentSkipListSet` can not be used, since it needs a `Comparator` (either provided or the class must be `Comparable`).

This is why we use a `List` to store the property sources. Sorting the list allows the use of `List.forEach`, which is less expensive than streams.

vy: I am not able to follow. What is wrong with *only* doing `sources.stream().sorted(Comparator.INSTANCE).forEach()`?

ppkarwasz: I didn't run any benchmark, but I would say that it is slower than `sources.forEach()`.

vy: How often `reload()` is called? Is the performance difference in the big picture notable enough to warrant the extra optimization? If not, I am inclined to go with the simplest approach.

I am not a big fan on these kind of discussions. Please address this conversation as you see fit, resolve it, and let's merge this PR.

ppkarwasz:
I am squashing the PR before a merge.

log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java
ppkarwasz: I ran some benchmarks of these methods:
```java

public final List<PropertySource> arrayList = new ArrayList<>();
public final Set<PropertySource> skipSet = ConcurrentHashMap.newKeySet();

@Benchmark
public void arrayList(final Blackhole bh) {
    sourceReadLock.lock();
    try {
        for (int i = 0; i < count; i++) {
            bh.consume(arrayList.get(i).getProperty("KEY"));
        }
    } finally {
        sourceReadLock.unlock();
    }
}

@Benchmark
public void skipList(final Blackhole bh) {
    skipSet.stream().sorted(comparator).forEach(source -> bh.consume(source.getProperty("KEY")));
}
```

with 8 property sources and the following results:

```
Benchmark                  (count)  Mode  Cnt    Score   Error  Units
StreamBenchmark.arrayList        8  avgt   25   20,696 ± 0,188  ns/op
StreamBenchmark.skipList         8  avgt   25  111,643 ± 4,113  ns/op
```

Assuming a 100 `getProperty` calls during initialization we are talking about a difference between `~2 µs` and `~10 µs`. It's probably not worth the additional code size of the `ArrayList`.

PR #2481, opened 2024-04-17
Deprecate `log4j2.defaultStatusLevel` property

ppkarwasz:
The Log4j system property `log4j2.defaultStatusLevel` has almost the same purpose as the `log4j2.statusLoggerLevel`:

* the `log4j2.statusLoggerLevel` property sets the level of the fallback status logger listener **before** a configuration is parsed,
* the `log4j2.defaultStatusLevel` property sets the level of the fallback status logger listener **after** a configuration is parsed (unless the configuration has a `status` attribute).

Removing one of them makes `-Dlog4j2.statusLoggerLevel=WARN` a viable less verbose alternative to `-Dlog4j2.debug`.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/xml/XmlConfigurationPropsTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/xml/XmlConfigurationPropsTest.java
@@ -72,9 +72,17 @@ class XmlConfigurationPropsTest {
     }
 
     @Test
+    @SetTestProperty(key = StatusLogger.DEFAULT_STATUS_LISTENER_LEVEL, value = "INFO")
     @SetTestProperty(key = Constants.LOG4J_DEFAULT_STATUS_LEVEL, value = "WARN")
     @LoggerContextSource(value = CONFIG1)
     void testDefaultStatus(final Configuration config) {
+        testConfiguration(config, CONFIG1_NAME, Level.INFO, null);
+    }
+
+    @Test
+    @SetTestProperty(key = Constants.LOG4J_DEFAULT_STATUS_LEVEL, value = "WARN")
+    @LoggerContextSource(value = CONFIG1)
+    void testDeprecatedDefaultStatus(final Configuration config) {
         testConfiguration(config, CONFIG1_NAME, Level.WARN, null);
     }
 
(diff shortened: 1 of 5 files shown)


src/site/antora/modules/ROOT/pages/manual/configuration.adoc
```diff
@@ -2219,18 +2216,6 @@ The following is a list of available global configuration properties. Note that
...
-  The StatusLogger logs events that occur in the logging system to the console.
-  During configuration, AbstractConfiguration registers a StatusConsoleListener with the StatusLogger that may
-  redirect status log events from the default console output to a file.
-  The listener also supports fine-grained filtering.
-  This system property specifies the default status log level for the listener to use if the configuration does not specify a status level.
-  Note: this property is used by the log4j-core implementation only after a configuration file has been found.
-
 | [[statusLoggerLevel]]log4j2.statusLoggerLevel
   ([[log4j2.StatusLogger.level]]log4j2.StatusLogger.level)
 | LOG4J_STATUS_LOGGER_LEVEL
```
vy, line 2221: This is incorrect, the default is `ERROR, right?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/xml/XmlConfigurationPropsTest.java
```diff
@@ -72,9 +72,17 @@ void testNoProps(final Configuration config) {
     }
 
     @Test
+    @SetTestProperty(key = StatusLogger.DEFAULT_STATUS_LISTENER_LEVEL, value = "INFO")
     @SetTestProperty(key = Constants.LOG4J_DEFAULT_STATUS_LEVEL, value = "WARN")
```
vy, line 76: In order to test `DEFAULT_` has precedence over `LOG4J_`, shouldn't the latter has a more specific level than `WARN`? That is, shouldn't `LOG4J_` level be, say, `DEBUG`?

vy (approved):

src/site/antora/modules/ROOT/pages/manual/configuration.adoc
ppkarwasz: Nice catch!

Somehow users never found this error in all these years, maybe because `log4j2.statusLoggerLevel` was always broken.

ppkarwasz: Done

PR #2500, opened 2024-04-23
Close stale issues

ppkarwasz:
We add a GitHub action to automatically close GitHub issues that:

- are marked as `waiting-for-user`,
- have been inactive for more than 60 days.

### DIFF
--- a/.github/workflows/close-stale.yaml
+++ b/.github/workflows/close-stale.yaml
@@ -30,6 +30,7 @@ jobs:
     steps:
       - uses: actions/stale@28ca1036281a5e5922ead5184a1bbf96e5fc984e    # v9.0.0
         with:
+          # Labels need to match with the ones in `labeler.yaml`!
           only-labels: waiting-for-user
           stale-issue-message: >
             This issue is stale because it has been waiting for your feedback for more than 60 days.
--- a/.github/workflows/close-stale.yaml
+++ b/.github/workflows/close-stale.yaml
@@ -1,0 +1,39 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to you under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


vy (commented):
@ppkarwasz, can you explain how this will work in practice?

Imagine a maintainer labels an issue, and the user promptly responds. Who will remove the label? If the maintainer doesn't react and remove the label, is the action smart enough to **not** close the issue?

I have the inkling that labeling should ideally be automated by an action too.

.github/workflows/close-stale.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          only-labels: waiting-for-user
+          stale-issue-message: >
+            This issue is stale because it has been waiting for your feedback for more than 60 days.
```
vy, line 35: Shall we explicitly add the `days-before-stale` parameter rather than relying on the default which might change after a version upgrade?

.github/workflows/close-stale.yaml
```diff
@@ -0,0 +1,39 @@
...
+            The Apache Logging Services community values every submitted issue, but without additional information from you,
+            we are unable to provide a solution to your problem.
+
+            Please comment on this issue or it will be closed in 7 days.
```
vy, line 39: Shall we explicitly add the `days-before-close` parameter?

.github/workflows/close-stale.yaml
```diff
@@ -0,0 +1,39 @@
...
+  workflow_dispatch:
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
```
vy, line 31:
```suggestion
      - uses: actions/stale@28ca1036281a5e5922ead5184a1bbf96e5fc984e    # v9.0.0
```

.github/workflows/close-stale.yaml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          only-labels: waiting-for-user
```
vy, line 33:
```suggestion
          # Labels need to match with the ones in `labeler.yaml`!
          only-labels: waiting-for-user
```

.github/workflows/labeler.yaml
```diff
@@ -0,0 +1,44 @@
...
+  issues: write
+
+jobs:
+  check-labels:
+    # Prevents the job from running on pull requests
+    if: ${{ ! github.event.issue.pull_request }}
+    runs-on: ubuntu-latest
+
+    steps:
+     - name: Add `waiting-for-maintainer` label
```
vy, line 38:
```suggestion

     # The `waiting-for-maintainer` label needs to match with the one in `close-stale.yaml`!
     - name: Add `waiting-for-maintainer` label
```

vy (approved):

PR #2505, opened 2024-04-24
Fix handling of `log4j2.messageFactory` and `log4j2.flowMessageFactory` properties

### LINKED ISSUE #2379: The `log4j2.messageFactory` property is ignored since 2.23.0
## Description

Due to the changes in version `2.23.0`, specifying an alternative default message factory through the `log4j2.messageFactory`
property does not work any more.

vy:
Fixes #2379.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
@@ -16,6 +16,8 @@
  */
 package org.apache.logging.log4j.core;
 
+import static java.util.Objects.requireNonNull;
+
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
```diff
@@ -63,16 +77,82 @@ public class Logger extends AbstractLogger implements Supplier<LoggerConfig> {
...
+            } catch (final Throwable error) {
+                StatusLogger.getLogger()
+                        .error(
+                                "failed instantiating the `{}` class obtained from the `{}` property",
+                                className,
+                                propertyName,
+                                error);
+            }
+        }
+        return fallbackInstanceSupplier.get();
```
ppkarwasz, line 155: You could use `LoaderUtil#newCheckedInstanceOfProperty` here.

ppkarwasz (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/Logger.java
vy: Nice! Fixed in 097853933f8a19aa6333dd62d7fa5c06250452be.

ppkarwasz (approved):

PR #2517, opened 2024-04-25
Prevent `log4j2.isWebapp` from overriding other constants

ppkarwasz:
Currently, if `log4j2.isWebapp` is enabled, the effective values of the following system properties:

* `log4j2.enableThreadlocals`,
* `log4j2.garbagefreeThreadLocalMap`,
* `log4j2.shutdownHookEnabled`.

is set to `false`.

After this PR:

* the above mentioned constants are set to the value **explicitly** set by the user,
* if the user didn't provide any value `!log4j2.isWebapp` is used.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
@@ -326,10 +326,10 @@ public class Provider {
             return threadContextMapClass;
         }
         // Default based on properties
-        if (props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY, !Constants.IS_WEB_APP)) {
+        if (props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY)) {
             return GARBAGE_FREE_CONTEXT_MAP;
         }
-        return Constants.IS_WEB_APP ? WEB_APP_CONTEXT_MAP : COPY_ON_WRITE_CONTEXT_MAP;
+        return Constants.ENABLE_THREADLOCALS ? COPY_ON_WRITE_CONTEXT_MAP : WEB_APP_CONTEXT_MAP;
     }
 
     /**
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
@@ -326,12 +326,10 @@ public class Provider {
             return threadContextMapClass;
         }
         // Default based on properties
-        if (Constants.ENABLE_THREADLOCALS) {
-            return props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY)
-                    ? GC_FREE_THREAD_CONTEXT_PROPERTY
-                    : COPY_ON_WRITE_CONTEXT_MAP;
+        if (props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY, !Constants.IS_WEB_APP)) {
+            return GARBAGE_FREE_CONTEXT_MAP;
         }
-        return WEB_APP_CONTEXT_MAP;
+        return Constants.IS_WEB_APP ? WEB_APP_CONTEXT_MAP : COPY_ON_WRITE_CONTEXT_MAP;
     }
 
     /**
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
@@ -30,7 +30,7 @@ public final class Constants {
      *     If {@code true} Log4j disables the features, which are incompatible with a typical servlet application:
      * </p>
      * <ol>
-     *     <li>It disables the usage of {@link ThreadLocal}s for object pooling (see {@link #ENABLE_THREADLOCALS},</li>
+     *     <li>It disables the usage of {@link ThreadLocal}s for object pooling (unless the user has explicitly provided a {@link #ENABLE_THREADLOCALS} property)</li>
      *     <li>It uses a web-application safe implementation of {@link org.apache.logging.log4j.spi.ThreadContextMap}
      *     (see {@link Provider#getThreadContextMap()}),</li>
      *     <li>It disables the shutdown hook,</li>
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
@@ -16,6 +16,8 @@
  */
 package org.apache.logging.log4j.util;
 
+import org.apache.logging.log4j.spi.Provider;
+
 /**
  * Log4j API Constants.
  *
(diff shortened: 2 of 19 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -326,12 +326,10 @@ public String getThreadContextMap() {
...
         // Default based on properties
-        if (Constants.ENABLE_THREADLOCALS) {
-            return props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY)
-                    ? GC_FREE_THREAD_CONTEXT_PROPERTY
-                    : COPY_ON_WRITE_CONTEXT_MAP;
+        if (props.getBooleanProperty(GC_FREE_THREAD_CONTEXT_PROPERTY, !Constants.IS_WEB_APP)) {
+            return GC_FREE_THREAD_CONTEXT_PROPERTY;
         }
-        return WEB_APP_CONTEXT_MAP;
+        return Constants.IS_WEB_APP ? WEB_APP_CONTEXT_MAP : COPY_ON_WRITE_CONTEXT_MAP;
```
ppkarwasz, line 332: This change probably merits a longer discussion.

If the user does not provide a value for `enableThreadlocals`, the type of thread context map is determined by `isWebapp` and `garbagefreeThreadContextMap` in the following way:

## Before

|          |       | garbagefreeThreadContextMap |             |             |
|----------|-------|-----------------------------|-------------|-------------|
|          |       | `false`                       | `true`        | `null`        |
| **isWebapp** | `false` | CopyOnWrite                 | GargageFree | CopyOnWrite |
|          | `true`  | WebApp                      | WebApp      | WebApp      |


## After

|          |       | garbagefreeThreadContextMap |             |             |
|----------|-------|-----------------------------|-------------|-------------|
|          |       | `false`                       | `true`        | `null`        |
| **isWebapp** | `false` | CopyOnWrite                 | GarbageFree | GarbageFree |
|          | `true`  | WebApp                      | GarbageFree      | WebApp      |

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
vy, line 1: Can you explain how the thread-local toggling property is taken into account in the new implementation?

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
```diff
@@ -16,31 +16,56 @@
...
     /**
-     * {@code true} if we think we are running in a web container, based on the boolean value of system property
-     * "log4j2.is.webapp", or (if this system property is not set) whether the  {@code javax.servlet.Servlet} class
-     * is present in the classpath.
+     * Specifies whether Log4j is used in a servlet container
+     * <p>
+     *     If {@code true} Log4j disables the features, which are incompatible with a typical servlet application:
+     * </p>
+     * <ol>
+     *     <li>It disables the usage of {@link ThreadLocal}s for object pooling (see {@link #ENABLE_THREADLOCALS},</li>
```
vy, line 33:
```suggestion
     *     <li>It disables the usage of {@link ThreadLocal}s for object pooling (unless the user has explicitly provided a {@link #ENABLE_THREADLOCALS} property)</li>
```

log4j-api/src/main/java/org/apache/logging/log4j/util/Constants.java
```diff
@@ -16,31 +16,56 @@
...
+     *     true}.
+     * </p>
      * <p>
-     * {@code True} for non-{@link #IS_WEB_APP web apps}, disable by setting system property
-     * "log4j2.enable.threadlocals" to "false".
+     *     <strong>Warning:</strong> This setting does <strong>not</strong> disable the usage of {@code ThreadLocal}s
+     *     for other purposes than object pooling. For example the {@link org.apache.logging.log4j.ThreadContext}
+     *     API, will user {@code ThreadLocal}s even if this constant is set to {@code false}.
      * </p>
+     * @see Provider#getThreadContextMap()
```
vy, line 65: 1. I think we should link to docs, not to a ticket
2. Narrowing the scope to only pooling is not a good idea. Pooling is already configurable in some places. This flag should be the TLA kill switch, regardless of where it is used. Is there a particular reason you took this route?

src/site/antora/modules/ROOT/pages/manual/json-template-layout.adoc
```diff
@@ -1085,7 +1085,7 @@ Resolves `logEvent.getMessage().getParameters()`.
 Regarding garbage footprint, `stringified` flag translates to
 `String.valueOf(value)`, hence mind not-`String`-typed values. Further,
 `logEvent.getMessage()` is expected to implement `ParameterVisitable` interface,
-which is the case if `log4j2.enableThreadLocals` property set to true.
+which is the case if `log4j2.enableThreadlocals` property set to true.
```
vy, line 1088: I'd be in favor of using `enableThreadLocals` (just like many other camelcase property names we have) and consistently using it everywhere in the code.

ppkarwasz: Me, too, except backward compatibility prevents me from doing it. :wink:

The original property is `log4j2.enable.threadlocals`, which has 2 tokens `enable` and `threadlocals`. Therefore the normalized form is `log4j2.enableThreadlocals`.

ppkarwasz: The matching is not so fuzzy. For a given legacy property name (`log4j2.enable.threadlocals`), `PropertiesUtil` looks for:

* a property of the same name,
* a property with a normalized name (`LOG4J_ENABLE_THREADLOCALS` for environment variables, `log4j2.enableThreadlocals` for all other property sources),
* a property that contains the same tokens (e.g. `org.apache.logging.log4j.enableThreadlocals`). The tokenizer is in `PropertySource.Util`.

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
ppkarwasz: I have improved the implementation:

- if the user sets `log4j2.garbagefreeThreadContextMap` to `true`, then a garbage free implementation of `ThreadContextMap` will be used regardless of other settings (a specific setting beats a generic one),
- otherwise if `log4j2.enableThreadlocals` is `true`, the copy-on-write implementation of `ThreadContextMap` will be used,
- otherwise an implementation safe from memory-leaks will be used.

**Remark**: after we fix #2523, the copy-on-write implementation can also be used with `log4j2.enableThreadlocals` set to `false`.

ppkarwasz: On the contrary: you pointed out a valid concern.

After looking at all `ThreadLocal` usages in Log4j, e.g. this one:

https://github.com/apache/logging-log4j2/blob/cc876ae7c629b698b92414b7f3c94213cdd0a427/log4j-core/src/main/java/org/apache/logging/log4j/core/config/AppenderControl.java#L39

I came to the conclusion that our `log4j2.enableThreadlocals` contract can be: "if the value is `false` we **do** use `ThreadLocal`s, but we clear them up before returning to user code".

The `ThreadContext` class is not different, but since the user has an almost direct access to the `ThreadLocal` the contract should be slightly different: "if you use `ThreadContext` properly and the value is `false` we **do** use `ThreadLocals`s, but you'll clear them up after your code executes".

Therefore I believe that the algorithm should be:

- if `ENABLE_THREADLOCALS` is `true`: use the garbage-free implementation (which does not clear the thread local after usage),
- otherwise we use either copy-on-write or the default implementation. I think the former is more performant, but can cause classloader leaks if the user does not use it properly.

vy: > https://github.com/apache/logging-log4j2/blob/cc876ae7c629b698b92414b7f3c94213cdd0a427/log4j-core/src/main/java/org/apache/logging/log4j/core/config/AppenderControl.java#L39
> 
> I came to the conclusion that our `log4j2.enableThreadlocals` contract can be: "if the value is `false` we **do** use `ThreadLocal`s, but we clear them up before returning to user code".

You are saying _"if `log4j2.enableThreadlocals=false` we **do** use `ThreadLocal`s but..."_ sorry but this makes no sense. No matter how much you stretch the meaning of the words, the intent is crystal clear.

> The `ThreadContext` class is not different,

It is completely different. You are comparing apples and oranges. If a user calls `ThreadContext`, by design, they expect the context propagation behaviour, and _the one and only way_ to achieve this is using `ThreadLocal`s. (Even the newly introduced `ScopedValue` uses `TL`s!) You cannot say _"I want to use `ThreadContext` but please don't store anything in the `Thread`"_.

ppkarwasz: > You are saying _"if `log4j2.enableThreadlocals=false` we **do** use `ThreadLocal`s but..."_ sorry but this makes no sense. No matter how much you stretch the meaning of the words, the intent is crystal clear.

The original intent in the JavaDoc is "enable thread locals for object pooling". This goal is reachable even if we didn't achieve it yet.

The goal implied by the naming to disable all `ThreadLocal` usage is and always was impossible.

> > The `ThreadContext` class is not different,
> 
> It is completely different. You are comparing apples and oranges. If a user calls `ThreadContext`, by design, they expect the context propagation behaviour, and _the one and only way_ to achieve this is using `ThreadLocal`s. (Even the newly introduced `ScopedValue` uses `TL`s!) You cannot say _"I want to use `ThreadContext` but please don't store anything in the `Thread`"_.

You can ask that code like this clears the thread local:

```java
try {
    ThreadLocal.put("key", "value");
    ...
} finally {
    ThreadLocal.remove("key");
}

Currently it doesn't.

Issue #2525, opened 2024-04-26
Revamp the `Configuration` page


vy:
Due to the fast development lifecycle of Log4j 2, the configuration options documented in the [configuration page](https://logging.apache.org/log4j/2.x/manual/configuration) and those effectively available in the code are currently out-of-sync.

The purpose of this task is to:

- [x] check the documentation of  [Log4j System Properties](https://logging.apache.org/log4j/2.x/manual/configuration.html#SystemProperties) with their usage in the code,
- [x] if they differ, fix either the documentation page or the code,
- [x] simplify and restructure the page.

## `Configuration` page outline

As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), revamp the _"Configuration"_ page such that

- Outline:
  - Supported formats
    - General structure of configuration file
    - How do formats map to each other?
  - Basics
    - Very briefly explain appender, logger, etc.
    - Refer to the “Architecture” page for details
  - Properties
    - Mention about fuzzy matching
    - List them in definition list (for main artifacts)
  - Recipes
    - Examples for common use cases
  -   Links to the “Guides” page contents
  - Programmatic configuration
- Fix [LOG4J2-3684](https://issues.apache.org/jira/browse/LOG4J2-3684)
- Target `2.x`, port to `main`

ppkarwasz:
The following "documentation" problems were fixed by adapting the code to the documented behavior:

- [ ] #2481,
- [ ] #2517,
- [ ] #2550

ppkarwasz:
The following documentation changes were applied:

- [ ] #2552

PR #2550, opened 2024-04-30
Improves support of `log4j2.enableThreadlocals` property

ppkarwasz:
If the `log4j2.enableThreadlocals` property is set to `false` Log4j should clear all `ThreadLocal`s from the calling thread before execution finishes.

The same rule should apply to user methods that **correctly** use the `ThreadLocal` interface. Code like this:

```
void method() {
    try {
        ThreadLocal.set("key", "newValue");
        ...
    } finally {
        ThreadLocal.remove("key");
    }
}
```

should perform a call to [`ThreadLocal#remove`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/ThreadLocal.html#remove()) before exiting.

While the difference between `threadLocal.remove()` and `threadLocal.set(null)` is almost imperceptible in a classical threading model, when virtual threads are used the difference might be considerable.

This PR:

* Fixes all the `ThreadContextMap` and `ThreadContextStack` implementations (except `GarbageFreeSortedArrayThreadContextMap`) so that the context map/stack is cleared if the map/stack is empty.
* Modifies the `ThreadLocal`s used to detect recursion to call `ThreadLocal#remove` when the recursion counter reaches 0.
* Fixes other `ThreadLocal` used for object pooling to respect the `log4j2.enableThreadlocals` property.

**Remark**: This PR does **not** address [LOG4J-2753](https://issues.apache.org/jira/browse/LOG4J2-2753) that will be fixed in another PR.

Part of #2525.


ppkarwasz:
> > While the difference between `remove()` and `set(null)` is almost imperceptible in a classical threading model, when virtual threads are used the difference **MIGHT BE** considerable.
> 
> This PR removes 94 and adds 425 LoC. Have we evaluated the assumption using benchmarks? It might be true that we will be shaving off some bytes from the thread stack, but aren't we also introducing an extra overhead since TLs need to be allocated again and again each time?

I ran some benchmarks a while ago. While I don't remember the numbers, the overhead due to TL deletion and creation was statistically significant.

The removal of TLs for the `ThreadContextMap` implementations should have an insignificant on performance, since the map should only become empty at the bottom of the call stack. Anyway we can fix this problem if users report an issue. I am closing this PR.

Issue #2539, opened 2024-04-29
[MS15] Documentation revamp 2/5


vy:
This is the 2<sup>nd</sup> of the 5 STF milestones planned to revamp the website and manual.

### Deliverable 1: Documentation

As the core deliverable of this STF milestone, we have implemented several documentation improvements:

- #2506 
- #2507 
- #2508 
- #2535
- #2520
- #2533

But probably the best way to view them is to see them in action:

- Resources
  - [F.A.Q.](https://logging.staged.apache.org/log4j/2.x/faq.html)
  - [Getting started](https://logging.staged.apache.org/log4j/2.x/getting-started.html)
  - [Development](https://logging.staged.apache.org/log4j/2.x/development.html)
- [API](https://logging.staged.apache.org/log4j/2.x/manual/api.html)
  - [Levels](https://logging.staged.apache.org/log4j/2.x/manual/customloglevels.html)
  - [Fluent API](https://logging.staged.apache.org/log4j/2.x/manual/logbuilder.html)
  - [Markers](https://logging.staged.apache.org/log4j/2.x/manual/markers.html)
  - [Thread Context](https://logging.staged.apache.org/log4j/2.x/manual/thread-context.html)
  - [Messages](https://logging.staged.apache.org/log4j/2.x/manual/messages.html)
  - [Flow Tracing](https://logging.staged.apache.org/log4j/2.x/manual/flowtracing.html)
- [Performance](https://logging.staged.apache.org/log4j/2.x/manual/performance.html)
  - [Garbage-free logging](https://logging.staged.apache.org/log4j/2.x/manual/garbagefree.html)

### Deliverable 2: Log4j Tools `0.9.0`

While working on the documentation, we encountered certain issues that required a fix in our infrastructure tooling. We addressed these in Log4j Tools `0.9.0` release. Related tickets from this release are as follows:

#### Log4j Docgen XSD generator bug

[...]

Issue #2538, opened 2024-04-29
[MS14] Documentation revamp 1/5


vy:
This is the 1<sup>st</sup> of the 5 STF milestones planned to revamp the website and manual.

### Tasks

- #2525 
- #2534

grobmeier:
Checked and completed

PR #2581, opened 2024-05-14
Rewrote manual intro

### LINKED ISSUE #1468: Revamp the manual introduction
[The manual's introduction page](https://logging.apache.org/log4j/2.x/manual/index.html) mostly consists of a _Log4j 2_ section. That section is somewhat long for an introduction, and it could be confusing. This ticket reports a few close issues:

1. The paragraph initially seems to suggest that Log4j 1 is still maintained, saying "However, through the years development on it has slowed down."
2. The paragraph implies that SLF4J+Logback and Log4j 2 are the only alternatives to Log4j 1, when several other logging frameworks (including Reload4j) exist.
3. It is not clear what the list contains. It appears to be a comparison with both Log4j 1 and SLF4J+Logback. For example, it is unclear if "Support for custom log levels" is a benefit when compared to Log4j 1 only, SLF4J+Logback too, or even just the latter.
4. The reasons for putting so much emphasis on a comparison with Log4j 1 are unclear, or even just outdated. It would be good to indicate that Log4j 1 has known vulnerabilities, but beyond that, I don't see why such an in-depth comparison would warrant inclusion in the introduction at this point.

I don't have a clear solution to propose, but I would first recommend to retitle the section "Log4j 1 and SLF4J+Logback" and reevaluate how prominent a comparison with Log4j 1 should remain. Another option would be to use a comparison matrix rather than a list.

grobmeier:
Rewriting of the manual intro.

Please comment if you see any missing "fantastic" features that need to be included in the intro or if anything should be removed.

closes #1468

### DIFF
--- a/src/site/antora/modules/ROOT/pages/manual/index.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/index.adoc
@@ -66,3 +66,7 @@ Log4j offers numerous advantages, including:
 7. **Modern Syntax**: Log4j supports xref:manual/api.adoc#LambdaSupport[lambda expressions]
 8. **Integration**: Log4j integrates with numerous other frameworks and libraries.
 9. **ASF Standards**: Developed at the Apache Software Foundation, adhering to rigorous standards and best practices.
+
+== How to learn more?
+
+To get started, learn more about the xref:manual/api.adoc[Log4j API] and it's xref:manual/architecture.adoc[architecture.] 
--- a/src/site/antora/modules/ROOT/pages/manual/index.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/index.adoc
@@ -25,12 +25,11 @@ The project is actively maintained by a {logging-services-url}/team-list.html[te
 Logging is an essential part of the software development process. 
 It provides a way to track the flow of execution in a program, allowing developers 
 to understand the application's behavior without needing a debugger. 
-This is particularly useful when trying to track down a bug or understand 
-why a particular piece of code is running slow.
+This is particularly useful when tracking bugs or understanding why a particular code runs slowly.
 
 The original concept for Log4j was conceived in early 1996 when the 
 E.U. SEMPER project decided to develop its own tracing API. 
-In 2003, the project was donated to the Apache Software Foundation, becoming Apache Log4j. 
+In 2003, the project was donated to the Apache Software Foundation, which became Apache Log4j. 
 
 Since then, Log4j has seen numerous releases and has become a widely adopted solution.
 
--- a/src/site/antora/modules/ROOT/pages/manual/index.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/index.adoc
@@ -54,18 +54,26 @@ to consider not using logging at all.
 
 Log4j offers numerous features, including:
 
-1. **Easy to use**: Log4j is easy to use and configure. You can get started
-   with just a few lines of code.
-2. **Speed**: Log4j is fast and has a low overhead. For more information see: xref:manual/async.adoc[Asynchronous Loggers]
-3. **Reliable**: Log4j is reliable and has been used in many production applications.
-4. **Audit-ready**: Log4j is designed to be a reliable audit logging framework that 
-   does not lose any events, unlike some competing frameworks.
-5. **Small footprint**: Log4j has a small impact on the garbage collector. For details, see: xref:manual/garbagefree.adoc[garbage-free].
-6. **Extensible**: Log4j is designed to be easily extended using a xref:manual/plugins.adoc[plugin system], 
-   allowing users to customize it to xref:manual/extending.adoc[fit their specific needs].
-7. **Modern Syntax**: Log4j supports xref:manual/api.adoc#LambdaSupport[lambda expressions]
-8. **Integration**: Log4j integrates with numerous other frameworks and libraries.
-9. **ASF Standards**: Developed at the Apache Software Foundation, adhering to rigorous standards and best practices.
+Easy to use::
+Log4j is easy to use and configure.
+You can xref:5min.adoc[get started] with just a few lines of code.
+
+Performance::
+When configured correctly, Log4j excels at speed.
+See the xref:manual/performance.adoc[Performance] page for details.
+
+Reliable::
+Log4j is reliable and has been used in many production applications.
+No messages will be lost even during reconfiguration.
+
+Extensible::
+Log4j is designed to be easily extended using a xref:manual/plugins.adoc[plugin system], allowing users to customize it to xref:manual/extending.adoc[fit their specific needs].
+
+Integration::
+Log4j integrates with numerous other frameworks and libraries.
+
+ASF Standards::
+Developed at the Apache Software Foundation, Log4j adheres to rigorous standards and best practices.
 
 == How to learn more?
 
--- a/src/site/antora/modules/ROOT/pages/manual/index.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/index.adoc
@@ -16,122 +16,53 @@
 ////
 = Manual
 
-== Welcome to Log4j!
+== Welcome to Apache Log4j!
 
-Almost every large application includes its own logging or tracing API.
-In conformance with this rule, the E.U. http://www.semper.org[SEMPER]
-project decided to write its own tracing API. This was in early 1996.
-After countless enhancements, several incarnations and much work that
-API has evolved to become log4j, a popular logging package for Java. The
-package is distributed under the https://www.apache.org/licenses/LICENSE-2.0[Apache Software
-License], a fully-fledged open source license certified by the
-http://www.opensource.org[open source] initiative. The latest log4j
-version, including full-source code, class files and documentation can
-be found at
-https://logging.apache.org/log4j/2.x/index.html[*https://logging.apache.org/log4j/2.x/index.html*].
+Welcome to Apache Log4j, one of the world's most widely used logging frameworks.
 
-Inserting log statements into code is a low-tech method for debugging
-it. It may also be the only way because debuggers are not always
-available or applicable. This is usually the case for multithreaded
-applications and distributed applications at large.
+Logging is an essential part of the software development process. 
+It provides a way to track the flow of execution in a program, allowing developers 
+to understand the application's behavior without needing a debugger. 
+This is particularly useful when trying to track down a bug or understand 
+why a particular piece of code is running slow.
 
-Experience indicates that logging was an important component of the
-development cycle. It offers several advantages. It provides precise
-_context_ about a run of the application. Once inserted into the code,
-the generation of logging output requires no human intervention.
-Moreover, log output can be saved in persistent medium to be studied at
-a later time. In addition to its use in the development cycle, a
-sufficiently rich logging package can also be viewed as an auditing
-tool.
+The original concept for Log4j was conceived in early 1996 when the 
+E.U. SEMPER project decided to develop its own tracing API. 
+In 2003, the project was donated to the Apache Software Foundation, becoming Apache Log4j. 
 
-As Brian W. Kernighan and Rob Pike put it in their truly excellent book
-_"The Practice of Programming":_
+Since then, Log4j has seen numerous releases and has become a widely adopted solution.
 
-____
-As personal choice, we tend not to use debuggers beyond getting a stack
-trace or the value of a variable or two. One reason is that it is easy
-to get lost in details of complicated data structures and control flow;
-we find stepping through a program less productive than thinking harder
-and adding output statements and self-checking code at critical places.
-Clicking over statements takes longer than scanning the output of
-judiciously-placed displays. It takes less time to decide where to put
-print statements than to single-step to the critical section of code,
-even assuming we know where that is. More important, debugging
-statements stay with the program; debugging sessions are transient.
-____
+== When should you use Log4j?
 
-Logging does have its drawbacks. It can slow down an application. If too
-verbose, it can cause scrolling blindness. To alleviate these concerns,
-log4j is designed to be reliable, fast and extensible. Since logging is
-rarely the main focus of an application, the log4j API strives to be
-simple to understand and to use.
+Log4j is an excellent choice for any Java application that needs logging capabilities. 
+It is user-friendly, fast, and flexible. You can use it to log messages at 
+different levels of severity, from debug to fatal, and you can configure it to 
+log messages to various destinations, such as files, databases, or the console.
 
-== Log4j 2
+== When not to use Log4j?
 
-Log4j 1 has been widely adopted and used in many applications.
-However, through the years development on it has slowed down. It has
-become more difficult to maintain due to its need to be compliant with
-very old versions of Java and became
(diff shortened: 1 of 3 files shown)


swebb2066:
> Please comment if you see any missing "fantastic" features that need to be included

What about being able to activate additional logging output or disable some output for one or more modules of the running application by changing the configuration file?

To me this is the super-power that a simpler logging solution misses.

src/site/antora/modules/ROOT/pages/manual/index.adoc
```diff
@@ -16,122 +16,57 @@
...
+2. **Speed**: Log4j is fast and has a low overhead. For more information see: xref:manual/async.adoc[Asynchronous Loggers]
+3. **Reliable**: Log4j is reliable and has been used in many production applications.
+4. **Audit-ready**: Log4j is designed to be a reliable audit logging framework that 
+   does not lose any events, unlike some competing frameworks.
+5. **Small footprint**: Log4j has a small impact on the garbage collector. For details, see: xref:manual/garbagefree.adoc[garbage-free].
+6. **Extensible**: Log4j is designed to be easily extended using a xref:manual/plugins.adoc[plugin system], 
+   allowing users to customize it to xref:manual/extending.adoc[fit their specific needs].
+7. **Modern Syntax**: Log4j supports xref:manual/api.adoc#LambdaSupport[lambda expressions]
+8. **Integration**: Log4j integrates with numerous other frameworks and libraries.
+9. **ASF Standards**: Developed at the Apache Software Foundation, adhering to rigorous standards and best practices.
```
vy, line 68: I am sort of hesitant about this entire section, since it duplicates [what we already have in the front page](https://logging.apache.org/log4j/2.x/#features). I suggest moving that content to an Antora `partial` and re-using it both here and in the front page.

---

I suggest:

```suggestion
Easy to use::
Log4j is easy to use and configure.
You can xref:5min.adoc[get started] with just a few lines of code.

Performance::
When configured correctly, Log4j excels at speed.
See the xref:manual/performance.adoc[Performance] page for details.

Reliable::
Log4j is reliable and has been used in many production applications.
No messages will be lost even during reconfiguration.

Extensible::
Log4j is designed to be easily extended using a xref:manual/plugins.adoc[plugin system], allowing users to customize it to xref:manual/extending.adoc[fit their specific needs].

Integration::
Log4j integrates with numerous other frameworks and libraries.

ASF Standards::
Developed at the Apache Software Foundation, Log4j adheres to rigorous standards and best practices.
```

Removed _small memory footprint_, since Log4j is the biggest logging API/implementation in the wild. Hence, it is not small by any means, AFAICT.

I am also hesitant to advertise garbage-free in the very first page, since
1. It is an advanced feature and we already indirectly share it in the performance page (which we reference above)
1. As @ppkarwasz's recent benchmarks hint, it is not a relevant feature anymore as much as it was in the past thanks to the continuous improvements in the JVM's GC.

I removed _modern syntax, lambda expressions_ too, since all popular logging APIs (SLF4J, Tinylog, etc.) support it. I don't think it qualifies as a selling point.

src/site/antora/modules/ROOT/pages/manual/index.adoc
```diff
@@ -16,122 +16,57 @@
...
+
+== How to learn more?
+
+To get started, learn more about the xref:manual/api.adoc[Log4j API] and it's xref:manual/architecture.adoc[architecture.]
```
vy, line 72: Instead of just linking to API and Architecture pages, I think a question-answer style quick links (just like we have [in the front page](https://logging.apache.org/log4j/2.x/#links)) will be more convenient for users:

```suggestion
* xref:manual/installation.adoc[How can I install Log4j?]
* xref:manual/configuration.adoc[How can I configure Log4j?]
* xref:manual/api.adoc[How can I use Log4j API?]
* xref:manual/performance.adoc[How can I tune my Log4j setup for performance?]
* xref:manual/migration.adoc[How can I migrate from Log4j 1 to Log4j 2]?
* xref:manual/plugins.adoc[What are Log4j plugins] and xref:manual/extending.adoc[how can I use them to extend Log4j?]
```

src/site/antora/modules/ROOT/pages/manual/index.adoc
```diff
@@ -16,122 +16,57 @@
...
+
+Logging is an essential part of the software development process. 
+It provides a way to track the flow of execution in a program, allowing developers 
+to understand the application's behavior without needing a debugger. 
+This is particularly useful when trying to track down a bug or understand 
+why a particular piece of code is running slow.
+
+The original concept for Log4j was conceived in early 1996 when the 
+E.U. SEMPER project decided to develop its own tracing API. 
+In 2003, the project was donated to the Apache Software Foundation, becoming Apache Log4j.
```
vy, line 31: **Tip:** I think it would be sweet if we can provide some links to these historical facts.

src/site/antora/modules/ROOT/pages/manual/index.adoc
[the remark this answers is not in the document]
grobmeier: I tried it, but it is a horrible mess and extremely hard for the author to write - at least for me. In my opinion, we should drop this convention. I prefer well formatted text instead of automatic linewrapping. We can talk about this on the mailing list

src/site/antora/modules/ROOT/pages/manual/index.adoc
grobmeier: Previously added links are dead. I suggest we are not adding them anymore, except we find other links

vy: I know, that is why I said _if we can_. :sweat_smile: I don't know of any, neither could find. :cry:

vy:
@grobmeier, I reused common content between `/index.adoc` and `manual/index.adoc` in 8d47758e5c3de70dc46b83dd2f2ee16e84eaaae0. I don't prefer pushing commits to other people's PR branches, but I couldn't think of a way to show the `include:partial$` concept I mentioned earlier. Feel free to revert it if you find the earlier version is better.

src/site/antora/modules/ROOT/partials/log4j-features.adoc
```diff
@@ -0,0 +1,43 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+////
+
+Batteries included::
+Log4j bundles a rich set of components to assist various use cases.
+* xref:manual/appenders.adoc[Appenders] targeting files, network sockets, databases, SMTP servers, etc.
+* xref:manual/layouts.adoc[Layouts] that can render CSV, HTML, JSON, Syslog, etc. formatted outputs
+* xref:manual/filters.adoc[Filters] that can be configured using log event rates, regular expressions, scripts, time, etc.
```
ppkarwasz, line 22: I don't understand the "can be configured using" expression.

Shouldn't it be "Filters based on log event rates, …"?

src/site/antora/modules/ROOT/partials/log4j-features.adoc
```diff
@@ -0,0 +1,43 @@
...
+(See xref:manual/api-separation.adoc[API Separation] for details.)
+The Log4j API also provides the most feature rich logging facade in the market; support for various `Message` types (`Object`, `Map`, etc.) besides plain `String`, lambda expressions, parametrized logging, markers, levels, diagnostic contexts (aka. MDC/NDC), etc.
+Check out the xref:manual/api.adoc[Java API], {logging-services-url}/log4j/kotlin[Kotlin API], and {logging-services-url}/log4j/scala[Scala API] pages for further information.
+
+No vendor lock-in::
+Even though the Log4j API is implemented by the Log4j at its fullest, users can choose to use another logging backend.
+This can be achieved by either using another backend implementing the Log4j API, or forwarding Log4j API calls to another logging facade (e.g., SLF4J) and using a backend for that particular facade.
+
+Performance::
+When configured correctly, Log4j can deliver excelling performance without almost any burden on the Java garbage collector.
```
ppkarwasz, line 36: I would add a **Reliability** point above this and modify this point to take into account configurability and reliability.

Do we offer the fastest Log4j API implementation possible, probably not: you can obtain a similar or _slightly_ better performance using components from two vendors (QOS.ch and [logfellow/logstash-logback-encoder](https://github.com/logfellow/logstash-logback-encoder)). If you use those solutions, however, you need to restart your application or risk losing log events if you need to make a configuration change (e.g. enable `DEBUG` logging for a single client).

ppkarwasz (changes requested):
We could expand the list of features we offer, otherwise it looks good to me.

src/site/antora/modules/ROOT/partials/log4j-features.adoc
vy: Fixed in 253e7a7aac1579fa1102489dccf3c77df2369716.

src/site/antora/modules/ROOT/partials/log4j-features.adoc
vy: Fixed in 253e7a7aac1579fa1102489dccf3c77df2369716.

ppkarwasz (approved):

ppkarwasz:
@grobmeier, can you merge this?

src/site/antora/modules/ROOT/pages/manual/index.adoc
```diff
@@ -16,122 +16,52 @@
...
+The project is actively maintained by a {logging-services-url}/team-list.html[team] of several volunteers and {logging-services-url}/support[support]ed by a big community.
+
+Logging is an essential part of the software development process. 
+It provides a way to track the flow of execution in a program, allowing developers 
+to understand the application's behavior without needing a debugger. 
+This is particularly useful when tracking bugs or understanding why a particular code runs slowly.
+
+The original concept for Log4j was conceived in early 1996 when the 
+E.U. SEMPER project decided to develop its own tracing API. 
+In 2003, the project was donated to the Apache Software Foundation, which became Apache Log4j.
```
Chealer, line 32: s/which became Apache Log4j/becoming _Apache Log4j_/

grobmeier: Thanks @Chealer . if you like, please create a PR and you can see your contribution in this file :-)

ppkarwasz: @Chealer,

You can follow our documentation rewrite on our [staging website](https://logging.staged.apache.org/log4j/2.x/index.html).
If we there are typos, grammar or other problems, feel free to use the "Edit" button on the top right side of the page to modify the page in Github and submit a PR.

**Note**: not all the pages have been rewritten yet. At this moment only the #2538 and #2539 pages are new. If you edit other pages, your changes might be lost, when we rewrite them.

src/site/antora/modules/ROOT/partials/log4j-features.adoc
```diff
@@ -0,0 +1,47 @@
...
+
+Batteries included::
+Log4j bundles a rich set of components to assist various use cases.
+* xref:manual/appenders.adoc[Appenders] targeting files, network sockets, databases, SMTP servers, etc.
+* xref:manual/layouts.adoc[Layouts] that can render CSV, HTML, JSON, Syslog, etc. formatted outputs
+* xref:manual/filters.adoc[Filters] based on log event rates, regular expressions, scripts, time, etc.
+* xref:manual/lookups.adoc[Lookups] for accessing system properties, environment variables, log event fields, etc.
+
+API separation::
+The API for Log4j (i.e., `log4j-api`) is separate from the implementation (i.e., `log4j-core`), making it straightforward for application developers which classes and methods to use while ensuring forward compatibility.
```
Chealer, line 26: Something is wrong in "making it straightforward for application developers which classes and methods to use while ensuring forward compatibility". Perhaps "straightforward to use in a correct and future-proof (forward-compatible) way"?

src/site/antora/modules/ROOT/pages/manual/index.adoc
Chealer: I made a few remarks which require someone more involved (ideally the author) to intervene, but if I still notice simple issues after that's solved, thanks, I will consider.
Thank you for the link @ppkarwasz !

grobmeier: No problem, I can work through the comments - thanks for adding them @Chealer 
Whenever you want to get more involved, please let me or @ppkarwasz know, we are happy about everyone who wants to get their hands dirty :)

src/site/antora/modules/ROOT/pages/manual/index.adoc
```diff
@@ -16,122 +16,52 @@
...
-issues. Many of these are fixed in Logback but many Logback classes
-still require synchronization at a fairly high level.
-14. It is an Apache Software Foundation project following the community
-and support model used by all ASF projects. If you want to contribute or
-gain the right to commit changes just follow the path outlined at
-http://jakarta.apache.org/site/contributing.html[Contributing].
+== Welcome to Apache Log4j!
+
+Apache Log4j is a versatile, industrial-grade Java logging framework composed of an API, its implementation,  and components to assist the deployment for various use cases.
+Log4j is https://security.googleblog.com/2021/12/apache-log4j-vulnerability.html[used by 8% of the Maven ecosystem] and listed as one of https://docs.google.com/spreadsheets/d/1ONZ4qeMq8xmeCHX03lIgIYE4MEXVfVL6oj05lbuXTDM/edit#gid=1024997528[the top 100 critical open source software projects].
```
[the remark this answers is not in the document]
vy: I can swear that the 1st link, added by me both to here and to the main page, was correct. I checked via Wayback Machine every single snapshot, but couldn't find it again. :cry: I guess we need to remove it.

I will massage the label of the 2nd link.

src/site/antora/modules/ROOT/partials/log4j-features.adoc
vy: I think this whole paragraph doesn't make sense at all on the benefits of API-vs-implementation separation. I will rewrite this.

ppkarwasz: Yes, I guess we should start with the advantages of Log4j API, followed by the advantages of Log4j Core. So in order:

1. "No vendor lock-in": the Log4j API and its reference implementation are separate and Log4j API is in permanent LTS-mode. People can use a stable API without breaking changes, while at the same time profiting of major rewrites of the reference implementation.
2. "A feature-rich API": the API provides 100+ **different** logging methods, even if we ignore the vararg expansions. These methods can even be repurposed to obtain different semantics, using the `Message` and `MessageFactory` abstractions.
3. "Batteries included", "Reliability" and so on.

Chealer: > I think this whole paragraph doesn't make sense at all on the benefits of API-vs-implementation separation. I will rewrite this.

I believe it can make sense, but you're right that it's at best unclear. I would suggest something closer to:

> In addition to the Log4j Core reference implementation, Log4j provides a stable facade (i.e., `log4j-api`) which is separate from that implementation, making it straightforward for developers to use Log4j while being confident on a high degree of API forward-compatibility.

Issue #2533, opened 2024-04-26
Revamp the `Performance` page


vy:
As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), revamp the `Performance` page such that

- Outline:
  - Asynchronous logging
  - Filtering
  - Source location information
  - Memory management (i.e., garbage-free & recyclers)
    - Add a `Performance` section to the `Pattern Layout` page and link to it
    - Add a `Performance` section to the `JSON Template Layout` page and link to it
- Revamp `garbagefree.adoc`
- Target `2.x`, port to `main`

PR #2583, opened 2024-05-15
Fixes recycler factory provider initialization

vy:
Discovered while porting #2533 to `main`.


ppkarwasz:
I think that the `RecyclerFactoryProvider.createForEnvironment` method should never return `null`.

If we want to deal with the specific case of `ThreadLocalRecyclerFactoryProvider` that should be disabled by **default** if `log4j2.enableThreadlocals` is set to `false`, we can play with its order:

https://github.com/apache/logging-log4j2/blob/8d19bc5e84ca74d01bc5e1ad7731d1b0063419f2/log4j-kit/src/main/java/org/apache/logging/log4j/kit/recycler/internal/ThreadLocalRecyclerFactoryProvider.java#L48-L51

If the order is high enough, the dummy provider will prevail unless the user **explicitly** requested `threadLocal` to be used.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/CoreDefaultBundle.java
```diff
@@ -103,24 +105,58 @@ public FlowMessageFactory defaultFlowMessageFactory() {
...
-        return provider.createForEnvironment(environment);
+        // Collect providers
+        @Nullable
+        final String providerName =
+                environment.getProperty(RecyclerProperties.class).factory();
+        final List<RecyclerFactoryProvider> providers = ServiceLoaderUtil.safeStream(
+                        RecyclerFactoryProvider.class,
+                        ServiceLoader.load(RecyclerFactoryProvider.class, loader),
+                        statusLogger)
+                .sorted(Comparator.comparing(RecyclerFactoryProvider::getOrder))
```
ppkarwasz, line 121: Can we use a composite comparator such that:

1. The provider supplied using `log4j.recycler.factory` comes first,
2. The remaining providers are ordered by `Order`.

```java
.sorted(Comparator.comparing(p -> p.getName().equals(providerName) ? Integer.MIN_VALUE : p.getOrder()
```

should work

vy:
> I think that the `RecyclerFactoryProvider.createForEnvironment` method should never return `null`.

Right. Fixed in b741632db09ad8cecb59bdf78a46eebfb0ac06c9.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/CoreDefaultBundle.java
vy: I prefer a more dumb-proof style: Have an explicit block for `if (providerName != null)`.

Issue #2535, opened 2024-04-26
Revamp the `API` page


vy:
As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), revamp the `API` page such that

- Revamp the content
- Link from/to 5min
- Address following tickets:
  - #1460
  - #1468 (make it visible in the manual intro) (3.x ca978bfcc71ee8303bd7c0e32ce44257a22515dc)
  - #1469 (make it visible in the menu)
  - #1918
- Make `api.adoc` the central entry point (merge `api-separation.adoc` into there)
- Cover content from:
  - [X] `flowtracing.adoc` -> (2.x: f1de65f8398b7f9aa8cb62df98f365d8c3ec7c7f, 28268e856c52bf3d76a3d75ec23343f49f255c3c), (3.x: 43181c292c6767bef62ff5c7be1042816ae184e5)
  - [X] `markers.adoc` -> https://github.com/apache/logging-log4j2/pull/2584 (Ported to 3.x with 891c08e76ae8ad3ad5119272d9df027599bf445b)
  - [x] `messages.adoc` -> https://github.com/apache/logging-log4j2/pull/2591
  - [x] `eventlogging.adoc` -> (2.x: 9ea02235afd880adb0fd2004b931d8ff81eaa9ff), (3.x: dac0ba84fa787f5a01b3faa20fb355448ec42ddc)
  - [x] `logbuilder.adoc` -> https://github.com/apache/logging-log4j2/pull/2595
- Target `2.x`, port to `main`

PR #2595, opened 2024-05-20
rewrote logbuilder

grobmeier:
Rewrote log builder.

When reviewing, please advise if we should remove all the performance tests at the end. I have not found accurate and up-to-date examples

#2535


vy (approved):

vy:
[As agreed with @grobmeier](https://the-asf.slack.com/archives/D02TQ7SQ9CY/p1716534032649039), merging this to wrap up the work on API docs.

Issue #2540, opened 2024-04-29
[MS16] Documentation revamp 3/5


vy:
This is the 3<sup>rd</sup> of the 5 STF milestones planned to revamp the website and manual.

### Tasks

- #2511 
- #2512 
- #2513 
- #2514 
- #2519
- #2527 
- #2619
  - Document `SimpleLogger` (#1474)
  - Explain [status messages](https://logging.apache.org/log4j/2.x/manual/configuration.html#status-messages) (since they are removed from the Configuration page in #2574)
  - There are several mentions of status messages, status logger, etc. in the code base. Link those to this new page/section.
- #2491

grobmeier:
completed

PR #2619, opened 2024-05-27
Add Simple Logger and Status Logger pages

vy:
This PR addresses `StatusLogger` and `SimpleLogger` tasks of #2540.

### DIFF
--- a/src/site/antora/modules/ROOT/pages/manual/api.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/api.adoc
@@ -226,7 +226,7 @@ xref:manual/eventlogging.adoc[Read more on event loggers...]
 
 Even though Log4j Core is the reference implementation of Log4j API, Log4j API itself also provides a very minimalist implementation: _Simple Logger_.
 This is a convenience for environments where either a fully-fledged logging implementation is missing, or cannot be included for other reasons.
-`SimpleLogger` is the default Log4j API implementation if no other is available in the classpath.
+`SimpleLogger` is the fallback Log4j API implementation if no other is available in the classpath.
 
 xref:manual/simple-logger.adoc[Read more on the simple logger...]
 
--- a/src/site/antora/modules/ROOT/pages/manual/api.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/api.adoc
@@ -49,9 +49,10 @@ You can also check out xref:manual/installation.adoc[] for the complete installa
 Log4j API provides
 
 * A logging API that libraries and applications can code to
+* xref:#simple-logger[A minimal logging implementation (aka. Simple logger)]
 * Adapter components to create a logging implementation
 
-This page tries to cover the most prominent Log4j API features that libraries and applications can code to.
+This page tries to cover the most prominent Log4j API features.
 
 [TIP]
 ====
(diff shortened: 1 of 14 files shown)


src/site/antora/modules/ROOT/pages/manual/api.adoc
```diff
@@ -220,7 +221,25 @@ link:../javadoc/log4j-api/org/apache/logging/log4j/LogManager.html[`EventLogger`
 
 xref:manual/eventlogging.adoc[Read more on event loggers...]
 
-[#feature-fluent-api]
+[#simple-logger]
+=== Simple logger
+
+Even though Log4j Core is the reference implementation of Log4j API, Log4j API itself also provides a very minimalist implementation: _Simple Logger_.
+This is a convenience for environments where either a fully-fledged logging implementation is missing, or cannot be included for other reasons.
+`SimpleLogger` is the default Log4j API implementation if no other is available in the classpath.
```
ppkarwasz, line 229: Maybe I would say it is the "fallback" Log4j API implementation, since unlike other implementations a warning is always logged, whenever it is used.

src/site/antora/modules/ROOT/pages/manual/simple-logger.adoc
```diff
@@ -0,0 +1,29 @@
...
+Even though Log4j Core is the reference implementation of xref:manual/api.adoc[Log4j API], Log4j API itself also provides a very minimalist implementation: link:../javadoc/log4j-api/org/apache/logging/log4j/simple/SimpleLogger.html[`SimpleLogger`].
+This is a convenience for environments where either a fully-fledged logging implementation is missing, or cannot be included for other reasons.
+`SimpleLogger` is the default Log4j API implementation if no other is available in the classpath.
+
+[#config]
+== Configuration
+
+`SimpleLogger` can be configured using the following system properties:
+
+include::partial$manual/systemproperties/properties-simple-logger.adoc[leveloffset=+1]
```
ppkarwasz, line 29: There are two components of `SimpleLogger` configuration:

* the configuration of `SimpleLogger`,
* the configuration of `ThreadContextMap`.

Currently the latter is common with `log4j-core`, but there is no reason for it to be this way.
For example if `log4j2.simpleLogShowContextMap` is set to `false` (default), the `ThreadContextMap` should probably be a no-op.

ppkarwasz (commented):
I think we should remove the documentation of `StatusLogger` and `SimpleLogger` configuration properties from `systemproperties.adoc`: these are common properties that can be useful to both Log4 Core 2 and Log4j Core 3 users, but should appear only once.

src/site/antora/modules/ROOT/pages/manual/api.adoc
vy: Fixed in 9a2815bb87b3bd0c8d7cde3ebefae095e948b049.

src/site/antora/modules/ROOT/pages/manual/simple-logger.adoc
vy: Fixed in 9a2815bb87b3bd0c8d7cde3ebefae095e948b049.

vy:
> I think we should remove the documentation of `StatusLogger` and `SimpleLogger` configuration properties from `systemproperties.adoc`: these are common properties that can be useful to both Log4 Core 2 and Log4j Core 3 users, but should appear only once.

Implemented in 9a2815bb87b3bd0c8d7cde3ebefae095e948b049.

Issue #2511, opened 2024-04-24
Create a `Using Log4j in Hibernate` guide


vy:
As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), create a _"Using Log4j in Hibernate"_ guide such that

- Check if the suggestion needs tweaking for different Hibernate major versions
- Target `2.x`, port to `main`

PR #2626, opened 2024-05-29
Add Hibernate integration page

vy:
This PR addresses #2511.


src/site/antora/modules/ROOT/examples/hibernate/log4j2.properties
```diff
@@ -0,0 +1,34 @@
...
+logger.0.name = org.hibernate.SQL
+logger.0.level = DEBUG
+
+# Log JDBC bind parameters and extracted values
+logger.1.name = org.hibernate.type
+logger.1.level = TRACE
+logger.2.name = org.hibernate.orm.jdbc.bind
+logger.2.level = TRACE
+logger.3.name = org.hibernate.orm.jdbc.bind
+logger.3.level = TRACE
```
ppkarwasz, line 31: I love that you also provided a `*.properties` file! :100:

src/site/antora/modules/ROOT/examples/hibernate/log4j2.xml
```diff
@@ -0,0 +1,46 @@
...
+
+  <Loggers>
+
+    <!-- Log just the SQL -->
+    <Logger name="org.hibernate.SQL" level="DEBUG"/>
+
+    <!-- Log JDBC bind parameters and extracted values -->
+    <Logger name="org.hibernate.type" level="TRACE"/>
+    <Logger name="org.hibernate.orm.jdbc.bind" level="TRACE"/>
+    <Logger name="org.hibernate.orm.jdbc.extract" level="TRACE"/>
```
ppkarwasz, line 38: I would keep this commented out with a warning:
> Be aware that JDBC statement parameters can contain sensitive data (passwords, credit card numbers, etc.).
> Uncomment this **only** in a dev environment (or start looking for a new job :wink:).
> See also "Arbiters" on how to switch config based on the environment.

ppkarwasz (changes requested):
LGTM

I would just add a warning about logging JDBC parameters.

IMHO sensitive data should **never** be logged. In this case it is impossible to filter out sensitive data from what Hibernate logs, so users should complement Hibernate logs with manually written log statements that only use non-sensitive data.

src/site/antora/modules/ROOT/examples/hibernate/log4j2.properties
vy: Friends don't let friends suffer alone. :sunglasses:

src/site/antora/modules/ROOT/examples/hibernate/log4j2.xml
vy: Fixed in 2d86aa07c5997d912e0c5dcc0243b0b79e9b967e.

PR #2630, opened 2024-05-30
Remove `log4j-iostreams` from `main`

ppkarwasz:
The Log4j IOStreams module is part of the Log4j API library. Since Log4j API 3 will not be release, Log4j IOStreams 3 shouldn't be released either.


vy (approved):

rgoers:
I have never considered log4j-iostreams to be part of Log4j-API. That said, it only has test dependencies on log4j-core and does not have any plugins so I see no reason why we need a 3.x specific version of this.

rgoers (approved):

garydgregory:
This is so confusing to me. So we are keeping 2.x for the API and main which is 3.x alive in parallel forever? Do we document this somewhere?

ppkarwasz:
> This is so confusing to me. So we are keeping 2.x for the API and main which is 3.x alive in parallel forever? Do we document this somewhere?

It is documented on the [`3.x` staging site](https://logging.staged.apache.org/log4j/3.x/manual/api.html).

rgoers:
@garydgregory This is why it would be less confusing if Log4j API and components only dependent on Log4j API (i.e. not dependent on log4j-core) were moved to a different repo.

garydgregory:
So eventually, once 3.x is out, how do I use the iostreams code?

ppkarwasz:
> So eventually, once 3.x is out, how do I use the iostreams code?

Nothing changes: `log4j-iostreams` only depends on `log4j-api` (same as all the bridges X-to-Log4j API).
Since `log4j-bom` version 3.x has been updated to contain the appropriate version of `log4j-iostreams`, you don't even need to know that the version number your are using is 2.x.

garydgregory:
Phew! 😄

PR #2641, opened 2024-06-04
Add Develocity integration

jprinet:
This PR will enable you to publish Build Scans to [ge.apache.org](https://ge.apache.org/) as discussed with @ppkarwasz

## Description
This PR publishes a build scan for CI builds from the build and deploy-site GH workflows and for every local build from an authenticated Apache committer. 
The build will not fail if publishing fails. 
Local cache is enabled on local builds but not on CI builds.

The build scans of the Apache Log4j2 project are published to the Develocity instance at [ge.apache.org](https://ge.apache.org/), hosted by the Apache Software Foundation and run in partnership between the ASF and Gradle. This Develocity instance has all features and extensions enabled and is freely available for use by the Apache Log4j2 project and all other Apache projects.

On this Develocity instance, Apache Log4j2 will have access not only to all of the published build scans but other aggregate data features such as:
- Dashboards to view all historical build scans, along with performance trends over time
- Build failure analytics for enhanced investigation and diagnosis of build failures
- Test failure analytics to better understand trends and causes around slow, failing, and flaky tests

Please let me know if there are any questions about the value of Develocity or the changes in this pull request and I’d be happy to address them.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,6 +34,9 @@ on:
 
 permissions: read-all
 
+env:
+  DEVELOCITY_ACCESS_KEY: ${{ secrets.GE_ACCESS_TOKEN }}
+
 jobs:
 
   build:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,8 +42,7 @@ jobs:
   build:
     if: github.actor != 'dependabot[bot]'
     # Temporary switch to `@fix/macos-java-8` to see how the situation evolves with macOS runners
-    uses: apache/logging-parent/.github/workflows/build-reusable.yaml@fix/macos-java-8
-    # uses: apache/logging-parent/.github/workflows/build-reusable.yaml@rel/11.0.0
+    uses: apache/logging-parent/.github/workflows/build-reusable.yaml@develocity
     with:
       java-version: |
         8
--- a/.github/workflows/deploy-site.yaml
+++ b/.github/workflows/deploy-site.yaml
@@ -29,6 +29,9 @@ on:
 
 permissions: read-all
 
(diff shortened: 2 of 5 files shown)


clayburn:
@ppkarwasz - I see your last commit to always publish from CI. This shouldn't be necessary. Your CI builds will be authenticated by accessing the `DEVELOCITY_ACCESS_KEY` env var. 

If you are expecting to see Build Scans from this build, unfortunately you will not, because @jprinet is creating this PR from a fork. Forks will not have access to the GE_ACCESS_TOKEN secret since they are not part of the ASF organization. In fact, what you might instead see is a warning about failing to publish a build scan.

I would remove that most recent commit. If you would like to verify the publication of the build scan, you may push @jprinet's branch to the `apache/logging-log4j2` and run it from there.

ppkarwasz:
@clayburn,

Thanks for the information.

We use reusable builds in our workflow, I am not entirely sure that the reusable workflow will have access to the environment variable, even if `build.yaml` has.

ppkarwasz (approved):

clayburn:
The reusable workflow may change what needs to happen here: https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

It looks like the reusable workflow needs to define that it will accept the access key as a secret and declare its use within the workflow. Then the workflow here needs to pass the secret in to the step.

ppkarwasz:
@clayburn,

Thanks, I fixed it in efb4a0e93b732710a30c9d0dcb643fad96fc6b8d.

We'll test the "Predictive Test Selection" feature for a couple of weeks and then we'll streamline it to our remaining repos. cc/ @garydgregory

Issue #2541, opened 2024-04-29
[MS17] Documentation revamp 4/5


vy:
This is the 4<sup>th</sup> of the 5 STF milestones planned to revamp the website and manual.

### Tasks

- #2510
- #2547
- #2530 
- #2531
- #2532

### Preview

You can view the implemented changes using following links:

* Manual
  * [Architecture](https://logging.staged.apache.org/log4j/2.x/manual/architecture.html)
  * [Extending](https://logging.staged.apache.org/log4j/2.x/manual/extending.html)
    * [Plugins](https://logging.staged.apache.org/log4j/2.x/manual/plugins.html)
* Resources
  * [Integrating with Jakarta EE](https://logging.staged.apache.org/log4j/2.x/jakarta.html)
  * [Integrating with service-oriented architectures](https://logging.staged.apache.org/log4j/2.x/soa.html)

grobmeier:
Closing as completed

PR #2660, opened 2024-06-13
Add Logback example to "Markers" page

ppkarwasz:
Since the Log4j API is independent from the Log4j implementation, the "API" chapter of the documentation should contain configuration examples for all major backends that support a certain API feature.

This PR adds a Logback configuration example to the "Markers" page.

Part of #2541.

### DIFF
--- a/src/site/antora/modules/ROOT/pages/manual/markers.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/markers.adoc
@@ -85,8 +85,7 @@ include::example$manual/markers/marker-filtering.log[tag=use-marker-parent]
 == Pitfalls
 
 It is important to note that marker names must be unique, as Log4j registers them permanently by name.
-Developers should avoid generic marker names, as they may conflict with
-those provided by third parties. Except, of course, if the user is aware of the implications.
+Developers are advised to avoid generic marker names, as they may conflict with those provided by third parties.
 
 While it is possible to add or remove parent markers dynamically through the `setParents()` method,
 it is generally advisable to set the parents when the marker is created for efficiency and performance reasons.
--- a/src/site/antora/modules/ROOT/pages/manual/markers.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/markers.adoc
@@ -87,8 +87,10 @@ include::example$manual/markers/marker-filtering.log[tag=use-marker-parent]
 It is important to note that marker names must be unique, as Log4j registers them permanently by name.
 Developers are advised to avoid generic marker names, as they may conflict with those provided by third parties.
 
-While it is possible to add or remove parent markers dynamically through the `setParents()` method,
-it is generally advisable to set the parents when the marker is created for efficiency and performance reasons.
+For technical reasons the
+link:../javadoc/log4j-api/org/apache/logging/log4j/Marker.html#setParents(org.apache.logging.log4j.Marker...)[`Marker.setParents(Marker...)`]
+method can be called at runtime to modify the list of parents of the current marker.
+However, we discourage such a practice and advise you to only use the method at initialization time.
 
 It is also worth noting that markers without parents are more efficient to evaluate
 than markers with multiple parents. It is generally a good idea to avoid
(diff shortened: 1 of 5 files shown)


src/site/antora/modules/ROOT/pages/manual/markers.adoc
```diff
@@ -61,88 +61,110 @@ include::example$manual/markers/marker-filtering.log[tag=use-marker]
...
 
+[#pitfalls]
+== Pitfalls
+
+It is important to note that marker names must be unique, as Log4j registers them permanently by name.
+Developers should avoid generic marker names, as they may conflict with
+those provided by third parties. Except, of course, if the user is aware of the implications.
+
+While it is possible to add or remove parent markers dynamically through the `setParents()` method,
+it is generally advisable to set the parents when the marker is created for efficiency and performance reasons.
```
vy, line 92: I don't see the point of this advice. By design, markers are expected to be seldom changed. Given this, the shared performance impact – I presume, due to `System.arraycopy()` – is insignificant.

src/site/antora/modules/ROOT/pages/manual/markers.adoc
```diff
@@ -61,88 +61,110 @@ include::example$manual/markers/marker-filtering.log[tag=use-marker]
...
 ----
 include::example$manual/markers/marker-filtering.log[tag=use-marker-parent]
 ----
 
+[#pitfalls]
+== Pitfalls
+
+It is important to note that marker names must be unique, as Log4j registers them permanently by name.
+Developers should avoid generic marker names, as they may conflict with
+those provided by third parties. Except, of course, if the user is aware of the implications.
```
vy, line 89: Removed last sentence, which, IMHO, doesn't add value:
```suggestion
Developers are advised to avoid generic marker names, as they may conflict with those provided by third parties.
```

src/site/antora/modules/ROOT/examples/manual/markers/log4j2.xml
```diff
@@ -25,16 +25,14 @@
       <PatternLayout pattern="%d{HH:mm:ss.SSS} (%marker) %m%n"/>
     </Console>
   </Appenders>
+  <MarkerFilter marker="SQL"
```
vy, line 28: Really liked this `MarkerFilter` usage at the root level. :star_struck: I stole it for #2662.

vy (changes requested):
In API pages, while stating anything in the context of the logging implementation, *I think* we should introduce the implementations. I tried to practice this as follows in #2662:

```
// Before introducing Log4j Core flow tracing implementation:

Log4j Core, the reference implementation of Log4j API, implements the flow tracing methods such that ...

...

// Before share Log4j Core configuration:

Log4j Core is the reference implementation of Log4j API. In the example configurations below..."_

...

// Before sharing Logback configuration:

{logback-url}[Logback] is another logging implementation for the {slf4j-url}[SLF4J] logging API, just like Log4j Core is a logging implementation for the xref:manual/api.adoc[Log4j API].
Using Log4j-to-SLF4J bridge, Logback can be xref:manual/installation.adoc#impl-logback[configured to consume Log4j API].
Below we will use a combination of `EvaluatorFilter`...
```

These kind of 1-2 introduction sentences I think set the stages well. I am curious of your thoughts. Since, you are the one who insisted on having non-Log4j examples in API pages.

src/site/antora/modules/ROOT/pages/manual/markers.adoc
ppkarwasz: This is a legacy comment. I guess we should replace it with something like:

> For technical reasons the `setParents` method can be called at runtime to modify the list of parents of the current marker.
> However, we discourage such a practice and advise you to only use the method at initialization time.

WDYT?

src/site/antora/modules/ROOT/examples/manual/markers/log4j2.xml
ppkarwasz: I introduced it to make the Logback example more complex/complete.

vy (approved):

PR #2662, opened 2024-06-14
Fix `flowtracing.adoc` examples

vy:


src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -14,109 +14,94 @@
...
+You can use them to track the input and output of the associated block:
 
+[#example-trace-args]
+.Example `traceEntry()` and `traceExit()` usage with arguments
 [source, java]
 ----
-public void someMethod(String param) {
-    logger.traceEntry(param); <1>
+public String someMethod(String input) {
+    logger.traceEntry(input); // <1>
```
ppkarwasz, line 61:
```suggestion
    logger.traceEntry(null, input); // <1>
```

This is a call to [`traceEntry(String format, Object... params)`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/Logger.html#traceEntry(java.lang.String,java.lang.Object...), so passing user-generated data as the format string is dangerous. :wink:

If the format is `null` the default `FlowMessageFactory` generates the appropriate format. In this case `params({})`.

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -14,109 +14,94 @@
...
     // method body
-    logger.traceExit(); <2>
+    String output = ...;
+    logger.traceExit(output); // <2>
+    return output;
```
ppkarwasz, line 65:
```suggestion
    return logger.traceExit(output); // <2>
```

I am not sure whether `traceExit(R result)` is the correct method for this example. Maybe `traceExit(String format, R result)` is better. If `format` is `null` it uses `with({})`.

Maybe we should also add an example, where the output of `traceEntry` is fed as parameter to `traceExit` to obtain a message like:
```
Exit: params(input): output
```

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -14,109 +14,94 @@
...
 public void someMethod() {
     try {
-        // Let's assume an exception is thrown here
-        String msg = messages[messages.length];        
-    } catch (Exception ex) {
-        logger.catching(ex); <1>
+        // Business logic
+    } catch (RuntimeException error) {
+        logger.throwing(error); // <1>
+        throw new RuntimeException(error); // <2>
```
ppkarwasz, line 99:
```suggestion
        throw logger.throwing(error); // <1>
```

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -131,83 +116,145 @@ public void someMethod() {
...
+Log4j Core, the reference implementation of Log4j API, implements the flow tracing methods such that
 
-[cols="3,3,3", options="header"]
+* <<impl-log4j-log-events,It generates specialized log events>>
+* <<impl-log4j-config,Its behaviour is configurable>>
+
+[#impl-log4j-log-events]
+==== Log events
+
+Log4j Core implements the flow tracing methods such that the generated log events are decorated to accommodate any need to selectively handle them:
```
ppkarwasz, line 144: Although I am not strongly opinionated about this, I would prefer to view this section as section that applies to all Log4j API implementations.

Although this is not part of the contract, all known implementation translate flow tracing calls into `log` calls with the given parameters.

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -131,83 +116,145 @@ public void someMethod() {
...
+|TRACE
+|`ENTER`, `FLOW`
+|link:../javadoc/log4j-api/org/apache/logging/log4j/message/EntryMessage.html[`EntryMessage`] extending from link:../javadoc/log4j-api/org/apache/logging/log4j/message/FlowMessage.html[`FlowMessage`]
+
+|traceExit()
+|TRACE
+|`EXIT`, `FLOW`
+|link:../javadoc/log4j-api/org/apache/logging/log4j/message/EntryMessage.html[`ExitMessage`] extending from link:../javadoc/log4j-api/org/apache/logging/log4j/message/FlowMessage.html[`FlowMessage`]
+
+|`throwing()`
```
ppkarwasz, line 163:
```suggestion
| throwing()
```

Already part of the column specification.

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
```diff
@@ -131,83 +116,145 @@ public void someMethod() {
...
+|ERROR
+|`THROWING`, `EXCEPTION`
+|Plain message
+
+|`catching()`
```
ppkarwasz, line 168:
```suggestion
| catching()
```

Same as above.

src/site/antora/modules/ROOT/examples/manual/flowtracing/log4j2.xml
```diff
@@ -0,0 +1,43 @@
...
+                   https://logging.apache.org/xml/ns
+                   https://logging.apache.org/xml/ns/log4j-config-2.xsd">
+
+  <MarkerFilter marker="FLOW" onMatch="ACCEPT" onMismatch="NEUTRAL"/><!--1-->
+
+  <Appenders>
+    <Console name="CONSOLE">
+      <PatternLayout>
+        <MarkerPatternSelector defaultPattern="%d %5p [%t] %c{1} - %m%n"><!--2-->
+          <PatternMatch key="ENTER" pattern="%d %5p [%t] %c{1} → %m%n"/><!--3-->
```
ppkarwasz, line 30: Graphically `→` is barely noticeable. Maybe we should:

- either choose a bigger Unicode arrow,
- or use `=>` instead.

The latter is used for example by Exim.

ppkarwasz (changes requested):
Nice job!

I leave some notes below.

src/site/antora/modules/ROOT/examples/manual/flowtracing/log4j2.xml
vy: Fixed in 3684d63934526d51633a0f69e3cccc6eedd94011.

src/site/antora/modules/ROOT/pages/manual/flowtracing.adoc
vy: I will kindly pass this Piotr. Though thanks for the suggestion.

PR #2667, opened 2024-06-17
Improve layout docs on property substitution

### LINKED ISSUE #2666: `PatternResolver` does not support substitution
I have an application use log4j2 template layout with below config

File `EcsLayout.json`
```json
{
  "message1": {
    "$resolver": "pattern",
    "pattern": "${env:TZ:-%d{yyyy-MM-dd HH:mm:ss.SSS}{UTC+07}}"
  },
  "message2": {
    "$resolver": "pattern",
    "pattern": "${env:TZ:-${TIMEZONE}}"
  }
}
```

File `log4j2.xml`

```xml
<Configuration name="Mask Library Test" xmlns:xi="http://www.w3.org/2001/XInclude">

  <Properties>
    <Property name="TIMEZONE">%d{yyyy-MM-dd HH:mm:ss.SSS}{UTC+07}</Property>
  </Properties>

  <Appenders>
      <PatternLayout
        disableAnsi="false"
        pattern="%style{${TIMEZONE}}{black} %highlight{%-5level}[%style{%t}{bright,blue}] %style{%c{1.}}{bright,yellow}: %m%n%throwable" />
    </Console>

    <RollingFile name="RollingFile"
      filePattern="logs/app-%d{yyyy-MM-dd-HH-mm}.log">
      <JsonTemplateLayout eventTemplateUri="classpath:EcsLayout.json" />

      <TimeBasedTriggeringPolicy />
    </RollingFile>
  </Appenders>

  <Loggers>
    <Root level="info">
      <AppenderRef ref="Console" />
      <AppenderRef ref="RollingFile" />
    </Root>
  </Loggers>
</Configuration>
```

And file log

```text
{"message1":"${env:TZ:-2024-06-14 15:00:33.561}","message2":"%d{yyyy-MM-dd HH:mm:ss.SSS}{UTC+07}"}
```

**Expected**

**message1**
If `env:TZ` do not exist, it will be replace by `{yyyy-MM-dd HH:mm:ss.SSS}{UTC+07}`

or

**message2**
log will be recorded date time instead `"%d{yyyy-MM-dd HH:mm:ss.SSS}{UTC+07}"`

I have a small demo, just run and go to file `logs/app-{yyyy-MM-dd-HH-mm}.log` and you will see result. Please tell me if I miss something.

https://github.com/ngocnhan-tran1996/log4j2-demo

vy:
Fixes #2666.


ppkarwasz (commented):
I am not really convinced that this is the right approach.

There are two ways to provide an event template to JTL:

1. You can inline the event template in the configuration file using the `eventTemplate` property (of type `String`). Since the template is inlined, property substitution **will occur** before control is passed to JTL. In this case JTL should **not** perform property substitution.

   Note that property substitution on the `eventTemplate` attribute does not guarantee that the result will be a valid JSON document.

   As far as I can see this is the only case you test.

3. You can provide an `eventTemplateUri` property and JTL will retrieve the value itself.
This is the case, when property substitution **should** be applied, possibly in JSON-safe way.

ppkarwasz:
Using the features currently provided by `PluginBuilder` a possible alternative approach to introduce lookups in event templates would be:

1. Replace the `@PluginAttribute(value = "eventTemplate")` annotation with `@PluginValue(value = "eventTemplate", substitute = false)` to prevent `PluginBuilder` from expanding the lookups in an inlined event template.
2. Apply property substitution manually to the event template in a JSON-safe way.

vy:
@ppkarwasz, as discussed and agreed on,

1. Added `substitute` flag to `@Plugin[Builder]Attribute`
2. Disabled [configuration file] property substitution for `eventTemplate`, `stackTraceTemplate`, and `EventTemplateAdditionalField#value` attributes

Could you review again, please?

ppkarwasz (approved):
Looks good to me, but I still find the usage of `StrSubstitutor` a little bit incoherent.

If you use lookups in literals they are expanded at **each** log event, e.g.:

```json
{
  "timestamp": "${date:hh:mm:ss.SSS}"
}
```

but if you wrap it into a `PatternResolver` it is also expanded at configuration time and requires escaping:

```json
{
  "timestamp": {
    "$resolver": "pattern",
    "pattern": "$${date:hh:mm:ss.SSS}"
  }
}
```

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/PatternResolver.java
```diff
@@ -71,6 +70,16 @@ public final class PatternResolver implements EventResolver {
...
     }
 
+    private static String readPattern(
+            final EventResolverStringSubstitutor substitutor, final TemplateResolverConfig config) {
+        final String pattern = config.getString("pattern");
+        if (Strings.isBlank(pattern)) {
+            throw new IllegalArgumentException("blank pattern: " + config);
+        }
+        final boolean substitutionNeeded = pattern.contains("${");
+        return substitutionNeeded ? substitutor.replace(null, pattern) : pattern;
```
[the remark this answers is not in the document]
vy: Removed in 03993a8eebfb88617d0c602781b07d7e3e05e41b.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TemplateResolverContext.java
```diff
@@ -30,7 +30,7 @@
  *
  * @see TemplateResolverFactory
  */
-interface TemplateResolverContext<V, C extends TemplateResolverContext<V, C>> {
+public interface TemplateResolverContext<V, C extends TemplateResolverContext<V, C>> {
```
ppkarwasz, line 33: Is this still needed?

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/package-info.java
```diff
@@ -16,7 +16,7 @@
  */
 @Export
 @Open("org.apache.logging.log4j.core")
-@Version("2.20.1")
+@Version("2.21.0")
```
ppkarwasz, line 19: ... if not, this change can also be reverted.

ppkarwasz (changes requested):

PR #2678, opened 2024-06-19
Switch to parallel testing by default

ppkarwasz:
Since `main` tests were more fragile, we kept the tests in `main` to run sequentially, while those in `2.x` run in parallel.

This PR switches `main` testing back to parallel mode (1 test per JVM).

### DIFF
--- a/log4j-parent/pom.xml
+++ b/log4j-parent/pom.xml
@@ -995,40 +995,12 @@
         </executions>
       </plugin>
 
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


vy (approved):

garydgregory:
That sounds like a bad idea and against the goal of reproducible builds.

ppkarwasz:
@garydgregory,

Tests still run in an isolated environment (each test uses a separate JVM), so this doesn't change anything except the time we need to wait for a test result.

Sure, we have bad tests that don't assert anything except the creation of a log file, which is shared with 7 other tests. These tests could have false negatives, but for most tests I would only expect more failures if the tests share filesystem resources with other tests.

PR #2686, opened 2024-06-20
Merge `build` and `merge-dependabot` workflows

ppkarwasz:
These two workflows are triggered by almost the same events and are mutually exclusive.
This causes a lot of "skipped" jobs to be shown in Github Actions.

IMHO merging these workflows will improve developer experience.


vy:
> These two workflows are triggered by almost the same events and are mutually exclusive.

The rationale for the decision of keeping them apart was to clearly and explicitly separate `merge-dependabot.yaml`, which listens on `pull_request_target` events indirectly granting write access to the PR author. This way, in `merge-dependabot.yaml`, we can laser-focus to not accept anything that can potentially be malicious. Merging regular build and dependabot handling logics weakens this security aspect, AFAICT. But I am open to be convinced the other way around.

vy:
In the current situation, there are `build` and `merge-dependabot` workflows, one is skipped. In your proposal, we will only have `build` workflow, it will show-up *twice* for each PR in the Actions tab, and one will be skipped. I find this more confusing.

@ppkarwasz, shall we start from scratch? **What the problem you are trying to solve?**

ppkarwasz:
Closing this to comply with [INFRA policy](https://infra.apache.org/github-actions-policy.html).

PR #2690, opened 2024-06-24
Clean up thread context map implementations

ppkarwasz:
We bring some order in the current variety of `ThreadContextMap` implementations by:

- replacing `DefaultThreadContextMap` with the more performant `StringArrayThreadContextMap` introduced in #2330,
- removing the `CopyOnWriteSortedArrayThreadContextMap`, since it has no advantages over the default one. It doesn't even guarantee garbage-free logging.
- moving the `GarbageFreeSortedArrayThreadContextMap` to `log4j-core`, since no other logging implementation can profit from it.

The documentation has been updated to include the fact that the configuration options for `ThreadContext` actually depend on the underlying implementation:

- Log4j Core users have access to all the `ThreadContext` configuration options,
- Simple Logger users have a more restricted list of config options (they can not use the garbage free map any more). We could even restrict this further by allowing only the [`log4j2.simplelogShowContextMap`](https://logging.staged.apache.org/log4j/2.x/manual/simple-logger.html#log4j2.simplelogShowContextMap) and [`log4j2.isThreadContextMapInheritable`](https://logging.staged.apache.org/log4j/2.x/manual/systemproperties.html#isThreadContextMapInheritable) options,
- Logback users have no configuration options, since the thread context map is hardcoded to use `MDC`,
- JUL users have no configuration options, since the thread context map is hardcoded to no-op.

This PR deprecates #2593.


rgoers:
Before merging this please ensure you look into the issues I had with the performance tests and StringArrayThreadContextMap.

ppkarwasz:
@rgoers,

I can not reproduce the issues in this branch. I suspect that they were due to the changes in the `ThreadContextDataInjector` (which I reverted).

The performance is similar to the `GcFreeSortedArrayThreadContextMap`, which seems reasonable:

```
Benchmark                            (count)  (threadContextMapAlias)  Mode  Cnt     Score    Error  Units
ThreadContextBenchmark.putAndRemove        5                  Default  avgt   25    40,337 ±  2,409  ns/op
ThreadContextBenchmark.putAndRemove        5             NoGcOpenHash  avgt   25    10,500 ±  0,446  ns/op
ThreadContextBenchmark.putAndRemove        5          NoGcSortedArray  avgt   25    46,233 ±  0,484  ns/op
ThreadContextBenchmark.putAndRemove       50                  Default  avgt   25   129,873 ±  2,131  ns/op
ThreadContextBenchmark.putAndRemove       50             NoGcOpenHash  avgt   25    12,334 ±  3,297  ns/op
ThreadContextBenchmark.putAndRemove       50          NoGcSortedArray  avgt   25    65,971 ±  0,742  ns/op
ThreadContextBenchmark.putAndRemove      500                  Default  avgt   25  1192,296 ± 21,985  ns/op
ThreadContextBenchmark.putAndRemove      500             NoGcOpenHash  avgt   25    15,351 ±  2,341  ns/op
ThreadContextBenchmark.putAndRemove      500          NoGcSortedArray  avgt   25   117,499 ±  1,333  ns/op
```

The poor performance when the context map has 50 or more entries is due to the `remove` operation, since the array is not sorted and the element was added at the end.

Large context maps and the `remove` operation itself are rather rare, but I can optimize it by reverting the order of the search.

ppkarwasz:
After reversing the lookup order in `remove()` the benchmark improves slightly:

```
Benchmark                            (count)  (threadContextMapAlias)  Mode  Cnt    Score    Error  Units
ThreadContextBenchmark.putAndRemove        5                  Default  avgt   25   37,141 ±  2,795  ns/op
ThreadContextBenchmark.putAndRemove        5             NoGcOpenHash  avgt   25   12,097 ±  2,007  ns/op
ThreadContextBenchmark.putAndRemove        5          NoGcSortedArray  avgt   25   46,352 ±  0,285  ns/op
ThreadContextBenchmark.putAndRemove       50                  Default  avgt   25   82,210 ±  1,496  ns/op
ThreadContextBenchmark.putAndRemove       50             NoGcOpenHash  avgt   25   15,974 ±  2,184  ns/op
ThreadContextBenchmark.putAndRemove       50          NoGcSortedArray  avgt   25   67,429 ±  1,369  ns/op
ThreadContextBenchmark.putAndRemove      500                  Default  avgt   25  731,671 ± 13,308  ns/op
ThreadContextBenchmark.putAndRemove      500             NoGcOpenHash  avgt   25   18,260 ±  5,095  ns/op
ThreadContextBenchmark.putAndRemove      500          NoGcSortedArray  avgt   25  117,670 ±  1,649  ns/op
```

ppkarwasz:
@rgoers,

Unless we find some problems, I will merge this PR next week.

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
vy, line 1: Instead of making this class public, we can still keep it package-private next to `DefaultThreadContextMap`. This 
will result in encapsulation for Java version 8 and earlier users out there in the wild. (Just an idea to think about. I am leaving the decision to you.)

log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
```diff
@@ -26,102 +27,86 @@
...
+ *  </p>
+ *  <p>
+ *      Performance of the underlying {@link org.apache.logging.log4j.internal.map.UnmodifiableArrayBackedMap} exceeds
+ *      {@link HashMap} in all supported operations other than {@code get()}. Note that {@code get()} performance scales
+ *      linearly with the current map size, and callers are advised to minimize this work.
+ * </p>
  */
-public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyStringMap {
-    private static final long serialVersionUID = 8218007901108944053L;
+public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyStringMap, CopyOnWrite {
```
vy, line 41: Can't we deprecate `CopyOnWrite` too?

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
```diff
@@ -271,75 +223,39 @@ public Class<? extends LoggerContextFactory> loadLoggerContextFactory() {
...
-        LOGGER.warn("Falling back to {}", SimpleLoggerContextFactory.INSTANCE);
-        return SimpleLoggerContextFactory.INSTANCE;
-    }
-
     /**
      * @return The logger context factory to be used by {@link org.apache.logging.log4j.LogManager}.
      * @since 2.24.0
      */
     public LoggerContextFactory getLoggerContextFactory() {
-        return loggerContextFactoryLazy.get();
```
vy, line 293: Any particular reasons we moved away from the `Lazy`? (I support the change, BTW.)

log4j-perf-test/src/main/java/org/apache/logging/log4j/perf/appender/StringAppender.java
vy, line 1: Can't we instead introduce a `capacity` property to `ListAppender` and use `ListAppender(capacity=1)` instead?

src/site/antora/modules/ROOT/pages/manual/thread-context.adoc
```diff
@@ -120,9 +120,25 @@ executor.submit(() -> {
...
-include::partial$manual/systemproperties/properties-thread-context.adoc[leveloffset=+1]
+Simple Logger::
++
+See xref:manual/simple-logger.adoc#thread-context[Thread context configuration of Simple Logger].
+
+Log4j Core::
++
+See xref:manual/systemproperties.adoc#properties-thread-context[Thread context configuration of Log4j Core].
+
+SLF4J implementations (Logback)::
```
vy, line 133: I find this a little bit confusing, since `log4j-slf4j2-impl` is also an SLF4J implementation. Can we (shall we?) clarify this a little bit more in the docs?

src/site/antora/modules/ROOT/partials/manual/systemproperties/properties-thread-context-core.adoc
```diff
@@ -67,16 +67,14 @@ or predefined constant
 | Default value
 | `WebApp`
 
-(GC-free mode: `CopyOnWrite`)
 |===
 
 Fully specified class name of a custom
 link:../javadoc/log4j-api/org/apache/logging/log4j/spi/ThreadContextMap.html[`ThreadContextMap`]
-implementation class or one of the predefined constants:
+implementation class or (since version 2.24.0) one of the predefined constants:
```
vy, line 74:
```suggestion
implementation class or (since version `2.24.0`) one of the predefined constants:
```

vy (commented):
Submitting my initial review. I liked the simplification in general.

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
ppkarwasz: Personally I find using a separate package for private classes less messy. I guess it is a matter of taste.

log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/2690/commits/50f6c36fc113bc835704ab0a9e3c594d4aaef917

log4j-api/src/main/java/org/apache/logging/log4j/spi/Provider.java
ppkarwasz: This method and `getThreadContextMapInstance()` are called only **once** in `log4j-api`.

`Lazy` is still used in the `Log4jProvider` implementation, so we don't need to access static methods to create a `ContextDataInjector`, for example.

In Log4j Core 3, `Lazy` is replaced by DI scopes.

src/site/antora/modules/ROOT/pages/manual/thread-context.adoc
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/2690/commits/33ba51e2dc74d2901f529f79d0deeffa69b47235

log4j-perf-test/src/main/java/org/apache/logging/log4j/perf/appender/StringAppender.java
ppkarwasz: This class was used in  #2834 for the `ScopedContext` vs `ThreadContextMap` performance test.
I am leaving it here so it is easier to refactor that PR.

In general I agree: many classes specific to `log4j-perf-test` could use equivalent classes in other modules. We can probably fix it when we start running benchmarks regularly.

vy (approved):
I've only reviewed functional and documentation changes, they LGTM. (Did not have time to check the tests.)

ppkarwasz:
> I've only reviewed functional and documentation changes, they LGTM. (Did not have time to check the tests.)

Thanks, I am merging it.

PR #2695, opened 2024-06-29
Set missing error-prone.version

timlueg:
`${error-prone.version}` property was referenced in pom.xml but never set.

This should (also) fix https://github.com/apache/logging-log4j2/blob/main/BUILDING.adoc#compilation-in-intellij-idea-fails-with-java-plug-in-not-found-errorprone

### DIFF
--- a/log4j-parent/pom.xml
+++ b/log4j-parent/pom.xml
@@ -107,6 +107,7 @@
     <disruptor.version>4.0.0</disruptor.version>
     <elasticsearch.version>7.17.22</elasticsearch.version>
     <embedded-ldap.version>0.9.0</embedded-ldap.version>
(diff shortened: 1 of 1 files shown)


ppkarwasz:
Hi @timlueg,

The `error-prone.version` property is inherited from the `org.apache.logging:logging-parent` POM file.
See [apache/logging-parent/pom.xml](https://github.com/apache/logging-parent/blob/cb849457e0cf849e5a5f1a2b8775988574b6cc17/pom.xml#L239).

Regarding IntelliJ IDEA: if you update the Maven configuration, sometimes IDEA gets the processor path right, but it doesn't show it in the GUI, you need to look into `.idea/compiler.xml` directly.

Issue #2530, opened 2024-04-26
Revamp the `Extending` page


vy:
As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), revamp the `Extending` page such that

- Outline:
  - `LoggerContextFactory`
  - `ContextSelector`
  - `ConfigurationFactory`
  - `LoggerConfig`
  - `LogEventFactory`
  - `MessageFactory`
  - Revamp `customloglevels.adoc`
    - Take apache/logging-log4j-transform#92 into account
  - Link to “Lookups > Extending lookups”
  - Link to “Layouts > Extending layouts”
  - Link to “Appenders > Extending appenders”
  - Link to “Filters > Extending filters”
- Target `2.x`, port to `main`

vy:
Fixed by #2696.

PR #2696, opened 2024-07-01
Revamp the `Extending`, `Plugins`, `Architecture`, and `Programmatic configuration` pages

vy:
This PR addresses #2530.


src/site/antora/modules/ROOT/pages/manual/customconfig.adoc
```diff
@@ -14,372 +14,232 @@
...
+updateLoggers(appender, config);
 ----
 
-The difference is the use of `OutputStreamAppender` instead of
-`WriterAppender`.
+This approach is prone several problems:
+
+* Your code relies on Log4j Core internals which don't have any backward compatibility guarantees.
+You not only risk breaking your build at a minor Log4j Core version upgrade, but also make the life of Log4j maintainers trying to evolve the project extremely difficult.
+* You move out from the safety zone, where Log4j Core takes care of components' life cycle (initialization, reconfiguration, etc.), and step into a minefield seriously undermining the reliability of your logging setup.
```
ppkarwasz, line 242: I think we could be more specific than that:

* the code will not fail to compile at a minor release, because we take BC (too?) seriously. We might however add new required arguments to the builders, which will make the `build()` call fail. More probably we could deprecate methods.
* we can explicitly state that by "reliability" we mean that during the execution of this code, log events might be lost, unless the method calls are carefully ordered. 

Personally I find building components directly a better and type-safe alternative to `ConfigurationBuilder`.
The only problem is that building a configuration from scratch should **not** be performed, while the configuration is active.
The code should only run in a configuration factory or the `doConfigure()` method of a custom configuration.
`Configurator.reconfigure()` can be used to swap the old and new configuration.

ppkarwasz (changes requested):
Thanks for revamping all these pages!

For now I reviewed only `customconfig.adoc`. Given the quality of the original, full of wrong or deprecated examples, I see a net improvement in the page.
Yet there are still some corrections to do.

src/site/resources/.htaccess
```diff
@@ -44,8 +44,13 @@ RewriteRule "^log4j-mongodb3(/index)?\.html$" "manual/appenders.html#NoSQLAppend
 RewriteRule "^log4j-mongodb4(/index)?\.html$" "manual/appenders.html#log4j-mongodb4" [R=permanent,NE]
 RewriteRule "^log4j-slf4j2?-impl(/index)?\.html$" "manual/installation.html#impl-core-bridge-slf4j" [R=permanent,NE]
 RewriteRule "^manual/api-separation\.html$" "manual/api.html" [R=permanent]
+RewriteRule "^manual/customconfig\.html#ConfigurationFactory$" "manual/configuration.html#ConfigurationFactory" [R=permanent]
+RewriteRule "^manual/extending\.html#Appenders$" "manual/appenders.html#extending" [R=permanent]
+RewriteRule "^manual/extending\.html#Filters$" "manual/filters.html#extending" [R=permanent]
 RewriteRule "^manual/extending\.html#Layouts$" "manual/layouts.html#extending" [R=permanent]
+RewriteRule "^manual/extending\.html#Lookups$" "manual/lookups.html#extending" [R=permanent]
 RewriteRule "^manual/extending\.html#PatternConverters$" "manual/pattern-layout.html#extending-converters" [R=permanent]
+RewriteRule "^manual/extending\.html#Plugin_Builders$" "manual/plugins.html#declare-plugin-factory" [R=permanent]
```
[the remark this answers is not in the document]
vy: Removed all invalid redirects in d1149f00f98ed2a4ece7418028e2ac420ccd7025.

src/site/antora/modules/ROOT/pages/manual/customconfig.adoc
```diff
@@ -14,372 +14,232 @@
...
 ----
+<1> The default `ConfigurationBuilder` instance is obtained using link:../javadoc/log4j-core/org/apache/logging/log4j/core/config/builder/api/ConfigurationBuilderFactory.html#newConfigurationBuilder()[`ConfigurationBuilderFactory.newConfigurationBuilder()`] static method
+<2> Add the appender along with the layout
+<3> Add the root logger along with a level and appender reference
+<4> Create the configuration, but *don't initialize* it
++
+[TIP]
+====
+It is a good practice to not initialize ``Configuration``s when they are constructed.
+This task should ideally be delegated to <<Configurator>>.
```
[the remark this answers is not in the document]
ppkarwasz: No, I am not.

We probably should add to `ConfigurationBuilder` a `build()` method equivalent to `build(false)` and deprecate the latter.
IDEA reminds me each time that `boolean`s in public methods are a code smell.

src/site/antora/modules/ROOT/pages/manual/customconfig.adoc
```diff
@@ -14,372 +14,232 @@
...
-            "true", refs, null, config, null );
-        loggerConfig.addAppender(appender, null, null);
-        config.addLogger("org.apache.logging.log4j", loggerConfig);
-        ctx.updateLoggers();
-}
+include::example$manual/customconfig/Usage.java[tag=loadConfigurationFile,indent=0]
 ----
+<1> Passing the optional `LoggerContext` argument as null, since this `Configuration` is not associated with a `LoggerContext` yet
+<2> Passing the optional configuration name argument as null, it will default to the configuration source location
+<3> URI pointing to the configuration file; `file://path/to/log4j2.xml`, `classpath:log4j2.xml`, etc.
```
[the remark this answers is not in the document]
ppkarwasz: Maybe we should shorten the explanations:

- we can pass `LoggerContext` as `null`,
- we can pass `name` as `null`, since it is not used if `configLocation` is provided.

"since this `Configuration` is not associated with a `LoggerContext` yet" is IMHO wrong: each configuration is not associated with a logger context upon creation.

src/site/antora/modules/ROOT/pages/manual/customconfig.adoc
ppkarwasz: I'll make a PR.

vy: I have seen #2702 and [commented on it](https://github.com/apache/logging-log4j2/pull/2702#pullrequestreview-2154646468). I don't see any rationale (why it is better, etc.), but just code. Sorry, I disagree with this proposal. Let's not encourage anyone on using `PatternLayout.createDefaultLayout()`.

src/site/antora/modules/ROOT/pages/manual/customconfig.adoc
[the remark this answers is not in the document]
vy: Fixed in a44fe6cca642eb16fe263c049a55c3b6c0a72eaf.

src/site/antora/modules/ROOT/pages/manual/extending.adoc
```diff
@@ -14,491 +14,132 @@
...
-3.  Setting the property "log4j2.loggerContextFactory" in a properties
-file named "log4j2.LogManager.properties" to the name of the
-LoggerContextFactory class to use. The properties file must be on the
-classpath.
+=== `LoggerContextFactory`
+
+link:../javadoc/log4j-api/org/apache/logging/log4j/spi/LoggerContextFactory.html[`LoggerContextFactory`] is the factory class used by Log4j API implementations to create xref:manual/architecture.adoc#LoggerContext[`LoggerContext`]s.
+If you are using Log4j Core, you can provide a custom implementation using xref:manual/systemproperties.adoc#log4j2.loggerContextFactory[the `log4j2.loggerContextFactory` property].
+Another way to provide a custom `LoggerContextFactory` is to <<Provider,provide a custom `Provider`>>.
+But this is generally discouraged, since it requires a more complicated setup and is intended mostly to be used by Log4j API implementations.
```
ppkarwasz, line 97: I disagree: if you are using Log4j Core, you should not provide a custom implementation of the `LoggerContextFactory`.
You should use `ContextSelector` below.

src/site/antora/modules/ROOT/pages/manual/extending.adoc
```diff
@@ -14,491 +14,132 @@
...
-                                            @PluginElement Filter filter) {
-        List<AppenderRef> appenderRefs = Arrays.asList(refs);
-        return new LoggerConfig(LogManager.ROOT_LOGGER_NAME, appenderRefs, filter, level, additivity);
-    }
-}
-----
+=== `LoggerConfig`
+
+link:../javadoc/log4j-core/org/apache/logging/log4j/core/config/LoggerConfig.html[`LoggerConfig`] denotes the `Logger` configurations in a `Configuration`.
+A custom `LoggerConfig` needs to suffice following conditions:
```
ppkarwasz, line 119:
```suggestion
A custom `LoggerConfig` needs to satisfy the following conditions:
```

src/site/antora/modules/ROOT/pages/manual/extending.adoc
```diff
@@ -14,491 +14,132 @@
...
-----
-ListAppender list1 = ListAppender.createAppender("List1", true, false, null, null);
-ListAppender list2 = ListAppender.newBuilder().setName("List1").setEntryPerNewLine(true).build();
-----
+=== `MessageFactory2`
 
-[#Custom_Plugins]
-== Custom Plugins
+Log4j Core uses link:../javadoc/log4j-api/org/apache/logging/log4j/message/MessageFactory2.html[`MessageFactory2`] to create link:../javadoc/log4j-api/org/apache/logging/log4j/message/Message.html[`Message`]s.
+You can replace the default `MessageFactory2` implementation with a custom one of yours by using xref:manual/systemproperties.adoc#log4j2.messageFactory[the `log4j2.messageFactory` property].
```
ppkarwasz, line 142: We should probably warn the user that the default message factory must interpret format strings that contain `{}` as placeholder.

If they want to use other formatters, they should use the `LogManager` methods that accept a `MessageFactory`.

src/site/antora/modules/ROOT/pages/manual/extending.adoc
```diff
@@ -14,491 +14,132 @@
...
 
-// TODO
-See the xref:manual/plugins.adoc[Plugins] section of the manual.
+In the case of xref:manual/flowtracing.adoc[], Log4j Core uses link:../javadoc/log4j-api/org/apache/logging/log4j/message/FlowMessageFactory.html[`FlowMessageFactory`].
+You can replace the default `FlowMessageFactory` implementation with a custom one of yours by using xref:manual/systemproperties.adoc#log4j2.flowMessageFactory[the `log4j2.flowMessageFactory` property].
```
ppkarwasz, line 145: Same remark as for `MessageFactory`: a `FlowMesssageFactory` needs to accept format string that use `{}` as placeholders.

ppkarwasz (changes requested):
Regarding the `Extending` page, I have a couple of comments.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -14,88 +14,274 @@
...
+A class can be declared as a plugin by adding a link:../javadoc/log4j-core/org/apache/logging/log4j/core/config/plugins/Plugin.html[`@Plugin`] annotation, which is essentially composed of following attributes:
+
+`name`::
+Name of the plugin.
+It is recommended to be distinct among plugins sharing the same `category`.
+`name` matching is case-insensitive.
+
+`category` (optional)::
+A name used for grouping a set of plugins.
+`category` matching is case-sensitive.
```
ppkarwasz, line 56:
```suggestion
The namespace to which this plugin belongs.
Plugin names within a namespace must be unique.
```

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -14,88 +14,274 @@
...
+
+`elementType` (optional)::
+Name of the corresponding category of the xref:manual/configuration.adoc[Log4j configuration] file element this plugin belongs under.
+You can omit this attribute if your plugin is not meant to be represented in a configuration file.
```
ppkarwasz, line 60: IMHO `elementType` is deprecated.

It was used to indicate the generic tag name for a configuration file element, e.g., instead of `<PatternLayout>` you could use `<appender type="PatternLayout">`.
Nowadays this is mostly useless.

src/site/antora/modules/ROOT/pages/manual/extending.adoc
vy: Fixed in 43f84a784ca11ae1f8b4a064fa397327aa4f7e66.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -14,88 +14,274 @@
...
 
-* Required: This annotation validates that a value is non-empty. This covers a check for null as well as several other scenarios: empty CharSequence objects, empty arrays, empty Collection instances, and empty Map instances.
-* ValidHost: This annotation validates that a value corresponds to a valid hostname. This uses the same validation as InetAddress::getByName.
-* ValidPort: This annotation validates that a value corresponds to a valid port number between 0 and 65535.
+link:../javadoc/log4j-core/org/apache/logging/log4j/core/config/plugins/util/PluginManager.html[`PluginManager`] is responsible for discovering plugins and loading their descriptions.
+It locates plugins by looking in following places in given order:
 
-[#key-providers]
-== Key providers
+. Plugin descriptor files on the classpath.
```
ppkarwasz, line 264: - We should probably give the location of the `Log4j2Plugins.dat` file.
- Since we are talking also about OSGi, we should specify that the `log4j-core` classloader is used.
- This category also has a fallback: if no `Log4j2Plugins.dat` file is found, the plugin manager scans the `org.apache.loggging.log4j.core` package. Users have been unfortunately exploiting this fallback: if the `Log4j2Plugins.dat` is corrupted due to shading, they just remove it from the shaded artifact.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -14,88 +14,274 @@
...
+These files are generated automatically at compile-time by the Log4j plugin annotation processor.
+See <<plugin-registry>> for details.
 
-Some components within Log4j may provide the ability to perform data encryption. These components require a secret key to perform the encryption. Applications may provide the key by creating a class that implements the SecretKeyProvider interface.
+. *[OSGi only]* Serialized plugin listing files in each active OSGi bundle.
+A `BundleListener` is added on activation to continue checking new bundles after Log4j Core has started.
```
ppkarwasz, line 269: Should we say that it is the same `Log4j2Plugins.dat` file that we use?

src/site/antora/modules/ROOT/pages/manual/extending.adoc
vy: Fixed in 8fd4b7e680a8f12dacd4333d578aefbc8ad755ee.

src/site/antora/modules/ROOT/pages/manual/extending.adoc
vy: Fixed in 8fd4b7e680a8f12dacd4333d578aefbc8ad755ee.

ppkarwasz (changes requested):
The `Plugins` page seems really nice! :100:

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
vy: Fixed in 242a7ed45a08a52ab79941cc9f83fafc171e7a33.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
vy: I will skip this suggestion, since
1. We introduce a new term (`namespace`) without defining it
2. Case-sensitive matching detail is important, but removed
3. `name` definition already mentions about uniqueness

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
vy: I explained which class loader is used in 0b8c835862e8327ba74777784dae6c85ec31b84d.

I will skip sharing the implementation details, e.g., `Log4j2Plugins.dat`, `PM` scans `o.a.l.l.core`, etc. These are not features, but _hacks_ we do to make something work. Users should not take this behaviour for granted and implement features relying on these assumptions. Yes, I know, they are already there, and we will keep them for backward compatibility reasons. But shouting out loud about them will make it impossible to deprecate them.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
vy: I prefer not to – see [this comment](https://github.com/apache/logging-log4j2/pull/2696#discussion_r1663730329).

PR #2702, opened 2024-07-02
Provide alternatives to config modifications

ppkarwasz:
Warns the user about the danger of modifying the current configuration and proposes alternative solutions to the problem.

### DIFF
--- a/src/site/antora/modules/ROOT/examples/manual/customconfig/ModifyConfiguration1.java
+++ b/src/site/antora/modules/ROOT/examples/manual/customconfig/ModifyConfiguration1.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


vy (commented):
Sorry Piotr, I am against the proposed changes:

1. `Your code relies on Log4j Core internals, which have lesser compatibility guarantees` statement is not correct. We don't have any guarantees on internals. We don't want to. Any expression hinting even slightly different than this encourages users. The reason we still try to keep things backward compatible is there is no decent encapsulation mechanism in Java 8.
2. Creating a component (in this case `PatternLayout`) manually is wrong. It is analogous to creating Spring beans manually instead of requesting them from the `ApplicationContext`.
3. We encourage the usage of `PatternLayout.createDefaultLayout()`, which should have not existed in the first place.

ppkarwasz:
>     1. `Your code relies on Log4j Core internals, which have lesser compatibility guarantees` statement is not correct. We don't have any guarantees on internals. We don't want to. Any expression hinting even slightly different than this encourages users. The reason we still try to keep things backward compatible is there is no decent encapsulation mechanism in Java 8.

We **do** guarantee that classes which are not marked as private or internal will evolve according to semantic versioning.
In the future we might be more aggressive with releasing major versions (e.g. once every couple of years like MongoDB), but in the context of a single major version we will not do breaking changes.

>     2. Creating a component (in this case `PatternLayout`) manually is wrong. It is analogous to creating Spring beans manually instead of requesting them from the `ApplicationContext`.

I see this as a sort of Spring Java-based configuration, where you call Java constructors and setters explicitly.
In this example the attributes of `PatternLayout` are hard-coded, but users could use other mechanisms to inject those.

>     3. We encourage the usage of `PatternLayout.createDefaultLayout()`, which should have not existed in the first place.

+1, I will remove it.

ppkarwasz:
After discussing it in a video call with Volkan, I am closing this for now.

Issue #2703, opened 2024-07-03
Unwanted `TRACE` output in v2.23.1 with `-Dlog4j.debug=false`


vkolomeyko:
## Description

The effect of `-Dlog4j.debug=false` has changed in v2.23.1

## Configuration

**Version:** 2.23.1

**Operating system:** Windows

**JDK:** zulu17.46.19-ca-jdk17.0.9-win_x64

## Reproduction

Consider the following Java program:
```
package org.example;

import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.Logger;

public class Main {

    public static void main(String[] args) {
        System.setProperty("log4j.debug", "false");

        Logger log = LogManager.getLogger(Main.class);
        log.error("Hello world!");
    }
}
```
With Log4j `v2.23.0` it prints:
```
15:20:02.690 [main] ERROR org.example.Main - Hello world!
```
as expected.

However, when version is upgraded to  `v2.23.1`, the output becomes a lot more verbose:
```
...
2024-07-03T14:23:42.658351400Z main TRACE Using default SystemClock for timestamps.
2024-07-03T14:23:42.658351400Z main DEBUG org.apache.logging.log4j.core.util.SystemClock supports precise timestamps.
2024-07-03T14:23:42.658351400Z main TRACE Using DummyNanoClock for nanosecond timestamps.
2024-07-03T14:23:42.658351400Z main DEBUG Reconfiguration complete for context[name=1d44bcfa] at URI NULL_SOURCE (org.apache.logging.log4j.core.LoggerContext@67e2d983) with optional ClassLoader: null
2024-07-03T14:23:42.658351400Z main DEBUG Shutdown hook enabled. Registering a new one.
2024-07-03T14:23:42.659352600Z main DEBUG LoggerContext[name=1d44bcfa, org.apache.logging.log4j.core.LoggerContext@67e2d983] started OK.
15:23:42.661 [main] ERROR org.example.Main - Hello world!
2024-07-03T14:23:42.663882800Z pool-1-thread-1 DEBUG Stopping LoggerContext[name=1d44bcfa, org.apache.logging.log4j.core.LoggerContext@67e2d983]
2024-07-03T14:23:42.664881900Z pool-1-thread-1 DEBUG Stopping LoggerContext[name=1d44bcfa, org.apache.logging.log4j.core.LoggerContext@67e2d983]...
[...]

PR #2704, opened 2024-07-03
Explicitly checking for `-Dlog4j.debug=false`

vkolomeyko:
As reported in:
https://github.com/apache/logging-log4j2/issues/2703


ppkarwasz (approved):
@vkolomeyko,

Thanks for catching this bug and your contribution!

Issue #2679, opened 2024-06-19
`StrSubstitutor` cannot parse nested braces


vy:
As reported in #2666,

```xml
<PatternLayout pattern="${env:TZ:-%d{yyyy-MM-dd HH:mm:ss}{GMT+00}}"/>
```

leaves behind a trailing `}` while formatting date. That is, `StrSubstitutor` doesn't correctly parse nested braces.

ppkarwasz:
To be more precise, the `StrLookup` parses the string `${env:TZ:-%d{yyyy-MM-dd HH:mm:ss}{GMT+00}}` as:

1. A lookup expression (`${env:TZ:-%d{yyyy-MM-dd HH:mm:ss}`), where `env:TZ` is the key used for the lookup and `%d{yyyy-MM-dd HH:mm:ss}` is the default value.
2. The literal `{GMT+00}}`.

The main problem here is to have a **coherent** set of escape rules for lookup expressions and pattern expressions.
Right now we have the following escaping rules:

* the special sequence `${` can be escaped as `$${`. This only works for the first level of lookups. If `${foo}` expands to `bar`, then `$${foo}` is not expanded, but `${$${foo}}` returns the expansion of `${$bar}`.
* the special sequence `:-` can be escaped as `:\-`, but if `\-` appears anywhere else, it will give two characters.
* the pattern sequence `%d` can be escaped as `%%d`.
* there is **no** way to escape a closing brace `}` neither in a lookup nor in the options of a pattern expression.

This resembles more the way [`cmd.exe` parses a command line](https://stackoverflow.com/a/4095133/11748454) than the simpler `bash` expansion.

ngocnhan-tran1996:
Hi @vy , @ppkarwasz 

I have a PR but I think this has more edge case. Please take a look and tell me if I miss something

https://github.com/apache/logging-log4j2/pull/2712

ppkarwasz:
Looking at [bash parameter expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html), I would propose to just introduce a simple escape mechanism that allows users to escape the closing `}` as `\}`.
The initial example would then need to be written as:
```
<PatternLayout pattern="${env:TZ:-%d{yyyy-MM-dd HH:mm:ss\}{GMT+00\}}"/>
```

This does not make the escaping rules coherent, but at least they are simpler:

- the sequence `${` is escaped as `$${`,
[...]

PR #2712, opened 2024-07-04
StrSubstitutor cannot parse nested braces

ngocnhan-tran1996:
As reported in:

[2679](https://github.com/apache/logging-log4j2/issues/2679)

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/lookup/ConverterKeysSubstitutorTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/lookup/ConverterKeysSubstitutorTest.java
@@ -83,4 +83,17 @@ class ConverterKeysSubstitutorTest {
 
   }
 
+  @Test
+  void test_missing_braces() {
+
+    final Map<String, String> map = new HashMap<>();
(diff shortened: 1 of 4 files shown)


ppkarwasz:
@ngocnhan-tran1996,

Thank you for the PR.
Unfortunately this does not solve #2679, but just your particular problem with the `%d{...}` pattern converter as default value of a lookup.

IMHO `StrLookup` should be aware of the existence and syntax of pattern specifiers `%<something>{...}`, it should just count the number of `{` and `}`.
We also need to have a mechanism to insert a `}` as default value of a lookup.

ngocnhan-tran1996:
@ppkarwasz 

I try to create class `ConverterKeysSubstitutor` for pattern specifiers ` %<something>{...}`. Please take a look and give me more test or some advice.

ppkarwasz:
> IMHO `StrLookup` should be aware of the existence and syntax of pattern specifiers `%<something>{...}`, it should just count the number of `{` and `}`. We also need to have a mechanism to insert a `}` as default value of a lookup.

Ooops, I am really sorry, I meant "should not be aware".

ngocnhan-tran1996:
@ppkarwasz 

I see, thanks for your help. I will close this PR

PR #2715, opened 2024-07-05
Revamp `cloud.adoc` for service-oriented architectures

### LINKED ISSUE #2510: Create a `Logging against ELK-stack` guide
As agreed in [the site structure document](https://docs.google.com/document/d/10Fu7oqDzdM_D6LbexzwX9arh51Tic7AGvkWTQrL6jjQ/edit), create a _"Logging against ELK-stack"_ guide such that

- Mention about JTL and [ecs-logging-java](/elastic/ecs-logging-java)
- Research on appenders – One can either use a `SocketAppender` targeted to Logstash (we have an example for this in `log4j-layout-template-json-test/pom.xml`) or write to files scraped by Beats. We need to evaluate the current status quo and share the most recent best practices.
- Ask for review from Elastic (We can ask using their Forums, it worked for @vy in the past.)
- If possible, retire/remove `cloud.adoc`
- Target `2.x`, port to `main`

vy:
Closes #2510.


src/site/antora/modules/ROOT/pages/manual/cloud.adoc
```diff
@@ -15,549 +15,267 @@
...
 
-Docker provides alternate https://docs.docker.com/config/containers/logging/configure/[logging drivers],
-such as https://docs.docker.com/config/containers/logging/fluentd/[fluentd], that
-can be used to redirect the standard output stream to a log forwarder or log aggregator.
+Platform stores logs differently::
 
-When routing to a log forwarder it is expected that the forwarder will have the same lifetime as the
-application. If the forwarder should fail the management tools would be expected to also terminate
-other containers dependent on the forwarder.
+There is no longer `/var/log/tomcat/app.log` combining all logs of a monolith.
```
ppkarwasz, line 46: _Nit_: Most Linux distributions use `/var/log/tomcat/catalina.out` for `stdout/stderr`.

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
```diff
@@ -15,549 +15,267 @@
...
-an ELK stack. Here we will briefly cover how one can forward Log4j generated
-events first to Logstash and then to Elasticsearch.
+Once you decide on <<proxy,the log proxy>> to use, the choice of appender pretty much becomes self-evident.
+Nevertheless, there are some tips we recommend you to practice:
 
-=== Log4j Configuration
+* *For writing to console*, use a xref:manual/appenders.adoc#ConsoleAppender[Console Appender] and make sure to configure its `direct` attribute to `true` for the maximum efficiency.
 
-==== JsonTemplateLayout
+* *For writing to an external service*, use a xref:manual/appenders.adoc#SocketAppender[Socket Appender] and make sure to configure the null delimiter of the associated layout.
```
ppkarwasz, line 101: Should we mention to users to prefer `TCP` over `UDP`?
At least if `SocketAppender` is concerned. We don't have an appender that uses `QUIC`.

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
```diff
@@ -15,549 +15,267 @@
...
+We strongly advise you to separate the logging configuration from the application and couple them in an environment-specific way.
+This will allow you to
+
+* Address environment-specific configurations (e.g., logging verbosity needs of test and production can be different)
+* Ensure Log4j configuration changes applies to all affected Log4j-using software without the need to manually update their Log4j configuration one by one
+
+How to implement this separation pretty much depends on your setup.
+We will share some recommended approaches to give you an idea:
+
+Ship configuration files along with the application, and choose a subset of them during deployment::
```
ppkarwasz, line 130: There is literally half a dozen ways to deploy the configuration.

1. Providing `log4j2.configurationFile` that points to a centralized configuration store,
2. Merging different configuration files (why not merge a local "fail-over" config with a remote one?),
3. Use arbiters to select the logging profile,
4. Use property sources to configure the logging level,
5. Mounting an external volume on the container,
6. Custom configurations.

I think we should point the user to the appropriate chapter and let them choose.

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
```diff
@@ -15,549 +15,267 @@
...
-        "$resolver": "mdc",
-        "flatten": {
-            "prefix": "_"
-        },
-        "stringified": true
-    }
-}
-----
+A common excuse for writing logs to files is to use them as a buffer prior to ingesting them either to the logging system proxy or to the log storage system.
+Even in this case, most of the time it is sign of bad architecture design, and costs extra resources and efficiency degradation.
```
ppkarwasz, line 116: Isn't this what proxies are doing?

Flume is out of shape, but it has such a functionality (as far as I understand it is the only functionality worth splitting).

ppkarwasz (approved):
LGTM

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
vy: Improved the text in e03faec93814ea5f57e5c8fdc3a2802801473cbf.

> * Use arbiters to select the logging profile
> * Custom configurations

This section is about _recommended_ approaches, I don't think these would qualify so. No modern SaaS, cloud, service mesh, etc. deployment works with arbiters-minded configuration – i.e., one configuration for all environments.

> Use property sources to configure the logging level

I don't understand how this is relevant.

> Mounting an external volume on the container,

I already explained this below.

While reviewing this section please keep in mind that we share recommended best practices, not every single possible way of deploying a Log4j configuration. Judging from my experience, *all* companies I know of use one of the two techniques I cited here.

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
ppkarwasz: What I mean is that the practice of buffering logs in logs files might still be valid.
However the logic that does that probably should not be in the application itself, but in the sidecar service that forwards the logs from the application.

vy: Removed the paragraph in fed8ac44975b212f60b070c3a6aef34fe0023d6b.

src/site/antora/modules/ROOT/pages/manual/cloud.adoc
ppkarwasz: > This section is about _recommended_ approaches, I don't think these would qualify so. No modern SaaS, cloud, service mesh, etc. deployment works with arbiters-minded configuration – i.e., one configuration for all environments.

I wouldn't be sure that arbiters (especially `SpringBootArbiter`) are not used in modern SaaS.
If they use Spring Boot profiles, they either use `SpringBootArbiter` or separate configurations.
I am not really sure, which one is more popular or which one should be recommended.

> While reviewing this section please keep in mind that we share recommended best practices, not every single possible way of deploying a Log4j configuration. Judging from my experience, _all_ companies I know of use one of the two techniques I cited here.

I rely on your experience here.
Most companies I know are SMEs and are not using cloud environments.

PR #2768, opened 2024-07-25
Fix `ClassCastException` in LMAX Disruptor 3 initialization

jackshirazi:
During startup with an agent, the resulting object can be in the wrong classloader leading to a ClassCastException. Catching that falls back to the old behaviour which works fine

The new behaviour was introduced in 2.23.0 with apache#2112

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
@@ -55,6 +55,7 @@ class AsyncLoggerDisruptor extends AbstractLifeCycle {
         if (DisruptorUtil.DISRUPTOR_MAJOR_VERSION == 3) {
             try {
                 return (EventHandler<RingBufferLogEvent>)
+                    // Avoid using `LoaderUtil`, which might choose an incorrect class loader – see #2768.
                     Class.forName("org.apache.logging.log4j.core.async.RingBufferLogEventHandler")
                         .getConstructor()
                         .newInstance();
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java
@@ -54,8 +54,11 @@ class AsyncLoggerDisruptor extends AbstractLifeCycle {
     private static EventHandler<RingBufferLogEvent> createEventHandler() {
         if (DisruptorUtil.DISRUPTOR_MAJOR_VERSION == 3) {
             try {
-                return LoaderUtil.newInstanceOf("org.apache.logging.log4j.core.async.RingBufferLogEventHandler");
-            } catch (final ClassCastException | ReflectiveOperationException | LinkageError e) {
+                return (EventHandler<RingBufferLogEvent>)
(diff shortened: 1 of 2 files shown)


ppkarwasz (changes requested):
@jackshirazi,

Thank You for bringing this problem to our attention.

Instead of ignoring the `CCE`, I would rather solve its cause.

While `LoaderUtil` might be useful to load classes specified by the user, in this case its usage is not appropriate: we exactly know were the class is and we don't need to look in the TCCL.

Can you replace the `LoaderUtil` invocation with a simple `Class.forName(...).getConstructor()...` chain of methods?

A test case would be useful, although in this case we can probably just add a comment **not** to use `LoaderUtil` with a link to this issue.

jackshirazi:
We don't use LoaderUtil. The linked issue is [this](https://github.com/elastic/apm-agent-java/issues/3691). The full sequence is pretty convoluted
1. Elastic agent opens a url connection, java.net.URL.openConnection which loads sun.net.www.protocol.http.HttpURLConnection
2. HttpURLConnection inits it's logger, a jul logger
3. The command line set -Djava.util.logging.manager=org.apache.logging.log4j.jul.LogManager so apache logger is used
4. Because this was loaded from the elastic agent, the current classloader is co.elastic.apm.agent.premain.ShadedClassLoader
5. ShadedClassLoader loads the log4j classes up to org.apache.logging.log4j.core.async.RingBufferLogEventHandler
6. The log4j2 class AsyncLoggerDisruptor casts the RingBufferLogEventHandler instance as a EventHandler, but EventHandler was loaded by log4j2 in the app classloader
7. hence the CCE

Because this is a class loading race condition with the JVM classes using the log4j2 logger before the JVM is initialized (agent's kick in at premain), this is complex to solve outside of log4j2, whereas this patch is pretty simple

ppkarwasz:
Did you try replacing `LoaderUtil.newInstanceOf` with:

```java
return (EventHandler<RingBufferLogEvent>)
        Class.forName("org.apache.logging.log4j.core.async.RingBufferLogEventHandler")
                .getConstructor()
                .newInstance();
```

This version should also be more GraalVM-friendly and shouldn't cause a `CCE`.

jackshirazi:
Again, we're not using LoaderUtil. This is also not in Graal (Graal doesn't yet support agents, and when it does it won't have this kind of issue because it flattens classloading)

ppkarwasz:
> Again, we're not using LoaderUtil. This is also not in Graal (Graal doesn't yet support agents, and when it does it won't have this kind of issue because it flattens classloading)

Sorry, I am not referring to the code of your application, but these lines in Log4j Core:

https://github.com/apache/logging-log4j2/blob/2d5f5828e6996ac330695ad35f2a713372d895d2/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLoggerDisruptor.java#L56-L60

I think that line 57 constitutes the problem, since it might load the `RingBufferLogEventHandler` class from a different classloader than the one that loaded `AsyncLoggerDisruptor`.

jackshirazi:
Thanks, I'll test that

jackshirazi:
that works, sorry took me a while to get what you were suggesting

vy:
@ppkarwasz, I cannot think of an easy way to test this. The rest LGTM. How shall we proceed?

ppkarwasz:
> @ppkarwasz, I cannot think of an easy way to test this. The rest LGTM. How shall we proceed?

We should at least add a code comment "Don't use `LoaderUtil`", otherwise we risk reverting the change. :wink:

PR #2783, opened 2024-07-31
Disable programmatic configuration if `log4j1.compatibility` is `false`

### LINKED ISSUE #2778: Disable programmatic configuration if `log4j1.compatibility` is `false` (LOG4J2-3467)
This issue is moved from the [LOG4J2-3467](https://issues.apache.org/jira/browse/LOG4J2-3467) JIRA issue.

The `log4j-1.2-api` artifact is a bridge, whose main purpose is to forward Log4j 1 "API" calls to Log4j 2 API.

Version 2.17.2 of the artifact introduced support for configuring Log4j Core using the Log4j 1 `PropertyConfigurator` and `DOMConfigurator` classes.
This feature should be disabled by default and depend on the opt-in `log4j1.compatibility` property, because:

* many third-party **libraries** (not related to logging) take upon themselves the reconfiguration of Log4j 1. This steals the control of logging configuration from application developers (see [How to disable log4j programmatic reconfiguration from a 3rd party jar](https://stackoverflow.com/questions/78801614/how-to-disable-log4j-programmatic-reconfiguration-from-a-3rd-party-jar) for example).
* if some **applications** actually use those classes, they can always set `log4j1.compatibility` to `true`. This configuration property is also used to enable the scanning of the classpath for `log4j.properties` and `log4j.xml` files.

ppkarwasz:
Closes #2778.


log4j-1.2-api/src/main/java/org/apache/log4j/xml/DOMConfigurator.java
```diff
@@ -66,20 +69,34 @@
...
 
+    private static boolean isFullCompatibilityEnabled() {
+        return PropertiesUtil.getProperties().getBooleanProperty(ConfigurationFactory.LOG4J1_EXPERIMENTAL);
+    }
+
+    private static void warnFullCompatibilityDisabled() {
+        LogLog.warn(
+                "Ignoring `DOMConfigurator` call, since `log4j1.compatibility` is not enabled.\n"
+                        + "See https://logging.staged.apache.org/log4j/2.x/migrate-from-log4j1.html#log4j1.compatibility for details.");
+    }
```
vy, line 80: :bulb: Idea: Instead of these two methods, you could have only introduced a wrapper method as follows:

```
private static void whenFullCompatibilityEnabled(ThrowingRunnable runnable) {
    // ...
}
```

vy (approved):

log4j-1.2-api/src/main/java/org/apache/log4j/xml/DOMConfigurator.java
ppkarwasz: Sure, but is it worth introducing `ThrowingRunnable`?

vy: @ppkarwasz, I'm leaving the decision to you. I am fine with either approaches.

src/changelog/.2.x.x/2778_configurator_no_op.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
```
ppkarwasz, line 5: Should be `changed`, since it changes the previous behavior.

vy: @ppkarwasz, maybe indeed `changed` is a better option to stress the change in the behaviour.

garydgregory: I agree that 'changed' is better.

ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/2783/commits/e6c7f197f2f9a3dda109b669ea9b2d73d5a77d24

Issue #2542, opened 2024-04-29
[MS18] Documentation revamp 5/5


vy:
This is the 5<sup>th</sup> of the 5 STF milestones planned to revamp the website and manual.

### Tasks

- #2526 
- #2528 
- #2529

grobmeier:
Thanks, checked and submitted

PR #2800, opened 2024-08-07
Fix links and remove per-module pages

ppkarwasz:
This PR:

- fixes internal links as far as IDEA can find.
- removes most `log4j-<something>.adoc` pages, if their content is available elsewhere. For example `log4j-docker` is completely described in `lookups.adoc`.

This is the last documentation touch of #2542.


src/site/antora/modules/ROOT/pages/manual/appenders/message-queue.adoc
vy, line 1: Don't these two corrections imply that associated features were never used by anyone?

vy (approved):

src/site/antora/modules/ROOT/pages/soa.adoc
```diff
@@ -167,7 +167,7 @@ In this section, we will share guides on some popular integration scenarios.
 [#docker]
 === Docker
 
-See xref:log4j-docker.adoc[Log4j Docker] for Docker-specific Log4j features, e.g., xref:manual/lookups.adoc#DockerLookup[Docker Lookup].
+See xref:manual/lookups.adoc#DockerLookup[Docker Lookup].
```
vy, line 170: Could you restore this sentence in #2806, please?

src/site/antora/modules/ROOT/pages/manual/appenders/message-queue.adoc
ppkarwasz: These instructions were in the current [`log4j-flume-ng.adoc`](
https://github.com/apache/logging-log4j2/blob/4b2f377114ec64f38eeadcab2d425dba48b94348/src/site/antora/modules/ROOT/pages/log4j-flume-ng.adoc) file.

Not sure if anybody actually uses them. @rgoers, any idea?
The `PERSISTENCE` manager uses a rarely updated `je` library, whose only advantage is that it is ASL licensed. If the dependency was not in the documentation we could switch to the better maintained, EPL licensed and smaller `h2` library.

PR #2846, opened 2024-08-16
Add support for custom context data in `ctx` lookup

### LINKED ISSUE #2331: ContextDataInjector ignores custom ContextDataProvider
## Description

Per documentation I should be able to create a [custom ContextDataProvider](https://logging.apache.org/log4j/2.x/manual/extending.html#ContextDataProvider). However, when following the documented steps, the context is not resolved. This is also reported [here](https://stackoverflow.com/questions/77455596/define-contextdataprovider-in-log4j2-at-creation-time-of-logger).

I created a [minimal reproducible example](https://github.com/lukasalexanderweber/ContextDataProviderMRE/) where I followed the steps described in the docs:

1. Custom implementations should implement the `org.apache.logging.log4j.core.util.ContextDataProvider` interface ([MyContextDataProvider](https://github.com/lukasalexanderweber/ContextDataProviderMRE/blob/main/src/main/java/custom/context/data/provider/MyContextDataProvider.java))
2. and declare it as a service by defining the implementation class in a file named `META-INF/services/org.apache.logging.log4j.core.util.ContextDataProvider` ([org.apache.logging.log4j.core.util.ContextDataProvider](https://github.com/lukasalexanderweber/ContextDataProviderMRE/blob/main/src/main/resources/META-INF/services/org.apache.logging.log4j.core.util.ContextDataProvider))

I created a [unit test](https://github.com/lukasalexanderweber/ContextDataProviderMRE/blob/main/src/test/java/custom/context/data/provider/MyContextDataProviderTest.java) with a [log4j2-test.properties setup](https://github.com/lukasalexanderweber/ContextDataProviderMRE/blob/main/src/test/resources/log4j2-test.properties) and expect for the path `src/test/logs/${ctx:tenant}` that the tenant is resolved by `MyContextDataProvider`. However, an error is printed that the file with the unresolved tenant cannot be created. It works as expected when the tenant is set via `ThreadContext.put("tenant", "tenant1");`

## Configuration

**Version:** 2.23.0

**JDK:** Corretto-17.0.8.8.1

## Logs

```
2024-03-01T08:40:19.499742800Z main ERROR Unable to create file src/test/logs/${ctx:tenant}/logs.log java.io.IOException: Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch
	at java.base/java.io.WinNTFileSystem.canonicalize0(Native Method)
	at java.base/java.io.WinNTFileSystem.canonicalize(WinNTFileSystem.java:462)
	at java.base/java.io.File.getCanonicalPath(File.java:626)
	at java.base/java.io.File.getCanonicalFile(File.java:651)
	at org.apache.logging.log4j.core.util.FileUtils.makeParentDirs(FileUtils.java:141)
	at org.apache.logging.log4j.core.appender.rolling.RollingFileManager$RollingFileManagerFactory.createManager(RollingFileManager.java:863)

```

## Reproduction

see [ContextDataProviderMRE](https://github.com/lukasalexanderweber/ContextDataProviderMRE)

ppkarwasz:
This backports (from the `2.25.x` branch), the support for custom context data providers in the `ctx` lookup.
The feature was first introduced in #2438.

Closes #2331.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
@@ -87,19 +87,33 @@ public class ThreadContextDataInjector {
         return Collections.unmodifiableList(providers);
     }
 
+    private abstract static class AbstractContextDataInjector implements ContextDataInjector {
+
+        final List<ContextDataProvider> providers;
+
+        AbstractContextDataInjector() {
+            this.providers = getProviders();
+        }
+
+        @Override
+        public Object getValue(String key) {
+            for (final ContextDataProvider provider : providers) {
+                final Object value = provider.getValue(key);
+                if (value != null) {
+                    return value;
+                }
+            }
+            return null;
+        }
+    }
+
     /**
      * Default {@code ContextDataInjector} for the legacy {@code Map<String, String>}-based ThreadContext (which is
      * also the ThreadContext implementation used for web applications).
      * <p>
      * This injector always puts key-value pairs into the specified reusable StringMap.
      */
-    public static class ForDefaultThreadContextMap implements ContextDataInjector {
-
-        private final List<ContextDataProvider> providers;
-
-        public ForDefaultThreadContextMap() {
-            providers = getProviders();
-        }
+    public static class ForDefaultThreadContextMap extends AbstractContextDataInjector {
 
         /**
          * Puts key-value pairs from both the specified list of properties as well as the thread context into the
@@ -171,12 +185,7 @@ public class ThreadContextDataInjector {
      * <p>
      * This injector always puts key-value pairs into the specified reusable StringMap.
      */
-    public static class ForGarbageFreeThreadContextMap implements ContextDataInjector {
-        private final List<ContextDataProvider> providers;
-
-        public ForGarbageFreeThreadContextMap() {
-            this.providers = getProviders();
-        }
+    public static class ForGarbageFreeThreadContextMap extends AbstractContextDataInjector {
 
         /**
          * Puts key-value pairs from both the specified list of properties as well as the thread context into the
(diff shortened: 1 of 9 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
```diff
@@ -87,19 +87,33 @@ private static List<ContextDataProvider> getServiceProviders() {
         return Collections.unmodifiableList(providers);
     }
 
+    private abstract static class AbstractContextDataInjector implements ContextDataInjector {
+
+        protected final List<ContextDataProvider> providers;
+
+        protected AbstractContextDataInjector() {
```
vy, line 94: Can't these two `protected` be `private` instead?

src/changelog/.2.x.x/2331_custom_context_data_in_lookup.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="added">
+  <issue id="2331" link="2330_add_faster_web_app_context_map.xml"/>
+  <description format="asciidoc">Add support for custom context data in `ctx` lookup.</description>
```
vy, line 7: 1. Shouldn't this rather be a `fixed`? Yes, we're _adding_ a new behaviour, but one that was missing. Nowhere we had stated that `ctx` lookup (or any particular component) won't work with custom `CDP`s.
2. `issue.link` needs to be fixed.
3. Could you rewrite the `description` a little bit more accurately, please? (Suggestion: _"Fix custom thread-context data provider handling in lookups and filters"_ – since it is not only the `ctx` lookup we fixed.)

vy (approved):
Requested minor changes, overall LGTM.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThreadContextDataInjector.java
ppkarwasz: I think this is mostly a question of taste, since the class is `private`.

I switched them to package private in https://github.com/apache/logging-log4j2/pull/2846/commits/c475dbf0af351bc30dc7f94e3f943ce4386ca0fa, so we don't need any synthetic accessors.

src/changelog/.2.x.x/2331_custom_context_data_in_lookup.xml
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/2846/commits/c475dbf0af351bc30dc7f94e3f943ce4386ca0fa

PR #2853, opened 2024-08-18
Add `useFqdn` attribute to RFC5424 Layout

### LINKED ISSUE #1740: With the parameter $PreserveFQDN on the logged-in host is not the full host(FQDN)
## Description

When on the client side you activate the **$PreserveFQDN** property on **/etc/rsyslog.conf** , the logs do not include the FQDN and only include the host 

## Configuration
My rsyslog.conf configuration (everything is default except the parameter)
```txt
$PreserveFQDN on
```
The /etc/hosts file
```txt
127.0.0.1 host.app.example host
```

And the log4j2 configuration
```xml
<Configuration>
	<Appenders>
		<Console name="LogToConsole" target="SYSTEM_OUT">
			<PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n" />
		</Console>
		<Syslog name="RFC5424" format="RFC5424" host="localhost" port="514" protocol="TCP" appName="app" includeMDC="true" facility="LOCAL0" enterpriseNumber="18060" newLine="true" messageId="app-ID" mdcId="mdc" id="app-MDCID" connectTimeoutMillis="1000" immediateFail="true" reconnectionDelayMillis="5000">
			<LoggerFields>
				<KeyValuePair key="thread" value="%t" />
				<KeyValuePair key="priority" value="%p" />
				<KeyValuePair key="category" value="%c" />
				<KeyValuePair key="exception" value="%ex" />
				<KeyValuePair key="message" value="%m" />
			</LoggerFields>
		</Syslog>
	</Appenders>
	<category name="org.springframework.beans">
		<priority value="info" />
	</category>
	<Loggers>
		<Logger name="foo.example.ErrorService" additivity="false">
			<AppenderRef ref="RFC5424" />
			<AppenderRef ref="LogToConsole" />
		</Logger>
	</Loggers>
</Configuration>
```

**Version:** [Log4j version] 

> 2.17.0

**Operating system:** [OS and version]

> SUSE Linux Enterprise Server 12 SP3

**JDK:** [JDK distribution and version]  

> OpenJDK 64-Bit Server VM (build 25.292-b10, mixed mode)

## Logs

```log
2023-08-24 14:01:49.569 host app[4539] [status] Starting...
```

## Reproduction

Activating the **$PreserveFQDN** on parameter and having in **/etc/hosts** an FQDN longer than the normal host should be enough, but ask any questions you may have

ppkarwasz:
Since RFC5424 specifies that the [`HOSTNAME`](https://datatracker.ietf.org/doc/html/rfc5424#section-6.2.4) field should contain the FQDN of the current host, we add a `useFqdn` configuration attribute to RFC5424 Layout that replaces the usage of `InetAddress.getHostName()` with `InetAddress.getCanonicalHostName()`.

Closes #1740.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
@@ -36,10 +36,9 @@ import org.apache.logging.log4j.core.appender.TlsSyslogFrame;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.Node;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
-import org.apache.logging.log4j.core.config.plugins.PluginAttribute;
-import org.apache.logging.log4j.core.config.plugins.PluginConfiguration;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
+import org.apache.logging.log4j.core.config.plugins.PluginBuilderFactory;
 import org.apache.logging.log4j.core.config.plugins.PluginElement;
-import org.apache.logging.log4j.core.config.plugins.PluginFactory;
 import org.apache.logging.log4j.core.layout.internal.ExcludeChecker;
 import org.apache.logging.log4j.core.layout.internal.IncludeChecker;
 import org.apache.logging.log4j.core.layout.internal.ListChecker;
@@ -167,7 +166,7 @@ public final class Rfc5424Layout extends AbstractStringLayout {
         this.appName = appName;
         this.messageId = messageId;
         this.useTlsMessageFormat = useTLSMessageFormat;
-        this.localHostName = NetUtils.getLocalHostname();
+        this.localHostName = NetUtils.getCanonicalLocalHostname();
         ListChecker checker = null;
         if (excludes != null) {
             final String[] array = excludes.split(Patterns.COMMA_SEPARATOR);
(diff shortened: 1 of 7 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/util/NetUtils.java
```diff
@@ -52,23 +53,35 @@ private NetUtils() {
...
      */
     public static String getLocalHostname() {
+        return getHostname(InetAddress::getHostName);
+    }
+
+    /**
+     * This method gets the FQDN of the machine we are running on. Returns
+     * "UNKNOWN_LOCALHOST" in the unlikely case where the host name cannot be found.
+     *
+     * @return String the canonical name of the local host
```
vy, line 63:
```suggestion
     * This method gets the FQDN of the machine we are running on.
     * It returns {@value UNKNOWN_LOCALHOST} if the host name cannot be found.
     *
     * @return The canonical name of the local host; or {@value UNKNOWN_LOCALHOST}, if cannot be found.
```

log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java
```diff
@@ -642,80 +642,128 @@ public String toString() {
...
         private String exceptionPattern;
+
+        @PluginBuilderAttribute
         private boolean useTLSMessageFormat;
+
+        /**
+         * If {@code true}, the FQDN of the current host will be used.
+         */
+        @PluginBuilderAttribute
+        private boolean useFqdn = false;
```
vy, line 756: Given that enabling this fixes our RFC5424 compliance, shouldn't this feature be a default instead of opt-in? I am aware that this will be a breaking change, yet, I'd favor an approach as follows:

1. Change the default
2. If we receive a bug report, introduce the flag as an opt-out (contrary to the current opt-in)

ppkarwasz: Excellent idea!

Fixed in https://github.com/apache/logging-log4j2/pull/2853/commits/082f3683dce1530c5d735ea413efecbe3e1fa320

vy (approved):

PR #2887, opened 2024-08-23
Add support for concatenated plugin caches

ppkarwasz:
`Log4j2Plugins.dat` descriptors cause problems while shading an application that uses Log4j Core, since they need a specialized resource transformer to be merged.

We maintain such a resource transformer for the Maven Shade Plugin and Gradle Up Shadow also contains a similar tool. Both of them have open issues.
The remaining shading build tools, such as SBT Assembly, however, don't have any resource transformer capable of dealing with the `Log4j2Plugins.dat` format.

This PR extends the `Log4j2Plugins.dat` format to allow it to contain `\n`-separated lists of serialized `PluginCache` objects.
This will allow users to merge plugin caches using more standard tools.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginCacheTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginCacheTest.java
@@ -16,29 +16,23 @@
  */
 package org.apache.logging.log4j.core.config.plugins.processor;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertEquals;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.List;
 import java.util.Map;
-import org.junit.Test;
-import org.junit.runner.RunWith;
-import org.junit.runners.JUnit4;
+import java.util.TreeMap;
+import org.junit.jupiter.api.Test;
 
-@RunWith(JUnit4.class)
-public class PluginCacheTest {
+class PluginCacheTest {
 
     @Test
-    public void testOutputIsReproducibleWhenInputOrderingChanges() throws IOException {
-        final PluginCache cacheA = new PluginCache();
-        createCategory(cacheA, "one", Arrays.asList("bravo", "alpha", "charlie"));
-        createCategory(cacheA, "two", Arrays.asList("alpha", "charlie", "bravo"));
-        assertEquals(cacheA.getAllCategories().size(), 2);
-        assertEquals(cacheA.getAllCategories().get("one").size(), 3);
-        assertEquals(cacheA.getAllCategories().get("two").size(), 3);
+    void testOutputIsReproducibleWhenInputOrderingChanges() throws IOException {
+        final PluginCache cacheA = createSampleCache();
         final PluginCache cacheB = new PluginCache();
         createCategory(cacheB, "two", Arrays.asList("bravo", "alpha", "charlie"));
         createCategory(cacheB, "one", Arrays.asList("alpha", "charlie", "bravo"));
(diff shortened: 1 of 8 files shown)


vy:
Together with this PR, we will have 3 plugin descriptor formats:

1. The Log4j 2 `Log4j2Plugins.dat` format
2. The Newline-separated Log4j 2 `Log4j2Plugins.dat` format (introduced by this PR)
3. Log4j 3 `ServiceLoader`s

Can we instead backport `ServiceLoader`s to `2.x` and be done with just two?

I understand the temptation for this easy fix and I am aware that backporting `ServiceLoader`s from `3.x` to `2.x` is non-trivial. That said, when it comes to the motivation for this new plugin descriptor file format:

- I don't see a pressing need (users are doing okay with the present tooling)
- Shading is a bad practice anyway

ppkarwasz:
> That said, when it comes to the motivation for this new plugin descriptor file format:
> 
> * I don't see a pressing need (users are doing okay with the present tooling)
> * Shading is a bad practice anyway

There is no pressing need for this PR, I only created it as a response to [this `log4j-user@logging` thread](https://lists.apache.org/thread/fk9zd3gh7068d7brq82x48r7jvf1nvqy), when I realized that not all build tools have a resource handler for `Log4j2Plugins.dat`.

I find shading a bad practice too:

* **libraries** should not be shaded, since this provides a software supply chain risk. The current SBOM tools do not support shading (see CycloneDX/cyclonedx-maven-plugin#472), so user might be unaware that they have a vulnerable copy of a library on the classpath. MvnRepository doesn't recognize shaded libraries either (see [`pax-logging-log4j2`](https://mvnrepository.com/artifact/org.ops4j.pax.logging/pax-logging-log4j2), which contains a copy of `log4j-core`) and does not show CVE warnings.
* **applications** should not be shaded, since there are better alternatives to create executable JARs, like the Spring Boot plugins.

jvz:
While I like the `ServiceLoader` approach best (and don't think it's _that_ hard to back port; the main challenges there revolve around updating the annotation processor to generate the service classes as the `main` version has diverged considerably from `2.x`. However, that only solves the problem for plugins compiled against the version of Log4j that introduces support for it, and 3.0 seemed like a solid point for that.

I agree with Volkan here that having three formats is not ideal, and even two formats is annoying enough (I had hoped to eliminate support for the .dat format in 3.0, but that's the only thing that allows for 2.x plugins to still be loadable due to annotation changes).

Could we have some sort of compromise where the .dat parser can work in a lax mode or something where it tries to continue parsing a concatenated .dat file?

ppkarwasz:
> Could we have some sort of compromise where the .dat parser can work in a lax mode or something where it tries to continue parsing a concatenated .dat file?

That is exactly what this PR does: the `.dat` parser continues reading the file until there are no bytes left.

jvz (approved):
I think this is a reasonable approach even though it reminds me of how we bundled everything in Core due to developers' ignorance of dependency management.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -281,3 +281,48 @@ See `@PluginElement("EventTemplateAdditionalField")` usage in {project-github-ur
 link:../javadoc/log4j-core/org/apache/logging/log4j/core/config/plugins/util/PluginUtil.html[`PluginUtil`],
 which is a convenient wrapper around <<#plugin-discovery,`PluginManager`>>, to discover and load plugins.
 See {project-github-url}/log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TemplateResolverFactories.java[`TemplateResolverFactories.java`] for example usages.
+
+[#plugin-merge]
+== Merging plugin descriptors
+
+The plugin descriptor is located in a JAR file at the following fixed location:
+----
```
vy, line 289:
```suggestion

[source]
----
```

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -281,3 +281,48 @@ See `@PluginElement("EventTemplateAdditionalField")` usage in {project-github-ur
...
+
+https://maven.apache.org/plugins/maven-shade-plugin/[Maven Shade Plugin]:::
+You can use the
+https://logging.staged.apache.org/log4j/transform/log4j-transform-maven-shade-plugin-extensions.html#log4j-plugin-cache-transformer[Log4j Plugin Descriptor Transformer].
+
+https://gradleup.com/shadow/[Gradle Shadow Plugin]:::
+You can use the
+https://github.com/GradleUp/shadow/blob/main/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/transformers/Log4j2PluginsCacheFileTransformer.groovy[`Log4j2PluginsCacheFileTransformer`].
+
+* Since version `2.24.0` you can any resource transformer capable of concatenating files using `\n` as separator.
```
vy, line 309:
```suggestion
* Since version `2.24.0`, you can employ any resource transformer capable of concatenating files using `\n` as separator.
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginCacheTest.java
```diff
@@ -48,9 +42,51 @@ public void testOutputIsReproducibleWhenInputOrderingChanges() throws IOExceptio
...
+                first = false;
+            }
+        }
+
+        final Map<String, Map<String, PluginEntry>> actual = new TreeMap<>();
+        PluginCache.loadInputStream(new ByteArrayInputStream(output.toByteArray()), actual);
+
+        assertThat(actual)
+                .as("Deserialized plugin cache")
+                .isEqualTo(createSampleCache().getAllCategories());
```
vy, line 74: _Nit:_ I'd prefer `createSampleCache()` accept categories and entries, instead of assuming hardcoded values will be identical in multiple places.

vy (commented):
Even though I find it helpful, I am not in favor of this ~hack~ relaxed-mode for following reasons:

1. PR is developed in response to a user trying to implement a JTL pretty-printer in a shaded application built using SBT. This is wrong in two major ways:
    1. She is doing wrong by trying to pretty-print JTL (already [explained this to her](https://lists.apache.org/thread/zz26313jbbz495qnokkwqjkwpos2llp7))
    1. She is doing wrong by using shading
1. PR further complicates the officially accepted Log4j plugin descriptor file format
1. We haven't received a single legitimate complaint from SBT users. I don't see a pressing need. This is important and a sensitive point, in particular, given Log4j is known for its bloat. We already massively revamped the docs. Let's first collect some user feedback before settling down on a solution.

These said, PR is technically sound. If there is a consensus from other maintainers to get this in, I won't stop in the way.

jvz:
I agree with Volkan still. No strong feelings from me on whether we should support this.

ppkarwasz:
> I agree with Volkan still. No strong feelings from me on whether we should support this.

Although I am the author of the PR, I also agree with Volkan. So we have consensus!

PR #2902, opened 2024-08-30
Move Flume Appender to `logging-log4j-flume` repo

ppkarwasz:
The code of the Flume Appender has been moved to a new [`logging-log4j-flume`](https://github.com/apache/logging-log4j-flume).

This PR remove the `log4j-flume-ng` module from the repository and cleans up the dependency management section.


vy (approved):
I presume you did not update `src/site` due to the missing `logging-log4j-flume` website, right?

ppkarwasz:
> I presume you did not update `src/site` due to the missing `logging-log4j-flume` website, right?

I think that the Flume Appender should be documented in the main site.
We can create a stub `/log4j/flume/2.x` site later that only refers the user to the main site.

vy:
> > I presume you did not update `src/site` due to the missing `logging-log4j-flume` website, right?
> 
> I think that the Flume Appender should be documented in the main site. We can create a stub `/log4j/flume/2.x` site later that only refers the user to the main site.

I prefer placing the docs next to the sources. This worked nicely for JMX GUI and Jakarta. Why do you think we need to create an exception for Flume?

ppkarwasz:
> I prefer placing the docs next to the sources. This worked nicely for JMX GUI and Jakarta. Why do you think we need to create an exception for Flume?

There must be **some** integration, so that users can find the appender.

Maybe we can move the Flume Appender documentation to its own website and replace the current description in `appenders.adoc` with a stub. I would prefer to do it, when the `/log4j/flume` website is ready though. What do you think?

vy:
> There must be **some** integration, so that users can find the appender.
>
> Maybe we can move the Flume Appender documentation to its own website and replace the current description in `appenders.adoc` with a stub. I would prefer to do it, when the `/log4j/flume` website is ready though. What do you think?

When `/log4j/flume` is up, we can place pointers to `nav.adoc`, `appenders.adoc`, etc. for Flume-related stuff. But I am not keen on sharing Flume-related technical details in `/log4j/2.x`, IMO, they should be situated in `/log4j/flume` instead.

Issue #2830, opened 2024-08-12
[MS21] Native Compilation Support #1


ppkarwasz:
This is the first milestone devoted to fixing GraalVM support in Log4j.

## Tasks

- [x] Fix GraalVM support in `log4j-api` (#2392),
- [x] Automatically check GraalVM compatibility for each Log4j API implementation (https://github.com/apache/logging-log4j-samples/commit/94c819ecdef430640b8a5367eacfdffca9cd5dc3)
- [x] Release Log4j API 2.24.0
- [x] Create and validate a plan to introduce GraalVM support in Log4j Core and the remaining modules. This is discussed in #2364.

ppkarwasz:
@vy,

I added the documentation task to #2831.

ppkarwasz:
A set of integration tests has been scheduled daily. See [CI run 10638659451](https://github.com/apache/logging-log4j-samples/actions/runs/10638659451) for example.

Note the the `simple` test case will fail until #2908 is merged.

ppkarwasz:
The [CI run 10647359122](https://github.com/apache/logging-log4j-samples/actions/runs/10647359122) gives a green light for a `2.24.0` release.

ppkarwasz:
The [2.24.0 release](https://github.com/apache/logging-log4j2/releases/tag/rel%2F2.24.0) has just shipped with GraalVM support out-of-the box in the API

Regarding GraalVM support in the rest of Log4j, the plan is as follows:

1. Extend `log4j-docgen` to be able to generate GraalVM reflection metadata.
2. Publish the reachability metadata on [oracle/graalvm-reachability-metadata](https://github.com/oracle/graalvm-reachability-metadata) with **all** the plugins.

vy:
> Extend `log4j-docgen` to be able to generate GraalVM reflection metadata.

Why do we use `log4j-docgen` but not `log4j-plugins` instead to generate this metadata?

ppkarwasz:
I guess we can extend `log4j-plugin-processor` to generate code for each configuration attribute, this way **no** reflection metadata is necessary. WDYT?

grobmeier:
Thanks for the hard work

PR #2908, opened 2024-08-30
Improve GraalVM support of `SimpleLoggerContext`

ppkarwasz:
This PR allows to create GraalVM applications that only use this `SimpleLoggerContext`.

For this purpose we create a `SimpleProvider` implementation of `Provider`, which is **not** instantiated by reflection.

Part of #2830

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
@@ -17,11 +17,10 @@
 package org.apache.logging.log4j.simple;
 
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
-import java.io.FileNotFoundException;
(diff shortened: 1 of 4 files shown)


vy (approved):
LGTM. I'm only concerned about the visibility of the `SimpleProvider`. It used to be a `private` member of `ProviderUtil`. Now it is _public_, that is, an API liability. Shall we instead move it to an internal package, e.g., `o.a.l.l.simple.internal`?

ppkarwasz:
> LGTM. I'm only concerned about the visibility of the `SimpleProvider`. It used to be a `private` member of `ProviderUtil`. Now it is _public_, that is, an API liability. Shall we instead move it to an internal package, e.g., `o.a.l.l.simple.internal`?

Good call, fixed in https://github.com/apache/logging-log4j2/pull/2908/commits/d20268cfe6bca3803fa95c686a75639aef7c253b.

PR #2909, opened 2024-08-30
chore: use `computIfAbsent` in `MarkerManager

JarvisCraft:
This is a small change which utlizes a more domain-specific and potentially performant `computeIfAbsent` in `MarkerManager`.

## Checklist

* [x] Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise
* [x] `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `./mvnw spotless:apply` and retry)
* [x] Non-trivial changes contain an entry file in the `src/changelog/.2.x.x` directory
* [x] ~~Tests for the changes are provided~~ There are no functional changes
* [x] [Commits are signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (optional, but highly recommended)

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/MarkerManager.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/MarkerManager.java
@@ -60,12 +60,7 @@ public final class MarkerManager {
      * @throws IllegalArgumentException if the argument is {@code null}
      */
     public static Marker getMarker(final String name) {
-        Marker result = MARKERS.get(name);
-        if (result == null) {
-            MARKERS.putIfAbsent(name, new Log4jMarker(name));
-            result = MARKERS.get(name);
-        }
-        return result;
+        return MARKERS.computeIfAbsent(name, Log4jMarker::new);
     }
 
     /**


ppkarwasz:
@JarvisCraft,

LGTM, thanks.

vy:
@JarvisCraft, thanks so much for the contribution! :bow: Curious: How did you encounter the problem? Do you observe a performance issue? If so, mind sharing some more details, please?

JarvisCraft:
@vy, hi! I was just looking through the sources of `MarkerManager` (interested in its implementation details) and, when looking at this specific method, just saw the common pattern which is almost always replaceable with the `computeIfAbsent`.

While I did need any practical performance improvement from it (our use case for `Markers` is initializing them once into a `static final` field) this generally should be better from microopt. point of view since `Map` implementation may (and commonly will) be able to perform single key lookup instead of multiple ones and, what's more important, `ConcurrentMap` explicitly [guarantees its atomicity][1].

But what's also important IMO is that the code is now more explicit now in what it does.

[1]: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#computeIfAbsent-K-java.util.function.Function-

Issue #1916, opened 2023-10-30
Rationalize the usage of `Connection` in the JDBC appender


ppkarwasz:
The JDBC appender takes a very conservative approach to SQL connection sharing:

1. as most appenders, each `append` call is synchronized,
2. each event (or batch of events) is logged using a different `Connection` object, which is closed at the end of the call,
3. to prevent the high performance hit from creating many `Connection`s, users are advised to use a connection pool like DBCP2.

This approach has many disadvantages:

- it does not profit from the connection pool, since at any time at **most** one connection will be borrowed from the pool,
- on the other hand users that use `DriverManager` get a performance hit, even if 1. ensures that no `Connection` object will be used concurrently on multiple threads.

garydgregory:
We already do sharing through the `log4j-jdbc-dbcp2` module. The default behavior should be as simple as possible IMO. Or, we can bake in the use of Commons DBCP and always offer pooling through that. I don't think we should provide a _third_ behavior. Either you get the stock KISS behavior or you get pooling through Commons DBCP.

To muddle the waters further, some JDBC Driver do their own connection pooling; I was the co-author of such a Driver.

ppkarwasz:
> We already do sharing through the `log4j-jdbc-dbcp2` module. The default behavior should be as simple as possible IMO.

I am OK with using **one** `Connection` per log event or batch, but in that case we should allow **multiple** `Connection`s to be used by multiple threads. Currently all database appenders have the same limitations that `OutputStream` appenders, with synchronization in `AbstractDatabaseManager.write()`:

https://github.com/apache/logging-log4j2/blob/d229eda077c313a8a2060caa42531f6b539b1a09/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManager.java#L297-L303

Currently this limitation is only due to the fact that `Connection` and `PreparedStatement` created in the `connectAndStart()` method are passed to `writeInternal` through **instance** fields of `JdbcDatabaseManager`:

https://github.com/apache/logging-log4j2/blob/d229eda077c313a8a2060caa42531f6b539b1a09/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManager.java#L325-L332

We can easily:

- Use `ThreadLocal` in `2.x`.
- Change the API in `3.x` to pass that context data as method parameters.

PR #2915, opened 2024-09-02
Move JNDI `DataSource` to separate module

### LINKED ISSUE #1914: Split JNDI support out of `log4j-jdbc`
Currently the `log4j-jdbc` appender has an **optional** dependency on `log4j-jndi`.

In order to remove this optional dependency, I propose to:

- create a new module `log4j-jdbc-jndi` and depends on `log4j-jdbc` and `log4j-jndi` and contains the [`DataSourceConnectionSource`](https://github.com/apache/logging-log4j2/blob/main/log4j-jdbc/src/main/java/org/apache/logging/log4j/jdbc/appender/DataSourceConnectionSource.java)  and inlined code from `JndiUtil`,
- remove the dependency on `log4j-jndi` from `log4j-jdbc`.

From a user perspective:

- most users will require just the `log4j-jdbc-dbcp2` module,
- some Java EE/Jakarta EE users will require the `log4j-jdbc-jndi` module,
- the `log4j-jdbc` module could be considered as an internal dependency.

ppkarwasz:
We move the [`DataSource` connection source](https://logging.staged.apache.org/log4j/3.x/manual/appenders/database.html#DataSourceConnectionSource) to a new module `log4j-jdbc-jndi`.

The main rationale for this change is to remove the **optional** dependency on `log4j-jndi` from `log4j-jdbc`.
This change forces users to opt-in for the JNDI `DataSource` by adding `log4j-jdbc-jndi` to their classpath.

After this change, the `log4j-jdbc` artifact contains:

* the `JDBC` Appender,
* a connection source based on `DriverManager`,
* a connection source based on a static factory method.

Closes #1914


src/changelog/.3.x.x/1914_split_JDBC_JNDI_connection_source.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="added">
+  <issue id="1914" link="https://github.com/apache/logging-log4j2/issues/1914"/>
+  <description format="asciidoc">Split off JDBC `DataSource` connection source into a new `log4j-jdbc-jndi` module.</description>
```
vy, line 7: Umm... I think this is not correct. Suggestion:

```suggestion
  <description format="asciidoc">Split off JNDI support of `log4j-jdbc` to a new `log4j-jdbc-jndi` module</description>
```

log4j-jdbc/pom.xml
```diff
@@ -30,89 +30,99 @@
...
-      <scope>test</scope>
+      <artifactId>log4j-plugins</artifactId>
     </dependency>
+
     <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-core-test</artifactId>
       <scope>test</scope>
+      <!--
+        ~ Exclude some unused dependencies of `log4j-core-test`.
```
vy, line 55: Why do we need this?

vy (commented):
Overall, LGTM.

**Question:** Is the DBCP detail of our `log4j-jdbc*` modules well-hidden? If so, I suggest

1. Removing all of its public mentions
2. Switching to HikariCP<sup>1</sup> instead

<sup>1</sup> HikariCP is, both performance- and feature-wise, a clear winner against any other alternative in the market as of today. It is the default JDBC connection pool in both Spring Boot 2 and [Spring Boot 3](https://docs.spring.io/spring-boot/reference/data/sql.html#data.sql.datasource.connection-pool).

Of course I am not suggesting this for this PR. But I feel like this is an okay place to discuss it.

log4j-jdbc/pom.xml
ppkarwasz: Right, we don't. The `log4j-core-test` artifact has a lot of dependencies that are added to the test classpath of each of our components. We probably should remove those that are not necessary.

Fixed in https://github.com/apache/logging-log4j2/pull/2915/commits/c9cc897bd2142eded6dfa475446cbe805064e1a8

vy: Removed from the wrong module.

vy: Fixed in af3b21b067a2546f089716d16a2a57ee0292d517.

vy (approved):

ppkarwasz:
> **Question:** Is the DBCP detail of our `log4j-jdbc*` modules well-hidden? If so, I suggest
> 
>     1. Removing all of its public mentions
> 
>     2. Switching to HikariCP instead

DBCP has a nice feature, that justifies its pool of _aficionados_: you can register a pool with the JDBC driver and retrieve it later using a `jdbc:apache:commons` URI.

Note that until #1916 is fixed, all pools perform the same: only a single connection is used at each time.
The main problem is the synchronization of `AbstractDatabaseManager`. Removing it might technically not be a breaking change, but it is a very **disruptive** one.

PR #2916, opened 2024-09-02
Split and modernize JANSI support.

### LINKED ISSUE #1736: Upgrade Jansi support to version 2.x
## Description

Although our POM file has an optional dependency on Jansi 2.4.0, the `ConsoleAppender` actually works only on Jansi 1.x.

We can modify `ConsoleAppender` to support both major versions of Jansi.

ppkarwasz:
This PR:

- Adds a new `ConsoleStreamProvider` plugin type that can decide how the Console Appender stream is built.
- Moves JAnsi support to a new `log4j-jansi` module.
- <s>Upgrades the JAnsi library to version 2.x</s> Replaces JAnsi with JLine 3, since JAnsi is deprecated (cf. fusesource/jansi#294).
- Rewrites `JAnsiTextRenderer` to use our internal `AnsiEscape` utility.

Fixes #1736.


ppkarwasz:
This is blocked by #2691: if the `TextRenderer` interface is removed from the `%ex` pattern converter, the `JAnsiTextRenderer` can be removed, otherwise it should be kept.

ppkarwasz:
I rewrite `JAnsiTextRenderer` to use our internal `AnsiEscape` ANSI sequence generator, instead of relying on JANSI.
Cf. https://github.com/apache/logging-log4j2/pull/2916/commits/5ae37df740d450ce457a732ecbd76bbd912d43ea

garydgregory (changes requested):
Module should not be named "log4j-jansi" since the PR no longer uses JAnsi. How about "log4j-ansi"?

ppkarwasz:
> Module should not be named "log4j-jansi" since the PR no longer uses JAnsi. How about "log4j-ansi"?

The module uses `org.jline:jansi-core` instead of `org.fusesource.jansi:jansi`, so I believe the name is appropriate.

The `JAnsiTextRenderer` (which is part of `log4j-core`, not `log4j-jansi`) effectively could be renamed `AnsiTextRenderer`. Keeping the same name, however will allow me to backport part of this PR to `2.x`.

garydgregory:
@ppkarwasz 
TY for the explanation: -)
+1

src/site/antora/modules/ROOT/pages/manual/appenders.adoc
```diff
@@ -234,38 +234,38 @@ This setting might be useful in multi-application environments.
...
+* Implement ANSI escape handling on old Windows consoles.
+* Strips ANSI escape codes if process output is being redirected and not attached to a terminal
+--
 +
-[tabs]
+See https://github.com/fusesource/jansi?tab=readme-ov-file#features[Features of JAnsi] for more details.
++
+[TIP]
+====
+Support for ANSI escape sequences was added to Windows 10 console around 2017.
```
vy, line 255: This translates to Log4j 3 will cater for **Windows 10 users that haven't updated their systems since 2017**. Instead of this, shouldn't we remove JAnsi support completely and direct such users, if they indeed do exist, to Log4j 2 instead?

/cc @garydgregory @rgoers

vy (commented):
See my remark about removing the JAnsi support, completely.

> This is blocked by #2691: if the `TextRenderer` interface is removed from the `%ex` pattern converter, the `JAnsiTextRenderer` can be removed, otherwise it should be kept.

#2691 cannot _remove_ anything, since it is against `2.x`. But be my guest to completely remove it while porting it to `main`. Hence, I see two paths we can take:

1. Wait for #2691 to get merged to `2.x`, port it to `main`, and then remove `TextRenderer` in `main`.
2. Remove `TextRenderer` in `main` without waiting for #2691.

[@alan0428a, I am CC'ing you, in case you're interested in.]

src/site/antora/modules/ROOT/pages/manual/appenders.adoc
garydgregory: Does this mean that my current Windows 10 OS is supported? It is "current" since it receives regular updates and shows version "10.0.19045.4780" when I run "ver".

ppkarwasz: Let me expand on that. The situation on a modern Windows 10+ is the following:

- All terminal windows are capable of interpreting ANSI escapes.
- The `cmd.exe` terminal by default has ANSI escape support switched off. You need to create the `HKEY_CURRENT_USER\Console\VirtualTerminalLevel` registry key and set it to `0x1` (as DWORD).
- Other applications have the feature switched on by default (I believe PowerShell and certainly the standalone Windows Terminal app).

Note also that if `VirtualTerminalLevel` is not enabled, a developer can enable it for his application by calling [`AnsiConsole.systemInstall()`](https://www.javadoc.io/doc/org.jline/jline/latest/org/jline/jansi/AnsiConsole.html).

Therefore:

- For a developer, `log4j-jansi` is useless. He can just call `AnsiConsole.systemInstall()` in his code.
- For a user, `log4j-jansi` can be used as an alternative to setting `VirtualTerminalLevel` to `0x1`.

All in all, I think we can drop `log4j-jansi` in `3.x`.

ppkarwasz:
> See my remark about removing the JAnsi support, completely.
> 
> > This is blocked by #2691: if the `TextRenderer` interface is removed from the `%ex` pattern converter, the `JAnsiTextRenderer` can be removed, otherwise it should be kept.
> 
> #2691 cannot _remove_ anything, since it is against `2.x`.

Of course you can not remove it, but as far as I can tell, you are bypassing it. Since `ThrowableProxy` and related classes are not used any more, `TextRenderer` isn't used either. Am I right?

ppkarwasz:
Note that in this PR we are discussing 3 related, but distinct features:

### Enabling ANSI escapes on Windows terminals

This feature is probably useless and can be dropped. It requires the JAnsi library.

### Detecting if the console is a terminal

The JAnsi library is very good at it. It can detect **separately** if `stdout` and `stderr` are TTY devices on UNIX or character devices on Windows. If the stream is not attached to a TTY, the ANSI escape sequences will be dropped.

In Log4j Core 2 we emulate this behavior by checking if `System.console()` exists (cf. [noConsoleNoAnsi](https://logging.staged.apache.org/log4j/2.x/manual/pattern-layout.html#plugin-attr-noConsoleNoAnsi). Iit exists if **both** `stdin` and `stdout` are connected to a terminal). Java 22 adds a new [`Console.isTerminal()`](https://docs.oracle.com/en/java/javase/22/docs/api/java.base/java/io/Console.html#isTerminal()) method, but I am unsure how it works.

### Coloring of stack traces and messages

In Log4j Core 2, JAnsi was also used to color stack traces and messages. This is enabled by adding `ansi` as an option to `%ex`, `%xEx`, `%rEx` or `%m`. The JAnsi library is not needed for `%style` or `%highlight`.

This PR rewrites `JAnsiTextRenderer` to work **without** JAnsi.

ppkarwasz:
@garydgregory,

How do you feel about removing JAnsi (but not ANSI) support as a test? For this PR it would mean:

- I remove the `log4j-jansi` module,
- I keep the modifications to `JAnsiTextRenderer` and I port them to `2.x`.

vy (approved):

ppkarwasz:
@vy,

Should I port the changes to `JAnsiTextRenderer` to `2.x`? It was used to support the `ansi` option in the `%m`, `%ex`, `%xEx` and `%rEx` pattern converters. Now only the `%m{ansi}` pattern converter requires it.

vy:
> Should I port the changes to `JAnsiTextRenderer` to `2.x`? It was used to support the `ansi` option in the `%m`, `%ex`, `%xEx` and `%rEx` pattern converters. Now only the `%m{ansi}` pattern converter requires it.

@ppkarwasz, yes, please do so.

Curious: Does that translate to no JAnsi dependency on `2.x`?

ppkarwasz:
> Curious: Does that translate to no JAnsi dependency on `2.x`?

Are you suggesting we drop an end-of-life dependency (Jansi 1.x) released by a project that does not exist any more (Jansi, which was replaced by JLine) and break backward compatibility for Windows platforms older than 2017? :stuck_out_tongue: 

Windows 7 and previous releases have a 3.5% market share, so I think we could do it.

vy:
> Are you suggesting we drop an end-of-life dependency (Jansi 1.x) released by a project that does not exist any more (Jansi, which was replaced by JLine) and break backward compatibility for Windows platforms older than 2017? 😛
> 
> Windows 7 and previous releases have a 3.5% market share, so I think we could do it.

@ppkarwasz, yes, let's please indeed drop the JAnsi dependency.

ppkarwasz:
Done in #3070

PR #2921, opened 2024-09-05
Move compression into new `log4j-compress` module

ppkarwasz:
We move the code that depends on Commons Compress
into a new `log4j-compress` module.

At the same time, we perform the following behavioral changes:

- Compression is performed by the `CompressActionFactoryProvider`, with the maximum order. In the current setting, if `log4j-compress` is on the classpath, it is used for everything, even GZ and ZIP.
- The `compressionLevel` parameter, specific to the GZ and ZIP compression factories in the JRE, is replaced by a more generic `compressionOptions` map. This will allow adding more compression options in the future if users request it.
- The available list of compression algorithms is no longer hardcoded, but depends entirely on the list of compressors supported by Apache Commons Compress.

> [!NOTE]
> Currently Commons Compress does not offer an algorithm-independent
> interface to supply additional compression options. However, each algorithm does provide such a way.
> If there are requests from users to support these options, we can extend the `RolloverStrategy` implementations.
> Until that happens I am reluctant to add additional features.

## Review guide

The main change revolves around the replacement of `FileExtension` with 2 new interfaces:

* `CompressActionFactory`: this class inherits the **instance** methods from `FileExtension` and allows to create lambdas that compress files.
* `CompressActionFactoryProvider`: a new Log4j plugin type that can be used to register new compression algorithms. In practice there are two implementations (one for the algos in the JRE and one for Apache Commons Compress). All the remaining compression algorithms should extend Commons Compress.


log4j-core/pom.xml
```diff
@@ -44,7 +44,6 @@
       <!-- Annotations only -->
       org.jspecify.*;resolution:=optional,
       <!-- External optional dependencies -->
-      org.apache.commons.compress.*;resolution:=optional,
```
vy, line 47: 1. Shouldn't we have `org.jspecify;transitive=false` in `bnd-extra-module-options`?
2. Why didn't we have `org.apache.commons.compress.*;transitive=false` in `bnd-extra-module-options`?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/CompressActionFactory.java
```diff
@@ -0,0 +1,84 @@
...
+@NullMarked
+public interface CompressActionFactory {
+
+    /**
+     * Option to specify the compression level.
+     * <p>
+     *     The value must be an {@code int} between -1 (default) and 9.
+     * </p>
+     */
+    String COMPRESSION_LEVEL = "compressionLevel";
```
vy, line 38: IMHO,

1. This is not the correct place for this configuration option name.
2. We shouldn't pass configurations in a `Map<String, ?>`, but in a well-typed object/record instead. Can't we adapt the code in that direction instead?

log4j-compress/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.apache.logging.log4j</groupId>
+    <artifactId>log4j</artifactId>
+    <version>${revision}</version>
+    <relativePath>../log4j-parent</relativePath>
+  </parent>
+
+  <artifactId>log4j-compress</artifactId>
```
vy, line 27: *Thinking out loudly:* Shouldn't we rather rename this to `log4j-compress-commons`, in case we want to support another 3rd party library in the future? :thinking:

src/site/antora/modules/ROOT/pages/components.adoc
```diff
@@ -53,6 +53,23 @@ See xref:manual/async.adoc[asynchronous loggers] for more details.
...
+|===
+| JPMS module
+| `org.apache.logging.log4j.compress`
+|===
+
+The `log4j-compress` artifact provides additional compression algorithms to the
+xref:manual/appenders/rolling-file.adoc[].
+
+See
+xref:manual/appenders/rolling-file.adoc#RolloverStrategy-compress-commons[Compressing archived files] for more details.
```
vy, line 69: Let's (try to?) emphasize that `log4j-compress` is not only for rolling file appenders.
```suggestion
The `log4j-compress` artifact provides additional compression algorithms.
Having this artifact in the classpath makes these automatically accessible to
xref:manual/appenders/rolling-file.adoc[].
See
xref:manual/appenders/rolling-file.adoc#RolloverStrategy-compress-commons[Compressing archived files] for more details.
```

src/site/antora/modules/ROOT/pages/manual/appenders/rolling-file.adoc
```diff
@@ -926,80 +926,56 @@ When the current log file is archived, the rolling file appender can compress it
...
++
+[NOTE]
+====
+Apache Commons Compress currently supports the following values of `algorithmName`:
 
-| [[RolloverStrategy-compress-xz]]`.xz` <<commons-compress-dep,^dep^>>
-| {x-mark}
-| https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/compressors/xz/package-summary.html[XZ] algorithm
+* `gzip`, `bzip2`, `deflate`, `snappy-framed`, `lz4-block`, `lz4-framed` are supported without **any** external dependency.
+* `lzma`, `.xz` and `.zst` extensions require **additional** Commons Compress dependencies.
```
vy, line 958:
```suggestion
* `lzma`, `.xz` and `.zst` extensions require **additional** dependencies.
```

src/site/antora/modules/ROOT/partials/features/compression.adoc
vy, line 1: Are we gonna port this to `2.x`? If not, I guess we should remove this one.

log4j-compress/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+            <configuration>
+              <additionalClasspathDependencies>
+                <dependency>
+                  <groupId>org.tukaani</groupId>
+                  <artifactId>xz</artifactId>
+                  <version>${xz.version}</version>
+                </dependency>
+              </additionalClasspathDependencies>
+              <includes>
+                <include>org/apache/logging/log4j/compress/commons/xz/*Test.class</include>
```
vy, line 156: Shouldn't we have LZMA here also?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/AbstractCompressAction.java
```diff
@@ -0,0 +1,93 @@
...
+            LOGGER.debug("Starting {} compression from {} to {}.", getAlgorithmName(), source, destination);
+            try (final OutputStream output = wrapOutputStream(Files.newOutputStream(destination))) {
+                Files.copy(source, output);
+            }
+            LOGGER.debug("Finished {} compression from {} to {}.", getAlgorithmName(), source, destination);
+            try {
+                Files.delete(source);
+                LOGGER.debug("File {} deleted successfully.", source);
+            } catch (final IOException ioe) {
+                LOGGER.warn("Unable to delete {}.", source);
```
vy, line 72:
```suggestion
                LOGGER.warn("Unable to delete {}.", source, ioe);
```

log4j-core/pom.xml
ppkarwasz: 1. Maybe, if it appears as `transitive`, I'll fix it when merging #2930
2. BND adds `transitive` only if a dependency type appears in the signature of a public method or extends/implements clauses of a public class/interface. This is not the case of Commons Compress.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/CompressActionFactory.java
ppkarwasz: Ad. 1
This option is hardcoded in the configuration of each `RolloverStrategy`, which currently provides a compression level as only way to configure configuration. This seems as a good intersection point between `RolloverStrategy` (which uses `CompressActionFactory`) and two `CompressActionFactory` implementations, which actually use the parameter. What better place would you suggest?

Ad. 2
Would you prefer `Map<String, String>`? There are multiple approaches to provide a generic way to pass parameters for an _a priori_ unknown `CompressActionFactory`:

- a `Map<String, String>`, which does not generate any `ClassCastException`s, but might generate parsing errors if a parameter can not be parsed (e.g. the value of `compressionLevel` can not be parsed as string).
- a `Map<String, ?>, which does not suffer from parsing errors, but causes `ClassCastException`s. :cry: 
- a marker interface, so that each algorithm can provide its own well-typed object/record. This will cause `IllegalArgumentException`s if the params for factory A are passed to factory B.
- parameterize `CompressActionFactory` with the type of options it uses. Everything works nicely in the code, but we are mostly using reflection.

I don't have a preferred strategy, but the `Map` approaches have the advantage that we can provide the user with a way to fill in a `Map`, pass it to Commons Compress and let it worry about interpreting them.
With this approach, if Commons Compress adds additional options to [GzipParameters](https://javadoc.io/doc/org.apache.commons/commons-compress/latest/org/apache/commons/compress/compressors/gzip/GzipParameters.html), users could immediately profit from them (assuming we add code to convert `Map` to `GzipParameters`.

log4j-compress/pom.xml
ppkarwasz: Maybe it would be prudent, but I would prefer 3rd party compression libraries to integrate with Commons Compress instead of us. I did add a custom `CompressorStreamProvider` to prove that 3rd party extensions to Commons Compress can be used by this module.

While writing this PR I also looked if we can integrate [CLP](https://github.com/y-scope/clp) through this mechanism. The answer is no: CLP would require us to compress each log event separately, it is not useful to compress already written log files (at least without parsing them again).

log4j-compress/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+        <groupId>org.apache.commons</groupId>
+        <artifactId>commons-compress</artifactId>
+        <version>${commons-compress.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>com.google.jimfs</groupId>
+        <artifactId>jimfs</artifactId>
+        <version>${jimfs.version}</version>
+      </dependency>
```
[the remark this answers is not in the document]
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/2921/commits/8c8c55bb264e2459465bb9674c9ec1d856bc5503

log4j-compress/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+        <groupId>org.tukaani</groupId>
+        <artifactId>xz</artifactId>
+        <version>${xz.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>com.github.luben</groupId>
+        <artifactId>zstd-jni</artifactId>
+        <version>${zstd.version}</version>
+      </dependency>
```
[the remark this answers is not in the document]
ppkarwasz: Dependabot must be able to upgrade these dependencies. I added comments in https://github.com/apache/logging-log4j2/pull/2921/commits/8c8c55bb264e2459465bb9674c9ec1d856bc5503.

src/site/antora/modules/ROOT/partials/features/compression.adoc
ppkarwasz: I wouldn't bother making 2.x compression algorithms extensible.

However I have already aligned the 2.x documentation to use a `features/compression.adoc` file in 8069c73c971f67dabcf01a6564ba0f91511be0f5.

log4j-compress/pom.xml
ppkarwasz: There is no `lzma` package.

I removed the misleading `lzma` exclusion in https://github.com/apache/logging-log4j2/pull/2921/commits/6627024993f48f1289184ec3ef04729068955b0b

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/CompressActionFactory.java
vy: I see. Given compression is activated based on the extension of the archived file name, can we instead

1. Remove/Deprecate the hardcoded `compressionLevel` attribute, and
2. Create a new `Compression` element holding arbitrary key-value pairs?

Example snippet:

```xml
<RollingFile name="FILE"
             fileName="/var/log/app.log"
             filePattern="/var/log/app.log.%i.gz"> 
  <JsonTemplateLayout/>
  <DefaultRolloverStrategy max="15"/> 
  <Policies>
    <CronTriggeringPolicy schedule="0 0 0 * * ?"/> 
    <SizeBasedTriggeringPolicy size="100k"/> 
  </Policies>
  <Compression>
    <level>10</level>
    <algorithm>gzip</algorithm>
    <bufferSize>8192</bufferSize>
  </Compression>
</RollingFile>
```

vy (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/CompressActionFactory.java
ppkarwasz: I created #2950 test for user support for this idea.

In the meantime I have implemented the removal of `compressionLevel` in https://github.com/apache/logging-log4j2/pull/2921/commits/806a0be16e43dff72b5f0645d704b594d30a8bbb

PR #2932, opened 2024-09-09
Reload key & trust stores on reconfiguration

vy:
Ports #2767 from `2.x` to `main`.


log4j-core/src/main/java/org/apache/logging/log4j/core/net/ssl/AbstractKeyStoreConfiguration.java
```diff
@@ -47,21 +48,20 @@ public AbstractKeyStoreConfiguration(
     @Override
     protected KeyStore load() throws StoreConfigurationException {
         final String loadLocation = this.getLocation();
+        final char[] password = this.getPassword();
```
ppkarwasz, line 51: I think you can make the `load` method `private`, so there is no way to reload the keystores.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/ssl/SslConfiguration.java
```diff
@@ -33,30 +29,43 @@
...
+
     private final boolean verifyHostName;
 
+    @Nullable
+    private final KeyStoreConfiguration keyStoreConfig;
+
+    @Nullable
+    private final TrustStoreConfiguration trustStoreConfig;
+
+    private final transient SSLContext sslContext;
```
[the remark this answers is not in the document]
vy: Actually I used the `transient` there to communicate that the value is derived. But if it is causing readers to raise eyebrows, I should indeed take it away. Fixed in 4c8d08bf706b6c9e1dedd5a5b7efd3990e474f4f.

log4j-core/src/main/java/org/apache/logging/log4j/core/net/ssl/AbstractKeyStoreConfiguration.java
vy: Fixed in 4c8d08bf706b6c9e1dedd5a5b7efd3990e474f4f.

PR #2936, opened 2024-09-09
Implement `MessageFactory`-namespaced logger registry

### LINKED ISSUE #2933: 2.24.0 status logger warns about message factory mismatch
## Description

After updating to 2.24.0 from 2.23.1, the status logger logs the following warning:

```
2024-09-09T12:14:29.525247Z main WARN The Logger eu.lampinen.log4j2test.Test was created with the message factory org.apache.logging.log4j.message.ReusableMessageFactory@3d680b5a and is now requested with a null message factory (defaults to org.apache.logging.log4j.message.ParameterizedMessageFactory), which may create log events with unexpected formatting.
```

The whole sample code is below

```java

public class Test {

    static {
        System.setProperty("log4j2.statusLoggerLevel", "WARN");
    }

        LogManager.getLogger(Test.class).error("Hello {}", "world!");
        LogManager.getLogger(Test.class).error("Hello {}", "world!");
    }

}
```

I think this is caused by a mismatch between
* `AbstractLogger.checkMessageFactory`, which uses constant `DEFAULT_MESSAGE_FACTORY_CLASS`
* `Logger.getEffectiveMessageFactory`, which uses either `ReusableMessageFactory` or `ParameterizedMessageFactory` depending on `Constants.ENABLE_THREADLOCALS`.

Setting `log4j2.enable.threadlocals=false` will disable the warning.

## Configuration

**Version:** 2.24.0

**JDK:** Azul 21

## Logs

2.23.1
```
ERROR - Hello world!
ERROR - Hello world!
```

2.24.0
```
ERROR - Hello world!
2024-09-09T12:16:29.792995Z main WARN The Logger eu.lampinen.log4j2test.Test was created with the message factory org.apache.logging.log4j.message.ReusableMessageFactory@49d904ec and is now requested with a null message factory (defaults to org.apache.logging.log4j.message.ParameterizedMessageFactory), which may create log events with unexpected formatting.
ERROR - Hello world!
```

## Reproduction

See above

vy:
Fixes #2933 by deprecating `AbstractLogger.checkMessageFactory()` and removing its usages.

### Rationale

In its current state, all `LoggerContext#getLogger(String, MessageFactory)` methods of all `LoggerContext` implementations delegate to `LoggerRegistry`, which in essence is a `Map<MessageFactory, Map<String, Logger>>`. That is, all `LoggerRegistry`-produced `Logger`s are `MessageFactory`-namespaced. This renders the `checkMessageFactory()` check redundant.

Put another way, `getLogger("foo", someMF)` will create a `Logger` associated with `someMF`. `getLogger("foo")` will create a `Logger` associated with the `MF` configured for the `LC`. In either case, if `someMF != the MF configured for the LC`, `Logger`s will differ anyway, hence `checkMessageFactory()` will always succeed.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
@@ -100,7 +100,6 @@ public class SimpleLoggerContext implements LoggerContext {
         // Note: This is the only method where we add entries to the 'loggerRegistry' ivar.
         final ExtendedLogger extendedLogger = loggerRegistry.getLogger(name, messageFactory);
         if (extendedLogger != null) {
-            AbstractLogger.checkMessageFactory(extendedLogger, messageFactory);
             return extendedLogger;
         }
         final SimpleLogger simpleLogger = new SimpleLogger(
--- a/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
@@ -20,11 +20,13 @@ import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.PrintStream;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.message.MessageFactory;
+import org.apache.logging.log4j.message.ParameterizedMessageFactory;
 import org.apache.logging.log4j.simple.internal.SimpleProvider;
 import org.apache.logging.log4j.spi.ExtendedLogger;
 import org.apache.logging.log4j.spi.LoggerContext;
 import org.apache.logging.log4j.spi.LoggerRegistry;
 import org.apache.logging.log4j.util.PropertiesUtil;
+import org.jspecify.annotations.Nullable;
 
 /**
  * A simple {@link LoggerContext} implementation.
@@ -39,6 +41,8 @@ public class SimpleLoggerContext implements LoggerContext {
 
     /** All system properties used by <code>SimpleLog</code> start with this */
     protected static final String SYSTEM_PREFIX = "org.apache.logging.log4j.simplelog.";
+
+    private static final MessageFactory DEFAULT_MESSAGE_FACTORY = ParameterizedMessageFactory.INSTANCE;
 
     private final PropertiesUtil props;
 
@@ -95,13 +99,14 @@ public class SimpleLoggerContext implements LoggerContext {
     }
 
     @Override
-    public ExtendedLogger getLogger(final String name, final MessageFactory messageFactory) {
-        // Note: This is the only method where we add entries to the 'loggerRegistry' ivar.
-        final ExtendedLogger extendedLogger = loggerRegistry.getLogger(name, messageFactory);
-        if (extendedLogger != null) {
-            return extendedLogger;
-        }
-        final SimpleLogger simpleLogger = new SimpleLogger(
+    public ExtendedLogger getLogger(final String name, @Nullable final MessageFactory messageFactory) {
+        final MessageFactory effectiveMessageFactory =
+                messageFactory != null ? messageFactory : DEFAULT_MESSAGE_FACTORY;
+        return loggerRegistry.computeIfAbsent(name, effectiveMessageFactory, this::createLogger);
+    }
+
+    private ExtendedLogger createLogger(final String name, @Nullable final MessageFactory messageFactory) {
+        return new SimpleLogger(
                 name,
                 defaultLevel,
                 showLogName,
@@ -112,8 +117,6 @@ public class SimpleLoggerContext implements LoggerContext {
                 messageFactory,
                 props,
                 stream);
-        loggerRegistry.putIfAbsent(name, messageFactory, simpleLogger);
-        return loggerRegistry.getLogger(name, messageFactory);
     }
 
     /**
@@ -129,16 +132,18 @@ public class SimpleLoggerContext implements LoggerContext {
 
     @Override
     public boolean hasLogger(final String name) {
-        return false;
+        return loggerRegistry.hasLogger(name, DEFAULT_MESSAGE_FACTORY);
     }
 
     @Override
     public boolean hasLogger(final String name, final Class<? extends MessageFactory> messageFactoryClass) {
-        return false;
+        return loggerRegistry.hasLogger(name, messageFactoryClass);
     }
 
     @Override
-    public boolean hasLogger(final String name, final MessageFactory messageFactory) {
-        return false;
+    public boolean hasLogger(final String name, @Nullable final MessageFactory messageFactory) {
+        final MessageFactory effectiveMessageFactory =
+                messageFactory != null ? messageFactory : DEFAULT_MESSAGE_FACTORY;
+        return loggerRegistry.hasLogger(name, effectiveMessageFactory);
     }
 }
--- a/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
@@ -21,7 +21,6 @@ import java.io.PrintStream;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.message.MessageFactory;
 import org.apache.logging.log4j.simple.internal.SimpleProvider;
-import org.apache.logging.log4j.spi.AbstractLogger;
 import org.apache.logging.log4j.spi.ExtendedLogger;
 import org.apache.logging.log4j.spi.LoggerContext;
 import org.apache.logging.log4j.spi.LoggerRegistry;
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -54,7 +54,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      * Data structure contract for the internal storage of admitted loggers.
      *
      * @param <T> subtype of {@code ExtendedLogger}
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      */
     @Deprecated
     public interface MapFactory<T extends ExtendedLogger> {
@@ -70,7 +70,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      * {@link MapFactory} implementation using {@link ConcurrentHashMap}.
      *
      * @param <T> subtype of {@code ExtendedLogger}
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      */
     @Deprecated
     public static class ConcurrentMapFactory<T extends ExtendedLogger> implements MapFactory<T> {
@@ -95,7 +95,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      * {@link MapFactory} implementation using {@link WeakHashMap}.
      *
      * @param <T> subtype of {@code ExtendedLogger}
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      */
     @Deprecated
     public static class WeakMapFactory<T extends ExtendedLogger> implements MapFactory<T> {
@@ -122,7 +122,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      * Constructs an instance <b>ignoring</b> the given the map factory.
      *
      * @param mapFactory a map factory
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      */
     @Deprecated
     public LoggerRegistry(@Nullable final MapFactory<T> mapFactory) {
@@ -138,7 +138,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      *
      * @param name a logger name
      * @return the logger associated with the name
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      * Use {@link #getLogger(String, MessageFactory)} instead.
      */
     @Deprecated
@@ -208,7 +208,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      *
      * @param name a logger name
      * @return {@code true}, if the logger exists; {@code false} otherwise.
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      * Use {@link #hasLogger(String, MessageFactory)} instead.
      */
     @Deprecated
@@ -263,7 +263,7 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      * @param name a logger name
      * @param messageFactory ignored – kept for backward compatibility
      * @param logger a logger instance
-     * @deprecated As of version {@code 2.24.1}, planned to be removed!
+     * @deprecated As of version {@code 2.25.0}, planned to be removed!
      * Use {@link #computeIfAbsent(String, MessageFactory, BiFunction)} instead.
      */
     @Deprecated
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -16,28 +16,47 @@
  */
 package org.apache.logging.log4j.spi;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.Objects;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.BiFunction;
 import org.apache.logging.log4j.message.MessageFactory;
+import org.apache.logging.log4j.status.StatusLogger;
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
 
 /**
- * Convenience class to be used by {@code LoggerContext} implementations.
+ * Convenience class to be used as an {@link ExtendedLogger} registry by {@code LoggerContext} implementations.
  */
+@NullMarked
 public class LoggerRegistry<T extends ExtendedLogger> {
-    private static final String DEFAULT_FACTORY_KEY = AbstractLogger.DEFAULT_MESSAGE_FACTORY_CLASS.getName();
-    private final MapFactory<T> factory;
-    private final Map<String, Map<String, T>> map;
+
+    private final Map<String, Map<MessageFactory, T>> loggerByMessageFactoryByName = new HashMap<>();
+
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+    private final Lock readLock = lock.readLock();
+
+    private final Lock writeLock = lock.writeLock();
 
     /**
-     * Interface to control the data structure used by the registry to store the Loggers.
+     * Data structure contract for the internal storage of admitted loggers.
+     *
      * @param <T> subtype of {@code ExtendedLogger}
+     * @deprecated As of version {@code 2.24.1}, planned to be removed!
      */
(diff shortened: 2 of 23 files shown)


vy:
@rgoers, would you mind reviewing this, please? (Checked with @ppkarwasz, he also agrees with deprecating `checkMF()`.)

log4j-api/src/main/java/org/apache/logging/log4j/simple/SimpleLoggerContext.java
```diff
@@ -100,7 +99,6 @@ public ExtendedLogger getLogger(final String name, final MessageFactory messageF
         // Note: This is the only method where we add entries to the 'loggerRegistry' ivar.
         final ExtendedLogger extendedLogger = loggerRegistry.getLogger(name, messageFactory);
         if (extendedLogger != null) {
-            AbstractLogger.checkMessageFactory(extendedLogger, messageFactory);
```
ppkarwasz, line 103: After the changes in [LOG4J2-1180](https://issues.apache.org/jira/browse/LOG4J2-1180) and [LOG4J2-1318](https://issues.apache.org/jira/browse/LOG4J2-1318) the `checkMessageFactory` method is mostly useless, but there is a corner case.

If I have a **stateful** message factory `MyMessageFactory` (e.g. it creates map messages with a predefined set of keys), then we still need a warning if we call:

```java
Logger logger1 = LogManager.getLogger("foo", new MyMessageFactory("bar", "baz"));
Logger logger2 = LogManager.getLogger("foo", new MyMessageFactory("bar", "car"));
```

The `LoggerRegistry` only uses the **class name** of `MyMessageFactory` as key, so we'll end up with `logger2 == logger1`, even if the user wanted two different loggers.

vy: Right. Implemented `MessageFactory`-namespaced logger registry in be6c229307fb907b1c70fbb421c3a8683907452e. Mind reviewing it, please?

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -87,97 +114,224 @@ public void putIfAbsent(final Map<String, T> innerMap, final String name, final
...
+            final Map<MessageFactory, T> loggerByMessageFactory = loggerByMessageFactoryByName.get(name);
+            if (loggerByMessageFactory == null) {
+                return null;
+            }
+            if (messageFactory != null) {
+                return loggerByMessageFactory.get(messageFactory);
+            }
+            return !loggerByMessageFactory.isEmpty()
+                    ? loggerByMessageFactory.values().iterator().next()
+                    : null;
```
ppkarwasz, line 173: I think the problem here is how to deal with a `null` message factory parameter. IMHO opinion we should use:

```
MessageFactory actualMessageFactory = messageFactory != null 
        ? messageFactory 
        : ParameterizedMessageFactory.INSTANCE;
```

This is due to the following reasoning:

- Before `2.23.0` the default message factory was determined by `log4j-api`,
- In `2.23.0` the code (in `AbstractLogger`) was removed, since it caused a circular initialization problem.
- Since `2.24.0` implementations decide, which message factory they want. If they do nothing, they get `ParameterizedMessageFactory`.

If `actualMessageFactory` is not null, we can use `Optional` to get rid of the `null` checks.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -87,97 +114,224 @@ public void putIfAbsent(final Map<String, T> innerMap, final String name, final
...
+            loggerByMessageFactory.putIfAbsent(loggerMessageFactory, logger);
+        } finally {
+            writeLock.unlock();
+        }
+    }
+
+    public T computeIfAbsent(
+            final String name,
+            final MessageFactory messageFactory,
+            final BiFunction<String, MessageFactory, T> loggerSupplier) {
```
ppkarwasz, line 285: I get the idea: you don't want multiple threads to create the same logger at the same time and **then** register it with the map.

Since all loggers with the same name and message factory are functionally equivalent, I don't think we really need `computeIfAbsent`, `putIfAbsent` is good enough here.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -91,6 +93,7 @@ public class LoggerContext extends AbstractLifeCycle
     private String contextName;
     private volatile URI configLocation;
     private Cancellable shutdownCallback;
+    private MessageFactory defaultMessageFactory = ParameterizedMessageFactory.INSTANCE;
```
ppkarwasz, line 96: The default message factory here should depend on `log4j2.messageFactory` and `log4j2.enableThreadlocals`.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -812,6 +812,22 @@ private void initApiModule() {
                 .init(); // Or make public and call ThreadContext.init() which calls ThreadContextMapFactory.init().
     }
 
+    private Logger newInstance(final String name, final MessageFactory messageFactory) {
+        final Logger logger = newInstance(this, name, messageFactory);
+        final MessageFactory loggerMessageFactory = logger.getMessageFactory();
+        if (!loggerMessageFactory.equals(messageFactory)) {
+            StatusLogger.getLogger()
+                    .error(
+                            "Newly created logger with name `{}` and message factory `{}` was actually requested to be created with a different message factory: `{}`.\n"
```
ppkarwasz, line 821: Why should these differ?

ppkarwasz (changes requested):
Nice work Volkan! 

This might be great for `2.25.0`, but I am not sure the problem requires such a drastical changes. By this I mean the deprecations in `LoggerRegistry`.

Note also that all message factories should implement `equals` and `hashCode` (I think this is technically a minor version change).

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: Fixed in e54dd2d976c36aac0f1d9e7b5dc557073fe813b9.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -62,15 +85,19 @@ public Map<String, Map<String, T>> createOuterMap() {
...
+     *
      * @param <T> subtype of {@code ExtendedLogger}
+     * @deprecated As of version {@code 2.24.1}, planned to be removed!
      */
+    @Deprecated
     public static class WeakMapFactory<T extends ExtendedLogger> implements MapFactory<T> {
+
         @Override
         public Map<String, T> createInnerMap() {
             return new WeakHashMap<>();
```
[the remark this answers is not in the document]
vy: Doh! Good catch! Fixed in e54dd2d976c36aac0f1d9e7b5dc557073fe813b9. Mind checking, please?

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -46,10 +65,14 @@ public interface MapFactory<T extends ExtendedLogger> {
     }
 
     /**
-     * Generates ConcurrentHashMaps for use by the registry to store the Loggers.
+     * {@link MapFactory} implementation using {@link ConcurrentHashMap}.
+     *
      * @param <T> subtype of {@code ExtendedLogger}
+     * @deprecated As of version {@code 2.24.1}, planned to be removed!
      */
+    @Deprecated
```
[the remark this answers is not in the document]
vy: Switched all version bumps, deprecations, etc. to version `2.25.0`. Let's merge this to `2.x` and I will cherry-pick this without deprecations to `2.24.x`.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: No, that is not the point – even though that is indeed a positive side effect. The reason I introduced `computeIfAbsent()` is:

1. It doesn't allow null `MF` and returns the effective `Logger`. These two cannot be implemented in a backward compatible manner using `putIfAbsent()`.
2. All existing `LC` implementations were doing: `var l = getLogger(); if (l == null) { putIfAbsent(newLogger); } return getLogger();` There were two redundant `getLogger()` calls – since `putIfAbsent()` is not returning the effective logger. `computeIfAbsent()` is not only more performant, but also results in less boilerplate. This is visible in the carried out `LC` changes too.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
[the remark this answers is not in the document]
ppkarwasz: Sorry, what I meant: you rightfully deprecated this code, since it didn't make any sense.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
vy: They shouldn't, but API allows this corner case. I wouldn't be surprised to see a `LC#newInstance()` implementation ignoring the passed `MF`.

vy:
> This might be great for `2.25.0`, but I am not sure the problem requires such a drastical changes. By this I mean the deprecations in `LoggerRegistry`.

I will merge these changes to `2.x` (targeting `2.25.0`) and then cherry-pick these to `2.24.x` without deprecation changes.

> Note also that all message factories should implement `equals` and `hashCode` (I think this is technically a minor version change).

AFAICT, `LocalizedMessageFactory` is the only stateful `MF`; implemented `equals()` and `hashCode()` for it in c747f4fe1edc7a0e83259462ce103afc39cbbee5.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -87,97 +116,233 @@ public void putIfAbsent(final Map<String, T> innerMap, final String name, final
...
+        try {
+            final Map<MessageFactory, WeakReference<T>> loggerRefByMessageFactory =
+                    loggerRefByMessageFactoryByName.get(name);
+            if (loggerRefByMessageFactory == null) {
+                return null;
+            }
+            final MessageFactory effectiveMessageFactory =
+                    messageFactory != null ? messageFactory : ParameterizedMessageFactory.INSTANCE;
+            final WeakReference<T> loggerRef = loggerRefByMessageFactory.get(effectiveMessageFactory);
+            return loggerRef.get();
```
ppkarwasz, line 174: Lack a `null` check (or a `computeIfAbsent(effectiveMessageFactory, ignored -> new WeakRefence<>(null))`).

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -87,97 +116,233 @@ public void putIfAbsent(final Map<String, T> innerMap, final String name, final
...
         return getLoggers(new ArrayList<T>());
     }
 
     public Collection<T> getLoggers(final Collection<T> destination) {
-        for (final Map<String, T> inner : map.values()) {
-            destination.addAll(inner.values());
+        requireNonNull(destination, "destination");
+        readLock.lock();
+        try {
+            loggerRefByMessageFactoryByName.values().forEach(loggerRefByMessageFactory -> loggerRefByMessageFactory
```
ppkarwasz, line 188: _Nit_: `stream().flatMap()`?

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
ppkarwasz: I think I would prefer to call `Logger.getEffectiveMessageFactory(null)` (which is in the same package).

Instantiating a logger just to see what factory will it use, seems like a big workaround.

vy: Unless you strongly object to it, I want to stick with the _big workaround_ approach.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: Good catch! Fixed.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: Fixed.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -92,6 +94,15 @@ public class LoggerContext extends AbstractLifeCycle
...
     private Cancellable shutdownCallback;
 
+    /**
+     * The default message factory to use while creating loggers.
+     * <p>
+     * The initial value is only set to avoid nullability.
+     * The actual value is populated by {@link #reloadDefaultMessageFactory()} during (re)configuration.
+     * </p>
+     */
+    private MessageFactory defaultMessageFactory = ParameterizedMessageFactory.INSTANCE;
```
ppkarwasz, line 104: Sorry, but this still doesn't make sense to me.

The default message factory is something that is fixed for at least each `LoggerContext` and can only be configured through configuration properties. It doesn't make sense to change default message factory each time a configuration file is loaded.

If we would implement such a feature, users would end up with half the loggers using the old MF and half the loggers using the new MF.

vy: How shall I fix it? I don't want to use `Logger.getEffectiveMessageFactory(null)`, since it will fix the value to a particular logger implementation, though the logger is subject to change via `newInstance()`. (Converting a `private static` method to package-local is already a clear sign of abstraction leakage...)

> The default message factory is something that is fixed for at least each `LoggerContext` and can only be configured through configuration properties. It doesn't make sense to change default message factory each time a configuration file is loaded.

A custom `LoggerContext` can change the `Logger` (and hence the `MF`) at will. I don't think _"[an `MF`] can only be configured through configuration properties"_ is correct. This is indeed what our `LC` implementations do, but that doesn't mean `LC` impls. out there in the wild follow the same rule. My point is: `LC` allows at-runtime change of `Logger` impl. (and hence, `MF` impl.), and we should cover this case.

> If we would implement such a feature, users would end up with half the loggers using the old MF and half the loggers using the new MF.

That's right. Is it bad? If users re-configure an `LC` with a new `MF`, shouldn't we be using this new `MF` as the default one?

vy: Anyway... I've put the default `MF` behind a `Lazy`: a8007767db6f80278a2045b39bba7faa4094ce4a. Let me know what you think.

ppkarwasz: > How shall I fix it? I don't want to use `Logger.getEffectiveMessageFactory(null)`, since it will fix the value to a particular logger implementation, though the logger is subject to change via `newInstance()`. (Converting a `private static` method to package-local is already a clear sign of abstraction leakage...)

IMHO, `LoggerContext#newInstance(LoggerContext, String, MessageFactory)` should not be called with `null` parameters, that's all. Then we have no doubt which MF is used by the `Logger`.

> > The default message factory is something that is fixed for at least each `LoggerContext` and can only be configured through configuration properties. It doesn't make sense to change default message factory each time a configuration file is loaded.
> 
> A custom `LoggerContext` can change the `Logger` (and hence the `MF`) at will. I don't think _"[an `MF`] can only be configured through configuration properties"_ is correct. This is indeed what our `LC` implementations do, but that doesn't mean `LC` impls. out there in the wild follow the same rule. My point is: `LC` allows at-runtime change of `Logger` impl. (and hence, `MF` impl.), and we should cover this case.

Let me be more precise: our `LoggerContext` implementations use Log4j configuration properties to determine the MF. Another `LC` implementation can do whatever it wants, as long as it doesn't pass `null` to the `LoggerRegistry`.

vy: > IMHO, `LoggerContext#newInstance(LoggerContext, String, MessageFactory)` should not be called with `null` parameters, that's all. Then we have no doubt which MF is used by the `Logger`.

Are you implying throwing an exception or something when the `MF` of `getLogger(String,MF)` is null? How shall we prevent this from happening?

> > > The default message factory is something that is fixed for at least each `LoggerContext` and can only be configured through configuration properties. It doesn't make sense to change default message factory each time a configuration file is loaded.
> > 
> > 
> > A custom `LoggerContext` can change the `Logger` (and hence the `MF`) at will. I don't think _"[an `MF`] can only be configured through configuration properties"_ is correct. This is indeed what our `LC` implementations do, but that doesn't mean `LC` impls. out there in the wild follow the same rule. My point is: `LC` allows at-runtime change of `Logger` impl. (and hence, `MF` impl.), and we should cover this case.
> 
> Let me be more precise: our `LoggerContext` implementations use Log4j configuration properties to determine the MF. Another `LC` implementation can do whatever it wants, as long as it doesn't pass `null` to the `LoggerRegistry`.

I don't think this would work for 3rd party classes extending from `LC`. You give them the impression that they have control over their choice of `Logger`, (and hence, the `MF`), but under the hood, you just ignore it.

ppkarwasz: > > IMHO, `LoggerContext#newInstance(LoggerContext, String, MessageFactory)` should not be called with `null` parameters, that's all. Then we have no doubt which MF is used by the `Logger`.
> 
> Are you implying throwing an exception or something when the `MF` of `getLogger(String,MF)` is null? How shall we prevent this from happening?

Let me remark, that nothing in the Javadoc allows users to call `getLogger(String, MF)` with a `null` second parameter.
The Javadoc didn't explicitly allow calling the `coreLogger` constructor with a `null` `MF` parameter either, until 2.23.0.
Of course we should protect against such calls by implicitly using the message factory computed using configuration properties.

> I don't think this would work for 3rd party classes extending from `LC`. You give them the impression that they have control over their choice of `Logger`, (and hence, the `MF`), but under the hood, you just ignore it.

`LC` implementations can override the `getLogger(String)` method to provide a different MF.
I really don't see a practical use case, where someone would want to do that.

vy: Have you seen the `Lazy` impl. committed? WDYT? If you don't like it, what is your suggestion besides using `Logger.getEffectiveMessageFactory(null)`?

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -92,6 +94,14 @@ public class LoggerContext extends AbstractLifeCycle
     private volatile URI configLocation;
     private Cancellable shutdownCallback;
 
+    /**
+     * The default message factory to use while creating loggers.
+     */
+    private final Lazy<MessageFactory> defaultMessageFactoryRef = Lazy.lazy(() -> {
+        final Logger throwawayLogger = newInstance(this, "throwaway-for-determining-MF", null);
+        return throwawayLogger.getMessageFactory();
+    });
```
ppkarwasz, line 103: Using this would mean that `newInstance` is called **once** with a `null` message factory and the message factory of that bogus `Logger` is used in every other `newInstance` invocation. 
So instead of allowing the user to determine the default message factory directly, we allow him to do it indirectly by placing some special logic in `newInstance`.

If you really believe that users want to change the default message factory programmatically, let's add a protected `getMessageFactory(String)` method.

Personally I don't think this is required. A default message factory must work exactly like `ParameterizedMessageFactory`, only the implementation details can change, but `getFormattedMessage()` of the messages it produces must be identical to `PMF`.

vy: Fixed in c1dff86c087c9e4ae5ee7e68ad16385d02dc894e as you wanted. (For the record, I don't agree with this direction, but there are more pressing issues than discussing this detail.)

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
vy: Fixed in c1dff86c087c9e4ae5ee7e68ad16385d02dc894e as you wanted. (For the record, I don't agree with this direction, but there are more pressing issues than discussing this detail.)

log4j-taglib/pom.xml
```diff
@@ -28,14 +28,19 @@
...
       ~ OSGi and JPMS options
       -->
+    <bnd-extra-package-options>
+      <!-- Annotations only -->
+      org.jspecify.*;resolution:=optional
+    </bnd-extra-package-options>
     <bnd-extra-module-options>
       <!-- Filebased module names: MUST be static -->
       javax.servlet.api;substitute="javax.servlet-api";static=true;transitive=false,
       javax.servlet.jsp.api;substitute="javax.servlet.jsp-api";static=true;transitive=false
```
ppkarwasz, line 41:
```suggestion
      javax.servlet.jsp.api;substitute="javax.servlet.jsp-api";static=true;transitive=false,
```

Issue #2813, opened 2024-08-08
Rewrite Jetty-based unit tests


ppkarwasz:
For our HTTP tests we have a mixture of Jetty and [WireMock](https://wiremock.org/) tests.

We should probably rewrite the Jetty tests to use the lighter WireMock.

vy:
Agreed with the proposal. I have an inkling that several of such tests can also be converted to a unit test that doesn't need any network infrastructure setup.

PR #2941, opened 2024-09-10
Fix reloading of the configuration from HTTP(S)

### LINKED ISSUE #2937: Log4j fails to honor `lastModifiedTime`
## Description

When using a monitorInterval setting on a remote/http log4j.xml the reconfigure-thread does not update the lastModifiedTime of the configuration source and reconfigures the loggers every time.

## Configuration

**Version:** 2.23 and 2.24

**JDK:** AdoptOpenJDK 21

## Logs

[log.txt](https://github.com/user-attachments/files/16942744/log.txt)

## Reproduction

I don't really have a completely isolated test. The problem does _not_ occur when using a file-based log-config, so you need something like a webserver.

My test-setup looks like this:

### Main
```java
  System.setProperty("log4j2.debug", "true");
  System.setProperty("log4j2.Configuration.allowedProtocols","http,https,file,jar");
		
    Configurator.reconfigure(URI.create("http://sfs00053/configs/t7/client/testlog.xml"));

    new Thread(() -> {
      while (true) {
      try {
        LogManager.getLogger().error("hello");
        TimeUnit.SECONDS.sleep(5);
      } catch (InterruptedException e) {
        throw new RuntimeException(e);
      }
    }
  }).start();
}
```
### Log-Config
```xml
<?xml version="1.0" encoding="UTF-8" ?>
<Configuration monitorInterval="10">
  <Appenders>

      <PatternLayout pattern="%d | %-5p | %c | %m%n" />
    </Console>
  </Appenders>
  <Loggers>

    <Root level="DEBUG">
      <AppenderRef ref="STDOUT" />
    </Root>
  </Loggers>
</Configuration>
```

### Steps to reproduce
1. Change the log4j.xml-location in the main to your own location
2. Start the main. log4j2 should configure itself with the given log-config and log every 5 seconds
3. Change the file on the webserver so the lastModified of the log-config is newer than the one log4j2 knows (e.g. by `touch testlog.xml`)
4. log4j2 will now reconfigure _every 10 seconds_ and not only once because it doesn't honor the new lastModified

See also the attached log. I changed the file on the webserver after the first watchmonitor run. After that it reconfigured itself every time.

I hope I filled out this issue correctly. If something's missing please let me know :)

ppkarwasz:
The `HttpWatcher` didn't propagate the observed last modification time back to the configuration.
As a result, each new `Configuration` had the same last modification date as the first one and the reconfiguration process looped.

Closes #2937


log4j-core/src/main/java/org/apache/logging/log4j/core/config/HttpWatcher.java
```diff
@@ -103,34 +102,47 @@ private boolean refreshConfiguration() {
         try {
             final LastModifiedSource source = new LastModifiedSource(url.toURI(), lastModifiedMillis);
             final HttpInputStreamUtil.Result result = HttpInputStreamUtil.getInputStream(source, authorizationProvider);
+            // Update lastModifiedMillis
+            // https://github.com/apache/logging-log4j2/issues/2937
+            lastModifiedMillis = source.getLastModified();
```
ppkarwasz, line 107: One bug was here: `lastModifiedMillis` didn't propagate from `source` the the `HttpWatcher`...

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -280,16 +274,23 @@ public InputStream getInputStream() {
...
-            return new ConfigurationSource(new FileInputStream(getFile()), getFile());
-        } else if (isURL() && data != null) {
+            return new ConfigurationSource(source, data, modifiedMillis);
+        }
+        File file = getFile();
+        if (file != null) {
+            return new ConfigurationSource(Files.newInputStream(file.toPath()), getFile());
+        }
+        URL url = getURL();
+        if (url != null && data != null) {
```
ppkarwasz, line 284: ...the second bug was here: the last modification date propagated correctly from `HttpWatcher` to the `modifiedMillis` mutable field of `ConfigurationSource`. However the code used the immutable `lastModified` field instead.

ppkarwasz (commented):
In general I took the liberty of bumping some status logger messages to `INFO`.
IMHO the `INFO` level should be used for important events like:

- The start and stop of configurations,
- The start and stop of logger contexts,
- Whenever `Watcher#isModified` returns true, which triggers a reconfiguration.

ppkarwasz:
Accidentally I had to rewrite `HttpThreadContextMapTest`. I believe it exploited the fact that its JSON configuration was reloaded every monitor interval. I'll double check that.

Since `HttpThreadContextMapTest` used Jetty, I rewrote also `UrlConfigurationFactoryTest` and removed Jetty from the dependencies of `log4j-core-test` (fixes #2813).

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/MutableThreadContextMapFilter.java
```diff
@@ -364,23 +367,29 @@ private class FileMonitor implements Runnable {
...
+                case NOT_FOUND:
+                case EMPTY:
                     for (FilterConfigUpdateListener listener : listeners) {
                         listener.onEvent();
                     }
-                }
-            } else if (result.status == Status.EMPTY) {
-                LOGGER.debug("Filter configuration is empty");
-                filter = new NoOpFilter();
             }
```
ppkarwasz, line 393: Here we have a small modification of behavior: an `EMPTY` result before didn't call the listeners.

However I don't consider this a minor change, since the listeners are basically only useful in tests.

ppkarwasz:
@vy, could you take a look at this?

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/MutableThreadContextMapFilter.java
vy: If so,

1. Why don't we call listeners for all `result.status` values?
2. Can we deprecate the listeners feature?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/WatchManager.java
```diff
@@ -112,15 +111,11 @@ public void run() {
...
-                        logger.info(
-                                "Source '{}' was modified on {} ({}), previous modification was on {} ({})",
-                                source,
-                                millisToString(lastModified),
-                                lastModified,
-                                millisToString(monitor.lastModifiedMillis),
-                                monitor.lastModifiedMillis);
-                    }
+                    logger.info(
+                            "Configuration source at {} was modified on {}, previous modification was on {}",
```
vy, line 115: There are several places in this PR where logging message is changed/improved. In those places, I'd really appreciate text parameters to be wrapped with backticks. For instance,

```suggestion
                            "Configuration source at `{}` was modified on `{}`, previous modification was on `{}`",
```

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/MutableThreadContextMapFilter.java
ppkarwasz: 1. In practice the listener is used to stop the unit test at certain moment. :cry: If we were to run the listener when the config file was not modified, it would complexify the listener: the listener does not have a `Result` parameter.
2. I would say yes, we can do it in `2.25.0`.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -128,9 +128,7 @@ public LoggerContext(final String name, final Object externalContext) {
      */
     public LoggerContext(final String name, final Object externalContext, final URI configLocn) {
         this.contextName = name;
-        if (externalContext == null) {
-            externalMap.remove(EXTERNAL_CONTEXT_KEY);
-        } else {
+        if (externalContext != null) {
```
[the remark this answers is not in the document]
ppkarwasz: The change was suggested by the IDE and actually it does not modify behavior since:

- this is a constructor,
- the map is initially empty.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -80,7 +84,7 @@ public ConfigurationSource(final InputStream stream, final File file) {
         } catch (Exception ex) {
             // There is a problem with the file. It will be handled somewhere else.
         }
-        this.lastModified = modified;
+        this.modifiedMillis = this.lastModified = modified;
```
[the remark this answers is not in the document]
ppkarwasz: I fixed this in https://github.com/apache/logging-log4j2/pull/2941/commits/35973f73ac4f9527a8a46b4b9a20246a6912eaa5.

ppkarwasz:
Regarding the `Source` and `ConfigurationSource` change I added nullability annotations to both.

`ConfigurationSource` is basically a triple (`Source`, `byte[]` and a `long` timestamp). The `source` and `data` fields can be `null`, but not at the same time. All the `ConfigurationSource#getUri()` and similar methods are delegated to `Source#getUri()`, **unless** the `source` field is `null`.

If we look at the `Source` class, the `uri` and `location` fields **can not** be `null`, since each constructor throws something if this happens. The only constructor that does not throw if `uri` or `location` are `null` is the one that takes `ConfigurationSource` as parameter. I would propose to deprecate that constructor in the next minor release, since it introduces an artificial cyclical dependency between `Source` and `ConfigurationSource`.

I have modified the 2 call sites that use the `Source(ConfigurationSource)` constructor to use more appropriate ones.

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/MutableThreadContextMapFilter.java
ppkarwasz: I added a comment explaining why we call the listeners for these statuses and not the other two.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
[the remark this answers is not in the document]
ppkarwasz: Can you double-check that I didn't modify the behavior of `ConfigurationSource`?

ppkarwasz:
While merging #2963, which was based on this PR, I accidentally merged part of these modifications.

It is fixed now, but a force-push was necessary, sorry.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -57,8 +63,8 @@ public class ConfigurationSource {
             new ConfigurationSource(Constants.EMPTY_BYTE_ARRAY, null, 0);
 
     private final InputStream stream;
-    private volatile byte[] data;
-    private volatile Source source;
+    private volatile byte /*@Nullable*/[] data;
+    private final /*@Nullable*/ Source source;
     private final long lastModified;
     // Set when the configuration has been updated so reset can use it for the next lastModified timestamp.
     private volatile long modifiedMillis;
```
vy, line 70: Could you document `lastModified` and `modifiedMillis` class fields, please? As a matter of fact, I find the names very confusing too. I would really appreciate an improvement and uniformity there too. For instance, does `lastModified` also contain millis? If so, why isn't it prefixed with `Millis`? etc. In short, please rename these two fields too.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -150,23 +141,23 @@ public ConfigurationSource(final InputStream stream) throws IOException {
...
+    }
+
+    /**
+     * @throws NullPointerException if {@code stream} is {@code null}.
+     */
+    private ConfigurationSource(
+            byte /*@Nullable*/[] data, /*@Nullable*/ Source source, InputStream stream, long lastModified) {
+        this.stream = requireNonNull(stream, "stream is null");
+        this.data = data;
+        this.source = source;
```
vy, line 159: You state the following:

> `ConfigurationSource` is basically a triple (`Source`, `byte[]` and a `long` timestamp). The `source` and `data` fields can be `null`, but not at the same time.

If so,

1. Would you mind checking here – i.e., at the canonical ctor – both `data` and `source` cannot be null?
2. Can't we do without a `stream` field? Can't we eagerly read the `stream` the first time it is (directly or indirectly via a `Source`) provided? (The only problem is `#getInputStream()`. There we can return `new ByteArrayInputStream(data)`.)

vy (commented):
> Can you double-check that I didn't modify the behavior of `ConfigurationSource?`

New canonical ctor and its usage LGTM, thanks!

I've dropped some more remarks.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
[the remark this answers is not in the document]
rgoers: I would recommend that you defer the change to the ConfigurationSource constructor and do that in a PR all by itself. It has too many variations of usage to be able to be checked with another PR IMO.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
ppkarwasz: Done in https://github.com/apache/logging-log4j2/pull/2941/commits/3598b6a863dfd6028e72c137936e490ef695a791

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
rgoers: I would have to check the code but as I recall the difference in what happens during initial configuration vs reconfiguration is why there are differences in behavior. Before trying to "simplify" you would really have to create a matrix of all the places the ConfigurationSource gets used and determine what data is available and whether the proposed simplifications are safe.

ppkarwasz: 1. Fixed in https://github.com/apache/logging-log4j2/pull/2941/commits/3598b6a863dfd6028e72c137936e490ef695a791
2. We could, but this would require a big modification of the behavior: I/O exceptions would appear in the constructors.

vy: > We could, but this would require a big modification of the behavior: I/O exceptions would appear in the constructors.

@ppkarwasz, we can wrap them into `UncheckedIOException`s?

ppkarwasz: I agree with @rgoers on this one: I merged the PR and have rolled back the changes to the `ConfigurationSource` file in 755497d3715efe80c543525a0fec3e7171c88dab.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -71,16 +78,7 @@ public class ConfigurationSource {
...
-        this.data = null;
-        this.source = new Source(file);
-        long modified = 0;
-        try {
-            modified = file.lastModified();
-        } catch (Exception ex) {
-            // There is a problem with the file. It will be handled somewhere else.
-        }
-        this.lastModified = modified;
+        this(null, new Source(file), stream, getLastModified(file.toPath()));
```
rgoers, line 81: If file is null this will result in an NPE in the previous version the NPE would be caught. That said, I don't recall if a null is ever passed here.

PR #2952, opened 2024-09-15
Enable `docker` profile in CI for `log4j-layout-template-json-test`

vy:


ppkarwasz:
I am wondering about a couple of things:

- Should we activate the `docker` profile by default on machines that have `/var/run/docker.sock`? Or maybe we should ask the reviewers of a release to activate `docker` in their Maven settings if they have Docker installed?
- The Docker image versions in these tests are somehow dated, but unfortunately Elastic Search does not provide a `latest` tag. Should we add [org.elastisearch:elasticsearch](https://mvnrepository.com/artifact/org.elasticsearch/elasticsearch/8.15.1) to the dependency management of this module, so Dependabot can update the version regularly? My guess is that the Docker images are published shortly after the `elasticsearch` artifact is out.

vy:
> Should we activate the `docker` profile by default on machines that have `/var/run/docker.sock`?

I think we should do neither. In particular, bootstrapping an ELK stack is a pretty resource intensive process, I doubt if it will succeed without an OOM on every developer machine.

> Or maybe we should ask the reviewers of a release to activate `docker` in their Maven settings if they have Docker installed?

I don't think we _should_, but I will update the review kit to reflect on the `docker` Maven profile – reviewers can choose for themselves.

> * The Docker image versions in these tests are somehow dated, but unfortunately Elastic Search does not provide a `latest` tag. Should we add [org.elastisearch:elasticsearch](https://mvnrepository.com/artifact/org.elasticsearch/elasticsearch/8.15.1) to the dependency management of this module, so Dependabot can update the version regularly? My guess is that the Docker images are published shortly after the `elasticsearch` artifact is out.

Addressed this issue.

log4j-parent/pom.xml
```diff
@@ -105,7 +105,14 @@
     <commons-lang3.version>3.17.0</commons-lang3.version>
     <commons-logging.version>1.3.4</commons-logging.version>
     <disruptor.version>4.0.0</disruptor.version>
-    <elasticsearch.version>7.17.24</elasticsearch.version>
+    <disruptor.version>3.4.4</disruptor.version>
```
ppkarwasz, line 108:
```suggestion
```

The correct version is already defined above.

PR #2954, opened 2024-09-16
Use a random `${test:logging.path}`

ppkarwasz:
The `@TempLoggingDir` JUnit 5 extensions create a totally deterministic directory to hold the log files for tests.

Since the directory is deleted only if all the tests succeed, this causes a problem with test reruns.


log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/TempLoggingDirectory.java
```diff
@@ -51,7 +51,8 @@ public void beforeAll(ExtensionContext context) throws Exception {
         Path loggingPath = null;
         for (final Field field : fields) {
             if (loggingPath != null) {
-                StatusLogger.getLogger().warn("Multiple fields with @TempLoggingDir annotation are not supported.");
+                StatusLogger.getLogger()
+                        .warn("Multiple static fields with @TempLoggingDir annotation are not supported.");
```
vy, line 55: If we're salting the generated paths, why not?

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/TempLoggingDirectory.java
```diff
@@ -78,7 +79,8 @@ public void beforeEach(ExtensionContext context) throws Exception {
         final Object instance = context.getRequiredTestInstance();
         for (final Field field : fields) {
             if (loggingPath != null) {
-                StatusLogger.getLogger().warn("Multiple fields with @TempLoggingDir annotation are not supported.");
+                StatusLogger.getLogger()
+                        .warn("Multiple instance fields with @TempLoggingDir annotation are not supported.");
```
vy, line 83: Again, why not?

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/TempLoggingDirectory.java
```diff
@@ -135,6 +132,29 @@ private PathHolder createLoggingPath(final ExtensionContext context, final Clean
...
+                final String baseDir = System.getProperty("basedir");
+                final Path basePath = (baseDir != null ? Paths.get(baseDir, "target") : Paths.get(".")).resolve("logs");
+                final Class<?> clazz = context.getRequiredTestClass();
+                final Package pkg = clazz.getPackage();
+                final String dir =
+                        pkg.getName().replaceAll("[.$]", File.separatorChar == '\\' ? "\\\\" : File.separator);
+                // Create a temporary directory that uses the simple class name as prefix
+                Path packagePath = basePath.resolve(dir);
+                Files.createDirectories(packagePath);
+                return Files.createTempDirectory(packagePath, clazz.getSimpleName());
```
vy, line 150: Can we suffix the path with something that will make directories lexicographically sorted by their creation time, please? For instance, `timestamp + counter + salt`, etc.

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/TempLoggingDirectory.java
ppkarwasz: Same as above: do we need to define multiple logging paths per test?
Which one would be used to resolve to `${test:logging.path}`?

log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/TempLoggingDirectory.java
ppkarwasz: I added a counter in https://github.com/apache/logging-log4j2/pull/2954/commits/463c5e8f0231f136685d5fb56c3cb418820b3157

vy: `Files.createTempDirectory()` already takes care of the distinctiveness – I just want to be able to spot the last effective folder. I don't see how adding just a counter helps, since counters get reset at reruns. Am I missing something?

ppkarwasz: Aren't those reruns happening inside the same JVM?

**Note**: the directories are deleted by default if the test is successful, so we are only looking at failed runs. I wouldn't want to have too long names.

PR #2961, opened 2024-09-16
Implement `MessageFactory`-namespaced logger registry

vy:
Ports #2936 fixing #2933 in `2.x` to `2.24.x`.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -255,7 +255,8 @@ public class LoggerRegistry<T extends ExtendedLogger> {
         writeLock.lock();
         try {
             final Map<MessageFactory, WeakReference<T>> loggerRefByMessageFactory =
-                    loggerRefByMessageFactoryByName.computeIfAbsent(logger.getName(), this::createLoggerRefByMessageFactoryMap);
+                    loggerRefByMessageFactoryByName.computeIfAbsent(
+                            logger.getName(), this::createLoggerRefByMessageFactoryMap);
             final MessageFactory loggerMessageFactory = logger.getMessageFactory();
             final WeakReference<T> loggerRef = loggerRefByMessageFactory.get(loggerMessageFactory);
             if (loggerRef == null || loggerRef.get() == null) {
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -242,21 +242,20 @@ public class LoggerRegistry<T extends ExtendedLogger> {
     /**
      * Registers the provided logger using the given name – <b>message factory parameter is ignored</b> and the one from the logger will be used instead.
      *
-     * @param name a logger name
+     * @param name ignored – kept for backward compatibility
      * @param messageFactory ignored – kept for backward compatibility
      * @param logger a logger instance
      */
     public void putIfAbsent(final String name, final MessageFactory messageFactory, final T logger) {
 
         // Check arguments
-        requireNonNull(name, "name");
         requireNonNull(logger, "logger");
 
         // Insert the logger
         writeLock.lock();
         try {
             final Map<MessageFactory, WeakReference<T>> loggerRefByMessageFactory =
-                    loggerRefByMessageFactoryByName.computeIfAbsent(name, this::createLoggerRefByMessageFactoryMap);
+                    loggerRefByMessageFactoryByName.computeIfAbsent(logger.getName(), this::createLoggerRefByMessageFactoryMap);
             final MessageFactory loggerMessageFactory = logger.getMessageFactory();
             final WeakReference<T> loggerRef = loggerRefByMessageFactory.get(loggerMessageFactory);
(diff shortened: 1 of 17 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -87,97 +106,168 @@ public void putIfAbsent(final Map<String, T> innerMap, final String name, final
...
+        }
     }
 
+    /**
+     * Registers the provided logger using the given name – <b>message factory parameter is ignored</b> and the one from the logger will be used instead.
+     *
+     * @param name a logger name
+     * @param messageFactory ignored – kept for backward compatibility
+     * @param logger a logger instance
+     */
```
ppkarwasz, line 248: I don't understand the different way `name` and `messageFactory` are treated: both can be derived from `logger`, but only `messageFactory` is.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -515,25 +522,23 @@ public Collection<Logger> getLoggers() {
...
-     * @param name The name of the Logger to return.
-     * @param messageFactory The message factory is used only when creating a logger, subsequent use does not change the
-     *            logger but will log a warning if mismatched.
-     * @return The Logger.
+     * @param name a logger name
+     * @param messageFactory a message factory to associate the logger with
+     * @return a logger matching the given name and message factory
      */
     @Override
     public Logger getLogger(final String name, final MessageFactory messageFactory) {
```
ppkarwasz, line 532: _Nit_ regarding `getLogger(String)`: that method could use:

```java
return getLogger(name, DEFAULT_MESSAGE_FACTORY);
```

instead of

```java
return getLogger(name, null);
```

If we enable nullability warnings, this change will remove one.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -515,25 +522,23 @@ public Collection<Logger> getLoggers() {
...
-        // Note: This is the only method where we add entries to the 'loggerRegistry' ivar.
-        Logger logger = loggerRegistry.getLogger(name, messageFactory);
-        if (logger != null) {
-            AbstractLogger.checkMessageFactory(logger, messageFactory);
-            return logger;
+        final MessageFactory effectiveMessageFactory =
+                messageFactory != null ? messageFactory : DEFAULT_MESSAGE_FACTORY;
+        final Logger oldLogger = loggerRegistry.getLogger(name, effectiveMessageFactory);
+        if (oldLogger != null) {
+            return oldLogger;
```
ppkarwasz, line 537: We should check if `oldLogger.getMessageFactory()` is equal to `effectiveMessageFactory` here...

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -812,6 +819,22 @@ private void initApiModule() {
...
+        final MessageFactory loggerMessageFactory = logger.getMessageFactory();
+        if (!loggerMessageFactory.equals(messageFactory)) {
+            StatusLogger.getLogger()
+                    .error(
+                            "Newly created logger with name `{}` and message factory `{}` was actually requested to be created with a different message factory: `{}`.\n"
+                                    + "This generally hints a problem.\n"
+                                    + "Please report this using the Log4j project issue tracker.",
+                            name,
+                            loggerMessageFactory,
+                            messageFactory);
```
ppkarwasz, line 833: ... instead of here, where it is basically impossible (unless the `newInstance` method is broken).

ppkarwasz (changes requested):
LGTM, except some minor remarks.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
ppkarwasz: Scratch that, I probably need a coffee.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
ppkarwasz: Can we put this test in a test suite for `LoggerContext` implementations?

If `newInstance` is called with a given `name` and `messageFactory`, we need to assert that the returned logger has that name and message factory.

vy: Fixed in 8a89e8c3903d4a5836459f19c6500938fd89b219. Mind checking it again, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
vy: Fixed in 8a89e8c3903d4a5836459f19c6500938fd89b219.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: Fixed in 8a89e8c3903d4a5836459f19c6500938fd89b219.

ppkarwasz (approved):

PR #2953, opened 2024-09-16
Enable `docker` profile in CI for `log4j-layout-template-json-test`

vy:
See #2952 for the `main` port.

PR #2930, opened 2024-09-08
Remove `transitive` modifier from optional dependencies

ppkarwasz:
This PR:

- Add a Groovy script that checks the generated `module-info.class` files for `static transitive` dependencies.
- Modifies the BND configuration not to mark `org.jspecify` as `transitive`.

We remove the `transitive` modifier from all optional dependencies.

Closes #2929.

**Note**: Once these changes are accepted, I'll move the Groovy script to `logging-parent` and implement apache/logging-parent#196 too.

PR #2963, opened 2024-09-18
Upgrade `logging-parent` to version `11.3.0`

ppkarwasz:
This PR upgrades to the latest version of `logging-parent`.

It's target branch is  #2941 and should be merged after it (and after `logging-parent` 11.3.0 is released).

It includes 3 additional changes:

- An upgrade from Groovy 3 to 4. Groovy is only used in tests and #2930. The new `logging-parent` already uses version 4.
- A back-port of #2953,
- A back-port of #2922.

All these changes involve only **tests**, by improving them. E.g. the MongoDB integration tests no longer use an old (possibly deprecated) version of MongoDB, but the latest compatible with the driver.


ppkarwasz:
Merged in 93483ef03e0643da95692b548311b7c0d1fcdf56

PR #3038, opened 2024-09-30
Remove `log4j-core` dependency from SLF4J bridges

ppkarwasz:
To prepare for the migration between Log4j Core 2 and Log4j Core 3 this PR removes the runtime dependency on `log4j-core` from both `log4j-slf4j-impl` and `log4j-slf4j2-impl`.

This explicitly reverts the changes in [LOG4J2-3601](https://issues.apache.org/jira/browse/LOG4J2-3601).

See #2924 for the discussion.


ppkarwasz:
Closing this, since we discussed an [alternative solution on `dev@logging`](https://lists.apache.org/thread/tk8k8bnmy11b08xftr4p37yzgghpf61j).

PR #3039, opened 2024-10-01
Remove ancient test fixtures and dependencies

ppkarwasz:
This removes the test dependency on `oro:oro` and the `VelocityTest`, which uses Apache Velocity 1.7 for no clear reason.


vy (approved):

ppkarwasz:
@vy,

I also removed the disabled tests in `DOMTestCase` and all the test fixtures they used.

Personally I find that enabling new tests in `log4j-1.2-api` is not worth the effort, except if users report a bug.

vy (approved):
> I also removed the disabled tests in `DOMTestCase` and all the test fixtures they used.

+1

> Personally I find that enabling new tests in `log4j-1.2-api` is not worth the effort, except if users report a bug

+1

PR #3043, opened 2024-10-01
Fix reloading of the configuration from HTTP(S) (`main`)

### LINKED ISSUE #2813: Rewrite Jetty-based unit tests

### LINKED ISSUE #2937: Log4j fails to honor `lastModifiedTime`

ppkarwasz:
This is the port of #2941 to `main`

The `HttpWatcher` didn't propagate the observed last modification time back to the configuration.
As a result, each new configuration was already deprecated when it started and the reconfiguration process looped.

Closes #2937

Rewrite Jetty tests using WireMock

Closes #2813


vy (approved):
Skimmed through the changes, LGTM.

PR #3050, opened 2024-10-02
Fix key removal issues in Thread Context

### LINKED ISSUE #2946: (Closeable)ThreadContext putAll removes entire context
## Description

The ThreadContext no longer works as intended. The new version closes/removes items from the stack when it shouldn't.

CloseableThreadContext calls `ThreadContext.putAll`.
``` java
if (!valuesToReplace.isEmpty()) {
    ThreadContext.putAll(valuesToReplace);
}
```
Once putAll has been called, the stack is gone, and only the newly inserted values are present.
Since it's used by `CloseableThreadContext#close()` it's quite destructive...

## Configuration

**Version:** 2.24.0

**Operating system:** All

**JDK:** All

## Reproduction

```java
	@Test
	public void testAutoCloseableThreadContext() {
		try (final CloseableThreadContext.Instance ctc1 = CloseableThreadContext.put("outer", "one")) {
			try (final CloseableThreadContext.Instance ctc2 = CloseableThreadContext.put("outer", "two")) {
				assertEquals("two", ThreadContext.get("outer"));

				try (final CloseableThreadContext.Instance ctc3 = CloseableThreadContext.put("inner", "one")) {
					assertEquals("one", ThreadContext.get("inner"));

					ThreadContext.put("not-in-closeable", "true");
					assertEquals("two", ThreadContext.get("outer"));
				}

				assertEquals("two", ThreadContext.get("outer"));
				assertNull(ThreadContext.get("inner"));
			}

			assertEquals("one", ThreadContext.get("outer"));
			assertNull(ThreadContext.get("inner"));
		}
		assertEquals("true", ThreadContext.get("not-in-closeable"));

		assertNull(ThreadContext.get("inner"));
		assertNull(ThreadContext.get("outer"));
	}
```

and

```java
	@Test
	public void testAutoCloseableThreadContextPutAll() {
		try (final CloseableThreadContext.Instance ctc1 = CloseableThreadContext.put("outer", "one")) {
			try (final CloseableThreadContext.Instance ctc2 = CloseableThreadContext.put("outer", "two")) {
				assertEquals("two", ThreadContext.get("outer"));

				try (final CloseableThreadContext.Instance ctc3 = CloseableThreadContext.put("inner", "one")) {
					assertEquals("one", ThreadContext.get("inner"));

					ThreadContext.put("not-in-closeable", "true");
					ThreadContext.putAll(Collections.singletonMap("inner", "two"));
					assertEquals("two", ThreadContext.get("inner"));
					assertEquals("two", ThreadContext.get("outer"));
				}

				assertEquals("two", ThreadContext.get("outer"));
				assertNull(ThreadContext.get("inner"));
			}

			assertEquals("one", ThreadContext.get("outer"));
			assertNull(ThreadContext.get("inner"));
		}
		assertEquals("true", ThreadContext.get("not-in-closeable"));

		assertNull(ThreadContext.get("inner"));
		assertNull(ThreadContext.get("outer"));
	}
```

vy:
Fixes #2946.


log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
```diff
@@ -350,82 +351,59 @@ public UnmodifiableArrayBackedMap copyAndRemove(String key) {
...
-            // iterate through the keys and build a set
-            keysToRemoveSet = new HashSet<>();
-            for (String key : keysToRemoveIterable) {
-                keysToRemoveSet.add(key);
-            }
-        }
+        // Collect distinct keys to remove
+        final Set<String> keysToRemove = keysToRemoveIterable instanceof Set
+                ? (Set<String>) keysToRemoveIterable
+                : StreamSupport.stream(keysToRemoveIterable.spliterator(), false)
```
jengebr, line 366: Streams create intermediate objects and add instructions.  Not a great idea in a performance-sensitive spot.

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
```diff
@@ -350,82 +351,59 @@ public UnmodifiableArrayBackedMap copyAndRemove(String key) {
...
+            newMap.numEntries = oldMapEntryCount;
+            return this;
         }
 
-        if (numEntriesKept == 0) {
-            return EMPTY_MAP;
+        // Iterate over old map entries
+        int newMapEntryIndex = 0;
+        for (int oldMapEntryIndex = 0; oldMapEntryIndex < oldMapEntryCount; oldMapEntryIndex++) {
+            final int oldMapKeyIndex = getArrayIndexForKey(oldMapEntryIndex);
```
jengebr, line 384: Much cleaner than the original, thanks.

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
vy: Fixed in 91ff32cec924a06611608bb8d86393c3e1c4be4e.

log4j-api-test/src/test/java/org/apache/logging/log4j/CloseableThreadContextTest.java
```diff
@@ -244,4 +246,71 @@ public void pushAllWillPushAllValues() {
...
         assertEquals("", ThreadContext.peek());
     }
+
+    /**
+     * User provided test stressing nesting using {@link CloseableThreadContext#put(String, String)}.
+     *
+     * @see <a href="https://github.com/apache/logging-log4j2/issues/2946#issuecomment-2382935426">#2946</a>
+     */
+    @Test
+    public void testAutoCloseableThreadContextPut() {
```
ppkarwasz, line 256: _Nit_:
```suggestion
    void testAutoCloseableThreadContextPut() {
```

log4j-api-test/src/test/java/org/apache/logging/log4j/CloseableThreadContextTest.java
```diff
@@ -244,4 +246,71 @@ public void pushAllWillPushAllValues() {
...
+        assertNull(ThreadContext.get("outer"));
+    }
+
+    /**
+     * User provided test stressing nesting using {@link CloseableThreadContext#putAll(Map)}.
+     *
+     * @see <a href="https://github.com/apache/logging-log4j2/issues/2946#issuecomment-2382935426">#2946</a>
+     */
+    @Test
+    public void testAutoCloseableThreadContextPutAll() {
```
ppkarwasz, line 288: _Nit_:
```suggestion
    void testAutoCloseableThreadContextPutAll() {
```

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
ppkarwasz: @vy,

Can you point out, where was the bug in the old implementation? It is not obvious to me.

log4j-api-test/src/test/java/org/apache/logging/log4j/CloseableThreadContextTest.java
```diff
@@ -244,4 +246,71 @@ public void pushAllWillPushAllValues() {
...
+            try (final CloseableThreadContext.Instance ctc2 = CloseableThreadContext.put("outer", "two")) {
+                assertEquals("two", ThreadContext.get("outer"));
+
+                try (final CloseableThreadContext.Instance ctc3 = CloseableThreadContext.put("inner", "one")) {
+                    assertEquals("one", ThreadContext.get("inner"));
+
+                    ThreadContext.put(
+                            "not-in-closeable", "true"); // Remove this line, and closing context behaves as expected
+                    ThreadContext.putAll(Collections.singletonMap("inner", "two")); // But this is not a problem
+                    System.err.println(ThreadContext.getContext());
```
ppkarwasz, line 299: I think that a proper usage of `ThreadContext` requires the removal of the "not-in-closeable" key before the end of the **try-with-resources** block.

Can we remove the assertions regarding "not-in-closeable" and find a different test to assert the correct behavior of `removeAll`?

ppkarwasz (changes requested):
LGTM, :100:

There is certainly a bug in `copyAndRemoveAll` implementation (although I don't seem to be able to pinpoint it), but the user-supplied test cases trigger it by using the `ThreadContext` API incorrectly (IMHO).

Can we trigger the bug with a different unit test?

log4j-api-test/src/test/java/org/apache/logging/log4j/CloseableThreadContextTest.java
tnleeuw: The `ThreadContext` keys like `"not-in-closeable"` are added by code that is not aware of the `CloseableThreadContext` for a number of reasons, and adds information to it that it wants logged independently of what level of nesting we're in.

It is only in the narrow context of a unit-test that all this code is together in a single block, and not spread out over a large codebase.

Why do you think that proper usage requires these keys to be removed before closing the `CloseableThreadContext`?

From our perspective, as users of the Log4J API, if we want keys to be removed by end of the try-with-resources we'd put them in the `CloseableThreadContext` but if we want them to be logged for the entire thread regardless of the scope of a `CloseableThreadContext` on that thread, we put them directly in the `ThreadContext`.

ppkarwasz: > The `ThreadContext` keys like `"not-in-closeable"` are added by code that is not aware of the `CloseableThreadContext` for a number of reasons, and adds information to it that it wants logged independently of what level of nesting we're in.
> 
> (...)
> 
> Why do you think that proper usage requires these keys to be removed before closing the `CloseableThreadContext`?

So, you find useful to leave some values in the thread context even if the execution on that thread has terminated and the thread is reused? Can you give an example of such a value?

In our future development of the Log4j API (e.g. #2438), we assume that thread context values should go out of scope once the execution of a method has finished. So, we are quite interested in use cases that do not follow this model.

> From our perspective, as users of the Log4J API, if we want keys to be removed by end of the try-with-resources we'd put them in the `CloseableThreadContext` but if we want them to be logged for the entire thread regardless of the scope of a `CloseableThreadContext` on that thread, we put them directly in the `ThreadContext`.

Got it

tnleeuw: We do manually clear the `ThreadContext` when we start processing on a new thread... Mostly, we often have nesting of `CloseableThreadContext`s but occassionally we want to add a value to the `ThreadContext` that should apply to the entire remainder of the thread, from a piece of code that doesn't have access to the `CloseableThreadContext`, certainly not to the outermost.

This could be something like an error-status from a nested sub-process, that runs in a `CloseableThreadContext` a few levels deeply nested, but that we want to see for the rest of the processing done by the thread.

It could also be a user-specified value from the nested level.

vy: @tnleeuw, I am having difficulty in understanding the use case you're describing. Could you give a concrete real-world (i.e., not contrived) example, please? Some pseudo code would be handy.

tnleeuw: This is some code copy pasted from one of our classes:
```
			ThreadContext.put(LogUtil.MDC_EXIT_STATE_KEY, result.getState().name());
			if (result.getExitCode() != null) {
				ThreadContext.put(LogUtil.MDC_EXIT_CODE_KEY, Integer.toString(result.getExitCode()));
			}
			ThreadContext.put("pipeline.duration", msgLogHumanReadable ? Misc.getAge(startTime) : Long.toString(duration));
```
https://github.com/frankframework/frankframework/blob/1bca5c0f3e8fb1b4760db0b8ee9e1ca8c526c44d/core/src/main/java/org/frankframework/core/Adapter.java#L636-640

Which does exactly what I described above: copy exit status information from a (sub)process to the ThreadContext. The method is called from another class and executed within scope of a `CloseableThreadContext`, but the exit state information should remain logged for the remainder of the process (along with other information not shown here, set at the higher levels of code execution) until end of the process, or overwritten with new information, so that in the log clients can see at every step what the state of the process is.

Another feature we have is client-configurable information that can be added to either the `CloseableThreadContext` or `ThreadContext`. We do not know what exactly clients want to put in the `ThreadContext`, but we do know there are clients that use this feature.
See: https://github.com/frankframework/frankframework/blob/1bca5c0f3e8fb1b4760db0b8ee9e1ca8c526c44d/core/src/main/java/org/frankframework/pipes/LogContextPipe.java#L84-L84

log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java
vy: Added a dedicated test for the `copyAndRemoveAll()` in df5ec8115974970b5625836d9c06a7d72a65b813. The bug was on this line:

https://github.com/apache/logging-log4j2/blob/2d61fe0e6a3b9c8f5d53dda1f5834752a2799138/log4j-api/src/main/java/org/apache/logging/log4j/internal/map/UnmodifiableArrayBackedMap.java#L394

It skips the removal if the slice-to-be-copied is `[0,0]`.

log4j-api-test/src/test/java/org/apache/logging/log4j/CloseableThreadContextTest.java
ppkarwasz: This discussion is a little bit transversal to this PR, which only fixes a bug in `removeAll()`.
Let us continue it in #2946.

ppkarwasz (approved):

PR #3065, opened 2024-10-07
Move BND configuration to `pom.xml`

ppkarwasz:
This removes the BND configuration in `log4j-1.2-api/bnd.bnd` and moves it to the POM file.

The resulting JPMS module has the same requirements as version `2.24.1`.

```
requires java.base
requires java.desktop
requires java.management transitive
requires java.xml transitive
requires javax.jms-api static
requires org.apache.logging.log4j transitive
requires org.apache.logging.log4j.core static
```


log4j-1.2-api/pom.xml
```diff
@@ -43,11 +43,12 @@
       com.sun.jdmk.comm;resolution:=optional,
       <!-- JMS is optional -->
       javax.jms;version="[1.1,3)";resolution:=optional,
-      <!-- Log4j Core is optional -->
-      org.apache.logging.log4j.core.*;resolution:=optional
     </bnd-extra-package-options>
+    <bnd-extra-module-options>
```
vy, line 47: Don't we also need an entry for `javax.jms:javax.jms-api` too?

ppkarwasz: No, setting the `javax.jms` module to `resolution:=optional` is enough for BND to mark the JPMS module as optional.

PR #3069, opened 2024-10-07
Upgrades JUnit to version `5.11.2`

### LINKED ISSUE #2843: Upgrade JUnit to 5.11.0
JUnit version 5.11.0 breaks some of our extensions, so we need to upgrade it manually.

I am creating this issue and closing Dependabots PRs #2839 and #2841.

ppkarwasz:
Due to a minor bug in JUnit 5.11.2 (junit-team/junit5#4054), fields meta-annotated with multiple `@ExtendWith` annotations are no longer recognized. We combine those annotations into single `@ExtendWith` annotations, which also looks better.

Closes #2843.

**Edit**: No changelog is necessary, since improvements to our test runtime are not advertised. One day we need to find a way to thank the suppliers of our test runtime.

### DIFF
--- a/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/package-info.java
+++ b/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/package-info.java
@@ -20,7 +20,7 @@
  * @see org.junit.rules.TestRule
  */
 @Export
-@Version("2.23.0")
+@Version("2.23.1")
 package org.apache.logging.log4j.core.test.junit;
 
 import org.osgi.annotation.bundle.Export;
--- a/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/SetTestProperty.java
+++ b/log4j-api-test/src/main/java/org/apache/logging/log4j/test/junit/SetTestProperty.java
@@ -39,8 +39,7 @@ import org.junitpioneer.jupiter.ReadsSystemProperty;
 @Target({TYPE, METHOD})
 @Inherited
 @Documented
-@ExtendWith(ExtensionContextAnchor.class)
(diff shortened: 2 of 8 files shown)


log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/package-info.java
```diff
@@ -20,7 +20,7 @@
  * @see org.junit.rules.TestRule
  */
 @Export
-@Version("2.23.0")
+@Version("2.23.1")
```
vy, line 23:
```suggestion
@Version("2.25.0")
```

vy (changes requested):

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/junit/package-info.java
ppkarwasz: Why? It is a `MICRO` change, so I only increased the patch number.

vy: Weren't you the one advising that we should use the version number of the release a change will be published with? (If I'm mistaken, you can just ignore my correction, but I'd appreciate an explanation.)

ppkarwasz: For `MINOR` changes yes.
What I would like to convey with the `@Version` annotation is:

- the stability of the API in each package,
- in which minor version was the API last modified (a sort of `max(@since)`).

In this case this package was last modified in `2.23.0`, since then users didn't get any additional functionality, only bug fixes. If they compile their code against `2.23.0`, they will get the same result as using the latest release.

PR #3070, opened 2024-10-08
Remove JANSI dependency in `2.x`

### LINKED ISSUE #1736: Upgrade Jansi support to version 2.x

ppkarwasz:
This commit:

- Removes support for the outdated [Jansi 1.x](http://fusesource.github.io/jansi/) version in `Console` appender.
- Rewrites `JAnsiTextRenderer`, use in the `%m{ansi}` and `%ex{ansi}` pattern converters to use our internal ANSI support instead of Jansi.

Fixes #1736.


ppkarwasz:
Since we already discussed these changes in #2916, I am merging this.

PR #3061, opened 2024-10-07
Migrate `log4j-core-test` to JUnit 5

ninetteadhikari:
### Introduction

We are from Neighbourhoodie, the implementation partner of the [STF](https://www.sovereigntechfund.de/programs/bug-resilience) Bug Resilience Program. This work is part of our agreed `Milestone 1. Upgrade from JUnit 4 to JUnit 5`. This PR migrates the tests located in `log4j-core-test` to JUnit5.c

Please note the `log4j-core-test` module contains over 150 files and this PR updates the tests for 40 of them. Further PRs will be made to update the rest of the tests.

## Checklist

- [x] Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise
- [x] `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `./mvnw spotless:apply` and retry)
- [ ] Non-trivial changes contain an entry file in the `src/changelog/.2.x.x` directory
- [ ] Tests for the changes are provided
- [x] [Commits are signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (optional, but highly recommended)

PR #3080, opened 2024-10-11
Migrate `log4j-slf4j2-impl` to JUnit 5

AlbaHerrerias:
Hello 👋

We are from Neighbourhoodie, the implementation partner of the [STF](https://www.sovereigntechfund.de/programs/bug-resilience) Bug Resilience Program. This work is part of our agreed Milestone 1. Upgrade from JUnit 4 to JUnit 5. This PR migrates the tests located in `log4j-slf4j2-impl`.

⚠️ We've found out that [`OverflowTest`](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/OverflowTest.java) is [set to run specifically with Junit 4](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-slf4j2-impl/pom.xml#L134). This behaviour was added not long ago so we wanted to ask, should we leave it like this or should we refactor `OverflowTest` to JUnit 5?

Thank you!

## Checklist

### DIFF
--- a/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/LoggerContextTest.java
+++ b/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/LoggerContextTest.java
@@ -16,12 +16,12 @@
  */
 package org.apache.logging.slf4j;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Set;
 import org.apache.logging.log4j.core.LifeCycle;
 import org.apache.logging.log4j.spi.LoggerContext;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
 import org.slf4j.LoggerFactory;
 
 /**
@@ -35,9 +35,9 @@ public class LoggerContextTest {
         factory.getLogger("test");
         Set<LoggerContext> set = factory.getLoggerContexts();
         final LoggerContext ctx1 = set.toArray(LoggerContext.EMPTY_ARRAY)[0];
-        assertTrue("LoggerContext is not enabled for shutdown", ctx1 instanceof LifeCycle);
+        assertTrue(ctx1 instanceof LifeCycle, "LoggerContext is not enabled for shutdown");
         ((LifeCycle) ctx1).stop();
         set = factory.getLoggerContexts();
-        assertTrue("Expected no LoggerContexts", set.isEmpty());
+        assertTrue(set.isEmpty(), "Expected no LoggerContexts");
     }
 }
--- a/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/LoggerContextTest.java
+++ b/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/LoggerContextTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.logging.slf4j;
 
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Set;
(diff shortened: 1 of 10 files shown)


log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/LoggerContextTest.java
```diff
@@ -35,9 +35,9 @@ public void testCleanup() throws Exception {
         factory.getLogger("test");
         Set<LoggerContext> set = factory.getLoggerContexts();
         final LoggerContext ctx1 = set.toArray(LoggerContext.EMPTY_ARRAY)[0];
-        assertTrue("LoggerContext is not enabled for shutdown", ctx1 instanceof LifeCycle);
+        assertTrue(ctx1 instanceof LifeCycle, "LoggerContext is not enabled for shutdown");
```
ppkarwasz, line 38:
```suggestion
        assertInstanceOf(LifeCycle.class, ctx1, "LoggerContext is not enabled for shutdown");
```

Could you change the assertions on `X instanceof Y` to use the new `assertInstanceOf` assertions?

ppkarwasz (changes requested):
This looks great, thanks!

Since the `log4j-slf4j-impl` module is mostly identical to this one (it has a couple of tests less). Could you apply the changes to that module too?

Could you also remove the `junit-vintage-engine` dependency in the `pom.xml` to show that not tests are using JUnit 4?

ppkarwasz:
> ⚠️ We've found out that [`OverflowTest`](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/OverflowTest.java) is [set to run specifically with Junit 4](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-slf4j2-impl/pom.xml#L134). This behaviour was added not long ago so we wanted to ask, should we leave it like this or should we refactor `OverflowTest` to JUnit 5?

As far as I can tell `OverflowTest` does not depend specifically on JUnit 4. It only requires an additional dependency (`log4j-to-slf4j`). The module should detect that Log4j API is redirected to SLF4J and should not forward SLF4J calls back to Log4j API.
Taking advantage of new Maven Surefire Plugin capabilities I believe that such a configuration should work:

```xml
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-surefire-plugin</artifactId>
        <executions>
          <execution>
            <id>loop-test</id>
            <goals>
              <goal>test</goal>
            </goals>
            <configuration>
             <additionalClasspathDependencies>
               <dependency>
                 <artifactId>log4j-to-slf4j</artifactId>
               </dependency>
             </additionalClasspathDependencies>
              <includes>
                <include>**/OverflowTest.java</include>
              </includes>
            </configuration>
          </execution>
          <execution>
            <id>default-test</id>
            <configuration>
              <includes>
                <include>**/*Test.java</include>
              </includes>
              <excludes>
                <exclude>**/OverflowTest.java</exclude>
              </excludes>
            </configuration>
          </execution>
        </executions>
      </plugin>
```

and `log4j-to-slf4j` can be removed from the `<dependencies>` section of the project.

AlbaHerrerias:
Thank you for the review and assistance, I have addressed now your requested changes. Please let me know if you'd like to request more.

I will make the refactor of the `log4j-slf4j-impl` module. My plan is to make it as a new PR (as we have been doing; a PR per module), but if you'd rather me doing it in this PR, please let me know and I will do so.

Also note that, with the removal of `junit-vintage-engine`, the tests of [`Log4j2_1482_Slf4jTest.java`](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-slf4j2-impl/src/test/java/org/apache/logging/slf4j/Log4j2_1482_Slf4jTest.java) are not being executed (total of 2 tests). They
extend from [`log4j-core-test/.../Log4j2_1482_Test.java`](https://github.com/apache/logging-log4j2/blob/cbe7e91304cd0d3681a3c6eedc1782b25cb96242/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/layout/Log4j2_1482_Test.java), currently written in JUnit 4. My coworker is refactoring those in https://github.com/apache/logging-log4j2/pull/3061.

ppkarwasz (approved):
Looks good to me, thanks! :rocket: 

No need to port this to `main`, since the 3.x artifact will be removed soon (#2924).

PR #3085, opened 2024-10-14
Avoid levels like `INFO#org.apache.log4j.Level`

jonasrutishauser:
If somebody (for example an external library) uses the deprecated `Priority` class there will be a custom level created even though there already exists a corresponding level.

### DIFF
--- a/log4j-1.2-api/src/test/java/org/apache/log4j/LevelTest.java
+++ b/log4j-1.2-api/src/test/java/org/apache/log4j/LevelTest.java
@@ -16,12 +16,15 @@
  */
 package org.apache.log4j;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Locale;
+import org.apache.log4j.helpers.OptionConverter;
 import org.apache.log4j.util.SerializationTestHelper;
-import org.junit.Test;
+import org.junit.jupiter.api.Test;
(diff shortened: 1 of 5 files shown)


ppkarwasz:
@jonasrutishauser,

Nice catch!

Looking at the `o.a.l.Priority` class there are other things that can go wrong:

1. `Priority.version2Level` can be `null`. This might lead to incorrect filtering of `Priority` instances (possibly even an NPE). Can you create a new package-private constructor that always sets this field to a non-`null` value:
   ```java
   this.version2Level = version2Equivalent != null ? version2Equivalent : OptionConverter.createLevel(this);
   ```
1. The `Priority.FATAL`, `Priority.ERROR` and so on fields reference the `Level` subclass, which might lead to class loading deadlocks on some JVMs (or they might just be `null` on others). Can you create them using the new constructor instead?

log4j-1.2-api/src/test/java/org/apache/log4j/LevelTest.java
```diff
@@ -205,6 +207,70 @@ public void testALL() {
...
+        assertEquals(org.apache.logging.log4j.Level.OFF, Level.OFF.getVersion2Level());
+    }
+
+    /**
+     * Tests Level.FATAL version2Level.
+     */
+    @Test
+    public void testFatalVersion2Level() {
+        assertEquals(org.apache.logging.log4j.Level.FATAL, Level.FATAL.getVersion2Level());
+    }
```
ppkarwasz, line 224: The test class uses unfortunately JUnit 4, but it should be easy to convert it to a JUnit 5 test class with a `@ParameterizedTest` instead of multiple single tests.

ppkarwasz:
@jonasrutishauser,

Thank You for your contribution!

PR #3118, opened 2024-10-24
Create `DefaultLayout` independent of `PatternLayout`

ppkarwasz:
Currently, `PatternLayout` and all its patterns are a required element of Log4j Core, since `DefaultConfiguration` uses it. The default configuration is used in all Log4j Core installation as failsafe to handle logging between the time a logger context is created and the real configuration starts (a couple of ms).

This PR creates a simple hardcoded layout to use with `DefaultConfiguration`.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
@@ -16,6 +16,18 @@
  */
 package org.apache.logging.log4j.core.config;
 
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.Writer;
+import java.nio.charset.Charset;
+import org.apache.logging.log4j.LoggingException;
+import org.apache.logging.log4j.core.Layout;
+import org.apache.logging.log4j.core.LogEvent;
+import org.apache.logging.log4j.core.layout.AbstractLayout;
+import org.apache.logging.log4j.core.util.StringBuilderWriter;
+import org.apache.logging.log4j.core.util.datetime.FixedDateFormat;
+import org.apache.logging.log4j.core.util.datetime.FixedDateFormat.FixedFormat;
+
 /**
  * The default configuration writes all output to the Console using the default logging level. You configure default
  * logging level by setting the system property "org.apache.logging.log4j.level" to a level name. If you do not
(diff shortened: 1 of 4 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
```diff
@@ -49,4 +61,60 @@ public DefaultConfiguration() {
...
+    /**
+     * A simple layout used only by {@link DefaultConfiguration}
+     * <p>
+     *   This layout allows to create applications that don't contain {@link org.apache.logging.log4j.core.layout.PatternLayout}
+     *   and all its patterns, e.g. GraalVM applications.
+     * </p>
+     */
+    private static final class DefaultLayout extends AbstractLayout<String> {
+
+        private static final byte[] EMPTY_BYTE_ARRAY = new byte[0];
```
vy, line 78: Do we really need this optimization? Can't we simply inline `new byte[0]` to call sites?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
```diff
@@ -49,4 +61,60 @@ public DefaultConfiguration() {
...
+
+        private final FixedDateFormat dateFormat = FixedDateFormat.create(FixedFormat.ABSOLUTE);
```
vy, line 80: Can you use `DateTimeFormatter`, please? I am working on nuking both `FixedDateFormat` and `FastDateFormat`.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
```diff
@@ -49,4 +61,60 @@ public DefaultConfiguration() {
 
     @Override
     protected void doConfigure() {}
+
+    static Layout<? extends String> createDefaultLayout() {
```
vy, line 65: Why do we need to publish `DefaultLayout` like this? Can't `DefaultLayout` simply be an individual package-private class in `.core.config` package where `AbstractConfiguration` accesses it via `DefaultLayout.INSTANCE`? I mean, given `DL` is only used by `AC`, I don't see the reason to contain it in `DC`.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
```diff
@@ -49,4 +61,60 @@ public DefaultConfiguration() {
...
+    }
+
+    /**
+     * A simple layout used only by {@link DefaultConfiguration}
+     * <p>
+     *   This layout allows to create applications that don't contain {@link org.apache.logging.log4j.core.layout.PatternLayout}
+     *   and all its patterns, e.g. GraalVM applications.
+     * </p>
+     */
+    private static final class DefaultLayout extends AbstractLayout<String> {
```
vy, line 76: Is there any way we can reuse logic from `StatusData#getFormattedStatus()` or `SimpleLogger`? That is, instead of creating yet another primitive event formatter, maybe we can reuse something already there in the `log4j-api`?

vy (commented):
I really like the idea @ppkarwasz! :100:

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
ppkarwasz: We can do better and use `null`.

Fixed in https://github.com/apache/logging-log4j2/pull/3118/commits/bf892e169e273d3e9c4e4812e4ffac4178460a3d

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
ppkarwasz: I replaced the formatter with whatever `StatusData` uses, so it will use DTF.

Fixed in https://github.com/apache/logging-log4j2/pull/3118/commits/bf892e169e273d3e9c4e4812e4ffac4178460a3d

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/3118/commits/bf892e169e273d3e9c4e4812e4ffac4178460a3d

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultConfiguration.java
ppkarwasz: Good call!

I went for `StatusData` in https://github.com/apache/logging-log4j2/pull/3118/commits/bf892e169e273d3e9c4e4812e4ffac4178460a3d
This slightly changes the format and removes the name of the logger, but I really doubt there is a real time situation were `DefaultConfiguration` is used. Logging is usually configured **before** a JVM starts new threads, so the concurrency problem in `LoggerContext` is mostly a worst case scenario.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/DefaultLayout.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.logging.log4j.core.layout.ByteBufferDestination;
+import org.apache.logging.log4j.status.StatusData;
+
+/**
+ * A simple layout used only by {@link DefaultConfiguration}
+ * <p>
+ *   This layout allows to create applications that don't contain {@link org.apache.logging.log4j.core.layout.PatternLayout}
+ *   and all its patterns, e.g. GraalVM applications.
+ * </p>
+ */
```
vy, line 33:
```suggestion
 *
 * @since 2.25.0
 */
```

vy (approved):

PR #3121, opened 2024-10-25
Rewrite date & time formatting

### LINKED ISSUE #2943: `FixedDateFormat` incorrectly calculates DST for `America/Santiago` time zone
## Description

After DST time change in Chile , the logs are not showing the correct time with plain %d date pattern on September 8th but after midnight of 9th September the logs are printing correct date without having to change anything 
but when we use date format with time zone it shows correct date on 8th September also .

## Configuration

**Version:** 2.24

**Operating system:** RHEL

**JDK:** openJDK17

## Logs

```
[Stacktraces, errors, etc. relevant applications logs.]
```

## Reproduction

Here’s the steps to reproduce it:
1.	Set the date on the server to September 8, with the time any time after 2am.
2.	Use a plain %d date format in log4j config file . 

Here is the java code and config file  : 
import java.util.TimeZone;

public class SimpleLog4jExample {

    // Create a logger instance
    private static final Logger logger = LogManager.getLogger(SimpleLog4jExample.class);

        // Log messages of different levels
         String tzid = "America/Santiago";
       // tzid = readln("Enter time zone ID [" + tzid + "]: ", tzid);
        TimeZone tz = TimeZone.getTimeZone(tzid);
        TimeZone.setDefault(tz);
        logger.info("This is an info message");
        logger.warn("This is a warning message");
        logger.error("This is an error message");
    }
}

log4j2.xml

<Configuration status="WARN">
    <Appenders>
            <PatternLayout pattern="%d %-5p %c{1}:%L - %m%n"/>
        </Console>
    </Appenders>
    <Loggers>
        <Root level="info">
            <AppenderRef ref="Console"/>
        </Root>
    </Loggers>
</Configuration>

vy:
In accordance with [this `dev@` discussion](https://lists.apache.org/thread/vmv7nzk2f263bdc3k1z0j1nfmos1f6dk), these changes

* Deprecate `FixedDateTime` and `FastDateTime`
* Adds the _internal_ `InstantFormatter` contract
  * Adds `InstantNumberFormatter` for epoch-related number-emitting formatting operations
  * Adds the `InstantPatternNumber` contract for formatting based on a specified date & time patter
    * Adds `InstantPatternThreadLocalCachedFormatter` which wraps any `InstantPatternNumber` and caches its values in a `ThreadLocal` for reusage
    * Adds `InstantPatternDynamicFormatter` that uses Java's `DateTimeFormatter` under the hood such that the pattern is analyzed and parts that require a precision lower than or equal to minute are precomputed, cached, and updated per minute. That is, `DateTimeFormatter` is effectively only used for sub-minute parts, e.g., `ss.SSS`.

## `InstantPatternDynamicFormatter` internals

For instance, given the pattern `yyyy-MM-dd'T'HH:mm:ss.SSSX`, the generated formatter will

1. Sequence the pattern and assign a time precision to each part (e.g., `MM` is of month precision)
2. Precompute and cache the output for parts that are of precision lower than or equal to minute (i.e., `yyyy-MM-dd'T'HH:mm:` and `X`) and cache it
3. Upon a formatting request, combine the cached outputs with the dynamic parts (i.e., `ss.SSS`)</li>

## GC implications

Since `DateTimeFormatter` is not garbage-free, neither the new `InstantPatternDynamicFormatter` is.

## Performance implications

**JDK:** Azul Zulu `17.0.12`
**OS:** GNU/Linux
**Date & time pattern:** `yyyy-MM-dd'T'HH:mm:ss.SSS`

### `InstantPatternFormatterBenchmark`

This benchmark stresses instant formatting efficiency. 1000 instants are formatted per operation.

```
DateTimeFormatter        2.844 ± 0.310  ops/ms
InstantPatternFormatter  4.807 ± 1.039  ops/ms
FastDateFormat           6.847 ± 1.302  ops/ms
FixedDateFormat         39.497 ± 6.423  ops/ms
```

**Interpretation:** When used in isolation, `FixedDF` yields a throughput of ~8x compared to `InstantPF`.

### `InstantPatternFormatterImpactBenchmark`

This benchmark stresses instant formatting efficiency when combined with `PatternLayout` configured with the `[%t] %p %-40.40c{1.} %notEmpty{%x }- %m%n` pattern. 1000 `LogEvent`s are formatted per operation.

**Using full log events (incl. exception, MDC & NDC):**
```
DateTimeFormatter       0.039 ± 0.001  ops/ms
InstantPatternFormatter 0.039 ± 0.001  ops/ms
FastDateFormat          0.039 ± 0.001  ops/ms
FixedDateFormat         0.040 ± 0.001  ops/ms
```

**Interpretation:** When `LogEvent` contains an exception, instant formatting efficiency becomes irrelevant.

**Using lite log events:**
```
DateTimeFormatter       1.888 ± 0.106  ops/ms
InstantPatternFormatter 2.404 ± 0.184  ops/ms
FastDateFormat          2.567 ± 0.032  ops/ms
FixedDateFormat         4.793 ± 0.019  ops/ms
```

**Interpretation:** When `LogEvent` doesn't contain an exception, MDC/NDC field, etc., `FixedDF` yields a throughput of ~2x compared to `InstantPF`.

## Bug implications

Fixes #2943.


ppkarwasz (commented):
> We don't really need DTF to print sub-day patterns (hours, minutes, seconds, etc.).

@ppkarwasz, what do you mean? If the pattern is, say, `yyyy-MM-dd HH:mm:ss.SSS`, we cache `yyyy-MM-dd HH:mm:` part and update it once per minute. I think this once-per-minute cost is negligible in a real-world setup.

> Are you sure `DTF` is garbage-free for these patterns? Can you add a test for it?

DTF is **not** garbage-free – no matter which method of DTF you use. Hence, neither our `InstantPatternDynamicFormatter` is garbage-free.

> the `PatternSequence`s can actually be pre-computed, which would improve the time to generate the formatters.

Agreed, but I doubt if this additional complexity brings any practical benefit. Even my JMH tests run for 30 seconds, that is, they don't even encounter any sequencing operation.

Allow me to remind that my aim with this PR is not to implement the fastest instant formatter on top of DTF, but implement the simplest one with acceptable performance. `FixedDF` and `FastDF` were enough of a lesson for what happens with unsolicited complexity.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
```diff
@@ -16,38 +16,50 @@
...
 
-        abstract String format(final Instant instant);
+    private abstract static class Formatter<F extends InstantFormatter> {
 
-        abstract void formatToBuffer(final Instant instant, StringBuilder destination);
+        final F delegate;
+
+        private Formatter(final F delegate) {
+            this.delegate = delegate;
+        }
```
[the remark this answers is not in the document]
vy: Good catch! Removed in ff4f6882b618ed65bb547d4bcb3c46a1eb4f810e.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java
```diff
@@ -18,13 +18,12 @@
...
 import java.util.Locale;
 import java.util.TimeZone;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.time.Instant;
-import org.apache.logging.log4j.core.time.MutableInstant;
+import org.apache.logging.log4j.core.util.internal.instant.InstantFormatter;
+import org.apache.logging.log4j.core.util.internal.instant.InstantNumberFormatter;
+import org.apache.logging.log4j.core.util.internal.instant.InstantPatternFormatter;
```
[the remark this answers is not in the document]
vy: > OSGi needs the package to be exported to anyone.

Isn't there any other way for OSGi? (Yes, I don't want to open up the new package & classes.)

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/util/InstantFormatter.java
```diff
@@ -37,7 +37,10 @@
  * Note that {@link FixedDateFormat} and {@link FastDateFormat} only support
  * millisecond precision. If the pattern asks for a higher precision,
  * {@link DateTimeFormatter} will be employed, which is significantly slower.
+ * </p>
+ * @deprecated Starting with version {@code 2.25.0}, this class is planned to be removed in the next major release.
  */
+@Deprecated
 public final class InstantFormatter {
```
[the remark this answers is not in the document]
vy: I deliberately do not want to. JTL's formatter was internal, it should not have been used. But there were no encapsulation mechanisms back then to enforce this. Users who are not comfortable with the current instant formatting should either submit a feature request or roll out their own instant formatters. But I am not willing to make ours public. (I will comment about JPMS/OSGi visibility in another conversation.)

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternDynamicFormatter.java
```diff
@@ -0,0 +1,807 @@
...
+            } else if (paddingRemovedContent.matches("[MLQq]+")) {
+                return ChronoUnit.MONTHS;
+            } else if (paddingRemovedContent.matches("[wW]+")) {
+                return ChronoUnit.WEEKS;
+            } else if (paddingRemovedContent.matches("[DdgEecF]+")) {
+                return ChronoUnit.DAYS;
+            } else if (paddingRemovedContent.matches("[aBhKkH]+")
+                    // Time-zone directives
+                    || paddingRemovedContent.matches("[ZxXOzvV]+")) {
+                return ChronoUnit.HOURS;
```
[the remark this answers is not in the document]
vy: I really do not want to have any assumptions on the time-zones anymore. Judging from the Wikipedia page you shared, all DST changes happen at the beginning of an hour. Hence, `HOURS` seem to be the correct precision to me. (Let me know if I misread the page.)

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java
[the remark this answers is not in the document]
ppkarwasz: IIRC, when the subject of `exports ... to` came out on `dev@logging` (or maybe the PMC meeting) the consensus was that we should either export a package to all users or don't export it at all.

As I stated above, we only need to put `InstantFormatter` in an exportable package. If the package has `spi` in it, users will know that it only should be used by Log4j Plugin implementors.

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/TimestampResolver.java
```diff
@@ -331,119 +273,48 @@ private static EventResolver createEpochResolver(final TemplateResolverConfig co
...
-    private static final EventResolver EPOCH_NANOS_RESOLVER = new EpochResolver() {
-        @Override
-        void resolve(final Instant logEventInstant, final JsonWriter jsonWriter) {
-            final long nanos = epochNanos(logEventInstant);
-            jsonWriter.writeNumber(nanos);
-        }
+    private static final EventResolver EPOCH_NANOS_RESOLVER = (logEvent, jsonWriter) -> {
+        final StringBuilder buffer = jsonWriter.getStringBuilder();
+        final Instant instant = logEvent.getInstant();
+        InstantNumberFormatter.EPOCH_NANOS.formatTo(buffer, instant);
```
ppkarwasz, line 279: Is this behavioral change intentional? The type of the field is changed from Number to String.

vy: The output is still a number, though this time directly written to the SB. Am I missing something?

ppkarwasz: Ah, so you are writing a **raw** string to JSON? I didn't know this was possible.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternFormatter.java
```diff
@@ -28,6 +28,12 @@
 
 /**
  * Contract for formatting {@link Instant}s using a date & time formatting pattern.
+ * <h2>Internal usage only!</h2>
+ * <p>
+ * This class is intended only for internal Log4j usage.
+ * <b>Log4j users should not use this class!</b>
+ * This class is not subject to any backward compatibility concerns.
+ * </p>
```
ppkarwasz, line 36: Do we really need access to this class from JTL?
Can't we just instantiate all the formatters from an `InstantFormatter` static factory method?

We could have for example an `InstantFormatter.newInstance(String pattern, TimeZone, Locale)` public method. I don't see why we would need a builder pattern: these 3 parameters were enough for a very long time. Of course, `TimeZone` will be ignored by `InstantNumberFormatter`, but `Locale` might apply to both formatters (if users request it of course).

vy: > Do we really need access to this class from JTL? Can't we just instantiate all the formatters from an `InstantFormatter` static factory method?
> 
> We could have for example an `InstantFormatter.newInstance(String pattern, TimeZone, Locale)` public method. I don't see why we would need a builder pattern: these 3 parameters were enough for a very long time. Of course, `TimeZone` will be ignored by `InstantNumberFormatter`, but `Locale` might apply to both formatters (if users request it of course).

Why shall we sacrifice from type safety to avoid exporting two classes, where we explicitly stated they are intended for internal usage only? If this approach would have helped to avoid exporting the entire package, I could have agreed with the sentiment, but that is not the case.

ppkarwasz (approved):
LGTM

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternDynamicFormatter.java
```diff
@@ -47,6 +47,30 @@
  * <li>Precompute and cache the output for parts that are of precision lower than or equal to {@value InstantPatternDynamicFormatter#PRECISION_THRESHOLD} (i.e., {@code yyyy-MM-dd'T'HH:mm:} and {@code X}) and cache it</li>
  * <li>Upon a formatting request, combine the cached outputs with the dynamic parts (i.e., {@code ss.SSS})</li>
  * </ol>
+ * <h2>Implementation note</h2>
+ * <p>
+ * Formatting can actually even be made faster and garbage-free by manually formatting sub-minute precision directives as follows:
```
ppkarwasz, line 52: I guess that the `DateTimeFormatter` does not become garbage-free on its own, when it is hot enough?

vy: @ppkarwasz, I indeed doubt that. If you'd skim through `DateTimeFormatter#formatTo()`, there are several new object instantiations.

Issue #3106, opened 2024-10-18
Performance regression in `2.25.0-SNAPSHOT`


ppkarwasz:
As shown by @hailshao in #3093, there might be a 60% loss of performance in the current `2.25.0-SNAPSHOT`, compared to `2.24.1`.
Before we publish `2.25.0` we need to run the benchmarks and fix the problem.

I consider this a blocker for the `2.25.0` release.

ppkarwasz:
The performance hit is probably due to the new `%xEx`. Performance could be improved by switching to `%ex` by default (since most users don't need packaging information anyway).

@jengebr, could you look at this? Does the performance regression also affect your usage?

jengebr:
@ppkarwasz our apps only rarely log exceptions, so this cost is present but quite small in comparison with other log4j work (especially MDC/ThreadContext).  I'll take a look at this in isolation though, and get back in a few days.

ppkarwasz:
Thanks, the statistics I mentioned above are obviously a little bit artificial: if we want to benchmark the formatting of an exception, we should take as a baseline the creation of exceptions without logging (which is probably very slow).

jengebr:
I recreated a large performance regression coming from ThrowablePatternConverter.format (2.25.0) vs ExtendedThrowablePatternConverter.format (2.24.1).  Does that sound right so far?

FWIW I was able to create one Exception as a static variable, then reuse that for every test iteration.  It's not a perfect reproduction but it certainly demonstrates the regression quite clearly.

ppkarwasz:
So even `ThrowablePatternConverter` (2.25.0) is slower than `ExtendedThrowablePatternConverter` (2.24.1)? That is **unexpected**, we'll need to check that.

jengebr:
## Problem
Data shows 2.24.x has nearly zero object allocation during the benchmark, but 2.25.0 has far more than that.  The allocation-heavy code in 2.25.x also comes with significant cpu load.

**Caveat**: I'm not 100% certain I reproduced the specific issue reported.


## Data
1. Async-profiler data shows that the overall hot path is`org.apache.logging.log4j.core.pattern.ThrowableExtendedStackTraceRenderer$ExtendedContext.lambda$createClassResourceInfoByName$0`.  Put another way, `ThrowableExtendedStackTraceRenderer.createClassResourceInfoByName()`. 
2. The benchmark (attached below) shows the old code allocates almost zero objects but the new code allocates rather more than that.
[...]

PR #3093, opened 2024-10-16
Disruptor + many logger.error("aa", new Exception("aaa"))，stuck on Th…

hailshao:
…rowableProxyHelperbak#loadClass

1、log4j2.component.properties
Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
2、for (int i = 0; i < 600000; i++) {
logger.error("aa", aa);
}
3、iotop（linux）only 13M/s
stuck on ThrowableProxyHelper#loadClass
4、 iotop（linux）new way:140M/s

PR #3123, opened 2024-10-26
Fix extended stack trace (i.e., `%xEx`) rendering performance regression

alan0428a:
* Fixes the extended stack trace rendering performance regression introduced in #2691. The issue is reported in #3106, due to #3093.
* Switches the default exception handler of Pattern Layout from `%xEx` to `%ex`


alan0428a:
@ppkarwasz @vy 
baseline
2.24.1 baseline: around 150MB

2 25 0-SNAPSHOT
2.25.0 SNAPSHOT: around 2GB

initial-fix
Initial fix [ea9944f](https://github.com/apache/logging-log4j2/pull/3123/commits/ea9944fea846aa40c261b8b975b6a3fec6c5c42c) reduce to around 300 MB

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ClassResourceInfo.java
vy, line 1: Could you instead keep the old version but replace `private final String text` with `private final Lazy<String> textRef` such that

* in ctor[0], use `Lazy.value("~[?:?]")`
* in ctor[2], use `Lazy.lazy(() -> getText())`

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableExtendedStackTraceRenderer.java
```diff
@@ -105,12 +107,8 @@ private static Map<String, ClassResourceInfo> createClassResourceInfoByName(
             // But we need the associated `Class<?>` to extract its resource information, i.e., JAR file and version.
             // We are capturing the current stack to find suitable class loaders.
             // We will use this as a bootstrap to go from a class name in a stack trace to a `Class<?>`.
-            final Map<String, ClassResourceInfo> classResourceInfoByName =
-                    StackLocatorUtil.getCurrentStackTrace().stream()
-                            .collect(Collectors.toMap(
-                                    Class::getName,
-                                    clazz -> new ClassResourceInfo(clazz, true),
-                                    (classResourceInfo1, classResourceInfo2) -> classResourceInfo1));
+            final Deque<Class<?>> rootStackTrace = StackLocatorUtil.getCurrentStackTrace();
```
vy, line 110: I really don't like this deque hack – I know you borrowed it from the earlier version, hence not your fault. It makes things pretty difficult to wrap one's mind around it. I will spend more time on this sometime tomorrow to see if we can do something else...

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ClassResourceInfo.java
alan0428a: Updated

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableExtendedStackTraceRenderer.java
alan0428a: Yeah...it is hard to read...

jengebr:
My benchmark shows that these changes are ~10% slower than 2.24.1, and use 3x the heap.  That is far better than previous 2.25.0, which was more than 2x slower and used ~15x the heap.

What level of performance regression is acceptable?

ppkarwasz:
> My benchmark shows that these changes are ~10% slower than 2.24.1, and use 3x the heap. That is far better than previous 2.25.0, which was more than 2x slower and used ~15x the heap.
> 
> What level of performance regression is acceptable?

They seem more than acceptable to me.

These are figures using synchronous logging, right? They seem more than acceptable to me. For asynchronous loggers, the performance regression is probably higher, since the `getCurrentStackTrace()` trick only gives a meaningful result if it is called near to the point, where the exception was generated.

Regardless of the results of asynchronous loggers, I think that the throwable chapter of `2.25.0` is almost over. I think that we only need to:

- [ ] Replace `%xEx` with the much improved `%ex` as default throwable formatter.
- [ ] Remove all the calls to [`LogEvent.getThrownProxy()`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/LogEvent.html#getThrownProxy()) from the codebase. They are useless and expensive.

If users complain about the asynchronous performance of `%xEx`, we could think about introducing an option that collects the packaging information on the calling thread and passes it in the `LogEvent`.

vy:
I am still observing a ~10x performance regression. I use the following crude script:

```java
import org.apache.logging.log4j.core.impl.Log4jLogEvent;
import org.apache.logging.log4j.core.pattern.ExtendedThrowablePatternConverter;

import java.net.Socket;

public class Log4j2Issue3123 {

        final Exception error = createException();
        final LogEvent logEvent = Log4jLogEvent.newBuilder().setThrown(error).build();
        final ExtendedThrowablePatternConverter converter = ExtendedThrowablePatternConverter.newInstance(null, null);
        long startInstantNanos = System.nanoTime();
        final StringBuilder buffer = new StringBuilder();
        for (int i = 0; i < 5_000_000; i++) {
            buffer.setLength(0);
            converter.format(logEvent, buffer);
        }
        double durationSeconds = (System.nanoTime() - startInstantNanos) / 1e9;
        System.out.format("took %.3f s%n", durationSeconds);
        // 2.24.1 - 1.753 s
        // 2.25.0 - 11.281 s
    }

    private static Exception createException() {
        try {
            new Socket("localhost", -1);
            throw new IllegalStateException("should not have reached here");
        } catch (final Exception error) {
            return error;
        }
    }

}
```

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableExtendedStackTraceRenderer.java
vy: Tried to improve the readability in a7daffed7951a8d995a3390017aaa27e7e0b3df7.

vy:
> My benchmark shows that ...

@jengebr, mind sharing the details about your benchmark, please?

jengebr:
Sure - my benchmark is attached.  Be sure to use `-Xmx3g -Xms3g -XX:+UseParallelGC`.

I vary the scenario by changing the classpath to include the desired log4j version.

[log4j_benchmark.zip](https://github.com/user-attachments/files/17599678/log4j_benchmark.zip)

jengebr:
@vy I confirm your results; profiling shows your script is bottlenecked on the `String.format` calls which are no longer called by my benchmark (after the Lazy objects were used).

alan0428a:
@vy @jengebr 
I also can confirm the regression by running vy's script. 
But in my profiling, the `String.format()` still takes a lot of time after we wrap it with Lazy object.
So I replace it with [3fa242a](https://github.com/apache/logging-log4j2/pull/3123/commits/3fa242a1d5fd0e9c1454db7320ebd66a13af5085). This can reduce the time for render part. The total time now reduce to around 9 seconds.

3fa242a_improvement

alan0428a:
And I noticed that compare to `2.24.1` baseline,  the context create part `2.25.0` is much more slower. 
In `2.24.1`, context creation is done in `Log4jLogEvent.getThrownProxy()`. And in `2.25.0`, it's done in `ThrowableExtendedStackTraceRenderer#createContext`. However, the same code takes way more time in `2.25.0`.

2 24 1_baseline
2.24.1

2.25.0 commit 3fa242a

alan0428a:
Oh...I think I know the reason. The test script is actually not fair. The `2.24.1` will store the `ThrowableProxy` in the `Log4jLogEvent`, so context creation only called once. But in the new implementation, the context creation gets called every time.
If we move this line `final LogEvent logEvent = Log4jLogEvent.newBuilder().setThrown(error).build();` into the for-loop in the script. I can get similar CPU time result compare to `2.24.1` (even better). But the memory allocation is 2x.

Screenshot 2024-11-02 at 2 52 07 PM
Screenshot 2024-11-02 at 2 52 42 PM

alan0428a:
I found that [current way](https://github.com/alan0428a/logging-log4j2/blob/3fa242a1d5fd0e9c1454db7320ebd66a13af5085/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableExtendedStackTraceRenderer.java#L185) to load class wastes 4GB more memory, compare to old [approach](https://github.com/alan0428a/logging-log4j2/blob/b55c4d3fa2489cf91d6569aac57b05ae2c9d4b06/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/ThrowableProxyHelper.java#L200)(simple if-else way) according to the profiler. I guess it's the overhead of Stream API

alan0428a:
Current way of using `Throwable` as a key for the metadata map might also be memory expensive. And this impacts to all render I guess.
https://github.com/alan0428a/logging-log4j2/blob/1e154d446fe9d56046281966c60091f0d5d2f48b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableStackTraceRenderer.java#L77

alan0428a:
Also the current way of using `Throwable` as a key for the map in metadata might be memeory expensive?
https://github.com/alan0428a/logging-log4j2/blob/1e154d446fe9d56046281966c60091f0d5d2f48b/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableStackTraceRenderer.java#L77

vy:
> in my profiling, the `String.format()` still takes a lot of time after we wrap it with `Lazy` object.
> ...
> The test script is actually not fair.
> ...
> I found that current way to load class wastes 4GB more memory, compare to old approach (simple if-else way)

Awesome findings @alan0428a! :star_struck:
Thanks so much for helping with the troubleshooting! :bow:
I will incorporate the changes you suggested and re-run the benchmark.

vy:
> Also the current way of using `Throwable` as a key for the map in metadata might be memory expensive?

What do you mean by this? What would be your suggestion instead?

vy:
I can confirm that with the `ClassResourceInfo` and `ThrowableExtendedStackTraceRenderer#loadClass()` fixes @alan0428a tipped, `2.25.0-SNAPSHOT` shows an (almost) identical performance to `2.24.1`. :partying_face: I think the performance regression hunt is over. Let's wrap up this work.

> - Replace %xEx with the much improved %ex as default throwable formatter.
> - Remove all the calls to [LogEvent.getThrownProxy()](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/LogEvent.html#getThrownProxy()) from the codebase. They are useless and expensive.

@ppkarwasz, agreed. I will implement the changes you suggested.

PR #3124, opened 2024-10-27
Corrected incorrect <Rogger> tag to <Logger> in XInclude logger example

sugandhitushar:
The `Logger` tag in XInclude logger example seems incorrect. Corrected the incorrect `Rogger` tag to `Logger`.


ppkarwasz:
@sugandhitushar,

The corrected [corrected example](https://logging.apache.org/log4j/2.x/manual/configuration.html#xinclude) is online, Thanks!

PR #3149, opened 2024-10-31
Prefix stack traces with a newline in Status Logger

vy:
Similar to what we have done for Pattern Layout in #3045, switch stack trace prefix of Status Logger from whitespace to newline.


ppkarwasz (approved):
LGTM

Issue #1416, opened 2023-04-08
Date and time patterns should follow `DateTimeFormatter`


ppkarwasz:
In release 2.x date time patterns follow [`SimpleDateFormat`](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) with additional extensions provided by [`FastDateFormat`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/FastDateFormat.html) and support for the `n` specifier for nanoseconds.

The exception is JSON Template Layout, which uses the format specifiers from [`DateTimeFormatter`](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).

In release 3.x we should switch all format specifiers to those documented `DateTimeFormatter`.

The major differences are:

 * `u` means day-of-week in SDF and era in DTF,
 * SDF lacks `L` meaning month as text,
 * SDF lacks `q` and `Q` as quarter,
 * in DTF `Z`, `ZZ` and `ZZZ` all mean the same format for the time zone (e.g. `-0800`, `+0000`). In SDF (Commons Land version) `ZZ` uses the ISO 8601 format with a colon (e.g. `-08:00`, `Z`) and `ZZZ` adds a colon to the original `Z` format (e.g. `-08:00`, `+00:00`).

vy:
Even though this will be a (slightly?) breaking change, can't we rather implement this on `2.x`? As long as we clearly indicate this in release notes, this feels legit to me. @rgoers, WDYT?

garydgregory:
SDT -> SDF?

ppkarwasz:
Fixed by #3150.

PR #3150, opened 2024-10-31
Port date & time formatter rewrite to `main`

vy:
Ports #3121 to `main`.


log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/LegacyPluginTest.java
```diff
@@ -1,55 +0,0 @@
...
-
-    @Test
-    public void testLegacy(final Configuration configuration) throws Exception {
-        assertThat(configuration, instanceOf(XmlConfiguration.class));
-        for (Map.Entry<String, Appender> entry : configuration.getAppenders().entrySet()) {
-            if (entry.getKey().equalsIgnoreCase("console")) {
-                final Layout layout = entry.getValue().getLayout();
-                assertNotNull("No layout for Console Appender");
-                final String name = layout.getClass().getSimpleName();
-                assertEquals("LogstashLayout", name, "Incorrect Layout class. Expected LogstashLayout, Actual " + name);
```
ppkarwasz, line 43: Just curious, why you removed this.

Out of the top of my head I would say that `LogstashLayout` has no chance to work, since `Layout` has breaking changes. Right?

ppkarwasz (approved):
LGTM

In the future we might expose some time formatting API in `log4j-kit`, but that would not require breaking changes.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/LegacyPluginTest.java
vy: No, the issue was `FixedDateFormat` & `FastDateFormat` is removed, which `LogstashLayout` depends on.

ppkarwasz:
Fixes #1416.

PR #3158, opened 2024-11-03
Improve validation for StringMatchFilter for null/empty text #3153

### LINKED ISSUE #3153: StringMatchFilter would return NPE if configured programmatically with 'null' text.
StringMatchFilter (Log4j 2.24.1)

When parsing from XML I *think* if the "text" attribute is missing it will be populated with an empty string due to the Builder field:

```
@PluginBuilderAttribute
private String text = "";
```

However, if for whatever reason, someone was programmatically creating this StringMatchFilter and passed *null* to "Builder#setMatchString", no validation is performed in the "build()" method or the constructor.

This would lead to a deferred NPE in the `StringMatchFilter#filter` method:

```
private Result filter(final String msg) {
   return msg.contains(this.text) ? onMatch : onMismatch;
}
```

Since String#contains(s) assumes `s` is NotNull.
```
public boolean contains(CharSequence s) {
   return indexOf(s.toString()) >= 0;  // <== NPE!
}
```

I thiink standard behaviour would be to log an error and return null in the 'build()' method if the 'text' field is null.

Also there seems to be a copy/paste error in the `StringMatchFilter.Builder#setMarkerText` javadoc:

```
/**
 * Sets the logging level to use.
 * @param text the logging level to use
 * @return this
 */
public StringMatchFilter.Builder setMatchString(final String text) {
    this.text = text;
    return this;
}
```

JWT007:
This should fix a potential NPE in StringMatchFilter if the 'text' field is not set or set to null/empty string. #3153
* Added '@Required' to @PluginAttribute
* removed default value of ""
* added notEmpty assert to Builder setText
* added notEmpty assert to StringMatchFilter constructor.
* added changelog XML file.

## Checklist

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
@@ -1,0 +1,70 @@
+package org.apache.logging.log4j.core.filter;
+
+import org.apache.logging.log4j.core.Filter;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.test.junit.LoggerContextSource;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+/**
+ * Unit test for {@link StringMatchFilter}.
+ */
+public class StringMatchFilterTest {
+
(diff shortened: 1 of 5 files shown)


ppkarwasz:
@JWT007,

Thank you for your contribution. Can you add some unit tests to the PR?

JWT007:
Hi @ppkarwasz I looked to see if I could extend an existing test for StringMatchFilter but there were no existing unit-tests.

ppkarwasz:
> Hi @ppkarwasz I looked to see if I could extend an existing test for StringMatchFilter but there were no existing unit-tests.

Yes, it happens. :wink: Can you create a `StringMatchFilterTest` in the `log4j-core-test` module and test the two conditions you fixed (`build()` returns `null` if `setMatchString` is not called and `setMatchString` throws if `null` or an empty string is used as paramter).

ppkarwasz (approved):
LGTM, thanks!

PR #3171, opened 2024-11-04
Improve implementations of `LogEvent.toImmutable()` and `ReusableMessage.memento()`

ppkarwasz:
The implementations of `LogEvent.toImmutable()` either use serialization or one of the `Log4jLogEvent.Builder` constructors and custom per class code.

This PR:

- Redirects all implementation of `LogEvent.toImmutable` to the `Builder(LogEvent)` constructor (except `Log4jLogEvent.toImmutable()`)
- Improve the constructor to create really immutable and thread-safe instances.
- Removes the usage of `ThrowableProxy` for purposes not related to serialization.
- Fixes some implementations of `ReusableMessage.memento()`.
- Uses `ReusableMessage.memento()` instead of `ImmutableMessage` where applicable.


log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
```diff
@@ -128,7 +128,9 @@ public <S> void forEachParameter(final ParameterConsumer<S> action, final S stat
 
     @Override
     public Message memento() {
-        return new ObjectMessage(obj);
+        Message message = new ObjectMessage(obj);
+        message.getFormattedMessage();
```
vy, line 132: I presume you make this call to trigger the stack trace capture. Would you mind documenting this, please? (Also in other `message/*Message.java` changes.)

log4j-core-test/src/test/java/org/apache/logging/log4j/core/async/RingBufferLogEventTest.java
```diff
@@ -222,25 +223,31 @@ void testSerializationDeserialization() {
...
-        assertThat(other.getThrown()).isNull();
-        assertThat(other.getThrownProxy()).isEqualTo(new ThrowableProxy(t));
-        assertThat(other.getContextData()).isEqualTo(evt.getContextData());
-        assertThat(other.getContextStack()).isEqualTo(contextStack);
-        assertThat(other.getThreadName()).isEqualTo(threadName);
-        assertThat(other.getSource()).isEqualTo(location);
-        assertThat(other.getTimeMillis()).isEqualTo(12345);
-        assertThat(other.getInstant().getNanoOfMillisecond()).isEqualTo(678);
+        final LogEvent other = SerialUtil.deserialize(SerialUtil.serialize(evt));
+        assertThat(other.getLoggerName()).as("Logger name").isEqualTo(loggerName);
```
vy, line 227: Nit: I wouldn't add to the complexity with extra `as()` calls, since the assertion failure line number precisely points to the associated field.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEventTranslator.java
```diff
@@ -124,6 +129,11 @@ public void setBasicValues(
         this.nanoClock = aNanoClock;
     }
 
+    /**
+     * @deprecated since 2.25.0. {@link RingBufferLogEventTranslator} instances should only be used on the thread that
+     * created it.
+     */
+    @Deprecated
```
vy, line 136: Since we removed `AsyncLogger::initTranslatorThreadValues`, usages of this method will result in undefined behavior. Shouldn't we throw an `UnsupportedOE` in this method?

vy (approved):

log4j-core-test/src/test/java/org/apache/logging/log4j/core/async/RingBufferLogEventTest.java
ppkarwasz: If this test fails in a CI, using `as()` allows to more easily understand what happened, but I agree that this kind of changes are not worth our time and I'll keep your remark in mind for the future.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/RingBufferLogEventTranslator.java
ppkarwasz: I don't believe this method is used outside our codebase, but there are a couple of potential use cases: for example, if `RingBufferLogEventTranslator` instances are stored in a queue instead of being thread-local, calling `updateThreadValues` might be necessary to prepare the translator for use.

I would rather keep the method's implementation as it is now.

log4j-api/src/main/java/org/apache/logging/log4j/message/ReusableObjectMessage.java
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/3171/commits/6812e5e3c4cb7181ebf9dd62f0c17e33ef07af2d

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jLogEvent.java
```diff
@@ -993,7 +1012,14 @@ public static LogEvent createMemento(final LogEvent logEvent) {
...
     public static Log4jLogEvent createMemento(final LogEvent event, final boolean includeLocation) {
-        return deserialize(serialize(event, includeLocation));
+        // In the case `includeLocation` is false, we temporarily disable its computation.
+        if (event.isIncludeLocation() && !includeLocation) {
+            event.setIncludeLocation(false);
+            Log4jLogEvent memento = (Log4jLogEvent) createMemento(event);
+            event.setIncludeLocation(true);
+            return memento;
+        }
+        return (Log4jLogEvent) createMemento(event);
```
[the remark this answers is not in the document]
ppkarwasz: Good point, except `setIncludeLocation` should be called only with `event.isIncludeLocation() && includeLocation` as parameter.

Fixed in https://github.com/apache/logging-log4j2/pull/3171/commits/1455a17787ecad8d63af49c817dc357bae597f1e

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jLogEvent.java
```diff
@@ -814,10 +833,8 @@ public Throwable getThrown() {
      */
     @Override
     public ThrowableProxy getThrownProxy() {
-        if (thrownProxy == null && thrown != null) {
-            thrownProxy = new ThrowableProxy(thrown);
-        }
-        return thrownProxy;
+        // The `thrownProxy` field is non-null only after deserialization.
+        return thrownProxy != null ? thrownProxy : thrown != null ? new ThrowableProxy(thrown) : null;
```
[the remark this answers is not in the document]
ppkarwasz: Readability does not seem to be a problem here.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jLogEvent.java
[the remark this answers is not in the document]
ppkarwasz: The suggestion was reverted in https://github.com/apache/logging-log4j2/pull/3171/commits/db3c0cf55c19c365d7ebc0448a620f111a41ff2d. Since most implementations of `LogEvent` are not thread-safe, race conditions are not a problem.

PR #3199, opened 2024-11-11
Removes weak references from `LoggerRepository`

### LINKED ISSUE #3143: Broken logger initialization in 2.24.1
## Description

LoggerContext / LoggerRegistry don't play well together in using weak references, which can result in a `null` logger if the GC kicks in at a bad timing.

## Configuration

**Version:** 2.24.1

**Operating system:** Linux / macOS

**JDK:** 21.0.4

## Logs

```

Cannot invoke "org.apache.logging.log4j.spi.ExtendedLogger.logIfEnabled(String, org.apache.logging.log4j.Level, org.apache.logging.log4j.Marker, String)" because "this.logger" is null

at org.apache.logging.slf4j.Log4jLogger.warn(Log4jLogger.java:233)
at <internal company code>
```

## Reproduction

- [`LoggerContext`](https://github.com/apache/logging-log4j2/blob/rel/2.24.1/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L533) keeps a local reference to `newLogger`, but this isn't safe to keep a reference (see below).
- [`LoggerRegistry.putIfAbsent`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java#L272) takes the instance and creates a `WeakReference` from it.
- Unfortunate GC timing might result in the weak reference being cleared and the [following `getLogger` call](https://github.com/apache/logging-log4j2/blob/8ee9387d9ec2063ab11f27eaa43e44a13f4c9935/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L535) returning `null`

> Optimizing transformations of a program can be designed that reduce the number of objects that are reachable to be less than those which would naively be considered reachable. For example, a Java compiler or code generator may choose to set a variable or parameter that will no longer be used to null to cause the storage for such an object to be potentially reclaimable sooner. 

https://docs.oracle.com/javase/specs/jls/se21/html/jls-12.html#jls-12.6.1

## Fix

https://github.com/apache/logging-log4j2/pull/2936 should fix this, but is currently unreleased.

[`Reference.reachabilityFence`](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/lang/ref/Reference.html#reachabilityFence(java.lang.Object)) would also fix this, but is JDK 9+ only.

See also: https://shipilev.net/jvm/anatomy-quarks/8-local-var-reachability/

ppkarwasz:
Removes weak references to `Logger`s in `LoggerRepository`. The usage of weak references in `LoggerRepository` might cause `null` to be returned by `LogManager.getLogger()` of all Log4j Core versions up to `2.24.1`. Versions of Log4j API up to `2.24.0` did hold **hard** references to all the registered loggers, so the change will not alter the previous behavior.

This PR also inverts the order of the `String` and `MessageFactory` keys to the `LoggerRepository` multi-map to limit the number of internal maps. The external map is a `WeakHashMap` to allow logger-specific message factories to be GC-ed.

Closes #3143.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -20,9 +20,9 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Optional;
 import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
@@ -39,7 +39,7 @@ import org.jspecify.annotations.Nullable;
 @NullMarked
 public class LoggerRegistry<T extends ExtendedLogger> {
 
-    private final Map<MessageFactory, Map<String, T>> loggerByNameByMessageFactory = new WeakHashMap<>();
+    private final Map<String, Map<MessageFactory, T>> loggerByMessageFactoryByName = new HashMap<>();
 
     private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
@@ -160,12 +160,10 @@ public class LoggerRegistry<T extends ExtendedLogger> {
         requireNonNull(name, "name");
         readLock.lock();
         try {
+            final @Nullable Map<MessageFactory, T> loggerByMessageFactory = loggerByMessageFactoryByName.get(name);
             final MessageFactory effectiveMessageFactory =
                     messageFactory != null ? messageFactory : ParameterizedMessageFactory.INSTANCE;
-            return Optional.of(loggerByNameByMessageFactory)
-                    .map(loggerByNameByMessageFactory -> loggerByNameByMessageFactory.get(effectiveMessageFactory))
-                    .map(loggerByName -> loggerByName.get(name))
-                    .orElse(null);
+            return loggerByMessageFactory == null ? null : loggerByMessageFactory.get(effectiveMessageFactory);
         } finally {
             readLock.unlock();
         }
@@ -179,8 +177,8 @@ public class LoggerRegistry<T extends ExtendedLogger> {
         requireNonNull(destination, "destination");
         readLock.lock();
         try {
-            loggerByNameByMessageFactory.values().stream()
-                    .flatMap(loggerByName -> loggerByName.values().stream())
+            loggerByMessageFactoryByName.values().stream()
+                    .flatMap(loggerByMessageFactory -> loggerByMessageFactory.values().stream())
                     .forEach(destination::add);
         } finally {
             readLock.unlock();
@@ -203,7 +201,8 @@ public class LoggerRegistry<T extends ExtendedLogger> {
     @Deprecated
     public boolean hasLogger(final String name) {
         requireNonNull(name, "name");
-        return getLogger(name) != null;
+        final @Nullable T logger = getLogger(name);
+        return logger != null;
     }
 
     /**
@@ -221,7 +220,8 @@ public class LoggerRegistry<T extends ExtendedLogger> {
      */
     public boolean hasLogger(final String name, @Nullable final MessageFactory messageFactory) {
         requireNonNull(name, "name");
-        return getLogger(name, messageFactory) != null;
+        final @Nullable T logger = getLogger(name, messageFactory);
+        return logger != null;
     }
 
     /**
@@ -237,9 +237,8 @@ public class LoggerRegistry<T extends ExtendedLogger> {
         requireNonNull(messageFactoryClass, "messageFactoryClass");
         readLock.lock();
         try {
-            return loggerByNameByMessageFactory.entrySet().stream()
-                    .filter(entry -> messageFactoryClass.equals(entry.getKey().getClass()))
-                    .anyMatch(entry -> entry.getValue().containsKey(name));
+            return loggerByMessageFactoryByName.getOrDefault(name, Collections.emptyMap()).keySet().stream()
+                    .anyMatch(messageFactory -> messageFactoryClass.equals(messageFactory.getClass()));
         } finally {
             readLock.unlock();
         }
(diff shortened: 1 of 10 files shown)


ppkarwasz:
The weak reference to `MessageFactory` might require additional explanations. Unlike a weak reference to `Logger`, it will not cause `null` to be returned by `LoggerContext#getLogger`, since the method keeps a hard reference to the factory:

https://github.com/apache/logging-log4j2/blob/8ee9387d9ec2063ab11f27eaa43e44a13f4c9935/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L526-L536

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
```diff
@@ -43,7 +41,7 @@
 @NullMarked
 public class LoggerRegistry<T extends ExtendedLogger> {
 
-    private final Map<String, Map<MessageFactory, WeakReference<T>>> loggerRefByMessageFactoryByName = new HashMap<>();
+    private final Map<MessageFactory, Map<String, T>> loggerByNameByMessageFactory = new WeakHashMap<>();
```
vy, line 44: Doesn't this change imply that a `Logger` will never be freed unless the associated `MF` becomes unreachable?

ppkarwasz: The entire `LoggerRegistry` will be freed, when the associated `LoggerContext` becomes unavailable.

The three lines:

```java
     logger = newInstance(this, name, messageFactory);
     loggerRegistry.putIfAbsent(name, effectiveMessageFactory, newLogger); 
     return loggerRegistry.getLogger(name, effectiveMessageFactory); 
```
have been in Log4j Core since 2016. These lines should not return `null`.

vy: Why aren't we instead trying to fix the existing code, which allows `Logger`s to be freed when they become unreachable?

ppkarwasz: The current code is fixed:

https://github.com/apache/logging-log4j2/blob/89c9b9f81cb12a6a217788a71832bfbd3b6d07b2/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L527-L531

The changes in this PR are mainly there to allow `log4j-api` version `2.24.2/2.25.0` to work with any `log4j-core` version `2.24.x`.

ppkarwasz:
In order to preserve backward compatibility and allow `Logger`s to be garbage-collected, I will:

- [ ] Copy `LoggerContext` to an internal package of `log4j-core`.
- [ ] Restore the usage of `WeakReference<T>` in the copied class.
- [ ] Remove the new methods in `LoggerRepository` in the original class.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: This PR, instead of fixing _"the immediate problem"_, rewrites the `LoggerRegistry` and reverts the behavior to scope the lifetime of `Logger`s with the associated `MF`'s lifetime. I prefer to preserve the old (yet buggy) behavior that `Logger`s will be GC'ed when they become unreachable. IMO, _the fix_ is two lines of code: adding `Reference.reachabilityFence(logger)` before `return logger` in `putIfAbsent()` and `computeIfAbsent()`. Though `Reference.reachabilityFence()` is introduced in Java 9. We can try imitating it with a usage of the value to force a strong reference (e.g., `logger.log(Level.OFF, "nada")`), but VM can get this inlined/eliminated, and we are back to square one. Anyway, I am looking at this. I will see what else can be done.

ppkarwasz:
> > In order to preserve backward compatibility and allow `Logger`s to be garbage-collected, I will:
> > 
> > * [ ]   Remove the new methods in `LoggerRepository` in the original class.
> 
> How will you implement this while preserving the backward compatibility?

`LoggerRepository#computeIfAbsent` is a new method added in `2.25.0`. There is no BC to preserve.

vy:
> `LoggerRepository#computeIfAbsent` is a new method added in `2.25.0`. There is no BC to preserve.

Isn't it released with `2.24.1`?

image

ppkarwasz:
> > `LoggerRepository#computeIfAbsent` is a new method added in `2.25.0`. There is no BC to preserve.
> 
> Isn't it released with `2.24.1`?

No, the backport of #2936 for `2.24.1` is in PR #2961, which did not introduce any API changes.

vy:
@ppkarwasz, in 8963f705886f2b6397c03e8787ed35dd920b2736, we removed 110 LoC and added 240 LoC, voiding the entire point of `computeIfAbsent()`. What is the rationale for this decision?

garydgregory:
A complex PR without tests? Isn't it possible to test this issue to avoid regressions in the future?

pjfanning:
Would it be possible to optionally support not having weak references? I understand why weak references are useful for edge cases like users running with dangerously low heap for their applications. There are also users who have applications that don't need weak references here and would appreciate not risking having NPEs.

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: > Currenlty the `Logger` is GC-ed between the call to `putIfAbsent()` and the call to `getLogger()`. Unless I am mistaken, `reachabilityFence` will not help with that.

No, it will. GC kicks in _after_ the `return logger`, i.e., _outside the scope_ where the reference returned by `WeakReference#get()` is assigned to a variable. Creating a strong reference to `logger` again within this scope will solve the problem – I just don't know of a stable way to this without using the reachability fence. This is in detailed explained [here](https://shipilev.net/jvm/anatomy-quarks/8-local-var-reachability/). You can even verify this behavior by creating a `private static final Set<Logger> LOGGERS` and calling `LOGGERS.add(logger)` just before every `return logger` statement.

> Therefore the refactoring from `Map<String, Map<MessageFactory, WeakReference<T>>>` to `Map<String, Map<MessageFactory, T>>` is required.

_Required_? I am not sure. As I stated, I am checking this. Could you give me a day, please?

> The refactoring from `Map<String, Map<MessageFactory, T>>` to `Map<MessageFactory, Map<String, T>>` is indeed not needed. Do you object to this change?

If this is indeed the only way to fix the problem, I don't object it. Please give me a day.

vy:
@pjfanning, making loggers garbage-collectible is not only a concern for environments with constrained memory resources. This will also greatly benefit users creating loggers programmatically.

ppkarwasz:
> A complex PR without tests? Isn't it possible to test this issue to avoid regressions in the future?

Fixed in https://github.com/apache/logging-log4j2/pull/3199/commits/041e5ebdfb77fc0fc4e5ee136d0d9a73fe5f6919

log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
vy: I'm not happy about this change, yet I couldn't come up with a better way in this limited amount of time. I leave the judgement to the rest of the maintainers.

ppkarwasz:
> There are also users who have applications that don't need weak references here and would appreciate not risking having NPEs.

There should be no such problem. The only place were weak references are used in this PR is the `WeakHashMap<MessageFactory, T extends ExtendedLogger>` used in `LoggerRegistry`.

Loggers _could_ become unreachable if the message factory becomes unreachable. However the message factory never ceases to be reachable. In a typical `getLogger()` call, the message factory is last used on line 166 of `LoggerRegistry`:

https://github.com/apache/logging-log4j2/blob/8ee9387d9ec2063ab11f27eaa43e44a13f4c9935/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L533-L535

https://github.com/apache/logging-log4j2/blob/bf79552c902d6638415c78441ddf46771ad5a27b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java#L159-L170

According to the [blog post shared by Volkan](https://shipilev.net/jvm/anatomy-quarks/8-local-var-reachability/) this means that the message factory will be reachable up to line 166. However, that line returns a `Logger` that has a reference to the message factory, so the message factory becomes reachable again. Worst case scenario, we might loose the entry in the `WeakHashMap`, but not the `Logger`.

ppkarwasz:
@vy,

Do you have any comments regarding [`InternalLoggerRepository`](https://github.com/apache/logging-log4j2/blob/bf79552c902d6638415c78441ddf46771ad5a27b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java)? The class should be similar to the current [`LoggerRegistry`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java) with two significant changes:

- All unused methods are removed.
- The order of the keys in inverted: the external map uses `MessageFactory`, since there are less message factories than logger names, which reduces the number of internal maps.

PR #3210, opened 2024-11-15
Fix key removal issues in Thread Context (`2.24.x` branch)

ppkarwasz:
This backports #3050 to the `2.24.x` branch.


vy (approved):
Assuming we didn't mess up something while cherry-pick'ing.

PR #3217, opened 2024-11-19
 Feat(SystemClock): Added PreciseClock implementation to the system clock class, for the benefit of Java 8 users.

jaykataria1111:
Straight forward change to override the behavior of System Clock, to implement the Precise Clock interface, for the benefit of Java 8 users. The PR is for issue #3198

## Checklist


ppkarwasz:
The build failure is due to the `2.24.2` release being published (BND checks your code against `2.24.2` now, instead of `2.24.1` before).

Please, rebase.

jvz:
Code changes look good to me. Can you add a changelog entry for this? You can copy `src/changelog/.2.x.x /3121_instant_format.xml` as an example of a `changed` entry.

jaykataria1111:
Hi @jvz , @ppkarwasz  was on vacation let me do it upcoming week.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/SystemClock.java
```diff
@@ -16,10 +16,15 @@
  */
 package org.apache.logging.log4j.core.util;
 
+import java.time.Instant;
+import org.apache.logging.log4j.core.time.MutableInstant;
+import org.apache.logging.log4j.core.time.PreciseClock;
+
 /**
  * Implementation of the {@code Clock} interface that returns the system time.
+ * @since 2.25
```
ppkarwasz, line 25: The class itself is older, so we should remove the `@since`.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/SystemClock.java
```diff
@@ -29,4 +34,13 @@ public final class SystemClock implements Clock {
     public long currentTimeMillis() {
         return System.currentTimeMillis();
     }
+
+    /**
+     * {@inheritDoc}
+     */
```
ppkarwasz, line 40: Personally I find `{@inheritDoc}` a little bit useless, since IDEs will look at the Javadoc of the parent class/implemented interface anyway, if the Javadoc is missing.

ppkarwasz (changes requested):
@jaykataria1111,

Since the Java 8 and Java 9+ versions of the `SystemClock` class are identical now, can you remove the Java 9+ version (in `log4j-core-java9`). The `log4j-core-java9` package has a lot of dummy classes, whose purpose is to make the module compilable, they are not packaged. Can you remove those dummy classes too?

jaykataria1111:
@ppkarwasz 

Great opportunity for me to learn here: 

>The log4j-core-java9 package has a lot of dummy classes, whose purpose is to make the module compilable, they are not packaged. Can you remove those dummy classes too

1. What does it mean to make the module "compatible"? 
2. they are not packaged - How do I identify the classes that are packaged. 
3. How do I identify these dummy classes?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/SystemClock.java
jaykataria1111: Sounds good!

log4j-core/src/main/java/org/apache/logging/log4j/core/util/SystemClock.java
jaykataria1111: Sounds good, will remove this too!

ppkarwasz:
@jaykataria1111 ,

> > The log4j-core-java9 package has a lot of dummy classes, whose purpose is to make the module compilable, they are not packaged. Can you remove those dummy classes too
> 
>     1. What does it mean to make the module "compatible"?

To make it "compilable".

The `log4j-core-java9` module is compiled **before** `log4j-core` and has no access to the classes of `log4j-core`.
In order to compile a class like [`SystemClock`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core-java9/src/main/java/org/apache/logging/log4j/core/util/SystemClock.java) we had to add to `log4j-core-java9` also the classes it depends upon, e.g., those in the [`core.time` directory](https://github.com/apache/logging-log4j2/tree/2.x/log4j-core-java9/src/main/java/org/apache/logging/log4j/core/time).

These are the classes that I call "dummy" classes.

>     2. they are not packaged - How do I identify the classes that are packaged.

We use the Maven Assembly plugin to package the classes in `log4j-core-java9`. There is [an explicit list of classes to package in the `java9.xml` descriptor](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core-java9/src/assembly/java9.xml).

>     3. How do I identify these dummy classes?

They should have a comment near the top that says that they are "dummy" or something equivalent.

jaykataria1111:
Thanks @ppkarwasz , for the answers, let me make a PR soon :) finally got back from vacation :)

jaykataria1111:
Hi @ppkarwas, finally got time to upload a revision, please check whenever you get time! 🙂

ppkarwasz (approved):
LGTM

PR #3219, opened 2024-11-19
Migrate `log4j-osgi-test` to JUnit 5

AlbaHerrerias:
Hello 👋

We are from Neighbourhoodie, the implementation partner of the STF Bug Resilience Program. This work is part of our agreed Milestone 1. Upgrade from JUnit 4 to JUnit 5. This PR migrates the tests located in `log4j-osgi-test`.

Thank you!

## Checklist

### DIFF
--- a/log4j-osgi-test/src/test/java/org/apache/logging/log4j/osgi/tests/AbstractLoadBundleTest.java
+++ b/log4j-osgi-test/src/test/java/org/apache/logging/log4j/osgi/tests/AbstractLoadBundleTest.java
@@ -16,18 +16,18 @@
  */
 package org.apache.logging.log4j.osgi.tests;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.lang.reflect.Method;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


AlbaHerrerias:
We have some questions we would like to ask. We've found that the test suites `CoreOsgiTest`, `JULProviderTest` and `SLF4JProviderTest` are using `PaxExam`, and the current version does not support JUnit 5. Before attempting an upgrade, we were wondering, does this feature need to be updated? We are missing some context, what is being checked by using this runner? Any help in this regard is greatly appreciated. Thank you!

ppkarwasz:
> Before attempting an upgrade, we were wondering, does this feature need to be updated? We are missing some context, what is being checked by using this runner? Any help in this regard is greatly appreciated. Thank you!

Since PAX Exam is basically dormant, it would be better to move those tests to a different framework, but honestly I don't know which. @HannesWell, any suggestions?

Frameworks like [Testcontainer for Java](https://java.testcontainers.org/) or [Arquillian](https://arquillian.org/) are not very OSGi friendly. The "new" [`osgi-test`](https://github.com/eclipse-osgi-technology/osgi-test) assumes that the test is already running in an OSGi environment. Maybe we could use something like [laeubisoft/osgi-test-framework](https://github.com/laeubisoft/osgi-test-framework) or the [BND Testing Plugin](https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-testing-maven-plugin/README.md). I will experiment with the latter. We already use BND plugins and they work well. I would like to keep the list of Open Source vendors we rely on minimal.

ppkarwasz:
@vy, should we move `log4j-osgi-test` to the Samples repository (which is our integration test repo).

jvz (approved):

Issue #3110, opened 2024-10-18
Upgrading to 2.24.1 produces a compile time warning: BaselineIgnore not found


lbergelson:
## Description
We're trying to update from 2.17.1 ->  2.24.1 but we've encountered compile warnings due to missing annotations.  It looks like essentially the same same issue as https://github.com/apache/logging-log4j2/issues/2144 and https://github.com/apache/logging-log4j2/issues/2232

## Configuration

**Version:**.  2.24.1
**Operating system:**   OSX Sonoma 14.4
**JDK:**
openjdk 17.0.12 2024-07-16
OpenJDK Runtime Environment Homebrew (build 17.0.12+0)
OpenJDK 64-Bit Server VM Homebrew (build 17.0.12+0, mixed mode, sharing)

## Logs
>/gradlew compileTestjava

> Task :compileJava
/Users/louisb/.gradle/caches/modules-2/files-2.1/org.apache.logging.log4j/log4j-api/2.24.1/7ebeb12c20606373005af4232cd0ecca72613dda/log4j-api-2.24.1.jar(/org/apache/logging/log4j/Level.class): warning: Cannot find annotation method 'value()' in type 'BaselineIgnore': class file for aQute.bnd.annotation.baseline.BaselineIgnore not found

## Reproduction

Update log4j to 2.24.1 in our build.gradle 
```
    platform('org.apache.logging.log4j:log4j-bom:2.24.1)
    implementation 'org.apache.logging.log4j:log4j-api'
    implementation 'org.apache.logging.log4j:log4j-core'
```

This was very surprising to me.  Have you considered changing the annotation dependencies to be included.  Or they processed out of the published jar?

We need to include `log4j-core` at compile time, because we want to change the default logging level at runtime based on command line input from the user.  If there is a way to do this through the API we would be happy to switch.

I've solved the warning by adding 2 additional dependencies
```
    implementation 'biz.aQute.bnd:biz.aQute.bnd.annotation'
    implementation 'org.osgi:org.osgi.annotation.bundle'
```

Just adding `biz.aQute.bnd:biz.aQute.bnd.annotation` causes additional failures because it includes OSGI bundle annotations which are not included in it's dependency list.  

I feel like I'm doing something wrong but I'm not really sure what I should be doing instead.

ppkarwasz:
Hi @lbergelson,

[...]

PR #3228, opened 2024-11-22
Move JSpecify from `provided` to `compile` scope

ppkarwasz:
[JSpecify](https://jspecify.dev/) is a nullability annotation library with a [large list of supporters](https://jspecify.dev/about/). It is the recommended library to solve [LOG4J2-1477](https://issues.apache.org/jira/browse/LOG4J2-1477).

The usage of JSpecify as `provided` library has however a drawback: if users don't have JSpecify on their stack, they'll get a Javac Linter warning whenever they use a class with nullability annotations (see [#3110](https://github.com/apache/logging-log4j2/issues/3110) for example). Since JSpecify is an annotation with `RUNTIME` retention and is not covered by [JDK-8342833](https://bugs.openjdk.org/browse/JDK-8342833), the easiest solution provided by this PR is:

- Move JSpecify from the `provided` to the `compile` scope. This will inflate users dependency stack by a whooping 3819 bytes.
- Mark JSpecify as optional for OSGi and JPMS, so users can exclude the dependency if they wish to do it.

In practice, this change should be neutral for users and will allow us to add nullability annotations to important class, such as `Logger` without breaking the build of users that use the [`-Werror` compiler flag](https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-Werror).

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,30 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to you under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+##
+# This file controls the "Sponsor" button in this repo.
+# For details see:
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/displaying-a-sponsor-button-in-your-repository
+#
+# This list includes in the same order, committers of Log4j listed on:
+# https://logging.apache.org/support.html#sponsors
+github:
+  - carterkozak
+  - garydgregory
+  - jvz
+  - ppkarwasz
+  - rgoers
+  - vy
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/Hierarchy.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/Hierarchy.java
@@ -16,6 +16,7 @@
  */
 package org.apache.log4j;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.Enumeration;
 import java.util.Hashtable;
 import java.util.Vector;
@@ -79,10 +80,16 @@ public class Hierarchy implements LoggerRepository, RendererSupport, ThrowableRe
     private static class PrivateLogManager extends org.apache.logging.log4j.LogManager {
         private static final String FQCN = Hierarchy.class.getName();
 
+        @SuppressFBWarnings(
+                value = "HSM_HIDING_METHOD",
+                justification = "The class is private, no confusion can arise.")
         public static LoggerContext getContext() {
             return getContext(FQCN, false);
         }
 
+        @SuppressFBWarnings(
+                value = "HSM_HIDING_METHOD",
+                justification = "The class is private, no confusion can arise.")
         public static org.apache.logging.log4j.Logger getLogger(final String name) {
             return getLogger(FQCN, name);
         }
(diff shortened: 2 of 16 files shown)


log4j-api/pom.xml
```diff
@@ -63,16 +63,19 @@
...
 
+    <!--
+      ~ Effectively optional, but included due to its size and the compilation warnings its absence causes.
+      -->
     <dependency>
-      <groupId>org.osgi</groupId>
-      <artifactId>org.osgi.core</artifactId>
-      <scope>provided</scope>
+      <groupId>org.jspecify</groupId>
+      <artifactId>jspecify</artifactId>
```
rgoers, line 76: -1

We have always had a rule that Log4j-API will have no required dependencies.

rgoers (changes requested):

log4j-api/pom.xml
ppkarwasz: Yes, I know this is against current policy, but can that policy be reevaluated?

The rationale is that keeping nullability annotations in the artifacts is mostly useful to `log4j-api` users. Log4j Core calls usually don't appear in user code, but Log4j API classes are more used.

While the implementation can certainly do null-checks, we might mark the `format` parameters of log calls as `@NonNull`: it doesn't make sense to log something, when the format string is `null`. Also the `Level` parameter should be marked `@NonNull` as well as many of the parameters and return types of `ThreadContext`.

rgoers:
Yes, I am aware of all these arguments as they were raised when the dependency was first added. I made the same objection then but was assured that by using provided scope users would have no problems if the dependency is missing. If that really isn’t the case then the dependency should be removed and a different solution should be found. Note that even using Jakarta.annotations would be a problem as we also have a policy to limit the dependencies to Java.base in JPMS, which is work I believe you worked on as part of 3.0.

ppkarwasz:
> I made the same objection then but was assured that by using provided scope users would have no problems if the dependency is missing. If that really isn’t the case then the dependency should be removed and a different solution should be found.

I don't consider a compiler warning (if the [`-Xlint:all`](https://docs.oracle.com/en/java/javase/21/docs/specs/man/javac.html#option-Xlint-custom) linter option is enabled) to be a real problem. Besides, users can always disable the [`classfile`] category in the linter: it contains only two warnings (see [this comment](https://github.com/apache/logging-log4j2/issues/3110#issuecomment-2423586754)) and both of them are pretty harmless.

If I understand correctly, your `-1` only applies to the additional dependency in `log4j-api`, while you are OK with adding it to the other artifacts?

rgoers:
Yes, the -1 is for the API only. While we have always tried to minimize the dependencies on Log4j-core that has never been as strict.  I believe we made some effort in 3.0 to remove many dependencies but I don't believe we got rid of them all. You would probably know that better than me since you have been actively working on it recently.

ppkarwasz:
> I believe we made some effort in 3.0 to remove many dependencies but I don't believe we got rid of them all.

Version `3.0.0-beta3` of Log4j Core has **no** external dependencies (see [MvnRepository](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-core/3.0.0-beta3)). Most importantly, it has no **optional** dependencies, which are always a problem. Even if we exclude JPMS and OSGi, normal Java users pin their optional Log4j Core dependency in the POM file and they forget to upgrade them, when they upgrade Log4j Core. This will no longer be an issue.

log4j-api/pom.xml
ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/3228/commits/a86d7ca480400de309623c22c7549da14f0c8008

ppkarwasz:
@rgoers,

I moved `jspecify` into the `provided` scope for `log4j-api`. Can you review this PR again?

**Note**: The build is expected to fail until #3601 is fixed. I opened #3602 that awaits a review.

ppkarwasz:
> will this have any [backward compatibility] implications on the fat JAR users?

I am not sure what implications are you thinking about. Yes, their fat JARs will be a couple of KiB bigger.

**Note**: I didn't replace the JPMS `requires static org.jspecify` with `requires org.jspecify`, so if JPMS or OSGi user don't want it, they can exclude it. Fat JAR users of course are not JPMS users.

vy:
> I am not sure what implications are you thinking about. Yes, their fat JARs will be a couple of KiB bigger.

No, what I mean is, there are users creating Fat JARs with an explicit list of dependencies. With the next minor Log4j 2 version, they will be missing a `compile`-scoped dependency of `log4j-core`. Will that have any implications on such users?

ppkarwasz:
> No, what I mean is, there are users creating Fat JARs with an explicit list of dependencies. With the next minor Log4j 2 version, they will be missing a `compile`-scoped dependency of `log4j-core`. Will that have any implications on such users?

The annotation is only used at compile time. It can be absent at runtime.

PR #3235, opened 2024-11-24
Fix `ConcurrentModificationException` in `updateLoggers()`

### LINKED ISSUE #3234: ConcurrentModificationException in LoggerContext due to race condition
## Description

Presumably due to https://github.com/apache/logging-log4j2/commit/a1dfa853c96604ded1f253da2ec25f8f2f95ef25 my application throws an error when loggers get changed from multiple threads.
This is likely because `InternalLoggerRegistry#getLoggers` does not return a thread-safe stream.
Streams are by nature evaluated on the fly, thus still using the values of `loggerRefByNameByMessageFactory` after `readLock` was unlocked.
When multiple threads modify `loggerRefByNameByMessageFactory`, sometimes it errors with the attached error.
This happens inconsistently because it's a race condition, but I've been able to get this error a few times.

## Configuration

**Version:** 2.24.2

**Operating system:** Windows

**JDK:** openjdk version "21.0.4" 2024-07-16 LTS

## Logs

```
java.util.ConcurrentModificationException: null
	at java.base/java.util.HashMap$ValueSpliterator.forEachRemaining(HashMap.java:1792) ~[?:?]
	at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762) ~[?:?]
	at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276) ~[?:?]
	at java.base/java.util.WeakHashMap$ValueSpliterator.forEachRemaining(WeakHashMap.java:1223) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
	at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151) ~[?:?]
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174) ~[?:?]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
	at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596) ~[?:?]
	at org.apache.logging.log4j.core.LoggerContext.updateLoggers(LoggerContext.java:776) ~[?:?]
	at org.apache.logging.log4j.core.LoggerContext.updateLoggers(LoggerContext.java:766) ~[?:?]
	at org.apache.logging.log4j.core.config.Configurator.setLevel(Configurator.java:379) ~[?:?]
	at org.apache.logging.log4j.core.config.Configurator.setLevel(Configurator.java:414) ~[?:?]
	at com.soulfiremc.server.SoulFireServer.setupLogging(SoulFireServer.java:213) ~[?:?]
	at com.soulfiremc.server.SoulFireServer.setupLoggingAndVia(SoulFireServer.java:208) ~[?:?]
	at com.soulfiremc.server.InstanceManager.start(InstanceManager.java:194) ~[?:?]
	at com.soulfiremc.server.SoulFireScheduler.lambda$wrapFuture$6(SoulFireScheduler.java:140) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1804) ~[?:?]
	at com.soulfiremc.server.SoulFireScheduler.runCommand(SoulFireScheduler.java:154) ~[?:?]
	at com.soulfiremc.server.SoulFireScheduler.lambda$schedule$2(SoulFireScheduler.java:79) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinTask$RunnableExecuteAction.exec(ForkJoinTask.java:1423) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:387) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1312) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1843) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1808) ~[?:?]
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:188) ~[?:?]
```

This shows the stream being collected due to the forEach call in `LoggerContext#updateLoggers`, but there being no read lock, therefore causing this race condition.

## Reproduction

I unfortunately cannot provide an example because I do not have much experience with reproducing race conditions.

ppkarwasz:
The `InternalLoggerRegistry` implementation introduced in version `2.24.2` did not return a copy of the registry, when `InternalLoggerRegistry.getLoggers()` was called. This could lead to a `ConcurrentModificationException` if a thread creates a new logger, while another thread calls `LoggerContext.updateLoggers()`.

Closes #3234

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
@@ -33,7 +33,6 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.stream.Collectors;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.ConfigurationFactory;
@@ -513,7 +512,7 @@ public class LoggerContext extends AbstractLifeCycle
      * @return a collection of the current loggers.
      */
     public Collection<Logger> getLoggers() {
-        return loggerRegistry.getLoggers().collect(Collectors.toList());
+        return loggerRegistry.getLoggers();
     }
 
     /**
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/LoggerRegistry.java
@@ -18,7 +18,6 @@ package org.apache.logging.log4j.spi;
 
 import static java.util.Objects.requireNonNull;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -28,6 +27,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.message.MessageFactory;
 import org.apache.logging.log4j.message.ParameterizedMessageFactory;
@@ -158,19 +158,19 @@ public class LoggerRegistry<T extends ExtendedLogger> {
     }
 
     public Collection<T> getLoggers() {
-        return getLoggers(new ArrayList<>());
+        readLock.lock();
+        try {
+            return loggerByMessageFactoryByName.values().stream()
+                    .flatMap(loggerByMessageFactory -> loggerByMessageFactory.values().stream())
+                    .collect(Collectors.toList());
+        } finally {
+            readLock.unlock();
+        }
     }
 
     public Collection<T> getLoggers(final Collection<T> destination) {
         requireNonNull(destination, "destination");
-        readLock.lock();
-        try {
-            loggerByMessageFactoryByName.values().stream()
-                    .flatMap(loggerByMessageFactory -> loggerByMessageFactory.values().stream())
-                    .forEach(destination::add);
-        } finally {
-            readLock.unlock();
-        }
+        destination.addAll(getLoggers());
         return destination;
     }
 
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerContextTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerContextTest.java
@@ -19,12 +19,22 @@ package org.apache.logging.log4j.core;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.mock;
 
+import java.util.Collection;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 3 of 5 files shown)


vy (approved):

garydgregory:
Hi all,

Does this mean a 2.24.3 is around the corner?

ppkarwasz:
> Does this mean a 2.24.3 is around the corner?

@garydgregory, I would wait a couple of weeks to see if other bugs are reported, but, yes, I think we should release a `2.24.3`.

We can also release `2.25.0`, but that will bring other problems and I have some long term plans for `2.24.x`.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -513,7 +512,7 @@ public Logger getLogger(final String name) {
      * @return a collection of the current loggers.
      */
     public Collection<Logger> getLoggers() {
-        return loggerRegistry.getLoggers().collect(Collectors.toList());
+        return loggerRegistry.getLoggers();
```
vy, line 515: Shouldn't we rather keep the old defensive copy around to avoid similar problems from integrators?

ppkarwasz: `Collectors.toList()` does make a copy, right?

I could add a test for that in case the implementation of `InternalLoggerRegistry.getLoggers()` changes

vy: I'm not concerned about `ILR`, I think it is covered, but integrators providing their own `LR` and such. Anyway... Checked the code and this line used to be `return loggerRegistry.getLoggers()`. :+1:

ppkarwasz:
@tnleeuw,

I removed the usage of `forEach()` in https://github.com/apache/logging-log4j2/pull/3235/commits/4fa2d548c9f704db37e3a6f0a50e06e3d2958d8a. Could you review the changes?

tnleeuw:
> @tnleeuw,
> 
> I removed the usage of `forEach()` in [4fa2d54](https://github.com/apache/logging-log4j2/commit/4fa2d548c9f704db37e3a6f0a50e06e3d2958d8a). Could you review the changes?

Looks good to me!

kelunik:
> @garydgregory, I would wait a couple of weeks to see if other bugs are reported, but, yes, I think we should release a 2.24.3.

Why? Can't you just release 2.24.4 if another bug is found?

ppkarwasz:
> > @garydgregory, I would wait a couple of weeks to see if other bugs are reported, but, yes, I think we should release a 2.24.3.
> 
> Why? Can't you just release 2.24.4 if another bug is found?

An Apache release process requires at least three maintainers and has a 72 hours voting period. Unless the reported problem affects most users, my personal policy is to wait a week or two to give users the time to report other problems.

I will start the release process for `2.24.3` tomorrow, which will contain a fix to this bug and the one reported in #3252 (a strange interaction between Spring Boot, ActiveMQ and Log4j API).

kelunik:
Thanks for providing these details!

PR #3258, opened 2024-11-30
Add `log4j.plugin.enableBndAnnotations` option to `PluginProcessor`

### LINKED ISSUE #3251: Generate `Log4jPlugins` class does not compile in JPMS
Currently `log4j-plugin-processor` in version `3.0.0-beta3` of Log4j Core adds [BND annotations](https://bnd.bndtools.org/chapters/230-manifest-annotations.html) to the generated code.

While these might be useful for OSGi users, they cause a compilation error if the `biz.aQute.bnd.annotation` module is not present:

- On the compiler classpath,
- A `static` dependency on the module is not declared in the `module-info.java` file.

ppkarwasz:
This adds a `log4j.plugin.enableBndAnnotations` option to the `PluginProcessor`. Its default value is inferred from the compiler classpath.

We also rename the `pluginPackage` option to a more coherent `log4j.plugin.package` option.

Fixes #3251


log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
```diff
@@ -282,6 +315,25 @@ private String commonPrefix(final String str1, final String str2) {
         return str1.substring(0, minLength);
     }
 
+    private static boolean isServiceConsumerClassPresent() {
+        try {
+            Class.forName("aQute.bnd.annotation.spi.ServiceConsumer");
+            return true;
```
ppkarwasz, line 321: This looks for `ServiceConsumer` on the plugin processor's classpath, while it should look for it on the application classpath.

vy:
@ppkarwasz, why do we need a toggle property instead of disabling it, always? This very much smells like leaking an internal detail through configuration properties.

ppkarwasz:
> @ppkarwasz, why do we need a toggle property instead of disabling it, always? This very much smells like leaking an internal detail through configuration properties.

It is an option that controls the output of the `PluginProcessor`, so I would not call it an internal detail. Should we disable BND annotations by default (for users) and keep the option for our own builds (the BND annotations are useful to generate the necessary `META-INF/services` files and similar)?

vy:
> Should we disable BND annotations by default (for users) and keep the option for our own builds (the BND annotations are useful to generate the necessary `META-INF/services` files and similar)?

Could you elaborate on how we can implement this? Does this also mean, there will be no need for a toggle property?

ppkarwasz:
> > Should we disable BND annotations by default (for users) and keep the option for our own builds (the BND annotations are useful to generate the necessary `META-INF/services` files and similar)?
> 
> Could you elaborate on how we can implement this? Does this also mean, there will be no need for a toggle property?

What I mean is, I just replace:

https://github.com/apache/logging-log4j2/blob/e6e6dddb41502edcbb89f8d23800449c142279f9/log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java#L318-L325

with `return false` and we add `-Alog4j.plugin.enabledBndAnnotations=true` to our builds. BND runs after `compile` and it registers the `PluginService` in the correct files.

src/changelog/.3.x.x/3151_plugin_processor_bnd_annotations.xml
vy, line 1: Since this is an internal change and not intended for public consumption, I am inclined to avoid a change log entry. WDYT?

ppkarwasz: In the current status, it will have external effects.

Users with BND Annotations on the "normal" classpath will get a different class that those without it.

vy (approved):

PR #3259, opened 2024-11-30
Add requiresLocation function for FailoverAppender #3257

### LINKED ISSUE #3257: File line numbers are not printed for the failover appender when using AsyncLoggerContextSelector
## Description

I was using a failover appender with an AsyncLoggerContextSelector and wanted to print the line number. I set includeLocation to true in the root logger, but there were still no line numbers in the logs. We encountered the same problem with the async appender, even with includeLocation set to true.

I believe this is because the requiresLocation function in FailoverAppender is the default implementation in AbstractAppender, which relies on the layout inside the appender. However, for FailoverAppender, we do not have a layout, and thus fail to generate the location field in the log event object. Additionally, the Async logger uses a RingBufferLogEvent, which directly reads the location field, so there's no location to print. This issue is also true for AsyncAppender.

## Configuration

**Version:** 2.24.2

**Operating system:** Windows11

**JDK:** OpenJDK23

## Reproduction

Log4j2.xml
```
<Configuration>
    <Appenders>
        <Console name="CONSOLE">
            <PatternLayout pattern="%m%L%n"/>
        </Console>

        <Console name="CONSOLE2">
            <PatternLayout pattern="%m%L%n"/>
        </Console>

        <Failover name="Failover" primary="CONSOLE">
            <Failovers>
                <AppenderRef ref="CONSOLE2"/>
            </Failovers>
        </Failover>
    </Appenders>
    <Loggers>
        <Root level="INFO" includeLocation="true">
            <AppenderRef ref="Failover"/>
        </Root>
    </Loggers>
</Configuration>
```

log4j2.component.properties
```
Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
```

eldwrjwt:
Fix #3257 

- Add `requiresLocation` implementation for `FailoverAppender`
- Isolate `LoggerContext` for each test case in `FailoverAppenderTest`

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/FailoverAppenderTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/FailoverAppenderTest.java
@@ -18,6 +18,7 @@ package org.apache.logging.log4j.core.appender;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.List;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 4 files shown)


eldwrjwt:
Hi @ppkarwasz , would you like to review this PR, and #3257 if possible?

ppkarwasz:
@eldwrjwt,

I am quite busy with my dayjob these days, but it is on my TODO list. Can you rebase the PR on the current `2.x`? All the tests were migrated to JUnit 5, so there might be some conflicts.

eldwrjwt:
Sure, I have push a new commit, and it should have no conflict.

ppkarwasz (approved):
LGTM, thanks!

PR #3260, opened 2024-12-01
Add requiresLocation function for AsyncAppender (#3257)

### LINKED ISSUE #3257: File line numbers are not printed for the failover appender when using AsyncLoggerContextSelector

eldwrjwt:
Fix #3257

- Add requiresLocation implementation for AsyncAppender

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/AsyncAppenderTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/AsyncAppenderTest.java
@@ -173,4 +173,11 @@ public class AsyncAppenderTest {
         context.getLogger("Logger").info("This is a test");
         context.stop();
     }
+
+    @Test
+    @LoggerContextSource("log4j-asynch-location.xml")
+    public void testRequiresLocation(final LoggerContext context) {
+        final AsyncAppender appender = context.getConfiguration().getAppender("Async");
+        assertTrue(appender.requiresLocation());
+    }
 }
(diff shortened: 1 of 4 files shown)


ppkarwasz (approved):
@eldwrjwt,

LGTM, thanks for your contribution.

Issue #3239, opened 2024-11-26
RegexFilter auto-boxing of Boolean useRawMsg can lead to NPE


JWT007:
In RegexFilter.Builder (2.24.1) the PluginAttribute for `useRawMsg` is a `Boolean` default `null`.

```
@PluginFactory
public static RegexFilter createFilter(
        // @formatter:off
        @PluginAttribute("regex") final String regex,
        @PluginElement("PatternFlags") final String[] patternFlags,
        @PluginAttribute("useRawMsg") final Boolean useRawMsg,
        @PluginAttribute("onMatch") final Result match,
        @PluginAttribute("onMismatch") final Result mismatch)
        // @formatter:on
        throws IllegalArgumentException, IllegalAccessException {
    if (regex == null) {
        LOGGER.error("A regular expression must be provided for RegexFilter");
        return null;
    }
    return new RegexFilter(useRawMsg, Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
}
```

The `RegexFilter` constructor expects a boolean primitive.

```
private RegexFilter(final boolean raw, final Pattern pattern, final Result onMatch, final Result onMismatch) {
    super(onMatch, onMismatch);
    this.pattern = pattern;
    this.useRawMessage = raw;
}
```

I believe, if the `useRawMsg` attribute is not set (undefined/null), auto-boxing from `Boolean` to `boolean` will cause/throw a NPE which is not caught and the builder will not return `null` as it *should*? 

Also, the 'regex' attribute should probably be defined as @Required for the `isValid()` builder-functionality to work correctly.

jaykataria1111:
@ppkarwasz , please assign this to me.

jaykataria1111:
@ppkarwasz just raised a fix for this, it was a one line change, so i did not discuss before, assuming the default value is false because that is the case for primitive types.

PR #3265, opened 2024-12-03
Fix(#3239) NPE, for RegexFilter creator.

jaykataria1111:
fix for #3239 

## Checklist

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
@@ -148,7 +148,8 @@ public final class RegexFilter extends AbstractFilter {
             LOGGER.error("A regular expression must be provided for RegexFilter");
             return null;
         }
-        return new RegexFilter(useRawMsg, Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
+        return new RegexFilter(
+                Boolean.TRUE.equals(useRawMsg), Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
     }
 
     private static int toPatternFlags(final String[] patternFlags)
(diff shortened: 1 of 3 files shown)


jvz (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -148,7 +148,8 @@ public static RegexFilter createFilter(
             LOGGER.error("A regular expression must be provided for RegexFilter");
             return null;
         }
-        return new RegexFilter(useRawMsg, Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
+        return new RegexFilter(
+                useRawMsg != null && useRawMsg, Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
```
ppkarwasz, line 152: _Nit_: Maybe comparing `useRawMsg` to `Boolean.TRUE` might be simpler to read:
```suggestion
        return new RegexFilter(
                Boolean.TRUE.equals(useRawMsg), Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
```

ppkarwasz (commented):
@jaykataria1111,

Can you add a unit test to see if `createFilter` does not throw with all parameters (except `regex`) equal to `null`?

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
jaykataria1111: +1, yeah I like the suggestion, let us do that!

jaykataria1111:
@ppkarwasz added the test, and updated the code :)

jaykataria1111:
@ppkarwasz , this one is still open I will rebase this just for tracking purposes, I do not recall any action items on this would you remember?

jaykataria1111:
@ppkarwasz I have rebased the commit, let me know if anything else is needed on this or we can close this out. Thanks for this!

jaykataria1111:
@ppkarwasz posted an update .

ppkarwasz:
Thanks for your contribution!

Issue #1867, opened 2023-10-18
[MS30] Fix API compatibility issues


ppkarwasz:
Since we had to abandon the [Revapi Maven Plugin](https://revapi.org/revapi-site/main/index.html) around version 2.17.2 due to incompatibilities with other Maven plugins, we do not have a way to automatically check for breaking API changes.

We need to introduce other tools to guarantee a stable API.

Possible candidates:

- [japicmp](https://siom79.github.io/japicmp/),
- [`bnd-baseline-maven-plugin`](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-baseline-maven-plugin).

garydgregory:
What about japicmp?

ppkarwasz:
> What about japicmp?

I added a list of candidates to the initial comment that we will evaluate in due time.

In version `2.21.0` I gave `bnd-baseline-maven-plugin` a test run. I am quite satisfied with the results. Also the two feature requests I submitted (bndtools/bnd#5792 and bndtools/bnd#5808) were granted in a very prompt manner. For the definitive decision we should probably also considerate the community that stands behind the tool.

ppkarwasz:
Over the past year the BND Baseline plugin prevented breaking API changes many times, e.g.:

- fe92961737ed3423763df6219e892b2d1a3ad49b
- 9f576e137b569b600cd086dd7522bc879bf16ff7
- f0e82d15e7f718059ecb4207c6f63396e3fb7421
- 9565d4b301f6ea4e422a1f3af1a661566b3dec20
- b8cb35fbb88baccdbde36f157a5ef737f2d6ed22

The only currently known limitation of the BND Baseline check is the case of the multi-release modules (`log4j-api-java9` and `log4j-core-java9`), because only their Java 9 equivalents are tested for API compatibility. We briefly explored the idea of using BND's multi-release support (cf. apache/logging-parent#93), but we decided to disable it (cf. apache/logging-parent#96), since it introduces a lot of complexity to the built artifacts.
I believe that commits like 945cbc886b9f78abdfa4e478babf242c5f7be199 should be enough for our use case.

ppkarwasz:
I tried `japicmp` and it can be useful to compare two already published JARs (I used this in #3339).
Otherwise it does not seem to offer anything that the BND Baseline plugin does not offer and BND seems to be maintained by a larger group of developers.
I would stick with BND.

PR #3339, opened 2024-12-29
Document all API changes in `log4j-api`

ppkarwasz:
This PR uses [japicmp](https://siom79.github.io/japicmp/) to compare all 53 stable 2.x releases and document the changes in the public API of the `log4j-api` artifact.
The basic invocation of `japicmp` is:

```
java -jar japicmp.jar \
  --old-classpath org.osgi.core.jar -o log4j-api-<old_version>.jar \
  --new-classpath org.osgi.core.jar -n log4j-api-<new_version>.jar \
  -m
```

The resulting report was used to create or fix Javadoc `@since` tags.

Part of #1867

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
@@ -142,6 +142,7 @@ public interface Logger {
      *
      * @param marker the marker data specific to this log statement
      * @param message the message CharSequence to log.
+     * @since 2.6
      */
     void debug(Marker marker, CharSequence message);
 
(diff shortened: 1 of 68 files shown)


vy (approved):
Great job @ppkarwasz! :100:

log4j-api/src/main/java/org/apache/logging/log4j/Logger.java
```diff
@@ -675,7 +699,7 @@ void debug(
     /**
      * Logs entry to a method. Used when the method in question has no parameters or when the parameters should not be
      * logged.
-     * @deprecated Use {@link #traceEntry()} instead which performs the same function.
+     * @deprecated :uUse {@link #traceEntry()} instead which performs the same function.
```
ppkarwasz, line 702: That is what happens if you switch too often from IDEA to Vim. :wink:

Issue #3369, opened 2025-01-07
LoggerConfig.RootLogger.Builder#withtFilter (typo in method name / only partially fixed)


JWT007:
**Log4j 2.24**

In LoggerConfig.Builder, an original typo existed in the '`withtFilter`' method (extra 't' in name).

This was fixed by deprecating the original method and adding the correct 'withFilter' method.

```
        /**  @deprecated Use {@link #withFilter(Filter)} instead  */
        @Deprecated
        public B withtFilter(final Filter filter) { ... }

        public B withFilter(final Filter filter) { ... }
```

However, it was not fixed in the nested `LoggerConfig.RootLogger.Builder` implementation - there the name is still incorrect.

The nested builder method should also be deprecated and '`withFilter`' added to replace it with the correct name.

ppkarwasz:
@JWT007,

Thanks! Do you mind making a PR with this change?

JWT007:
@ppkarwasz - made quick change - PR workflow awaiting approval:

https://github.com/apache/logging-log4j2/pull/3372

JWT007:
@ppkarwasz - it took me a while but I added tthe change log entry for 2.25.0

I also fixed spotless problem and updated package versions for baseline

PR #3372, opened 2025-01-08
Fix typo in LoggerConfig.RootLogger.Builder#withtFilter (#3369)

JWT007:
In LoggerConfig.RootLogger.Builder:
* Deprecated method with typo in name 'withtFilter'.
* Added correct method 'withFilter'.


ppkarwasz:
@JWT007,

Can you add a changelog entry (of type `added`)?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
```diff
@@ -945,11 +944,20 @@ public Filter getFilter() {
...
+            /**
+             * @deprecated Use {@link #withFilter(Filter)} instead
+             */
+            @Deprecated
             public B withtFilter(final Filter filter) {
                 this.filter = filter;
                 return asBuilder();
             }
 
+            public B withFilter(final Filter filter) {
```
ppkarwasz, line 956: Since we settled on using "setters" for new builder methods instead of "withers" (see e.g. #1206), could you change it to `setFilter`?

ppkarwasz (changes requested):
We should use `setFilter` instead of `withFilter`.

I also noticed that **none** of our tests use the `LoggerConfig` builder and they call the deprecated `LoggerConfig#createLogger` method instead. Could you fix this?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/LoggerConfig.java
JWT007: sure... will do

JWT007: @ppkarwasz all the other methods are withers which is why I stuck with it.  Should I add only the one setter or all?

JWT007: @ppkarwasz 

I changed the LoggerConfig.RootLogger.Builder#withFilter to #setFilter.

The change of LoggerConfig.Builder#withFilter was not part of my original change, but I have now deprecated that too and added a 'setFilter' there as well. I couldn't change the name of the existing #withFilter there without triggering a major baseline change.

I updated the tests using the #createLoggerConfig.  Fixed other code in same test methods to use builders as well (i.e. FileAppender).

NOTE I did not update test 'JiraLog4j2_2134Test' since it seems to *explicitly* be testing deprecated methods.

ppkarwasz: > Should I add only the one setter or all?

As you prefer, you can replace all "withers" in `LoggerConfig.Builder` with "setters".

ppkarwasz (approved):
LGTM

norrisjeremy:
> We should use `setFilter` instead of `withFilter`.
> 
> I also noticed that **none** of our tests use the `LoggerConfig` builder and they call the deprecated `LoggerConfig#createLogger` method instead. Could you fix this?

@ppkarwasz 
What is the justification for deprecating `withFilter()`?
Nothing else in this builder class is configured with a setter, but rather they are all withers, so this seems to introduce an API inconsistency.

ppkarwasz:
Hi @norrisjeremy,

Thanks for your thoughtful and detailed questions—these are all excellent points, and I really appreciate you taking the time to raise them.

> 1. So why weren't all the other wither methods replaced with setters and marked as deprecated too?

You're absolutely right to highlight the inconsistency. The focus of the PR was to correct a specific issue with `withtFilter` (with an additional `t`), and the introduction of `setFilter` was a side-effect driven by the fact that `withFilter` is already deprecated in `3.x`. So it made sense to align it preemptively here.

Replacing all wither methods in the same Log4j Core plugin—or across all plugins—would indeed be more consistent. But that would have significantly expanded the scope of this PR. Since there are around 300 plugins, we felt it was better to handle that broader change separately. To address this, I’ve opened issue #3750 to track the work of adding setters and deprecating all remaining withers. If you're interested, contributions are always welcome!

> 2. And why wasn't [issue #1206](https://github.com/apache/logging-log4j2/issues/1206) referenced in the PR?**

It was actually mentioned in a resolved conversation here: https://github.com/apache/logging-log4j2/pull/3372#discussion_r1957283823

> 3. Why wasn’t this change documented in the `Deprecated` section of the Release Notes?

That’s a fair point. When preparing release notes, we tend to prioritize high-impact changes for the majority of users, most of whom use Log4j via configuration files. The Log4j Core internal API is mostly used by Log4j Plugin developers and integrators. As a result, API-level changes like this one can get less visibility, especially if they're part of a narrowly scoped fix.

That said, you're absolutely right: this deprecation should have been classified as `deprecated`, not just `changed`. Feel free to submit a PR correcting that in the release notes—those are always welcome, even after the release, and can help future users.

> From a user's perspective, this change seems poorly implemented and documented, which is extremely frustrating.

I hear you, and I’m sorry the process came across that way. We always aim to provide a stable and predictable experience, and clearly we can improve how we communicate these kinds of changes.

Out of curiosity—what kind of impact did this deprecation have on your application? Was it mainly the compiler warnings, or something else? Understanding that might help us improve how we handle and communicate such changes going forward.

Thanks again for your feedback—it’s valuable and helps us improve the project for everyone.

norrisjeremy:
Hi @ppkarwasz,

Thank you for your detailed response.

To answer your question about the impact: we deliberately run compiler warnings & deprecations set to errors for our internal apps, so that they immediately cause a compilation failure, in order for our team to notice and deal with API deprecations immediately and avoid having them pile up over time.

So we immediately noticed the deprecation when a Dependabot PR for one of our internal apps failed to build yesterday.
Upon seeing the deprecation warning, we first turned to the release notes for the new 2.25.0 release, but did not note any mention of the `withFilter()` method being deprecated.

And then when trying to load up the latest Javadocs, noticed that they had not yet been published on the website.
(And in fact updated Javadocs are still not published, as all the pages linked [here](https://logging.apache.org/log4j/2.x/javadoc.html) lead to 404 page not found errors...)

This then led us to having to go dumpster diving through Github, to determine how to handle the deprecation, as well as why it had been occurred.

On top of this, this is the second time in past year that this API method has been touched.
In the 2.24.0 release, the original `withtFilter()` was marked as deprecated in lieu of the new `withFilter()` method (see #788).

So all told, this is just frustrating: it results in lost time & productivity for what is amounting to whack-a-mole operation, chasing the whims of upstream library developers: deprecating and changing withers to setters in a Builder class just doesn't seem to offer tangible benefit to any existing users.

Thanks,
Jeremy

Issue #3074, opened 2024-10-10
Reload of key/trustsore when re-establishing a connection


MichaelMorrisEst:
https://github.com/apache/logging-log4j2/pull/2767 introduces functionality to enable reloading key/trustore when the certs are renewed. However a manual step of triggering a reconfiguration (e.g. by touching the config file) is needed for the key/trust store to be reloaded. While this is a big improvement on having no reload, it is still not ideal to have to trigger a reconfiguration.

The cert renewal has no impact on existing established connections (as the handshake is done when the connection is established) so there is no need for the key/trust store to be reloaded for existing connections to continue working.
However, when an error occurs in writing to the socket a retry is attempted which includes the creation of a new socket and connection. Using a no longer valid cert here will prohibit the connection being re-established. If, during the retry, the key/truststore are reloaded, then the latest certs would always be used in re-establishing the connection and would effectively remove the need to trigger the reconfiguration.

Is this something the community would be open accepting a PR on? If so I can work on it and submit

vy:
> If, during the retry, the key/truststore are reloaded, then the latest certs would always be used in re-establishing the connection and would effectively remove the need to trigger the reconfiguration.

@MichaelMorrisEst, this is something crossed my mind while working on #2767, but I am not sure about the implications of trying to reload key/trust store on every new socket creation attempt. [Thinking out loud here...] If the server has indeed gained a new identity, reloading key/trust stores will solve the problem swiftly. But what if the failure is due to something else? If we decide to implement this, shall this be the default, or opt-in via a new attribute?

Another concern I have regarding this approach is we will be introducing an exception to the reconfiguration logic. That is, when a Log4j configuration is loaded, it is treated more or less immutable – obviously, except `LoggerConfig`s. Now we will introduce a _"please reload only this little configuration element"_ exception. I am not saying this is something bad per se. But exceptions generally bite us in the long run.

@ppkarwasz, any thoughts?

MichaelMorrisEst:
Hi @vy, thanks for the feedback 
> @MichaelMorrisEst, this is something crossed my mind while working on #2767, but I am not sure about the implications of trying to reload key/trust store on every new socket creation attempt. [Thinking out loud here...] If the server has indeed gained a new identity, reloading key/trust stores will solve the problem swiftly. But what if the failure is due to something else? If we decide to implement this, shall this be the default, or opt-in via a new attribute?

[...]

PR #3378, opened 2025-01-10
Implements an event service and demonstrates its use for sending/receiving SSL key/trust store changes

MichaelMorrisEst:
Intended as a means to discuss what the implementation for https://github.com/apache/logging-log4j2/issues/3074#issuecomment-2446935514 should look like

## Checklist


ppkarwasz:
The PR looks nice, but I need some time to analyze the existing code to see if we can reuse something. [`WatchManager`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/util/WatchManager.html) looks reusable, but it works in a complex way.

vy:
Instead of a fully-blown event bus that is publicly exposed, can we make this private and as small as possible, please? I agree with @ppkarwasz that we shall seek to use existing file watching mechanisms available, if possible.

ppkarwasz:
Hi @MichaelMorrisEst,

Since #2767 was merged instead, I think we can close this PR. Any objections?

MichaelMorrisEst:
That's fine, not needed now we have #2767

PR #3394, opened 2025-01-14
Bug Fix: Script Resolution fix

### LINKED ISSUE #3336: Script resolution depends on order of element in config file
The resolution of `ScriptRef` elements depends on the order of elements in the configuration file: the `Scripts` element must appear **before** any `ScriptRef` element.

Suvrat1629:
#3336
Refactored the ScriptFilter class to enhance script resolution and error handling. This update ensures that ScriptRef scripts are resolved properly using the ScriptManager. Improved logging for scenarios where scripts fail to execute or resolve.

## Checklist

* `.Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise


log4j-core/src/main/java/org/apache/logging/log4j/core/filter/ScriptFilter.java
```diff
@@ -57,58 +57,67 @@ private ScriptFilter(
...
+                // Optionally: mark the filter as unusable or handle failure gracefully
+                return;
+            }
+            this.script = resolvedScript; // Update to resolved script
+        } else {
+            if (!configuration.getScriptManager().addScript(script)) {
+                LOGGER.error("Failed to add script {} to the ScriptManager.", script.getName());
+            }
+        }
+    }
```
ppkarwasz, line 76: You can not use the same code as in the `createFilter` factory method: if the `script` variable is `null` it will create problems further down in the call hierarchy.

ppkarwasz (changes requested):
@Suvrat1629,

Thank you for the PR.

Unfortunately it does not solve #3336, since we have a lot of elements that use scripting (`ScriptFilter` was just an example):

- [`ScriptPatternSelector`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/layout/ScriptPatternSelector.html)
- [`Routes`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/routing/Routes.html)
- `ScriptFilter`
- [`ScriptAppenderSelector`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/ScriptAppenderSelector.html)
- [`ScriptCondition`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/rolling/action/ScriptCondition.html)
- [`ScriptArbiter`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/config/arbiters/ScriptArbiter.html)
- [`RoutingAppender`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/appender/routing/RoutingAppender.html)

To fix the resolution of scripts in all of these, I think we should take a different approach: we could alter the order in which the children of a `<Configuration>` element are processed. Right now we process them in order of appearance, with a single exception: the `<Properties>` element must be processed first to allow the values of the remaining elements to be properly interpolated:

https://github.com/apache/logging-log4j2/blob/6e3f2e23437d2876f5db2b72bd82b8a34a61b285/log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java#L668-L688

We should probably process the `<Scripts>` element just after `<Properties>`.

Note: the `ScriptArbiter` should probably be processed differently. [Arbiters](https://logging.apache.org/log4j/2.x/manual/configuration.html#arbiters) are special plugins that modify the structure of the configuration file **before** it is processed. Therefore the `ScriptArbiter` should probably just signal an `ERROR` (or throw), if it contains a `ScriptRef`. At the stage, when the arbiter is used, the `Scripts` element is not yet processed.

log4j-core-test/src/test/resources/log4j-script-filters.xml
```diff
@@ -18,7 +18,13 @@
...
   <Appenders>
     <List name="List">
-      <PatternLayout pattern="[%-5level] %c{1.} %msg%n"/>
+      <PatternLayout pattern="[%-5level] %c{1.} %msg%n"/>[suvrat@suvrat logging-log4j2]$ git push
+Username for 'https://github.com': suvrat1629
+Password for 'https://suvrat1629@github.com':
+remote: Support for password authentication was removed on August 13, 2021.
+remote: Please see https://docs.github.com/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
+fatal: Authentication failed for 'https://github.com/Suvrat1629/logging-log4j2.git/'
+
```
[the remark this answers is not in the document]
Suvrat1629: Well i just shifted to arch and was trying to push to code from intellij and this log ended up here after i pushed the code and i realised it just after :D

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/ScriptFilter.java
Suvrat1629: so should i just revert back the changes i made to the ScriptFilter.java file and and make changes to the processRootNode method in the AbstractConfiguration and update the logic?
The logic would be:
1.Process <Properties> first, as it currently does.
2.Process script-related components (ScriptPatternSelector, Routes, ScriptFilter, etc.) after <Properties> to ensure their dependencies are resolved.
3.Process all remaining components afterward.

is this how i should keep my approach?

Also thanks for your patience and guidance :)

ppkarwasz:
> so should i just revert back the changes i made to the ScriptFilter.java file and and make changes to the processRootNode method in the AbstractConfiguration and update the logic?

The `AbstractConfiguration.doConfigure()` method seems to be a better choice.

> The logic would be:
> 1. Process first, as it currently does.
> 2. Process script-related components (ScriptPatternSelector, Routes, ScriptFilter, etc.) after to ensure their dependencies are resolved.
> 3. Process all remaining components afterward.

It is hard to identify the script-related components. It is easier to make sure that the [`Scripts`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/config/ScriptsPlugin.html) container is processed at the beginning, so that the scripts are already registered, when `ScriptFilter` and company are instantiated. This means:

1. Process the nodes of type `Arbiter` to modify the configuration tree. This is what `doConfigure()` already does. 
2. Process the "Properties" plugin first, as the current code does. If the "Properties" plugin is not present, the current code uses something else to initialize the interpolator.
3. Process the "Scripts" plugin next, to register all the scripts.
4. Process the remaining components (skipping "Properties" and "Scripts").

> is this how i should keep my approach?

It is one of the possible approaches, probably the easiest to implement.

Suvrat1629:
i have made the necessary fix to the code please review the code.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/HttpAppenderBuilderTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.jupiter.api.condition.DisabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import java.net.URL;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@StatusLoggerLevel(Level.ERROR) // Ensure the logger captures ERROR level messages
+class HttpAppenderBuilderTest {
```
ppkarwasz, line 48: You added another copy of the test to the `log4j-core/src/main`. Please remove it.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/HttpAppenderBuilderTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .setLayout(new JsonLayout.Builder().build())
+                .build();
+
+        // Assert that the appender is null
+        assertNull(appender, "Appender should be null when URL is missing");
+
+        // Verify that an ERROR log message was recorded
+        assertTrue(
+                listener.findStatusData(Level.ERROR)
+                        .stream()
```
ppkarwasz, line 87: This does not compile: `findStatusData` already returns a `Stream`!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/HttpAppenderBuilderTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .setLayout(new JsonLayout.Builder().build())
+                .build();
+
+        // Assert that the appender is null
+        assertNull(appender, "Appender should be null when name is missing");
+
+        // Verify that an ERROR log message was recorded
+        assertTrue(
+                listener.findStatusData(Level.ERROR)
+                        .stream()
```
ppkarwasz, line 109: Same here

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/HttpAppenderBuilderTest.java
```diff
@@ -0,0 +1,128 @@
...
+                .setUrl(new URL("https://localhost"))
+                .build();
+
+        // Assert that the appender is null
+        assertNull(appender, "Appender should be null when layout is missing");
+
+        // Verify that an ERROR log message was recorded
+        assertTrue(
+                listener.findStatusData(Level.ERROR)
+                        .stream()
```
ppkarwasz, line 65: Same here.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -687,28 +687,37 @@ protected void doConfigure() {
             configurationStrSubstitutor.setVariableResolver(interpolator);
         }
 
+        Node scriptsNode = null;
+        for (final Node node : rootNode.getChildren()) {
+            if ("Scripts".equalsIgnoreCase(node.getName())) {
+                scriptsNode = node;
+                createConfiguration(scriptsNode, null);
+                if (scriptsNode.getObject() != null) {
+                    for (final AbstractScript script : scriptsNode.getObject(AbstractScript[].class)) {
```
ppkarwasz, line 696: The `scriptsNode` variable is not really useful here. Otherwise it looks good to me.

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/ScriptFilter.java
```diff
@@ -145,16 +154,6 @@ public static ScriptFilter createFilter(
...
-        if (script instanceof ScriptRef) {
-            if (configuration.getScriptManager().getScript(script.getName()) == null) {
-                logger.error("No script with name {} has been declared.", script.getName());
-                return null;
-            }
-        } else {
-            if (!configuration.getScriptManager().addScript(script)) {
-                return null;
-            }
-        }
```
ppkarwasz, line 157: Now you can restore this code and remove the one in `start()`.

ppkarwasz (changes requested):
@Suvrat1629,

Please review the files you are committing before you commit them. A lot of garbage/temporary files made their way into your PR.

Also make sure to push the **minimal** amount of changes, e.g. configure you IDE **not** to reformat comments or the lines you didn't modify.

In this project we use the Palantir Java Formatter, which is [available as IntelliJ IDEA plugin](https://plugins.jetbrains.com/plugin/13180-palantir-java-format), but you can also run it using:

```bash
./mvnw spotless:apply
```

Regarding Java imports, we **don't** use wildcard (`*`) imports and we follow the order:

```
all `import static` statements
<empty line>
all `import` statements
```

The changes to `AbstractConfiguration` look almost OK, but the unit test does not currently compile.

log4j-core-test/${test:logging.path}/AsyncLoggerTest.log
```diff
@@ -0,0 +1,75 @@
+INFO c.f.Bar mapvalue [stackvalue] {KEY=mapvalue, configProp=configValue, configProp2=configValue2} Async logger msg
```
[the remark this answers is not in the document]
Suvrat1629: Apologies for this, I think my intellij ide is misbehaving a bit i will look into this.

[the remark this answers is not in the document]
ppkarwasz: When you make a commit in IntelliJ, you have an explicit list of files that you can add or remove to the commit.

See [Jetbrains documentation](https://www.jetbrains.com/help/idea/commit-and-push-changes.html#commit).

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/HttpAppenderBuilderTest.java
Suvrat1629: Oh this was for the Http issue i was working on i accidentally committed that code in this pr, I will just delete the file and that would solve most of the issues.

Suvrat1629:
I have tried to fix the changes you have asked me and keep the pr as clean as possible however for the test cases I am not so sure because I lack knowledge about testing but if there are anymore further changes please let me know.

Thank you.

ppkarwasz:
> I have tried to fix the changes you have asked me and keep the pr as clean as possible however for the test cases I am not so sure because I lack knowledge about testing but if there are anymore further changes please let me know.

For testing we have a JUnit 5 extension to create a new logger context (and configuration) for each test case. See for example:

https://github.com/apache/logging-log4j2/blob/fce6ce9bd4a8ecb7a70229e0d6182e296b9b86b7/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/ScriptAppenderSelectorTest.java#L42-L60

On line 43 we use the `@LoggerContextSource` annotation to point to a configuration file on the classpath. To test that your PR solves #3336, you just need to create a configuration file that uses `ScriptRef` elements everywhere possible, but the `Scripts` configuration element is at the end.

On line 44 our extension automatically injects an instance of `Configuration` to your test case. You can then check the structure of the configuration to see if the script-related elements are there, e.g.:

```java
assertThat(configuration.getFilter()).isNotNull().isInstanceOf(ScriptFilter.class);
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/AbstractConfigurationTest.java
```diff
@@ -79,17 +79,30 @@ public void setup() {
...
+            // Add a Scripts Node if required (Example for flexibility in testing)
+            final Node scripts = newNode(rootNode, "Scripts");
+            rootNode.getChildren().add(scripts);
+
+            // Add sample script node for testing purposes
+            final Node script = newNode(scripts, "Script");
+            script.getAttributes().put("name", "TestScript");
+            script.getAttributes().put("language", "JavaScript");
+            script.getAttributes().put("scriptText", "print('Hello, Log4j!');");
+            scripts.getChildren().add(script);
```
ppkarwasz, line 105: I created `AbstractConfigurationTest` mostly as a proof of concept that I don't need an XML configuration file.

Sure, writing a unit test this way will guarantee that your fix will work for all configuration file formats, but if you write a test based on an XML file it should be proof enough that the patch works. :wink:

Suvrat1629: Well I followed what you said and made a log4j-script-ref-test.xml file and wrote a ScriptConfigurationTest.java file, used a bit of ai for help and this is the best i could come up with based on my understanding.

Suvrat1629: I also just restored the AbstractConfigurationTest.java file.

Suvrat1629:
@ppkarwasz 
Well I just checked that the test cases arent compiling but could you verify if the log4j-script-ref-test.xml file is right?

Suvrat1629:
I am closing this pr, will open a new one when i am done with the issue.

ppkarwasz:
@Suvrat1629,

I would prefer if you keep this PR open. Due to FOSDEM, I didn't have time to look at this PR, but it is going in the right direction.

Suvrat1629:
Could you verify the log4j-script-ref-test.xml?
Also I did try writing the test cases my self and I was getting the ScriptManager == null error everytime. How do i handle this?

ppkarwasz:
> Could you verify the log4j-script-ref-test.xml?

The `log4j-script-ref-test.xml` does not contain any Log4j plugin that uses scripts (see [the list in my comment above](https://github.com/apache/logging-log4j2/pull/3394#pullrequestreview-2555043675), so we can not test if `ScriptRef`s are resolved correctly.

Issue #3252, opened 2024-11-28
apparent deadlock in InternalLoggerRegistry.getLogger()


jpcmonster:
## Description

SpringBoot 3.x application experiences deadlock/hang at startup with ActiveMQ Artemis and log4j2.

Deadlock occurs if SpringBoot `backend-preinit` is enabled (the default); if disabled, it causes a stackoverflow instead.  Both stacktraces are attached.

Best guess from our team is that it may relate to [weak references](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java) in `InternalLoggerRegistry` and `LoggerContext` [use of](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java#L531) `computeIfAbsent()`.  The speculation is that if a registered logger is released as a weak reference during initialization, then it may be repeatedly initialized - causing aconcurrency/deadlock issue.

## Configuration

**Version:** 2.24.2

**Operating system:** Windows 10 (observed on Alma Linux as well)

**JDK:** Eclipse Temurin 17.0.8.1

## Logs

with `backend-preinit` enabled:
```
Java stack information for the threads listed above:
===================================================
"main":
        at jdk.internal.misc.Unsafe.park(java.base@17.0.8.1/Native Method)
        - parking to wait for  <0x0000000613400050> (a java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(java.base@17.0.8.1/LockSupport.java:211)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@17.0.8.1/AbstractQueuedSynchronizer.java:715)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(java.base@17.0.8.1/AbstractQueuedSynchronizer.java:1027)
        at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lock(java.base@17.0.8.1/ReentrantReadWriteLock.java:738)
        at org.apache.logging.log4j.core.util.internal.InternalLoggerRegistry.getLogger(InternalLoggerRegistry.java:69)
        at org.apache.logging.log4j.core.util.internal.InternalLoggerRegistry.computeIfAbsent(InternalLoggerRegistry.java:139)
[...]

Issue #3161, opened 2024-11-04
Synchronization with `main` backlog


ppkarwasz:
This meta-issue lists the PRs or modifications that didn't make it to `main` yet. Feel free to modify it after each merged PR that you don't have time or can not port immediately:

- [ ]  #2716
- [ ]  #2962
- [x] #3055
- [x]  #3052
- [x] #3118
- [ ]  #3158
- [x] #3133
- [x] #3175 (not needed before first `3.0.0` release)
- [x] #3226
- [ ] #3195
- [x] #3338
- [x] #3259
- [x] #3260
- [x] #3265
- [ ] #3217 (documentation part)
- [ ] #3381
- [ ] #3418
- [x] #3438
- [x] #3469
- [ ] #3467
- [x] https://github.com/apache/logging-log4j2/commit/e9168137ab4463013aa8d3007812b4a1cf569dac
- [ ] #4184 

Ported PRs:
- [ ] #3583
- [x] #3584
- [x] #3586
- [ ] #3171
- [x] #3372
- [x] #3322

ramanathan1504:
Hi @ppkarwasz, I’d like to help with porting some of the changes listed here from 2.x to main.

I’m planning to start with:
	•	[ ] Remove outdated @BaselineIgnore annotations (#3133)

I’ll create separate PRs for each, referencing this issue. Please let me know if this approach works or if you’d prefer I prioritize other items.

ppkarwasz:
Hi @ramanathan1504,

Thank you for the help. Please coordinate with @jvz, since he is also working on these ports.

ramanathan1504:
@jvz 
PR #3381 
issue #3170
synced with main, can I take pr #3199

vy:
> can I take pr [#3199](https://github.com/apache/logging-log4j2/pull/3199)

@ramanathan1504, go ahead. (AFAIK, @jvz is busy with DI- and plugins-related tasks.)

ramanathan1504:
@vy 
ok

ramanathan1504:
Hi @vy,

While working on this PR I noticed that the failing tests are related to earlier work in PR #3418 (issues #3252, #3399, and the discussion in #3209). That some change seems to have landed in the 2.24 branch, but I don’t see it yet in 2.x or main.

Am I correct that the fix needs to be ported forward (from 2.24 → 2.x → main) before this PR can pass cleanly?

[...]

PR #3418, opened 2025-01-28
Minimize lock usage in `InternalLoggerRegistry`

### LINKED ISSUE #3399: Suspect deadlock in InternalLoggerRegistry (intermittent)
## Description

Intermittent deadlock/thread starvation on startup in `InternalLoggerRegistry#getLoggers` on Spring Boot 3.3.5 application.

## Configuration

**Version:** 
2.24.3

**Operating system:** 
Red Hat Enterprise Linux 8.10 (Ootpa)
Kernel: Linux 4.18.0-553.27.1.el8_10.x86_64

**JDK:** 
openjdk version "21.0.5" 2024-10-15 LTS
OpenJDK Runtime Environment (Red_Hat-21.0.5.0.10-1) (build 21.0.5+10-LTS)
OpenJDK 64-Bit Server VM (Red_Hat-21.0.5.0.10-1) (build 21.0.5+10-LTS, mixed mode, sharing)

## Logs

I can't find the thread that has locked `<0x00000000e01f0be0>`.

```
"Log4j2-TF-3-ConfigurationFileWatcher-3" #275 [1024631] daemon prio=5 os_prio=0 cpu=9,61ms elapsed=1901,81s tid=0x00007f4164037f80 nid=1024631 waiting on condition  [0x00007f414c0bf000]
   java.lang.Thread.State: WAITING (parking)
	at jdk.internal.misc.Unsafe.park(java.base@21.0.5/Native Method)
	- parking to wait for  <0x00000000e01f0be0> (a java.util.concurrent.locks.ReentrantReadWriteLock$NonfairSync)
	at java.util.concurrent.locks.LockSupport.park(java.base@21.0.5/LockSupport.java:221)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@21.0.5/AbstractQueuedSynchronizer.java:754)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(java.base@21.0.5/AbstractQueuedSynchronizer.java:1079)
	at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lock(java.base@21.0.5/ReentrantReadWriteLock.java:738)
	at org.apache.logging.log4j.core.util.internal.InternalLoggerRegistry.getLoggers(InternalLoggerRegistry.java:84)
	at org.apache.logging.log4j.core.LoggerContext.updateLoggers(LoggerContext.java:812)
	at org.apache.logging.log4j.core.LoggerContext.updateLoggers(LoggerContext.java:802)
	at org.apache.logging.log4j.core.LoggerContext.setConfiguration(LoggerContext.java:675)
	at org.apache.logging.log4j.core.LoggerContext.onChange(LoggerContext.java:828)
	- locked <0x00000000e01f0a20> (a org.apache.logging.log4j.core.LoggerContext)
	at org.apache.logging.log4j.core.util.AbstractWatcher$ReconfigurationRunnable.run(AbstractWatcher.java:97)
	at java.lang.Thread.runWith(java.base@21.0.5/Thread.java:1596)
	at java.lang.Thread.run(java.base@21.0.5/Thread.java:1583)
```

## Reproduction
Happens intermittently on deploy. Can't reproduce it at will.

ppkarwasz:
It has been reported that holding a lock on `InternalLoggerRegistry` during the creation of a logger can cause performance loss and deadlocks. The logger constructor can trigger property lookups and other pluggable operations, we don't entirely control. The fix to #3252 only covered one of these cases.

This change moves the instantiation of new `Logger`s outside the write lock. While in some cases, this will cause multiple instantiations of loggers with the same parameters, all such loggers are functionally equivalent. On the other hand, the change allows the creation of different loggers in parallel.

Closes #3399

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
@@ -147,44 +147,43 @@ public final class InternalLoggerRegistry {
             return logger;
         }
 
+        // Creating a logger is expensive and might cause lookups and locks, possibly deadlocks:
+        // https://github.com/apache/logging-log4j2/issues/3252
+        // https://github.com/apache/logging-log4j2/issues/3399
+        //
+        // Creating loggers without a lock, allows multiple threads to create loggers in parallel, which also
+        // improves performance.
+        // Since all loggers with the same parameters are equivalent, we can safely return the logger from the
+        // thread that finishes first.
+        Logger newLogger = loggerSupplier.apply(name, messageFactory);
+
+        // Report name and message factory mismatch if there are any
+        final String loggerName = newLogger.getName();
+        final MessageFactory loggerMessageFactory = newLogger.getMessageFactory();
+        if (!loggerName.equals(name) || !loggerMessageFactory.equals(messageFactory)) {
+            StatusLogger.getLogger()
+                    .error(
+                            "Newly registered logger with name `{}` and message factory `{}`, is requested to be associated with a different name `{}` or message factory `{}`.\n"
+                                    + "Effectively the message factory of the logger will be used and the other one will be ignored.\n"
+                                    + "This generally hints a problem at the logger context implementation.\n"
+                                    + "Please report this using the Log4j project issue tracker.",
+                            loggerName,
+                            loggerMessageFactory,
+                            name,
+                            messageFactory);
+        }
+
         // Write lock slow path: Insert the logger
         writeLock.lock();
         try {
 
-            // See if the logger is created by another thread in the meantime
-            final Map<String, WeakReference<Logger>> loggerRefByName =
-                    loggerRefByNameByMessageFactory.computeIfAbsent(messageFactory, ignored -> new HashMap<>());
-            WeakReference<Logger> loggerRef = loggerRefByName.get(name);
-            if (loggerRef != null && (logger = loggerRef.get()) != null) {
-                return logger;
-            }
-
-            // Create the logger
-            logger = loggerSupplier.apply(name, messageFactory);
-
-            // Report name and message factory mismatch if there are any
-            final String loggerName = logger.getName();
-            final MessageFactory loggerMessageFactory = logger.getMessageFactory();
-            if (!loggerMessageFactory.equals(messageFactory)) {
-                StatusLogger.getLogger()
-                        .error(
-                                "Newly registered logger with name `{}` and message factory `{}`, is requested to be associated with a different name `{}` or message factory `{}`.\n"
-                                        + "Effectively the message factory of the logger will be used and the other one will be ignored.\n"
-                                        + "This generally hints a problem at the logger context implementation.\n"
-                                        + "Please report this using the Log4j project issue tracker.",
-                                loggerName,
-                                loggerMessageFactory,
-                                name,
-                                messageFactory);
-                // Register logger under alternative keys
-                loggerRefByNameByMessageFactory
-                        .computeIfAbsent(loggerMessageFactory, ignored -> new HashMap<>())
-                        .putIfAbsent(loggerName, new WeakReference<>(logger));
-            }
-
-            // Insert the logger
-            loggerRefByName.put(name, new WeakReference<>(logger));
-            return logger;
+            Logger currentLogger = loggerRefByNameByMessageFactory
+                    .computeIfAbsent(messageFactory, ignored -> new HashMap<>())
+                    .computeIfAbsent(name, ignored -> new WeakReference<>(newLogger))
+                    .get();
+            // A replacement for Reference.reachabilityFence() from Java 9.
+            // Prevents `newLogger` to become unreachable in the lines above.
+            return currentLogger != null ? currentLogger : newLogger;
         } finally {
             writeLock.unlock();
         }
--- a/src/changelog/.2.x.x/3399_logger_registry.xml
+++ b/src/changelog/.2.x.x/3399_logger_registry.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="3399" link="https://github.com/apache/logging-log4j2/issues/3399"/>
+  <description format="asciidoc">
+    Minimize lock usage in `InternalLoggerRegistry`.
+  </description>
+</entry>


log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -147,44 +147,43 @@ public Logger computeIfAbsent(
...
-                        .computeIfAbsent(loggerMessageFactory, ignored -> new HashMap<>())
-                        .putIfAbsent(loggerName, new WeakReference<>(logger));
-            }
-
-            // Insert the logger
-            loggerRefByName.put(name, new WeakReference<>(logger));
-            return logger;
+            Logger currentLogger = loggerRefByNameByMessageFactory
+                    .computeIfAbsent(messageFactory, ignored -> new HashMap<>())
+                    .computeIfAbsent(name, ignored -> new WeakReference<>(newLogger))
```
vy, line 182:
```suggestion
                    .computeIfAbsent(loggerMessageFactory, ignored -> new HashMap<>())
                    .computeIfAbsent(loggerName, ignored -> new WeakReference<>(newLogger))
```

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -147,44 +147,43 @@ public Logger computeIfAbsent(
...
+                    .get();
+            // A replacement for Reference.reachabilityFence() from Java 9.
+            // Prevents `newLogger` to become unreachable in the lines above.
+            return currentLogger != null ? currentLogger : newLogger;
```
vy, line 186: Are you sure this is necessary? If `WR` returns _a non-null value_, and you hold a reference to that, I don't think `WR` can be emptied in the meantime. Am I mistaken?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -147,44 +147,43 @@ public Logger computeIfAbsent(
...
 
+        // Creating a logger is expensive and might cause lookups and locks, possibly deadlocks:
+        // https://github.com/apache/logging-log4j2/issues/3252
+        // https://github.com/apache/logging-log4j2/issues/3399
+        //
+        // Creating loggers without a lock, allows multiple threads to create loggers in parallel, which also
+        // improves performance.
+        // Since all loggers with the same parameters are equivalent, we can safely return the logger from the
+        // thread that finishes first.
+        Logger newLogger = loggerSupplier.apply(name, messageFactory);
```
vy, line 158: _Nit:_ I prefer the old `logger = loggerSupplier.apply()`, this makes it clear that the variable names sharing the same prefix and introduced further down (`loggerName`, `loggerMessageFactory`, etc.) are of same origin.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -147,44 +147,43 @@ public Logger computeIfAbsent(
...
+
+        // Report name and message factory mismatch if there are any
+        final String loggerName = newLogger.getName();
+        final MessageFactory loggerMessageFactory = newLogger.getMessageFactory();
+        if (!loggerName.equals(name) || !loggerMessageFactory.equals(messageFactory)) {
+            StatusLogger.getLogger()
```
vy, line 164: Having this outside the write-lock is a sweet win.

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
ppkarwasz: I prefer to register the logger using the parameters provided by the user. Otherwise `log1` and `log2` in the code below, could point to different instances:

```
Logger log1 = LogManager.getLogger("lala", messageFactory);
Logger log2 = LogManager.getLogger("lala", messageFactory);
```

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
ppkarwasz: You might be right, but I am not eager to risk another `2.24.1` :smile: 

My reasoning was:

- We get a `WR` to `newLogger`.
- At that point in code `newLogger` has already been referenced for the last time. The `() -> new WeakReference<>(newLogger)` has already been used and can be reclaimed.
- If a GC occurs just before the `WR.get()`, the reference might be emptied.

Since I reference `newLogger` on line 186, `currentLogger` will never be `null`.

vy: > At that point in code `newLogger` has already been referenced for the last time. The `() -> new WeakReference<>(newLogger)` has already been used and can be reclaimed.

No, it cannot; because there is still a strong reference to the instance.

> You might be right, but I am not eager to risk another `2.24.1` 😄

Could you at least document that this is a _"speculative measure"_?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
vy, line 182: This logic is wrong. It can be that there is a `WR` registered for the `(messageFactory, name)` key, yet the `WR`s reference is reclaimed. I think you need to revert this block back to my `LoggerRegistry::computeIfAbsent` implementation in 2062676c:

```
Map<String, WeakReference<Logger>> loggerRefByName = loggerRefByNameByMessageFactory.get(messageFactory);
// noinspection Java8MapApi (avoid the allocation of lambda passed to `Map::computeIfAbsent`)
if (loggerRefByName == null) {
    loggerRefByNameByMessageFactory.put(messageFactory, loggerRefByName = new HashMap<>());
}
final WeakReference<Logger> loggerRef = loggerRefByName.get(name);
if (loggerRef == null || (logger = loggerRef.get()) == null) {
    loggerRefByName.put(name, new WeakReference<>(logger = newLogger));
}
return logger;
```

vy: Implemented this in 14c6df0d8053e4f312bf72cd117ce8488f0a43df.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
vy: Removed this in 14c6df0d8053e4f312bf72cd117ce8488f0a43df – even IntelliJ warned that this is redundant.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -147,43 +147,51 @@ public Logger computeIfAbsent(
...
         try {
-
-            // See if the logger is created by another thread in the meantime
-            final Map<String, WeakReference<Logger>> loggerRefByName =
-                    loggerRefByNameByMessageFactory.computeIfAbsent(messageFactory, ignored -> new HashMap<>());
-            WeakReference<Logger> loggerRef = loggerRefByName.get(name);
-            if (loggerRef != null && (logger = loggerRef.get()) != null) {
-                return logger;
+            Map<String, WeakReference<Logger>> loggerRefByName = loggerRefByNameByMessageFactory.get(messageFactory);
+            // noinspection Java8MapApi (avoid the allocation of lambda passed to `Map::computeIfAbsent`)
```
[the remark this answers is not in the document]
vy: Very fair point. Implemented in 71b3b1c053da87d0877c3c85f263bbc9f3e1d599.

ppkarwasz (commented):
@vy, looks good to me!

vy:
Tried implementing a test<sup>1</sup> where `ILR` read and write methods are stressed using multiple threads – see the attachment. Even after several different configuration attempts – increased test duration, logger name cardinality, played around with ro-vs-rw operation ratios, etc. – had no luck.

<sup>1</sup> [`InternalLoggerRegistryConcurrencyTest.java.txt`](https://github.com/user-attachments/files/18655356/InternalLoggerRegistryConcurrencyTest.java.txt)

ppkarwasz:
@vy,

Thanks for merging this. Could you cherry-pick it to `2.x` and `main` too? :pleading_face:

vy:
@ppkarwasz, ported to `2.x`. `main` needs _some_ work, since #3199 needs to be ported too.

ppkarwasz:
Thanks, I added it to the backlog #3161

jvz:
I don't see anywhere in `main` this even applies. We don't have an `InternalLoggerRegistry` there and rely on `LoggerRegistry` from `log4j-api`.

PR #3448, opened 2025-02-10
Fix log4j2.properties sample

mz0:
Attempt to use the original log4j2.properties results in
org.apache.logging.log4j.core.config.ConfigurationException: No type attribute provided for Layout on Appender CONSOLE

Apparently `.type` is missing on respective line. This commit fixes that documentation error.


ppkarwasz:
@mz0,

Thanks for the fix, it [is online](https://logging.apache.org/log4j/2.x/manual/installation.html#impl-core-config).

Issue #3437, opened 2025-02-06
Compiling on Module Path with -Werror and -Xlint:all


jjohannes:
This is about getting an error like the following when compiling with a `module-info.java` and the compiler flags `-Xlint:all` and `-Werror`:

```
> Compilation failed; see the compiler output below.
log4j-api-2.24.3.jar(/org/apache/logging/log4j/message/ParameterizedMessage.class): warning: Cannot find annotation method 'replacement()' in type 'InlineMe': class file for com.google.errorprone.annotations.InlineMe not found
log4j-api-2.24.3.jar(/org/apache/logging/log4j/message/ParameterizedMessage.class): warning: Cannot find annotation method 'imports()' in type 'InlineMe'
log4j-api-2.24.3.jar(/org/apache/logging/log4j/status/StatusData.class): warning: Cannot find annotation method 'value()' in type 'SuppressFBWarnings': class file for edu.umd.cs.findbugs.annotations.SuppressFBWarnings not found
log4j-api-2.24.3.jar(/org/apache/logging/log4j/status/StatusData.class): warning: Cannot find annotation method 'justification()' in type 'SuppressFBWarnings'
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/SystemPropertiesPropertySource.class): warning: Cannot find annotation method 'value()' in type 'ServiceProvider': class file for aQute.bnd.annotation.spi.ServiceProvider not found
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/SystemPropertiesPropertySource.class): warning: Cannot find annotation method 'resolution()' in type 'ServiceProvider'
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/Activator.class): warning: Cannot find annotation method 'value()' in type 'Headers': class file for org.osgi.annotation.bundle.Headers not found
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/Activator.class): warning: Cannot find annotation method 'name()' in type 'Header': class file for org.osgi.annotation.bundle.Header not found
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/Activator.class): warning: Cannot find annotation method 'value()' in type 'Header'
log4j-api-2.24.3.jar(/org/apache/logging/log4j/util/Activator.class): warning: Cannot find annotation method 'name()' in type 'Header'
[...]

PR #3450, opened 2025-02-10
Add transitive compileOnlyApi (requires static) dependencies

jjohannes:
This pull requests adds `compileOnlyApi` dependencies (Gradle Module Metadata) and `requires static` (module-info) for all annotation libraries used in Log4j. See #3437 for details.

The following example can be used to test the effects of the change in a Gradle build on a locally built snapshot:
https://github.com/jjohannes/gradle-demos/tree/main/log4j-metadata

## Checklist

### DIFF
--- a/log4j-parent/pom.xml
+++ b/log4j-parent/pom.xml
@@ -154,6 +154,16 @@
     <httpcore.version>4.4.16</httpcore.version>
     <jna.version>5.17.0</jna.version>
 
+    <!-- Add 'requires static' for annotation library dependencies (defined as <scope>provided</scope> in this file)
+         to allow them to be visible at compile time to make the '-Xlint:classfile' check of javac pass.
+         Overrides parent's 'bnd-jpms-module-info' and hence re-defines '$[bnd-module-name];access=0' -->
+    <bnd-jpms-module-info>$[bnd-module-name];access=0;modules="
+      biz.aQute.bnd.annotation,
+      com.google.errorprone.annotations,
+      com.github.spotbugs.annotations,
+      org.osgi.annotation.bundle,
+      org.osgi.annotation.versioning
+      "</bnd-jpms-module-info>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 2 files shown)


vy:
@jjohannes, I've just released `logging-parent` version `12.0.0`. It contains breaking changes touching these dependencies. `dependabot` should automatically update `logging-parent` version in `/pom.xml`. Would you mind keeping an eye on it, and once the update kicks in, make necessary adjustments to this PR, please?

vy:
@ppkarwasz, shall we also add a test for this to `log4j-samples`?

ppkarwasz:
> @ppkarwasz, shall we also add a test for this to `log4j-samples`?

What test do you have in mind? Do we test that a certain dependency is in the Gradle `compileClasspath` configuration, but is not in the Gradle `runtimeClasspath`?

jjohannes:
@vy I rebased the PR and re-tested. I also checked the changes in `logging-parent:12.0.0` - there were no conflicts.

If it fits, the [sample Gradle project](https://github.com/jjohannes/gradle-demos/tree/main/log4j-metadata) I created could be moved to `log4j-samples`. I added a check that asserts what is ending up on the classpaths if you only add the `log4j-api`. So running `./gradlew build` on that sample would be a quite good "end-to-end" test for the metadata publishing.

vy:
> If it fits, the [sample Gradle project](https://github.com/jjohannes/gradle-demos/tree/main/log4j-metadata) I created could be moved to `log4j-samples`. I added a check that asserts what is ending up on the classpaths if you only add the `log4j-api`. So running `./gradlew build` on that sample would be a quite good "end-to-end" test for the metadata publishing.

@jjohannes, sounds plausible, thanks! We can run `./gradlew build` either as a part of the `integration-test.yaml` workflow, or add it under a Maven module and run the command (using `antrun-maven-plugin`?), etc. @ppkarwasz, WDYT?

ppkarwasz:
> @jjohannes, sounds plausible, thanks! We can run `./gradlew build` either as a part of the `integration-test.yaml` workflow, or add it under a Maven module and run the command (using `antrun-maven-plugin`?), etc. @ppkarwasz, WDYT?

I would prefer not to use `antrun-maven-plugin`, lets have a separate Gradle run instead. We could think about merging some jobs in `integration-test.yaml` workflow. Currently we have:

1. a Maven-based run for most of the example modules.
2. a Maven-based run for the GraalVM test.
3. a Gradle-based run for the Android test.

Since 1 and 2 only differ by the JDK used, we could use the GraalVM JDK for both and merge them.
I would be also nice to merge the new test module with the Android test, although I don't how to properly do it.

jjohannes:
What you could do is rename [workflows/android-reusable-test.yaml](https://github.com/apache/logging-log4j-samples/blob/main/.github/workflows/android-reusable-test.yaml) to something like `gradle-tests.yaml`.

Then you can add another step that runs Gradle against the new example [similar to this one](https://github.com/apache/logging-log4j-samples/blob/main/.github/workflows/android-reusable-test.yaml#L115-L126) that runs the Android example.

If that sounds like a good way forward to you, I can provide a (Draft) PR with the new example.

ppkarwasz:
> If that sounds like a good way forward to you, I can provide a (Draft) PR with the new example.

Sounds good to me.

jjohannes:
Here is the PR for the samples repo: https://github.com/apache/logging-log4j-samples/pull/279
Hope I did not miss anything important there.

@vy no problem. There is no urgency from my side. Let me know if I can help with anything else.

src/changelog/.2.x.x/3437_transtive_compile_only_dependencies.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="3437" link="https://github.com/apache/logging-log4j2/issues/3437"/>
+  <description format="asciidoc">
+    Make annotation libraries visible transitively at compile time for Gradle users through Gradle Module Metadata
+    and for JPMS users through 'requires static' directives in module-info files.
```
vy, line 9:
```suggestion
    and for JPMS users through `requires static` directives in `module-info.java` files.
```

vy (commented):
@jjohannes, thanks so much for the fix!

@ppkarwasz, changes LGTM. Would you mind also skimming through, please?

log4j-parent/pom.xml
```diff
@@ -154,6 +154,15 @@
...
     <jna.version>5.16.0</jna.version>
 
+    <!-- Add 'requires static' for annotation library dependencies (defined as <scope>provided</scope> in this file)
+         to allow them to be visible at compile time to make the '-Xlint:classfile' check of javac pass.
+         Overrides parent's 'bnd-jpms-module-info' and hence re-defines '$[bnd-module-name];access=0' -->
+    <bnd-jpms-module-info>$[bnd-module-name];access=0;modules="
+      biz.aQute.bnd.annotation,
+      com.google.errorprone.annotations,
+      com.github.spotbugs.annotations,
+      org.osgi.annotation.bundle
```
ppkarwasz, line 164: Can you add `org.osgi.annotation.versioning` too?

ppkarwasz (changes requested):
Looks good to me.

There is one missing annotation library and we are good to go!

log4j-parent/pom.xml
jjohannes: Good point. I think this is automatically visible transitively through `org.osgi.annotation.bundle` but we should still add it explicitly as it is directly used.

Added it.

ppkarwasz (approved):
LGTM

ppkarwasz:
@jjohannes,

The CI build no. [14389096405](https://github.com/apache/logging-log4j2/actions/runs/14389096405), which includes the integration tests in `logging-log4j-samples` succeeded. Thanks! :100:

Issue #3431, opened 2025-02-03
AbstractConfiguration clears Configuration name if some conditions are met (i.e. no defined loggers)


JWT007:
Log4j 2.24.3 - AbstractConfiguration

If I create a BuiltConfiguration that defines a Configuration 'name'  (i.e. "FooBar") and no loggers, the `AbstractConfigurat#doConfigure()` method calls the `setToDefault()` method for example if the following conditions are met:

* no loggers are defined
* no root logger is defined

For example,

```
if (!setLoggers) {
  LOGGER.warn("No Loggers were configured, using default. Is the Loggers element missing?");
  setToDefault();
  return;
}
```

The first thing the `setToDefault` method does is clear the given name and replace it with a default.

```
protected void setToDefault() {
        // LOG4J2-1176 facilitate memory leak investigation
        setName(DefaultConfiguration.DEFAULT_NAME + "@" + Integer.toHexString(hashCode()));
       ...
}
```

It should however (IMHO) be a perfectly valid scenario to create a configuration with no content but with a custom name - specifically where I discovered this in testing.  But maybe I want to only define global filters and otherwise rely on default behaviour.

In my opinion the name should only be set here if no name has yet been set.

Generally it should be OK to set a configuration with no loggers and no root logger.  Consider a CompositeConfiguration named "FooBar" with an initially empty BuiltConfiguration.  Some runtime event might create appenders and loggers dynamically (ie. for a new service).  But it is no longer possible to lookup that configuration by name because a default has been applied.

NOTE: The referenced bug/fix (https://issues.apache.org/jira/browse/LOG4J2-1176) doesn't seem to make any mention of the name being a problem.

JWT007:
@ppkarwasz  - trying to take care of some of my many small bug reports - created a PR for this issue.

Under the assumption that the above mentioned change for LOG4J-1176 was only added to ensure that a configuration gets a default name for testing the fix, I wrapped it so that the default name is only set if no name has already been set.

PR #3454, opened 2025-02-11
Only set default configuration name if none has been previously set (…

JWT007:
[#3431]

AbstractConfiguration#setToDefault() - only sets default configuration name if none has been previously set,

+ added test for this change
+ added changelog entry


log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -774,8 +774,11 @@ public static Level getDefaultLevel() {
     }
 
     protected void setToDefault() {
-        // LOG4J2-1176 facilitate memory leak investigation
-        setName(DefaultConfiguration.DEFAULT_NAME + "@" + Integer.toHexString(hashCode()));
+        // LOG4J2-3431 don't set a default name if one has already been set
+        if (this.getName() == null || this.getName().trim().isEmpty()) {
```
ppkarwasz, line 778: We have a `Strings.isEmpty` utility method for this.

JWT007: @ppkarwasz should I update the PR?

JWT007: @ppkarwasz - Fixed and moved changelog file as mentioned in other PR.

ppkarwasz (approved):

Issue #3086, opened 2024-10-14
Log4j 2.x Documentation - RegexFilter does not document the possible nested element PatternFlags 


JWT007:
In the Log4j 2.x documentation for the RegexFilter:

https://logging.apache.org/log4j/2.x/manual/filters.html#RegexFilter

The only two configuration attributes are "regex" and "useRawMsg" (and the common filter attributes onMatch/onMismatch).

In the RegExFilter @PluginFactory there is a nested element for a set of Strings which is not mentioned in the documentation:

```
@PluginFactory
    public static RegexFilter createFilter(
            // @formatter:off
            @PluginAttribute("regex") final String regex,
            @PluginElement("PatternFlags") final String[] patternFlags,
            @PluginAttribute("useRawMsg") final Boolean useRawMsg,
            @PluginAttribute("onMatch") final Result match,
            @PluginAttribute("onMismatch") final Result mismatch)
            // @formatter:on
            throws IllegalArgumentException, IllegalAccessException {
```

If I understand correctly one can provide field names from the Pattern class here: (CASE_INSENSITIVE, COMMENTS, etc.).

The example in the Plugin Reference for the RegexFilter shows an example of nested strings but is not very clear with "a-String-implementation":

```
<RegexFilter onMatch=""
             onMismatch=""
             regex=""
             useRawMsg="">
    <a-String-implementation/><!-- multiple occurrences allowed -->
</RegexFilter>
```

Is it:

```
<RegexFilter ...>
  <CASE_INSENSITIVE>
  <COMMENTS>
</RegexFilter>
```

which the plugin reference implies.

Or:

```
<RegExFilter ...>
  <PatternFlags>CASE_INSENTIVE, COMMENTS</PatternFlags>
</RegExFilter>
```
?

A usage  example would be helpful.

Maybe this API is not really "public"? In the log4j-core-test RegexFilter test it seems the RegexFilter is never tested with anything but a null value for the patternFlags.

ppkarwasz:
Hi @JWT007,

Thank you for helping us checking our documentation against the automatically-generated plugin reference. We created the plugin reference exactly for this purpose and we are happy you find it useful.

[...]

PR #3463, opened 2025-02-12
Removed @PluginAttribute 'patternFlags' argument from @PluginFactory for RegexFilter. (#3086)

JWT007:
(#3086)

Removed @PluginAttribute 'patternFlags' argument from @PluginFactory for RegexFilter.

Actually created new @PluginFactory without the argument and deprecated old 'createFilter' method.

See ticket for discussion of the reasoning - short version - pattern flags can be passed as part of RegEx expression.

Also guarded the "PatternCompile" method against excepttions - logging and returning null if the pattern fails to compile.


log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -123,6 +115,40 @@ public String toString() {
...
+     * @param mismatch
+     *        The action to perform when a mismatch occurs.
+     * @return The RegexFilter.
+     * @throws IllegalAccessException  When there is no access to the definition of the specified member.
+     * @throws IllegalArgumentException When passed an illegal or inappropriate argument.
+     * @deprecated use {@link #createFilter(String, Boolean, Result, Result)}
+     */
+    @Deprecated
+    // TODO Consider refactoring to use AbstractFilter.AbstractFilterBuilder
+    public static RegexFilter createFilter(
```
ppkarwasz, line 132: Whenever we need to modify a factory method, we usually create a `Builder` class instead.

JWT007: @ppkarwasz I significantly reworked this one - please review if OK

I reworked it so that all config parameters are accessible via getters (i.e. 'regex') - this sort of comes from my own prroblems with these classes where config parameters are not accessible later making it impossible to see *how* the element was configured (i.e. to recreate config at runtime).  If this is OK I would probably make some new tickets to do the same for some other configuration items.

JWT007:
Problems rebasing this PR on 2.x branch - recreated new PR #3512 - closing this one.

PR #3467, opened 2025-02-15
Add `collectionName` and `databaseName` attributes to `MongoDbProvider`

jesmith17:
Ports #3322 to `2.x`


jesmith17:
@vy  I went ahead and did a new PR to merge with 2.x branch because the classes, projects, and packages already had changes in the 3.x stuff that did seem to make sense to back-port.  So tried to just back-port the specific code for this change and leave the rest alone.

vy:
Thanks Josh. I was about to suggest you creating separate PRs for 2.x and
main, but you were already ahead of me. 💯 I will review this one on Monday.

Op za 15 feb 2025 om 06:22 schreef Josh Smith ***@***.***>

ppkarwasz (approved):

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
```diff
@@ -119,4 +121,12 @@ public String toString() {
                 "Mongo4Connection [connectionString=%s, collection=%s, mongoClient=%s]",
                 connectionString, collection, mongoClient);
     }
+
+    /*
+     * This method is exposed to help support unit tests for the MongoDbProvider class.
+     *
+     */
+    public MongoCollection<Document> getCollection() {
```
vy, line 129: Please don't add a public method (which becomes a liability for decades) unless there is a practical runtime use for it. You can simply extract this value for tests using reflection, it is okay.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Provider.java
```diff
@@ -60,14 +61,47 @@ public static class Builder<B extends Builder<B>> extends AbstractFilterable.Bui
...
+                LOGGER.error("Invalid MongoDB database name `{}`.", effectiveDatabaseName, e);
+                return null;
+            }
+            // Validate the provided collectionName property
+            try {
+                MongoNamespace.checkCollectionNameValidity(effectiveCollectionName);
+            } catch (final IllegalArgumentException e) {
+                LOGGER.error("Invalid MongoDB collection name `{}`.", effectiveCollectionName, e);
+                return null;
+            }
```
vy, line 99: Shouldn't this check be better placed to `MongoDb4Provider::new`, which receives both the connection URI and the database/collection name from different code paths?

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
```diff
@@ -67,21 +67,23 @@ public MongoDb4Connection(
             final ConnectionString connectionString,
             final MongoClient mongoClient,
             final MongoDatabase mongoDatabase,
+            final String collectionName,
```
vy, line 70: This is a breaking change, try running `./mvnw verify -DskipTests -pl :log4j-mongodb4`. Would you mind

1. Creating a new _canonical_ ctor that accepts all arguments
2. Delegate from all other ctors to the new canonical ctor using `this(connectionString, ...)`
3. Mark all other ctors `@Deprecated`
4. Migrate places using `@Deprecated` ctors to the new canonical ctor

vy (changes requested):
@jesmith17, pushed some minor fixes and dropped some remarks. I'd appreciate it if you can address them.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
jesmith17: Their is no way that i could identify through reflection to find this value. It seems to be a potential limitation to the MongoDB driver itself. but there isn't a way to get at this. I looked for all other options before adding this method. We can omit it, that means removing test cases as well.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Provider.java
jesmith17: When I made these changes to the main branch, was specifically asked to make them to the build method. So making these change in this format keep the approach in line with the same code changes in the main branch.  It seems odd to want to solve the same challenge in 2 different ways across the branches.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
jesmith17: So there are 2 challenges here. 

1. The build system that log4j uses is something that I have not been able to get to work on my machine. Even if I solve odd errors like distributionSha256Sum challenges and downgrade the java version, it still gets blocked by issues with some of the docker steps due to ARM architecture of my Mac, so running that command isn't currently possible. 
2. I understand your point about the constructor, but the constructor is only called from inside the build method. And with the builder pattern people should be using the build method to handle all of those pieces anyway and not directly invoking the constructor.  This approach is inline with previously requested changes in the `main` branch and moving these would just make the change bigger and make it a larger difference between 2.x and the 3.x approaches.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Provider.java
ppkarwasz: > Shouldn't this check be better placed to `MongoDb4Provider::new`, which receives both the connection URI and the database/collection name from different code paths?

The constructor can only throw if the parameters are invalid and then each call to the constructor must be in a `try...catch`. The builder can return `null`.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
vy: This method is only used by `MongoDb4ProviderTest` to get the collection name from a `MongoDb4Provider`. You can use `MongoDb4Provider::getConnection` to obtain the `MongoDb4Connection` instance, and use reflection on that instance to get the value of its `collection` field:

```java
private static MongoNamespace findProviderNamespace(final MongoDb4Provider provider) {
    MongoDb4Connection connection = provider.getConnection();
    try {
        Field collectionField = connection.getClass().getDeclaredField("collection");
        collectionField.setAccessible(true);
        MongoCollection<?> collection = (MongoCollection<?>) collectionField.get(connection);
        return collection.getNamespace();
    } catch (Exception exception) {
        throw new RuntimeException(exception);
    }
}
```

Then you can replace

```java
assertEquals(
        collectionName,
        provider.getConnection().getCollection().getNamespace().getCollectionName());
```

with

```java
MongoNamespace namespace = findProviderNamespace(provider);
assertThat(namespace.getCollectionName()).isEqualTo(collectionName);
```

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Provider.java
vy: > The constructor can only throw if the parameters are invalid 

@ppkarwasz, right, this is a scheme practiced by several other Log4j components: ctor fails on invalid arguments with an explanatory message – this is not a problem.

> then each call to the constructor must be in a `try...catch`

Assuming unchecked exceptions are used, this is not needed, since `PluginBuilder::build` already does this:

https://github.com/apache/logging-log4j2/blob/8c0e3c6c4f32ba97985efc05286e5425bfe36742/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/util/PluginBuilder.java#L126

Moving these checks to `MongoDb4Provider::new` will provide complete coverage independent of the call site.

> When I made these changes to the main branch, was specifically asked to make them to the build method. So making these change in this format keep the approach in line with the same code changes in the main branch. It seems odd to want to solve the same challenge in 2 different ways across the branches.

@jesmith17, apologies for the inconvenience. PRs, should target `2.x`, not `main`, unless they target to-be-released Log4j 3. This is explicitly stated in the PR template:

https://github.com/apache/logging-log4j2/blob/8c0e3c6c4f32ba97985efc05286e5425bfe36742/.github/pull_request_template.md#L5

> It seems odd to want to solve the same challenge in 2 different ways across the branches.

You can also throw from a component ctor in `main` too, no need to deviate from the scheme that we're implementing in `2.x`.

jesmith17: For people just trying to figure out how to add functionality its really hard when disputes about coding approach between maintainers are getting done inside the comments for a specific PR.   I would recommend that this becomes something you document better to make this easier for others.

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
garydgregory: IMO, saying that everything (IOW public and protected) in all of Log4j is bound by binary compatibility is not a good thing and a bit of a fantasy we've not strictly adhered to in the past. The public API IMO is `log4j-api` and that's what's binary compatible, always.

More than that are just handcuffs we're putting on ourselves. Log4j API call sites can happily hum along from release to release. Third-party Implementors (as few as there may be) of these APIs need to be aware of _all_ the changes from release to release, where behavioral changes will be trickier to deal with for them anyway.

jesmith17: So the first problem is that the maven-wrapper breaks when running it on my machine using maven 3.9.9

Specifically it complains about needing to update the `distributionSha256Sum` which is impossible because 3.9.9. comes with a sha-512 sum. In order to get any of the `./mvnw` commands to run I had to rebuild the wrapper, but I suspect you don't want me to check that in. 

Seems like there is an undocumented dependent on a specific version of maven that needs to be addressed.

ppkarwasz: > So the first problem is that the maven-wrapper breaks when running it on my machine using maven 3.9.9
> 
> Specifically it complains about needing to update the `distributionSha256Sum` which is impossible because 3.9.9. comes with a sha-512 sum.

Can you provide the exact wrapper error? For example you can run:

```shell
MVNW_VERBOSE=true ./mvnw validate -N
```

The fact that Apache Maven does not publish a SHA256 checksum to Maven Central doesn't matter, because the value is hardcoded in the `wrapper.properties` file and committed to the Git repo. Each time I bump the version of Maven used by the wrapper, I verify the signature of the archive and compute the SHA256 checksum locally.

BTW: our wrapper currently uses Maven 3.9.8, not 3.9.9. I should probably upgrade that.

jesmith17:
@vy  Just did a new clone of the full repo, into a new directory. Verified that it's the 2.x branch. Running `./mvnw install -DskipTests` produced the following errror. 

```
oshsmith@Joshs-Macbook-Air-2024 log4j-temp % ./mvnw install -DskipTests
Error: Failed to validate Maven distribution SHA-256, your Maven distribution might be compromised.
If you updated your Maven version, you need to update the specified distributionSha256Sum property.
joshsmith@Joshs-Macbook-Air-2024 log4j-temp %
```

even trying to run a command like `./mvnw --version` returns the same error. 

In trying to see if I could manually view the sha256 for the 3.9.8 download, it appears to also only offer a sha512. 

Here is the content of my maven-wrapper.properties file included with the checkout.  Ommitting the comments at the top for brevity. 

```
distributionSha256Sum=8351955a9acf2f83c136c4eee0f6db894ab6265fdbe0a94b32a380307dbaa3e1
distributionType=only-script
distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.8/apache-maven-3.9.8-bin.zip
wrapperVersion=3.3.2
~
```

ppkarwasz:
> @vy Just did a new clone of the full repo, into a new directory. Verified that it's the 2.x branch. Running `./mvnw install -DskipTests` produced the following errror.
> 
> ```
> oshsmith@Joshs-Macbook-Air-2024 log4j-temp % ./mvnw install -DskipTests
> Error: Failed to validate Maven distribution SHA-256, your Maven distribution might be compromised.
> If you updated your Maven version, you need to update the specified distributionSha256Sum property.
> joshsmith@Joshs-Macbook-Air-2024 log4j-temp %
> ```

Try removing the `~/.m2/wrapper` folder. It is just a cache. In `main` we did have a `script` Maven Wrapper shell script, in `2.x` we have the more recent `only-script` version. Maybe they are incompatible.

jesmith17:
@ppkarwasz 
Removing the .mvn/wrapper folder all together just changes it to an error complaining that it can't find the maven-wrapper.properties. 

```
./mvnw: line 114: ./.mvn/wrapper/maven-wrapper.properties: No such file or directory
cannot read distributionUrl property in ./.mvn/wrapper/maven-wrapper.properties
joshsmith@Joshs-Macbook-Air-2024 log4j-temp %
```

ppkarwasz:
> Removing the .mvn/wrapper folder all together just changes it to an error complaining that it can't find the maven-wrapper.properties.

By `~/.m2/wrapper` folder I meant the `.m2/wrapper` subfolder of your home directory. The problematic data is there and this is why the problem persists, even if you make a clean Git checkout.

jesmith17:
@ppkarwasz  There is no `~/.m2/wrapper` folder in my home mvn install.  I just verified. 

But even if that was the offending file then deleting the one in the code base wouldn't have trigger an error about the file missing as it would have found the file in my local home directory, correct?

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
jesmith17: Trying the VERBOSE command above produced this

```
joshsmith@Joshs-Macbook-Air-2024 log4j-temp % MVNW_VERBOSE=true ./mvnw validate -N
Couldn't find MAVEN_HOME, downloading and installing it ...
Downloading from: https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.8/apache-maven-3.9.8-bin.zip
Downloading to: /var/folders/wm/96zyw_3x4cdbvycjl7btg8sc0000gn/T/tmp.4xLeCVRCI2/apache-maven-3.9.8-bin.zip
Found curl ... using curl
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 8979k  100 8979k    0     0  1584k      0  0:00:05  0:00:05 --:--:-- 1402k
Error: Failed to validate Maven distribution SHA-256, your Maven distribution might be compromised.
If you updated your Maven version, you need to update the specified distributionSha256Sum property.
joshsmith@Joshs-Macbook-Air-2024 log4j-temp %
```

that's after doing a `git reset` to bring back the deleted maven-wrapper.properties file. 

I suspect that this needs to be moved to an issue so we can separate the resolution of this from the approval of the code changes for the appender.

vy:
Thanks so much Josh for bearing with us to deliver this feature. There are
some last mile wrinkles to be ironed out in the tests, I will take care of
them. (Please keep the PR open, I will continue working on it.)

Op wo 19 feb 2025 om 17:59 schreef Josh Smith ***@***.***>

log4j-mongodb4/src/main/java/org/apache/logging/log4j/mongodb4/MongoDb4Connection.java
ppkarwasz: Yes, either it is a bug in Maven Wrapper or something is wrong with the downloaded file.

PR #3468, opened 2025-02-16
Fix `RollingRandomAppenderDirectWriteAndSwitchDirectoryTest`

ppkarwasz:
Attempt to fix the flaky `RollingRandomAppenderDirectWriteAndSwitchDirectoryTest`.


ppkarwasz:
The test expects a `1/1.log` file, but there is only a `01/01.log` file.

PR #3469, opened 2025-02-16
Fix formatting of `s` pattern

ppkarwasz:
The specialized `SecondPatternSequence` formatter is optimized to format the `ss.?S*` patterns, where seconds are padded. For the not padded (and uncommon) `s` pattern, we should use the usual `DynamicPatternSequence`.

This is the main reason for the failing #3468 test.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternDynamicFormatter.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternDynamicFormatter.java
@@ -702,11 +702,19 @@ final class InstantPatternDynamicFormatter implements InstantPatternFormatter {
             super(
                     createPattern(secondDigits, separator, fractionalDigits),
                     determinePrecision(secondDigits, fractionalDigits));
-            if (secondDigits < 0 || secondDigits > 2) {
-                throw new IllegalArgumentException("Unsupported number of `s` pattern letters.");
+            final int maxSecondDigits = 2;
+            if (secondDigits > maxSecondDigits) {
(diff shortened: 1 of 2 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternDynamicFormatter.java
```diff
@@ -239,7 +239,11 @@ private static List<PatternSequence> sequencePattern(final String pattern) {
                 final PatternSequence sequence;
                 switch (c) {
                     case 's':
-                        sequence = new SecondPatternSequence(true, "", 0);
+                        if (sequenceContent.length() == 2) {
+                            sequence = new SecondPatternSequence(true, "", 0);
+                        } else {
+                            sequence = new DynamicPatternSequence(sequenceContent);
+                        }
```
vy, line 246: @ppkarwasz, in #3338, you wanted keep the garbage-free behavior for subminute precision at the cost of increasing the complexity. I think we can still adhere to your motivation by adding a `secondPaddingLength` argument to `SecondPatternSequence::new` and updating `SecondPatternSequence::formatSeconds` accordingly. I am not keen on creating an exception for second patterns that are not padded. WDYT?

ppkarwasz: Fixed in https://github.com/apache/logging-log4j2/pull/3469/commits/34597ddf1c87294da6b1e7e6a544fc355de66d9d

vy (approved):

PR #3474, opened 2025-02-19
Expunge stale loggers in InternalLoggerRegistry on method invocation

Suvrat1629:
This PR updates InternalLoggerRegistry to automatically remove stale loggers when its methods are invoked.
Changes Introduced:
    Introduced a ReferenceQueue<Logger> to track loggers that have been reclaimed by the garbage collector.
    Implemented expungeStaleEntries(), which removes stale loggers before executing registry operations.
    Updated methods like getLogger() and computeIfAbsent() to invoke expungeStaleEntries() before proceeding.
    Added tests to verify stale loggers are expunged without relying on private methods.

## Checklist

### DIFF
--- a/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
+++ b/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.test.util;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.lang.ref.WeakReference;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.util.internal.InternalLoggerRegistry;
+import org.apache.logging.log4j.message.MessageFactory;
+import org.apache.logging.log4j.message.SimpleMessageFactory;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class InternalLoggerRegistryTest {
+    private InternalLoggerRegistry registry;
+    private MessageFactory messageFactory;
+
+    @BeforeEach
+    void setUp() {
+        registry = new InternalLoggerRegistry();
+        messageFactory = new SimpleMessageFactory();
+    }
+
+    @Test
+    void testGetLoggerReturnsNullForNonExistentLogger() {
+        assertNull(registry.getLogger("nonExistent", messageFactory));
+    }
+
+    @Test
+    void testComputeIfAbsentCreatesLogger() {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+        assertNotNull(logger);
+        assertEquals("testLogger", logger.getName());
+    }
+
+    @Test
+    void testGetLoggerRetrievesExistingLogger() {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+        assertSame(logger, registry.getLogger("testLogger", messageFactory));
+    }
+
+    @Test
+    void testHasLoggerReturnsCorrectStatus() {
+        assertFalse(registry.hasLogger("testLogger", messageFactory));
+        registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                .getLogger(name, factory));
+        assertTrue(registry.hasLogger("testLogger", messageFactory));
+    }
+
+    @Test
+    void testExpungeStaleEntriesRemovesGarbageCollectedLoggers() throws InterruptedException {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+
+        WeakReference<Logger> weakRef = new WeakReference<>(logger);
+        logger = null; // Dereference to allow GC
+
+        // Retry loop to give GC time to collect
+        for (int i = 0; i < 10; i++) {
+            System.gc();
+            Thread.sleep(100);
+            if (weakRef.get() == null) {
+                break;
+            }
+        }
+
+        // Access the registry to potentially trigger cleanup
+        registry.computeIfAbsent("tempLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                .getLogger(name, factory));
+
+        assertNull(weakRef.get(), "Logger should have been garbage collected");
+        assertNull(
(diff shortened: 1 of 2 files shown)


Suvrat1629:
@vy I have always had a hard time with test cases, I have opened a pr regardless please take look and provide further insights on how i can improve it.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -39,10 +41,8 @@
 /**
  * A registry of {@link Logger}s namespaced by name and message factory.
  * This class is internally used by {@link LoggerContext}.
- * <p>
- * We don't use {@linkplain org.apache.logging.log4j.spi.LoggerRegistry the registry from Log4j API} to keep Log4j Core independent from the version of Log4j API at runtime.
- * This also allows Log4j Core to evolve independently from Log4j API.
- * </p>
```
vy, line 45: Why do you remove these?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -53,23 +53,44 @@ public final class InternalLoggerRegistry {
...
+            writeLock.unlock();
+        }
+    }
+
     /**
      * Returns the logger associated with the given name and message factory.
-     *
-     * @param name a logger name
-     * @param messageFactory a message factory
-     * @return the logger associated with the given name and message factory
```
vy, line 68: Why do you remove these?

vy (changes requested):
Please put back _all_ the removed comment blocks.

Suvrat1629:
Ok sorry about that, I have put back all the comment blocks as they were.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -53,23 +57,48 @@ public final class InternalLoggerRegistry {
...
+    /**
+     * Removes a logger from the registry.
+     */
+    private void removeLogger(Reference<? extends Logger> loggerRef) {
+        writeLock.lock();
+        try {
+            loggerRefByNameByMessageFactory.values().forEach(map -> map.values().removeIf(ref -> ref == loggerRef));
+        } finally {
+            writeLock.unlock();
+        }
```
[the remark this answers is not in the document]
Suvrat1629: Well you are right this is a bit inefficient. I have made some changes to the `removeLogger`, please take a look if it fits fine.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -142,12 +188,12 @@ public Logger computeIfAbsent(
             final MessageFactory messageFactory,
             final BiFunction<String, MessageFactory, Logger> loggerSupplier) {
 
-        // Check arguments
```
vy, line 145: @Suvrat1629, please revert all changes that are irrelevant in the context of this PR – just like the one on this line. To make sure you indeed implemented what I asked for, please compare your branch with `2.x` (i.e., `git diff origin/2.x`) before submitting them.

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
Suvrat1629: Done!

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -53,13 +57,48 @@ public final class InternalLoggerRegistry {
...
+
+        writeLock.lock();
+        try {
+            Map<String, WeakReference<Logger>> loggerRefByName = loggerRefByNameByMessageFactory.get(messageFactory);
+            if (loggerRefByName != null) {
+                loggerRefByName.remove(name);
+                if (loggerRefByName.isEmpty()) {
+                    loggerRefByNameByMessageFactory.remove(messageFactory); // Cleanup
+                }
+            }
```
vy, line 96: Assume I move this logic (and only this!) to a method called `unsafeRemoveLogger(Logger)`. Note that this method receives a `Logger`, not a `Reference`.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -53,13 +57,48 @@ public final class InternalLoggerRegistry {
...
     public InternalLoggerRegistry() {}
 
+    /**
+     * Expunges stale logger references from the registry.
+     */
+    private void expungeStaleEntries() {
+        Reference<? extends Logger> loggerRef;
+        while ((loggerRef = staleLoggerRefs.poll()) != null) {
+            removeLogger(loggerRef);
+        }
```
vy, line 75: Would you mind changing this loop as follows, please?
```
boolean locked = false;
Reference<? extends Logger> loggerRef;
while ((loggerRef = staleLoggerRefs.poll()) != null) {
    Logger logger = loggerRef.get();
    if (logger != null) {
        if (!locked) {
            writeLock.lock();
        }
        unsafeRemoveLogger(logger);
    }
}
if (locked) {
    writeLock.unlock();
}
```
See my other comment regarding the `unsafeRemoveLogger(Logger)` method.

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
vy, line 1: Can you replace `LoggerContext.getContext()` calls as follows, please?
```
@Test
void someTest(TestInfo testInfo) {
    try (final LoggerContext loggerContext = new LoggerContext(testInfo.getDisplayName)) {
        // Use `loggerContext`
    }
}
```
That is, don't disrupt the global one, use ephemeral `LC`s instead.

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+        }
+
+        // Access the registry to potentially trigger cleanup
+        registry.computeIfAbsent("tempLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                .getLogger(name, factory));
+
+        assertNull(weakRef.get(), "Logger should have been garbage collected");
+        assertNull(
+                registry.getLogger("testLogger", messageFactory), "Stale logger should be removed from the registry");
```
vy, line 102: Can we instead replace this as follows, please?

1. Use an ephemeral `LoggerContext` (as I described in another comment)
2. Create 1000 `Logger`s (should be many enough to create GC pressure)
3. Log using all `Logger`s (so we imitate a real-world setup where `Logger`s are used at some point)
4. Use `Awaitility.waitAtMost(...).until(...)` to verify that after a `System.gc()`, `InternalLoggerRegistry::loggerRefByNameByMessageFactory` is emptied

You can access to `InternalLoggerRegistry::loggerRefByNameByMessageFactory` using reflection. That is, first extract the `LoggerContext::loggerRegistry` field, and then the `InternalLoggerRegistry::loggerRefByNameByMessageFactory` field.

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
ppkarwasz: Note that if `loggerRef.get()` returns `null`, we still need to remove the entry that holds the `WeakReference`.

What do you think about such an approach:

1. We `poll()` the queue until its empty.
2. If there was any reference in the queue we iterate **once** over all the entries of the map and we remove **all** the invalidated `WeakReference`s, not just those that were in the queue.

As of [my previous comment](https://github.com/apache/logging-log4j2/pull/3474#discussion_r1965597651), this seems a little bit inefficient if `loggerRef.get()` is not null, but is more efficient than iterating over all entries **once-per-reference**.

vy: @ppkarwasz, you're right. @Suvrat1629, please proceed with the suggestion of @ppkarwasz. Though I suggest having double-checking:

```
if (!refQueue.isEmpty) {
    lock();
    try {
        if (!refQueue.isEmpty) {
            // Clear `refQueue`
            // Clean up `loggerRefByNameByMessageFactory`
        }
    } finally {
       unlock();
    }
}
```

PR #3501, opened 2025-02-26
Add option to provide URIs to monitor in addition to the config file

MichaelMorrisEst:
A change detected in either the config file itself or the provided additional URIs shall result in a reconfigure
See #3074 

## Checklist


log4j-core-test/src/test/java/org/apache/logging/log4j/core/PropertiesFileConfigTest.java
```diff
@@ -64,4 +64,25 @@ void testReconfiguration(final LoggerContext context) throws Exception {
         } while (newConfig == oldConfig && loopCount++ < 5);
         assertNotSame(newConfig, oldConfig, "Reconfiguration failed");
     }
+
+    @Test
```
vy, line 68: We can remove this if we can have a dedicated `MonitorUriTest` I hinted earlier.

log4j-core-test/src/test/resources/log4j-test2.xml
```diff
@@ -15,7 +15,7 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<Configuration status="OFF" name="XMLConfigTest" monitorInterval="1">
+<Configuration status="OFF" name="XMLConfigTest" monitorInterval="1" monitorUris="target/test-classes/org/apache/logging/log4j/core/net/ssl/keyStore.p12">
```
vy, line 18: I'm reading through the changes, I don't know if it is supported, but please use a dedicated element instead:

```
<Configuration monitorInterval="1">
  <MonitorUris>
    <Uri>file:///path/to/file</Uri> <!-- Note that this needs to be a URI, not just a path! -->
  <MonitorUris>
<Configuration>
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -275,6 +275,14 @@ protected void initializeWatchers(
             final Reconfigurable reconfigurable,
             final ConfigurationSource configSource,
             final int monitorIntervalSeconds) {
+        initializeWatchers(reconfigurable, configSource, Collections.emptySet(), monitorIntervalSeconds);
+    }
+
+    protected void initializeWatchers(
+            final Reconfigurable reconfigurable,
+            final ConfigurationSource configSource,
+            final Collection<Source> auxiliarySources,
```
vy, line 284: Why don't we just provide a single `Set<Source> monitorSources`?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/api/package-info.java
```diff
@@ -20,7 +20,7 @@
  * @since 2.4
  */
 @Export
-@Version("2.20.1")
+@Version("2.21.0")
```
vy, line 23: Always target the version this feature will (supposedly) be shipped with:

```suggestion
@Version("2.25.0")
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/BuiltConfiguration.java
```diff
@@ -148,9 +154,28 @@ public void setShutdownTimeoutMillis(final long shutdownTimeoutMillis) {
     }
 
     public void setMonitorInterval(final int intervalSeconds) {
+        initializeMonitoring(intervalSeconds, "");
+    }
+
+    public void initializeMonitoring(final int intervalSeconds, final String monitorUris) {
```
vy, line 160: I think this should be `Set<URI> monitorUris`

vy (changes requested):
@MichaelMorrisEst, thanks so much for putting effort into this, much appreciated! :heart_eyes: 

I've dropped some remarks. I'd appreciate it if you can update the PR with requested changes. I will have some more remarks – e.g., adding docs – but I will share them last.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
```diff
@@ -544,6 +544,30 @@ void testReconfiguration(final LoggerContext context) throws Exception {
...
+        for (int i = 0; i < 17; ++i) {
+            logger.debug("Reconfigure");
+        }
+        Thread.sleep(100);
+        for (int i = 0; i < 20; i++) {
+            if (context.getConfiguration() != oldConfig) {
+                break;
+            }
+            Thread.sleep(50);
+        }
```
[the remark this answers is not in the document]
MichaelMorrisEst: Used Awatility in new test class

log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
```diff
@@ -544,6 +544,30 @@ void testReconfiguration(final LoggerContext context) throws Exception {
...
 
+    @Test
+    void testReconfigurationMonitorUris(final LoggerContext context) throws Exception {
+        final Configuration oldConfig = context.getConfiguration();
+        final int MONITOR_INTERVAL_SECONDS = 5;
+        final File file = new File("target/test-classes/org/apache/logging/log4j/core/net/ssl/keyStore.p12");
+        final long orig = file.lastModified();
+        final long newTime = orig + 10000;
+        assertTrue(file.setLastModified(newTime), "setLastModified should have succeeded.");
+        TimeUnit.SECONDS.sleep(MONITOR_INTERVAL_SECONDS + 1);
```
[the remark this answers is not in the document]
MichaelMorrisEst: Done

log4j-core-test/src/test/java/org/apache/logging/log4j/core/PropertiesFileConfigTest.java
MichaelMorrisEst: Done

log4j-core-test/src/test/resources/log4j-test2.xml
MichaelMorrisEst: Changed the implementation to use a dedicated element instead

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/api/package-info.java
MichaelMorrisEst: Done

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/BuiltConfiguration.java
MichaelMorrisEst: Used URIs in reworked impl

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
MichaelMorrisEst: Not applicable anymore with reworked impl

MichaelMorrisEst:
Thanks for your comments @vy. The comment to use a dedicated element instead of an attribute changes the implementation quiet a bit. Please take a look when you have time, feedback gratefully received. Thanks!

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
```diff
@@ -28,38 +31,57 @@
...
+        }
+        return lastModifiedMillis;
     }
 
     @Override
     public void fileModified(final File file) {
-        lastModifiedMillis = file.lastModified();
+        monitoredFiles.entrySet().stream()
+                .forEach(monitoredFile ->
+                        monitoredFile.setValue(monitoredFile.getKey().lastModified()));
```
vy, line 59: Shouldn't you be `monitoredFiles.get(file)` and then...?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
```diff
@@ -28,38 +31,57 @@
...
     @Override
     public long getLastModified() {
-        return file != null ? file.lastModified() : 0;
+        Long lastModifiedMillis = 0L;
+        for (final File monitoredFile : monitoredFiles.keySet()) {
+            if (monitoredFile.lastModified() > lastModifiedMillis) {
+                lastModifiedMillis = monitoredFile.lastModified();
+            }
+        }
+        return lastModifiedMillis;
```
vy, line 52: I think this can be simplified with a `stream()` and `min()`.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
```diff
@@ -28,38 +31,57 @@
...
+     * Add the given URIs to be watched.
+     *
+     * @param monitorUris URIs to also watch
+     */
+    public void addMonitorUris(final List<URI> monitorUris) {
+        monitorUris.forEach(uri -> {
+            File additionalFile = new Source(uri).getFile();
+            monitoredFiles.put(additionalFile, additionalFile.lastModified());
+        });
+    }
```
vy, line 79: I'm not really sure if `ConfigurationFW` should have a concept of _monitor URIs_. Instead, the call-site can call `watching(Source)` multiple times, but this time `super.watching(source)` will be overriden incorrectly. I have the impression that there is an impedance mismatch. That is, we're trying to extend `CFW` to support multiple files. Instead, in `AbstractConfiguration`, can't we create a `CFW` for each monitor URI?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -767,6 +785,18 @@ protected void doConfigure() {
         setParents();
     }
 
+    private List<URI> convertToJavaNetUris(final List<Uri> uris) {
+        final List<URI> javaNetUris = new ArrayList<>();
+        for (Uri uri : uris) {
+            try {
+                javaNetUris.add(new URI(uri.getUri()));
+            } catch (URISyntaxException e) {
+                LOGGER.error("Error parsing monitor URI: " + uri, e);
```
vy, line 794: Are we sure we want to proceed with (re)configuration obtained from an invalid configuration? I think we should propagate the exception and let it fail.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -323,10 +325,19 @@ public void start() {
...
-                    "Start watching for changes to {} every {} seconds",
-                    getConfigurationSource(),
-                    watchManager.getIntervalSeconds());
+            if (uris != null && uris.size() > 0) {
+                LOGGER.info(
+                        "Start watching for changes to {} and {} every {} seconds",
+                        getConfigurationSource(),
+                        uris,
+                        watchManager.getIntervalSeconds());
+                watchManager.addMonitorUris(configurationSource.getSource(), uris);
```
vy, line 334: Can't we instead for each URI create a new `ConfigurationFileWatcher` and register it using `watchManager.watch(cfgSource, watcher)`?

vy (commented):
@MichaelMorrisEst, I am extremely thankful for your effort and patience. I've some questions. I'd appreciate it if you can address them.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
MichaelMorrisEst: My thinking here was to be consistent with other error scenarios, for example lines 726, 764, 790.
Also to let the exception propagate we would need to declare it on the doConfigure() method which could cause an impact on anyone with their own custom implementation which extends AbstractConfiguration

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
MichaelMorrisEst: Yes that works too, updated accordingly

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
MichaelMorrisEst: Not applicable any more with changes from other comments

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFileWatcher.java
MichaelMorrisEst: Not applicable any more with changes from other comments

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -323,10 +325,19 @@ public void start() {
...
+                        getConfigurationSource(),
+                        uris,
+                        watchManager.getIntervalSeconds());
+                watchMonitorUris();
+            } else {
+                LOGGER.info(
+                        "Start watching for changes to {} every {} seconds",
+                        getConfigurationSource(),
+                        watchManager.getIntervalSeconds());
+            }
```
vy, line 340: Can't we simplify this as follows? (Can `uris` be null at all?)

```suggestion
            LOGGER.info(
                    "Start watching for changes to {} and {} every {} seconds",
                    getConfigurationSource(),
                    uris,
                    watchManager.getIntervalSeconds());
            watchMonitorUris();
```

vy: ... and, this is the wrong place and time to do admit monitor URIs to the `WatchManager`. Would you move this admission operation (i.e., the logic in `watchMonitorUris`) to `initializeWatchers`, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
vy: See [this recent discussion on how to handle configuration failures](https://lists.apache.org/thread/h2oydyk6xld47ljttqvflbt4530o73vw). In short, configuration failures should result in the entire configuration to fail, instead of applying in partially. Existing code that doesn't adhere to this needs to be fixed, but that is another issue.

vy (changes requested):
Would you mind also updating `configuration.adoc` such that

1. Just before `Arbiters` section, create a `Monitor URIs` section.
2. Link `Monitor URIs` and `#configuration-attribute-monitorInterval` to each other with sufficient text and explanation.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
MichaelMorrisEst: Hi @vy 
The reason I did not put it in initializeWatchers() is because initializeWatchers() is called before the elements are read. The top level attributes (including monitorInterval) have been read at this stage and can therefore be used, but the elements (including monitorUris) have not been.

Would it be ok to move invoking watchMonitorUris() to intialization() which is earlier than here, but after the elements have been read? This would avoid having to read the monitorUri element earlier and separately from the other elements.

I can move it to initializeWatchers() either if you still prefer that, and add code to read the monitorUri element separately to the other elements, but I think the above would be cleaner in the code.

MichaelMorrisEst:
> @MichaelMorrisEst, would you mind giving me write access to your fork, please? I want to push some changes and get this merged for `2.25.0`, which we will be releasing in 1-2 weeks.

Hi @vy, I think I gave you access a few months ago for another PR, can you check and let me know if there is a problem

vy:
> I think I gave you access a few months ago for another PR, can you check and let me know if there is a problem

@MichaelMorrisEst, see the following failure. Am I missing something?

```
$ git push git@github.com:Nordix/logging-log4j2.git Nordix-issue-3074:issue-3074
ERROR: Permission to Nordix/logging-log4j2.git denied to vy.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

(`Nordix-issue-3074` is my local branch, where I also happen to execute the command from.)

MichaelMorrisEst:
Ive added you there now @vy , you should get notification

vy:
@MichaelMorrisEst, nope, I haven't gotten it yet. Can you merge (not rebase!) the upstream `2.x` to your local `issue-3074`, and push your changes, please? I will base my work on that tip and send you a patch.

vy:
Have just seen the invitation in an inbox I wasn't expecting. :face_exhaling:

Pushed my changes. In short, renamed `MonitorUri` to `MonitorResource`, improved tests, and some other minor changes. @MichaelMorrisEst, do you have any further remarks?

MichaelMorrisEst:
> Have just seen the invitation in an inbox I wasn't expecting. 😮‍💨
> 
> Pushed my changes. In short, renamed `MonitorUri` to `MonitorResource`, improved tests, and some other minor changes. @MichaelMorrisEst, do you have any further remarks?

Looks good, I've no further remarks

ppkarwasz (changes requested):
Thank you all for the work you've put into this contribution — it's shaping up to be a valuable addition for users!

I’ve reviewed the code and it looks great overall. I do have a few concerns about the proposed syntax for the Java Properties configuration format, and I’d love to discuss them further to ensure we’re aligned with the existing conventions and the future goals for 3.x.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -347,6 +349,17 @@ public void start() {
         LOGGER.info("Configuration {} started.", this);
     }
 
+    private void watchMonitorResources() {
+        if (this instanceof Reconfigurable && watchManager.getIntervalSeconds() >= 0) {
```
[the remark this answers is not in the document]
MichaelMorrisEst: Done

log4j-core/src/main/java/org/apache/logging/log4j/core/config/AbstractConfiguration.java
```diff
@@ -325,7 +327,7 @@ public void start() {
         if (watchManager.getIntervalSeconds() >= 0) {
             LOGGER.info(
                     "Start watching for changes to {} every {} seconds",
-                    getConfigurationSource(),
+                    watchManager.getConfigurationWatchers().keySet(),
```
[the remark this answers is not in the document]
MichaelMorrisEst: Extracted to new private method for use in both locations

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -124,6 +125,12 @@ public PropertiesConfiguration build() {
             }
         }
 
+        final Map<String, Properties> monitorResources = PropertiesUtil.partitionOnCommonPrefixes(
+                PropertiesUtil.extractSubset(rootProperties, "monitorResources"));
+        for (final Map.Entry<String, Properties> entry : monitorResources.entrySet()) {
+            builder.add(createMonitorResource(entry.getKey().trim(), entry.getValue()));
+        }
```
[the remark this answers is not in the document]
MichaelMorrisEst: If I understand your comment correctly the configuration should then look like:

```
monitorResources.type = MonitorResources
monitorResources.0.type = MonitorResource
monitorResources.0.uri = file://path/to/external-file-1.txt
monitorResources.1.type = MonitorResource
monitorResources.1.uri = file://path/to/external-file-2.txt
```
I've update the code accordingly, let me know if this is not correct

log4j-core/src/main/java/org/apache/logging/log4j/core/config/MonitorResource.java
```diff
@@ -0,0 +1,74 @@
...
+        this.uri = requireNonNull(uri, "uri");
+        if (!"file".equals(uri.getScheme())) {
+            final String message =
+                    String.format("Only `file` scheme is supported in monitor resource URIs! Illegal URI: `%s`", uri);
+            throw new IllegalArgumentException(message);
+        }
+    }
+
+    @PluginFactory
+    public static MonitorResource createMonitorResource(@PluginAttribute("uri") final URI uri) {
```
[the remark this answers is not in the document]
ppkarwasz: URIs can be problematic for some Windows users, which is likely why `log4j2.configurationFile` supports both URIs and native file paths. In Log4j Core 3, I used the term `location` for this kind of attribute (see [`log4j.configuration.location`](https://logging.apache.org/log4j/3.x/manual/systemproperties.html#log4j.configuration.location)).

For `MonitorResource`, we could consider two approaches:

- **Use a single `location` attribute** that accepts both file paths and URIs.
- **Introduce two separate attributes** (`path` and `uri`) to allow users to be explicit about the format.

At this point, I'm unsure which approach is best. Since it's unclear whether users will request support for Windows paths in the future, adopting the builder pattern now seems like a reasonable way to postpone this decision without blocking the current feature.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/api/ConfigurationBuilder.java
```diff
@@ -61,6 +61,10 @@ public interface ConfigurationBuilder<T extends Configuration> extends Builder<T
      */
     ConfigurationBuilder<T> add(CustomLevelComponentBuilder builder);
 
+    default ConfigurationBuilder<T> add(MonitorResourceComponentBuilder builder) {
+        throw new UnsupportedOperationException();
+    }
```
ppkarwasz, line 66: 1. Add a generic method for top-level elements in `ConfigurationBuilder`:
    ```java
    ConfigurationBuilder<T> addComponent(ComponentBuilder<?> builder);
    ```
    It mirrors `ComponentBuilder#addComponent`, but uses a distinct name to differentiate from specialized `add` methods.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -124,6 +125,26 @@ public PropertiesConfiguration build() {
...
+                final Properties monitorResourceProps = entry.getValue();
+                final String monitorResourceType = (String) monitorResourceProps.remove("type");
+                if (!"MonitorResource".equals(monitorResourceType)) {
+                    throw new ConfigurationException(
+                            "No or invalid type provided for monitorResouce - must be MonitorResource");
+                }
+                builder.add(createMonitorResource(entry.getKey().trim(), entry.getValue()));
+            }
+        }
+
```
ppkarwasz, line 147: 2. Add an overload of `processRemainingProperties` that accepts a `ConfigurationBuilder`:
    ```java
    private void processRemainingProperties(ConfigurationBuilder builder, Properties properties);
    ```
3. Invoke this new method at the end of the `build()` method, replacing the less generic implementation above.
4. One remaining issue: the `build()` method contains multiple `Properties.getProperty()` calls, which do not remove entries from the `Properties` object.
To prevent duplicate processing in `processRemainingProperties`, replace these calls with `Properties.remove()`.

ppkarwasz (changes requested):
Thanks for addressing my suggestions on processing the `*.properties` format!

Your changes cover the newly introduced `monitorResources` element well. However, there’s another lesser-known configuration element—[`AsyncWaitStrategyFactory`](https://logging.apache.org/log4j/2.x/manual/async.html#custom-waitstrategy)—that’s still unsupported in the Java Properties format.

What do you think about the approach below to handle both elements together?

MichaelMorrisEst:
> Thanks for addressing my suggestions on processing the `*.properties` format!
> 
> Your changes cover the newly introduced `monitorResources` element well. However, there’s another lesser-known configuration element—[`AsyncWaitStrategyFactory`](https://logging.apache.org/log4j/2.x/manual/async.html#custom-waitstrategy)—that’s still unsupported in the Java Properties format.
> 
> What do you think about the approach below to handle both elements together?

I played around with it and think it worked out quiet well I think, was able to re-use some existing methods, so I've pushed with the changes

vy (approved):

ppkarwasz (changes requested):
This looks great! :100:

There is only one task remaining: the removal of the unused `MonitorResourceComponentBuilder` class.

ppkarwasz (approved):

vy:
Created #3703 to fix the commit signatures. (I did not want to force-push into this PR to avoid truncating git and GitHub conversations history.)

Issue #2791, opened 2024-08-02
Fix handling of `onMatch` and `onMismatch` attributes in the properties configuration format


linghengqian:
## Description

- Since apache/logging-log4j2:2.23.0, `onMismatch` of `MarkerFilter` is no longer available. This was actually found from https://github.com/apache/hive/pull/5375 . It can be reproduced in both `2.23.0` and `2.23.1`. This will be reflected in a log similar to `ERROR MarkerFilter contains an invalid element or attribute "onMismatch"`.

## Configuration

**Version:** [Log4j version] `2.23.0`

**Operating system:** [OS and version] `Ubuntu WSL 22.04.4`

**JDK:** [JDK distribution and version] `8.0.422-tem` on `SDKMAN!`

## Logs

```shell
-------------------------------------------------------------------------------
Test set: org.apache.hive.hplsql.TestHplsqlLocal
-------------------------------------------------------------------------------
Tests run: 86, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.359 s <<< FAILURE! - in org.apache.hive.hplsql.TestHplsqlLocal
org.apache.hive.hplsql.TestHplsqlLocal.testException  Time elapsed: 0.676 s  <<< FAILURE!
org.junit.ComparisonFailure: 
expected:<[]Ln:2 PRINT
Correct
L...> but was:<[2024-08-02T03:55:09.225Z main ERROR MarkerFilter contains an invalid element or attribute "onMismatch"
]Ln:2 PRINT
Correct
L...>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at org.apache.hive.hplsql.TestHplsqlLocal.run(TestHplsqlLocal.java:482)
	at org.apache.hive.hplsql.TestHplsqlLocal.testException(TestHplsqlLocal.java:229)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
[...]

PR #3505, opened 2025-02-27
2791 Fix handling of `onMatch` and `onMismatch` attributes in the properties configuration format

JWT007:
#2791
---
+ Updated DefaultComponentBuilder `put(String, String)` to remove attributte if value is null
+ Updated DefaultComponentBuilder guarded `putAttribute(...)` implementations with object values against NPE
+ updated javadoc
+ added changelog

(edit)
+ per @ppkarwasz review added JVerify annotations to `config.builder` and `config.builder.api` packages
+ optimized builder APIs and Default***ComponentBuilder implementations
+ changed (deprecated) 'addAttribute' methods to 'setAttribute' to better match other builder APIs
+ added convencience APIs for setting general attributes on ***ComponentBuilder interfaces
+ javadoc'd both packages 100%
+ updated callers of now deprecated APIs
+ updated changelog

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
@@ -20,6 +20,8 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.builder.api.Component;
@@ -57,36 +59,43 @@ class DefaultComponentBuilder<T extends ComponentBuilder<T>, CB extends Configur
         this.value = value;
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final boolean value) {
         return put(key, Boolean.toString(value));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final Enum<?> value) {
-        return put(key, value.name());
+        return put(key, Optional.ofNullable(value).map(Enum::name).orElse(null));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final int value) {
         return put(key, Integer.toString(value));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final Level level) {
-        return put(key, level.toString());
+        return put(key, Optional.ofNullable(level).map(Level::toString).orElse(null));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final Object value) {
-        return put(key, value.toString());
+        return put(key, Optional.ofNullable(value).map(Object::toString).orElse(null));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final String value) {
         return put(key, value);
     }
 
+    /** {@inheritDoc} */
     @Override
     @SuppressWarnings("unchecked")
     public T addComponent(final ComponentBuilder<?> builder) {
@@ -94,6 +103,7 @@ class DefaultComponentBuilder<T extends ComponentBuilder<T>, CB extends Configur
         return (T) this;
     }
 
+    /** {@inheritDoc} */
     @Override
     public Component build() {
         final Component component = new Component(type, name, value);
@@ -102,19 +112,37 @@ class DefaultComponentBuilder<T extends ComponentBuilder<T>, CB extends Configur
         return component;
     }
 
+    /** {@inheritDoc} */
     @Override
     public CB getBuilder() {
         return builder;
     }
 
+    /** {@inheritDoc} */
     @Override
     public String getName() {
         return name;
     }
 
+    /**
+     * Puts the given key/value pair to the attribute map.
+     * <p>
+     *   If the value is {@code null} the component attribute with the given {@code key} will
+     *   instead be removed from the map.
+     * </p>
+     * @param key the key
+     * @param value the value
+     * @return this builder (for chaining)
+     */
     @SuppressWarnings("unchecked")
     protected T put(final String key, final String value) {
-        attributes.put(key, value);
+
+        if (value != null) {
+            attributes.put(key, value);
+        } else {
+            attributes.remove(key);
+        }
+
         return (T) this;
     }
 }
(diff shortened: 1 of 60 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
```diff
@@ -57,43 +59,51 @@ public DefaultComponentBuilder(final CB builder, final String name, final String
...
     @Override
     public T addAttribute(final String key, final boolean value) {
         return put(key, Boolean.toString(value));
     }
 
+    /** {@inheritDoc} */
     @Override
     public T addAttribute(final String key, final Enum<?> value) {
-        return put(key, value.name());
+        return put(key, Optional.ofNullable(value).map(Enum::name).orElse(null));
```
ppkarwasz, line 71:
```suggestion
        return put(key, value != null ? value.name() : null);
```
We tend to use more primitive expressions in Log4j to keep it garbage-free. For a configuration it doesn't matter, but you'll find a lot of simple `null` checks and iterations on arrays in the code.

Personally I wouldn't use `Optional` for a single method call.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
```diff
@@ -57,43 +59,51 @@ public DefaultComponentBuilder(final CB builder, final String name, final String
         this.value = value;
     }
 
+    /** {@inheritDoc} */
```
ppkarwasz, line 62: Do we need it? Since we don't add anything, we can leave the Javadoc empty.

ppkarwasz (changes requested):
@JWT007,

Thanks! :100:

I think we could profit from the occasion and add JSpecify annotation to all classes in this package. As far as I can tell only a few parameters and fields can be `null`, mostly those named `name` and `value`.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
JWT007: oh I guess I just do that as a matter of course... I asked my AI Assistant - best practice, etc. :)

  - Including `/** {@inheritDoc} */` as a rule for all overridden methods ensures that all methods are documented in the same manner, even if no additional details are needed.
  - This can be particularly useful in libraries used by others (e.g., open-source or enterprise frameworks) where clear documentation is critical.
  -  If you're building a library or public-facing tool, adding `{@inheritDoc}` for every method provides clarity when generating documentation and ensures consistency
-  Adding `/** {@inheritDoc} */` to methods that don't add any documentation is often considered a **good practice for clarity**, especially for public APIs or widely used libraries.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
```diff
@@ -17,104 +17,217 @@
...
+        Objects.requireNonNull(builder, "The 'builder' argument must not be null.");
+        components.add(builder.build());
+        return self();
     }
 
+    /** {@inheritDoc} */
     @Override
-    public T addAttribute(final String key, final int value) {
-        return put(key, Integer.toString(value));
+    public @NonNull CB getBuilder() {
```
ppkarwasz, line 110:
```suggestion
    public CB getBuilder() {
```

Isn't `@NonNull` the default?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
```diff
@@ -17,104 +17,217 @@
...
+     * Clears the internal state removing all configured attributes and components.
+     * <p>
+     *   This method is primarily intended to be used in testing.
+     * </p>
+     */
+    protected void clear() {
+        synchronized (this) {
+            attributes.clear();
+            components.clear();
+        }
```
ppkarwasz, line 166: Since the class is not thread-safe anyway, do we need this?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/impl/DefaultComponentBuilder.java
```diff
@@ -17,104 +17,217 @@
...
+     */
+    protected Map<String, String> getAttributes() {
+        return Collections.unmodifiableMap(attributes);
+    }
+
+    /**
+     * Gets the list of child components.
+     * @return an <i>immutable</i> list of the child components
+     */
+    protected List<Component> getComponents() {
```
ppkarwasz, line 197:
```suggestion
    protected List<? extends Component> getComponents() {
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/PropertiesConfigurationBuilder.java
```diff
@@ -334,7 +332,7 @@ private static <B extends ComponentBuilder<B>> ComponentBuilder<B> createCompone
             final ComponentBuilder<?> parent, final String key, final Properties properties) {
         final String name = (String) properties.remove(CONFIG_NAME);
         final String type = (String) properties.remove(CONFIG_TYPE);
-        if (Strings.isEmpty(type)) {
+        if (type == null || Strings.isEmpty(type)) {
```
ppkarwasz, line 335:
```suggestion
        if (Strings.isEmpty(type)) {
```

`Strings.isEmpty` is `null`-safe.

src/changelog/.2.x.x/2791_rework_ComponentBuilder_API.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="changed">
+  <issue id="2791" link="https://github.com/apache/logging-log4j2/issues/2791"/>
+  <description format="asciidoc">
+    Fix problem when null attribute values are set on DefaultComponentBuilder. GitHub issue #2791.
+    Added JVerify annotations to config.builder.* packages.
```
ppkarwasz, line 9:
```suggestion
    Added JSpecify annotations to config.builder.* packages.
```

ppkarwasz (changes requested):
Thanks for the PR! :100:

Overall it looks OK, I added some notes below:

vy:
@ramanathan1504, is this something you might want to pick up?

ramanathan1504:
@vy, Yes, I can pick this up. I will review this PR and the issue discussion and come up with updated one.

vy:
Superseded by #4116.

Issue #3509, opened 2025-03-02
Log4j 3.x - StringMatchFilter - guard potential NPEs and builder validation/API


JWT007:
This is a parallel of Ticket #3153 for the main (3.x) branch.

StringMatchFilter:

+ add validation on 'text' attribute to class constructor and builder (non-blank)
+ add JVerify nullability annotations to class
+ replace 'setMatchString' builder method with 'setText'
+ add javadoc details about the specific StringMatchFilter implementation/behaviour
+ update tests

PR #3510, opened 2025-03-02
Update StringMatchFilter builder API (#3509)

JWT007:
#3509 
---
Fix StringMatchFilter Validation
* Guard against NPEs
* added private constructor to builder
* added getText() accessor
* add JVerify nullability annotations
* added detailed javadoc with implementation details
* optimized AbstractLifeCycle and AbstractFilter "equalsImpl" implementations
* added changelog

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
@@ -1,0 +1,103 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.filter;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.apache.logging.log4j.core.Filter;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.test.junit.LoggerContextSource;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for {@link StringMatchFilter}.
+ */
+class StringMatchFilterTest {
+
+    /**
+     * Test the normal valid programmatic instantiation of a {@link StringMatchFilter} via its builder.
+     */
+    @Test
+    void testFilterBuilderOK() {
+        StringMatchFilter.Builder stringMatchFilterBuilder = StringMatchFilter.newBuilder();
+        stringMatchFilterBuilder.setText("foo");
+        StringMatchFilter stringMatchFilter = stringMatchFilterBuilder.build();
+        assertNotNull(stringMatchFilter, "The filter should not be null.");
+        assertEquals("foo", stringMatchFilter.getText());
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


JWT007:
NOTE: This PR replaces the 2.x PR #3510 (#3509) for the main 3.x branch.
A cherry pick was not possible due to missing API.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Test that if a {@code null} string is set as a match-pattern, an {@code IllegalArgumentExeption} is thrown.
+     */
+    @Test
+    @SuppressWarnings({"DataFlowIssue" // invalid null parameter explicitly being tested
+    })
+    void testFilterBuilderFailsWithExceptionOnNullText() {
+        StringMatchFilter.Builder stringMatchFilterBuilder = StringMatchFilter.newBuilder();
+        Assertions.assertThrows(IllegalArgumentException.class, () -> stringMatchFilterBuilder.setText(null));
```
ppkarwasz, line 65: This is certainly debatable, but I would rather expect NPE to be thrown here.

The way I see it, there are two levels of validation:

- The parameter is annotated as `@NonNull` and we expect the programmer to check for that. In this case NPE seems appropriate.
- There is a more involved validation process that requires a non-empty string. In this case `IllegalArgumentException` seems appropriate.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
```diff
@@ -0,0 +1,103 @@
...
+     * Test that if a {@link StringMatchFilter} is specified without a 'text' attribute it is not instantiated.
+     *
+     * @param configuration the configuration
+     */
+    @Test
+    @LoggerContextSource("log4j2-stringmatchfilter-3153-nok.xml")
+    void testConfigurationWithTextNEG(final Configuration configuration) {
+        final Filter filter = configuration.getFilter();
+        assertNull(filter, "The filter should be null.");
+    }
```
ppkarwasz, line 102: This would rather belong in a unit test for `PluginBuilder` or similar.
The fact that the user misspells `StringMatchFilter` or another name, doesn't seem relevant.

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/StringMatchFilter.java
```diff
@@ -26,232 +27,551 @@
...
+     *
+     * @throws NullPointerException if the given {@code event} is {@code null}
+     */
     @Override
-    public Result filter(
-            final Logger logger, final Level level, final Marker marker, final Object msg, final Throwable t) {
-        return filter(logger.getMessageFactory().newMessage(msg).getFormattedMessage());
+    public Result filter(final LogEvent event) {
+        Objects.requireNonNull(event, "The 'event' argument must not be null.");
+        return filter(event.getMessage().getFormattedMessage());
```
ppkarwasz, line 89: Please don't sort methods. It makes it much harder to review the code.

ppkarwasz (changes requested):
My 2 cents:

log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
JWT007: this is only because I used the Log4j Assert which throws IllegalArgumentException.

```
    public Builder setText(final String text) {
            this.text = Assert.requireNonEmpty(text, "The 'text' argument must not be null or empty.");
            return this;
        }
```

I could of course add an extra Objects.requiresNonNull if you think that is better?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
JWT007: this simply tests that a misconfigured (or not at all) fails on build and returns null I think.

A "<StringMooFilter>" invalid name would throw in an entirely different locattion I think - unresolved plugin type.

ppkarwasz: Unless I am mistaken this tests that a misspelled plugin name (`StringMatchFfilter`) will be ignored.

Since the plugin name is misspelled `PluginBuilder` will not find the plugin and return `null`. No code in `StringMatchFilter.java` will be executed.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/StringMatchFilterTest.java
ppkarwasz: Maybe we can modify `Assert.requireNonEmpty`? What do you think?

It is mostly a question of convention (and we have none), but I think that it is worth differentiating the `null` and empty cases, since there are plenty of tools that do static analysis on `null`s and the caller could use one.

PR #3512, opened 2025-03-02
Removed @PluginAttribute 'patternFlags' argument from @PluginFactory for RegexFilter. (#3086)

### LINKED ISSUE #3086: Log4j 2.x Documentation - RegexFilter does not document the possible nested element PatternFlags

JWT007:
(#3086)

Removed @PluginAttribute 'patternFlags' argument from https://github.com/pluginfactory for RegexFilter.

Actually created new https://github.com/pluginfactory without the argument and deprecated old 'createFilter' method.

See ticket for discussion of the reasoning - short version - pattern flags can be passed as part of RegEx expression.

Also guarded the "PatternCompile" method against excepttions - logging and returning null if the pattern fails to compile.

Recreated from 2.x and superceding PR #3463


JWT007:
New branch based on 2.x branch - replacing PR #3463

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -29,68 +29,210 @@
...
+    /**
+     * {@inheritDoc}
+     *
+     * @throws NullPointerException if the {@code event} argument is {@code null}
+     */
     @Override
     public Result filter(final LogEvent event) {
-        final String text = targetMessageTest(event.getMessage());
-        return filter(text);
+        Objects.requireNonNull(event, "The 'event' argument must not be null.");
```
vy, line 192:
```suggestion
        Objects.requireNonNull(event, "event");
```

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
+    public static final class Builder extends AbstractFilterBuilder<RegexFilter.Builder>
+            implements org.apache.logging.log4j.core.util.Builder<RegexFilter> {
+
+        /* NOTE: LOG4J-3086 - No patternFlags in builder - this functionality has been deprecated/removed. */
+
+        /**
+         * The regular expression to match.
+         */
+        @PluginBuilderAttribute
+        @Required(message = "No 'regex' provided for RegexFilter")
```
vy, line 272: Please don't provide a failure message for cases where _the failure system_ should be providing the necessary context.

```suggestion
        @Required
```

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
+         */
+        public Builder setUseRawMsg(final boolean useRawMsg) {
+            this.useRawMsg = useRawMsg;
+            return this;
+        }
+
+        /** {@inheritDoc} */
+        @Override
+        public boolean isValid() {
+            return (Strings.isNotEmpty(this.regex));
```
vy, line 314:
```suggestion
            return Strings.isNotEmpty(this.regex);
```

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
+            }
+
+            // build with *safety* to not throw exceptions
+            try {
+                return new RegexFilter(this);
+            } catch (final Exception ex) {
+                LOGGER.error("Unable to create RegexFilter. {}", ex.getMessage(), ex);
+                return null;
+            }
         }
```
vy, line 338: We had [a `dev@` discussion about this](https://lists.apache.org/thread/h2oydyk6xld47ljttqvflbt4530o73vw), and the conclusion is to throw exception on failures while building components. Would you mind updating this `build()` logic accordingly, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
+     * The constructor/fields/methods below have been deprecated.
+     * - the 'create***' factory methods should no longer be used - use the builder instead
+     * - pattern-flags should now be passed via the regular expression itself
+     */
+
+    /**
+     * @deprecated pattern flags have been deprecated - they can just be included in the regex-expression.
+     */
+    @Deprecated
+    private static final int DEFAULT_PATTERN_FLAGS = 0;
```
vy, line 352: AFAIU, this constant is only used in `toPatternFlags()`, could you move it into that method, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
+    /**
+     * Returns the pattern-flags applied to the regular-expression when compiling the pattern.
+     *
+     * @return the pattern-flags (maybe empty but never {@code null}
+     * @deprecated pattern-flags are no longer supported
+     */
+    @Deprecated
+    public String[] getPatternFlags() {
+        return this.patternFlags.clone();
     }
```
vy, line 392: We're adding a new deprecated method? Why don't we simply not add it?

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -100,59 +242,191 @@ private String targetMessageTest(final Message message) {
...
             throws IllegalArgumentException, IllegalAccessException {
-        if (regex == null) {
-            LOGGER.error("A regular expression must be provided for RegexFilter");
-            return null;
-        }
-        return new RegexFilter(
-                Boolean.TRUE.equals(useRawMsg), Pattern.compile(regex, toPatternFlags(patternFlags)), match, mismatch);
+
+        // LOG4J-3086 - pattern-flags can be embedded in RegEx expression
+        Objects.requireNonNull(regex, "The 'regex' argument must not be null.");
```
vy, line 422:
```suggestion
        Objects.requireNonNull(regex, "regex");
```

vy (changes requested):
Thanks so much for the hard work @JWT007! Dropped some remarks.

ppkarwasz (changes requested):

vy:
I've resurrected this in #4119. Thanks so much for the ground work @JWT007, and apologies for the late processing.

PR #3513, opened 2025-03-02
Updated RegexFilter implementtation with changes from 2.x branch merge to main [3.x] (#3086) 

### LINKED ISSUE #3086: Log4j 2.x Documentation - RegexFilter does not document the possible nested element PatternFlags

JWT007:
Fixes #3086 
---
* add `RegexFilter(Builder)` private constructor
* add RegexFilter#getPattern()
* add RegexFilter#getRegex()
* add RegexFilter#isUseRawMessage()
* add RegexFiltter changes from #2773 from 2.x branch that were only partially merged to main
* add JVerify nullability annotations
* add protected `AbstractFilter(Builder)` constructor
* removed deprecated API
* updated all RegexFilter javadoc
* updated RegexFilterTest
* updated AbstractFilterableTest - fixed JUnit5 warnings
* updated AbstractFilterTestt

NOTE: cherry-picking changes directly from 2.x branch not possible due to missing APIs, changed 3.x Plugin design/imports, etc.


log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -28,140 +25,312 @@
...
-            @PluginAttribute final Boolean useRawMsg,
-            @PluginAttribute final Result onMatch,
-            @PluginAttribute final Result onMismatch)
-            // @formatter:on
-            throws IllegalArgumentException, IllegalAccessException {
-        if (regex == null) {
-            LOGGER.error("A regular expression must be provided for RegexFilter");
-            return null;
-        }
-        return new RegexFilter(useRawMsg, Pattern.compile(regex, toPatternFlags(patternFlags)), onMatch, onMismatch);
```
ppkarwasz, line 142: For backward compatibility, we can not remove this method, but we can:

- Remove the `@PluginFactory` annotation.
- Deprecated it.
- Rewrite it in terms of builder calls.

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -28,140 +25,312 @@
...
+    public static Builder newBuilder() {
+        return new Builder();
     }
```
ppkarwasz, line 260: A plugin builder factory must be annotated with `@PluginBuilderFactory` instead of `@PluginFactory`.

PR #3532, opened 2025-03-12
Deterministic UUID option for PatternLayout

kazariad:
Draft implementation of a deterministic UUID generator based on a LogEvent hash.  Would enable Appenders to produce repeatable UUIDs for the same LogEvent and correlate log outputs.

[Discussion](https://github.com/apache/logging-log4j2/discussions/3503)


log4j-core/src/main/java/org/apache/logging/log4j/core/util/UuidUtil.java
```diff
@@ -140,4 +142,47 @@ public static UUID getTimeBasedUuid() {
...
+        // Scramble
+        hash ^= (hash >>> 33);
+        hash *= 0xff51afd7ed558ccdL;
+        hash ^= (hash >>> 33);
+        hash *= 0xc4ceb9fe1a85ec53L;
+        hash ^= (hash >>> 33);
+        // Add salt
+        hash ^= SALT;
+        return hash;
+    }
```
ppkarwasz, line 187: Can you explain the purpose of this mangling? This could actually decrease the number of possible hashes, not increase it.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/UuidUtil.java
```diff
@@ -140,4 +142,47 @@ public static UUID getTimeBasedUuid() {
...
+        long nanoOfSecond = logEvent.getInstant().getNanoOfSecond();
+        // Thread IDs typically increment from 0 producing a narrow range
+        long threadId = logEvent.getThreadId();
+
+        // Increase entropy
+        long most = mix(epochSecond, nanoOfSecond, threadId);
+        long least = mix(threadId, ~nanoOfSecond, epochSecond);
+        // Set UUID v4 bits
+        most &= 0xFFFFFFFFFFFF0FFFL;
+        most |= 0x0000000000004000L;
```
ppkarwasz, line 166: This looks more like a [UUID Type 8](https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-8) (vendor-specific or experimental), not a UUID Type 4 (purely random).

log4j-core/src/main/java/org/apache/logging/log4j/core/util/UuidUtil.java
```diff
@@ -50,6 +51,7 @@ public final class UuidUtil {
     private static final int HUNDRED_NANOS_PER_MILLI = 10000;
 
     private static final long LEAST = initialize(NetUtils.getMacAddress());
+    private static final long SALT = new SecureRandom().nextLong();
```
ppkarwasz, line 54: What is the impact of this on startup time, e.g. in a VM without a physical source of entropy?

I think that `Random` might be enough, since this is not used for cryptographic purposes.

ppkarwasz (changes requested):
@kazariad,

In general this looks fine to me, thanks!
A couple of notes:

kazariad:
Thanks for your feedback, @ppkarwasz.  I was originally using a SecureRandom to comply with the UUID guidelines for using crypto-secure RNG, but I've experimented and decided to switch to a straightforward murmurhash of the LogEvent fields instead.  It's also formatted as a UUID v8 now.

vy:
@kazariad, I internally discussed this issue with @ppkarwasz, and we decided to not proceed with this PR due to following reasons:

1. Putting aside your changes, UUID Pattern Converter is not in its ideal shape. The way it uses properties to pass temporal state, branches at runtime on `UuidPatternConverter::isRandom`, etc. needs to be improved. Further, with some minor effort, it can even be made garbage-free.
2. I'm against copy-pasting a full-blown hashing algorithm – this is error prone and can pose security risks if used unwisely. In short, it is a liability that I prefer to avoid.
3. There is more to this issue than it appears and, as the only active maintainers, @ppkarwasz and I would rather be busy with other more pressing issues.

In [our earlier discussion](/apache/logging-log4j2/discussions/3503), @ppkarwasz and I pointed you several potential paths. If you decide to go that route, and publish the code yourself (or your organization), then we would be happy to refer to your project and its Maven artifacts in our official documentation.

Another route I can personally suggest you is to seek for [commercial support offered by Log4j maintainers](https://logging.apache.org/support.html#commercial).

PR #3553, opened 2025-03-19
docsImprove String concatenation best practice

ppkarwasz:
This change splits the "don't use String concatenation" best practice into two parts:

- A recommendation concerning performance: string concatenation is not efficient if the logger is off.
- A security recommendation: format string must be constants to prevent `{}` placeholder injection.

The `${dangerousLookup}` part of the example is removed, since lookups are not executed since version `2.15.0`

### DIFF
--- a/src/site/antora/modules/ROOT/pages/manual/api.adoc
+++ b/src/site/antora/modules/ROOT/pages/manual/api.adoc
@@ -78,6 +78,8 @@ include::partial$manual/api-best-practice-exception-as-last-argument.adoc[]
 
 include::partial$manual/api-best-practice-dont-use-string-concat.adoc[]
 
+include::partial$manual/api-best-practice-dont-mix-concat-and-params.adoc[]
+
 [#best-practice-supplier]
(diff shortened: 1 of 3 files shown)


src/site/antora/modules/ROOT/partials/manual/api-best-practice-dont-use-string-concat.adoc
```diff
@@ -34,3 +29,10 @@ Imagine `userId` being provided by the user with the following content:
 ----
 /* GOOD */ LOGGER.info("failed for user ID `{}`", userId);
 ----
+
+* [x] Use message lambdas
++
+[source,java]
+----
+/* GOOD */ LOGGER.info(() -> "failed for user ID: " + userId);
```
vy, line 37: @ppkarwasz, I think this is a foot gun. It'd be very hard to distinguish for inexperienced users, which is the chief audience of this guide, why it is fine to do string concatenation here, whereas before it was not. IMHO, this should not be supported by the API in the first place, but anyway.

Issue #3423, opened 2025-01-29
Log4J2: PatternLayout "charset" in documentation but forbitten by xml schema


palbr:
## Description

I have a config file for Log4j 2.x in XML format:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Configuration xmlns="https://logging.apache.org/xml/ns"
               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
               xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-config-2.xsd">
    <Appenders>
        <Console name="console" target="SYSTEM_OUT">
            <PatternLayout charset="UTF-8" pattern="test"/>    <!-- error here -->
        </Console>
    </Appenders>

    <Loggers>
        <Root level="INFO">
            <AppenderRef ref="console"/>
        </Root>
    </Loggers>
</Configuration>
```

The XML validation is telling me:
> Attribute charset is not allowed here

The documentation for Log4j 2.x is telling me that an attribut `charset` exists in `PatternLayout`: https://logging.apache.org/log4j/2.x/manual/pattern-layout.html#plugin-attr-charset

Reading the website published at https://logging.apache.org/xml/ns, I get the impression that https://logging.apache.org/xml/ns/log4j-config-2.xsd is the official "configuration file xml schema".

## Configuration

**Version:** Log4j 2.x

palbr:
If I use https://logging.apache.org/xml/ns/log4j-config-2.23.1.xsd instead of  https://logging.apache.org/xml/ns/log4j-config-2.xsd, `charset` is invalid, too.

jbb01:
After digging through some code I found, that the plugin descriptor `log4j-core-plugins.xml` still contains the `charset` attribute and the issue seems to be with the generation of the XSD from said descriptor. In particular, the issue seem to be https://github.com/apache/logging-log4j2/blob/f2efdd5a33e869869528d727fd26c1d2a4f0754c/pom.xml#L368 which prevents the generation of XML types for classes in the `java` package, which then prevents attributes with those types from being included in the generated XSD in `SchemaGenerator::writePluginAttribute`. After removing the exclusion, the generated XSD contains the `charset` attribute.

A reasonable thing would probably be to allow generation of XML types for all classes for which a `TypeConverter` is available, even if those classes are in the `java` package.

ppkarwasz:
@jbb01,

Could you submit a PR with the solution?

[...]

PR #3568, opened 2025-03-22
Exclude certain standard library classes from docgen type exclusion (#3423)

jbb01:
This will allow the log4j-docgen-maven-plugin to generate xsd schema attributes for plugin fields with a type of `Class`, `InetAddress`, `URL`, `Charset` or `Pattern`. This addresses #3423.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -366,7 +366,7 @@
          `log4j-docgen`-specific properties
          ================================== -->
     <!-- `log4j-docgen` qualified class name exclude pattern -->
-    <log4j.docgen.typeFilter.excludePattern>^java\..+</log4j.docgen.typeFilter.excludePattern>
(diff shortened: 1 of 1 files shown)


vy (approved):

jbb01:
I am somewhat confused as to why the build fails. If I'm reading the logs correctly the spotbugs plugin is complaining about some Java code, but this PR does not change any Java code. Is this an issue with the PR or with the build pipeline?

ppkarwasz:
> I am somewhat confused as to why the build fails. If I'm reading the logs correctly the spotbugs plugin is complaining about some Java code, but this PR does not change any Java code. Is this an issue with the PR or with the build pipeline?

Please merge the changes from `2.x`.
The failure is due to an upgrade of Spotbugs, which has more rules now.

vy (approved):

PR #3603, opened 2025-04-11
Adapt `merge-dependabot` to RTC

ppkarwasz:
This PR removes the `merge-dependabot` workflow and replaces it with a new `add-dependabot-changelog` workflow that:

- Runs on each push to a `dependabot/*` branch.
- If the current reference is a Dependabot PR it pushes a new changelog entry to the Dependabot branch.


vy:
@ppkarwasz, please do not proceed with this change – I am DM'ing you.

ppkarwasz:
> @ppkarwasz, please do not proceed with this change – I am DM'ing you.

This is the magic of RTC: I can not merge any PR unless you or another committer agrees.

.github/workflows/add-dependabot-changelog.yaml
vy, line 1: Shouldn't this be a reusable in `logging-parent`? Does every project need to roll out there own `a-d-c.yaml`?

ppkarwasz:
Currently all Dependabot workflows fail (see [job 40576197611](https://github.com/apache/logging-log4j2/actions/runs/14467978596/job/40576197611) and the new commit does not trigger any workflow, so the required checks can not pass.

.github/workflows/add-dependabot-changelog.yaml
ppkarwasz: Yes, it should, be we need to test it first. The hard part is probably to trigger a build on the new commit that the `merge-dependabot` creates. This is probably the only way to trigger the required checks.

vy: I prefer this to be made a part of `logging-parent` first. If we you can make that happen, I can help with making the next `logging-parent` release.

vy (changes requested):

ppkarwasz:
Maybe it is easier to write it in TypeScript and reference it from `logging-parent`.

ppkarwasz:
I am closing this, since it will be implemented in `logging-parent`.

PR #3624, opened 2025-04-17
Disable (temporarily) Dependabot updates

ppkarwasz:
This is a short workaround for the compatibility issues between `merge-dependabot` and branch protection rules.

Commits pushed by a GitHub workflow don't trigger other workflows, so Dependabot PRs cannot be merged if the workflow manages to add a changelog file.


vy:
@ppkarwasz, what is long-term solution to this problem?

ppkarwasz:
> what is long-term solution to this problem?

The main problem to solve is how to trigger a build on the commit with the changelog. I haven't test any of these, which is why I call the "longterm" solutions:

1. We could try to trigger the build using `workflow_run` like in [`develocity-publish-build-scans.yaml`](https://github.com/apache/logging-log4j2/blob/2.x/.github/workflows/develocity-publish-build-scans.yaml).
2. We could try to trigger the build, when the PR is approved.
3. We could try using labels.

PR #3071, opened 2024-10-09
Fix Android-related issues in Log4j Core

ppkarwasz:
This fixes three issues encountered in the [`log4j-samples-android`](https://github.com/apache/logging-log4j-samples/tree/main/log4j-samples-android) test project:

1. Disables the `jvmrunargs` lookup on Android and fixes it on the other platforms. Previously, the lookup always returned `null`.
2. Switches the default context selector to `BasicContextSelector` on Android. `StackLocator` is broken on Android: it cannot use our JDK 8 code (missing `sun.reflect` classes), but also it cannot use our JDK 11+ code (missing multi-release JAR support). This causes `ClassLoaderContextSelector` to use two different logger contexts for the same classloader.
3. Fixes a `ParserConfigurationException` caused by the lack of XInclude capabilities in Android's XML parser. The fix to [LOG4J2-3531](https://issues.apache.org/jira/browse/LOG4J2-3531) didn't cover all the cases.

Closes #3056.
Part of #2832.

PR #3638, opened 2025-04-25
A way to disable the optimization for filling the stack traces

neboskreb:
This PR allows to disable the optimization for filling the stack traces on platforms where it is not available (e.g. on Android). 

The developer of the port to such platform shall call `PrivateSecurityManagerStackTraceUtil.disable()` before any loggins is printed, e.g. during the configuration phase.

For this reason, the visibility of `PrivateSecurityManagerStackTraceUtil` was increased to `public`.


ppkarwasz (changes requested):
@neboskreb,

Thank you for your contribution!

I think that `PrivateSecurityManagerStackTraceUtil` should **always** be disabled on Android. Similarly to #3071 I would simply disable the class if the `java.vendor` system property contains `Android`.

neboskreb:
@ppkarwasz , as you mentioned in the other ticket, this PR seems to be on a critical path. 

But I'm not quite sure I understand what is the blocker here, sorry. What can I do to unblock the progress?

ppkarwasz:
> But I'm not quite sure I understand what is the blocker here, sorry. What can I do to unblock the progress?

Currently you are proposing a new `PrivateSecurityManagerStackTraceUtil#disable()` method that can be used by other libraries to disable the security manager.

My point is that `PrivateSecurityManagerStackTraceUtil` will **never** work on Android, so we might as well always disable the class on that platform. More concretely the static initialization code:

https://github.com/apache/logging-log4j2/blob/7acbc486854cd9b62184883ef9ee5973e1ef1b8b/log4j-api/src/main/java/org/apache/logging/log4j/util/PrivateSecurityManagerStackTraceUtil.java#L31-L44

should:

- check if the Java property `java.vendor` contains the word `Android`.
- if we are running on Android set `SECURITY_MANAGER` to `null`.
- otherwise proceed as before.

neboskreb:
Ah, I see your point. Let me quickly draft the solution.

neboskreb:
Okay, so checking the runtime platform appears ugly, because it requires copying the code from `SystemUtils` (which is available in `core` but not here in `api`) or referring to `core` as dependency - neither being a great idea... 

Therefore I decided, instead, to go with checking the capability of the instantiated `PrivateSecurityManager`. This looks more robust, because it's not the platform (we wouldn't be excited to add check `isOsXxxx` for every possible beast in the zoo) but the _capability_ we're interested in.

I have completely rewritten the solution, and replaced the PR description.
@ppkarwasz please check and say what you think?

ppkarwasz:
> Okay, so checking the runtime platform appears ugly, because it requires copying the code from `SystemUtils` (which is available in `core` but not here in `api`) or referring to `core` as dependency - neither being a great idea...

I think that this should be enough:

```java
boolean isAndroid = System.getProperty("java.vendor", "").contains("Android");
```

If we already know the `System.getSecurityManager()` is `null`, we don't even have to use a `try...catch`, since the `getProperty()` method can not throw.

neboskreb:
Re-formatted the code to make Spotless happy. Hopefully, this turns the build green. 

This is the most I can do, I'm sorry.

ppkarwasz (approved):
LGTM, thanks!

ppkarwasz:
@neboskreb,

Thanks! In order to merge the PR:

- All the commits must be signed with a GPG recognized by Github. If you didn't do it before, could you [add a GPG key to your Github account](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account) and then rebase this PR, so all commits get signed?
- Could you [add a changelog file](https://logging.apache.org/log4j/tools/log4j-changelog.html#changelog-entry-file) to `src/changelog/.2.x.x`? You can use one of the files already there as template.

neboskreb:
> * Could you [add a changelog file](https://logging.apache.org/log4j/tools/log4j-changelog.html#changelog-entry-file) to `src/changelog/.2.x.x`? You can use one of the files already there as template.

Done

> * All the commits must be signed with a GPG recognized by Github. If you didn't do it before, could you [add a GPG key to your Github account](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account) and then rebase this PR, so all commits get signed?

Hmmm, `git log --show-signature 40d66ed7` shows that the commit is signed but I don't see the checkmark next to it here on Github... Is this normal? If not, what can I do to rectify?

ppkarwasz:
I don't see the signatures, I even checked the raw data.
Try

```sh
git rebase -f -i -S 2.x
git push -f
```

neboskreb:
This helped!
Thanks a lot!

ppkarwasz:
@neboskreb,

Thanks a lot! Now the Android support is almost as good as the JVM support! :100:

IIRC there is still one location-related feature that breaks on Android: `LogManager.getLogger()` will fail, since the Java 8 version of `StackLocator` relies on the absent `sun.reflect.Reflection` class and Android does not support Multi-Release JARs, so it doesn't use the Java 11 version of `StackLocator`.

neboskreb:
> Thanks a lot! Now the Android support is almost as good as the JVM support!

It was my honour to contribute!


> IIRC there is still one location-related feature that breaks on Android: `LogManager.getLogger()` will fail, since the Java 8 version of `StackLocator` relies on the absent `sun.reflect.Reflection` class and Android does not support Multi-Release JARs, so it doesn't use the Java 11 version of `StackLocator`.
> 
> If you know of a smart way to fix that, let me know.

If you mean [this place](https://github.com/apache/logging-log4j2/blob/db0360d4ce1e5ba5a3d9409da79695f54e857857/log4j-api/src/main/java/org/apache/logging/log4j/LogManager.java#L543) then, in my humble opinion, the culprit is lack of flexibility in `StackLocatorUtil`. 

If I was to refactor it, I would employ the Strategy pattern in `StackLocatorUtil` rather than adding more code to `StackLocator`. This way, while initializing `StackLocatorUtil`, I would inject the existing `StackLocator` for JVM platforms and `AndroidStackLocator` for Android.

The easiest way to fill the stack is `new Throwable().getStackTrace()`, which is already used this way in `StackLocator#calcLocation` - though this can be slow. I'm not aware of faster ways - I'm not deep enough into Java mechanics, sorry.

I see 2 implementations of `StackLocator`, with the newer one relying on `StackWalker`. It is [available on Android](https://developer.android.com/reference/java/lang/StackWalker), though on the most recent versions only (Android 14+). This way, the `StackLocator` strategy can be used on newer devices and on JVM while falling back to the slower `AndroidStackLocator` on older devices.

Not sure how this solution aligns with your current packaging strategy (because you mentioned multi-release JAR). Take it as a wild idea.

ppkarwasz:
> I see 2 implementations of `StackLocator`, with the newer one relying on `StackWalker`. It is [available on Android](https://developer.android.com/reference/java/lang/StackWalker), though on the most recent versions only (Android 14+). This way, the `StackLocator` strategy can be used on JVM and newer devices while falling back to the slower `AndroidStackLocator` on older devices.
> 
> Not sure how this solution aligns with your current packaging strategy (because you mentioned multi-release JAR). Take it as a wild idea.

One of the two versions of `StackLocator` is in `META-INF/versions/9`, so Android can not use it. Maybe there is an option to use that class when compiling, but I never found it.

> If I was to refactor it, I would employ the Strategy pattern in `StackLocatorUtil` rather than adding more code to `StackLocator`. This way, while initializing `StackLocatorUtil`, I would inject the existing `StackLocator` for JVM platforms and `AndroidStackLocator` for Android.

Yes, probably this is what we should have used from the start.

Issue #3206, opened 2024-11-14
NPE when using `rootLogger` shorthand properties notation


ayush11-96:
Currently we are using log4j2 version 2.17.0 with spring boot but while updating log4j2 to 2.17.2 we are getting below error and which is because of the appender of root is initialise to null
`Caused by: java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "org.apache.logging.log4j.core.config.LoggerConfig.getAppenderRefs()" is null`

We are passing below properties:
```
rootLogger.appenderRef.console.ref=console
rootLogger.appenderRefs=console
rootLogger=INFO
```

Please help us to understand if we require some changes along with version upgrade?

ppkarwasz:
@ayush11-96,

Does the problem also occur with version `2.24.1`?
Can you provide the full stack trace of the exception?

ayush11-96:
@ppkarwasz 
I tried till 2.23.1 version and this is still happening
**Stack Trace:**
`Caused by: java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "org.apache.logging.log4j.core.config.LoggerConfig.getAppenderRefs()" is null
	at org.apache.logging.log4j.core.config.AbstractConfiguration.doConfigure(AbstractConfiguration.java:740) ~[log4j-core-2.23.1.jar:2.23.1]
	at org.apache.logging.log4j.core.config.AbstractConfiguration.initialize(AbstractConfiguration.java:264) ~[log4j-core-2.23.1.jar:2.23.1]
	at org.apache.logging.log4j.core.config.AbstractConfiguration.start(AbstractConfiguration.java:313) ~[log4j-core-2.23.1.jar:2.23.1]
	at org.apache.logging.log4j.core.LoggerContext.setConfiguration(LoggerContext.java:631) ~[log4j-core-2.23.1.jar:2.23.1]
	at org.apache.logging.log4j.core.LoggerContext.start(LoggerContext.java:292) ~[log4j-core-2.23.1.jar:2.23.1]`

ayush11-96:
@ppkarwasz could u please elaborate the issue if u got that?

ppkarwasz:
@ayush11-96,

[...]

PR #3650, opened 2025-05-06
Prevent NPE with rootLogger shorthand when no appenders are specified

jhl221123:
This pull request addresses issue #3206, which could lead to a `NullPointerException` when using the shorthand notation for the root logger (e.g., rootLogger = INFO) in `log4j2.properties` without explicitly defining any appender references.

To resolve this, the static helper method `LoggerConfig.getLevelAndRefs()` has been updated to ensure that the `LevelAndRefs` container always holds a non-null list of `AppenderRef` objects. When no appenders are specified in the shorthand string, the method has been updated to return an empty list instead of `null`.

## Checklist

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/LoggerConfigTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/LoggerConfigTest.java
@@ -17,9 +17,11 @@
 package org.apache.logging.log4j.core.config;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNotSame;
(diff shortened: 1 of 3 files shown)


ppkarwasz (approved):
Looks good to me!

Can you add a changelog entry to the `src/changelog/.2.x.x` folder. The `type` of the entry should be `fixed`.

jhl221123:
Thanks, @ppkarwasz! I've added the changelog entry. I appreciate your guidance and review!

ppkarwasz:
> I've added the changelog entry. I appreciate your guidance and review!

Thanks!

For some reason I don't fully understand the required GitHub Action workflows didn't start on your last commit.

- @vy any idea why?
- @jhl221123, could you try updating the branch again, e.g. `git rebase -f origin/2.x; git push -f`? Maybe it is a temporary problem and it will work for the next commit.

PR #3657, opened 2025-05-09
Add Max Base as collaborator

ppkarwasz:
I propose to add @basemax as collaborator, to help with bug triage and project management.
Max is a very prolific Open Source contributor and have helped many projects (see [his profile](https://github.com/basemax)).

**Note**: To prevent misunderstandings, "contributor" is not the same thing as "committer" and IMHO this PR does not require a PMC vote.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -55,6 +55,9 @@ github:
     - logger
     - api
     - syslog
+  # People helping with triaging and Project Management
+  collaborators:
+    - basemax
 
   # Enforce squashing while merging PRs.
   # Otherwise, the git log gets polluted severely.


Relaxx422 (approved):

ppkarwasz:
The triaging role in GitHub is so limited, that I don't think we should set the bar too high to assign it.

However, since triagers [can not do many simple things like creating labels](https://github.com/orgs/community/discussions/5773), we should probably first sit down and add some order to the labels we are using, before wasting time assigning labels that will be removed.

If we want to use GitHub Projects and the new GitHub features:

- labels that indicate the type of issue report (`bug`, `enhancement`) can be replaced with GitHub issue types.
- labels that indicate the affected Log4j component, can be replaced with a custom field in GitHub Projects.
- labels that indicate the state of the PR or issue report can be replaced with fields in GitHub Projects.

We should probably start by specifying the set of all possible states for an issue report and a PR. I'll try to make a draft state diagram and then we'll schedule a call with @BaseMax for some feedback.

ppkarwasz:
So much time has passed on this, that we should probably close this.

@BaseMax: thanks for your help offer last year. It really lifted our spirits in a difficult moment.

PR #3658, opened 2025-05-11
fix toString() method in configuration implementations (apache#3599)

anindita-sarkarArray:
[A clear and concise description of what the pull request is for along with a reference to the associated issue IDs, if they exist.]

Added toString implementation to AbstractConfiguration and removed from all derived classes.

## Checklist


garydgregory:
Why? This makes debugging worse as we loosing information.

anindita-sarkarArray:
> Why? This makes debugging worse as we loosing information.

hi @garydgregory , instead of AbstractConfiguration  , shall I put my changes to all derived classes so that it can retain the additional info of subclasses.

garydgregory:
I am still -1 here: `toString()` is for debugging. If a call site needs a different view on the object, then it can call another method on that object or build that view (like a String) itself.

PR #3668, opened 2025-05-16
Don't test reproducibility on Windows

ppkarwasz:
Due to CRLF vs. LF differences, artifacts generated on Linux cannot easily be reproduced on Windows.

We disable the reproducibility test on Windows.


FreeAndNil (approved):
LGTM

ppkarwasz:
@FreeAndNil, thanks!

PR #3675, opened 2025-05-19
StringMap: Generalize `equals`/`hashCode` across implementations

### LINKED ISSUE #3669: Different `SimpleMap` implementations are never equal
## Description

An equality comparison between a `SortedArrayStringMap` and a `JdkMapAdapterStringMap` is guaranteed to return `false`:

```java
public class JdkMapAdapterStringMap implements StringMap {
    @Override
    public boolean equals(final Object object) {
        if (object == this) {
            return true;
        }
        if (!(object instanceof JdkMapAdapterStringMap)) {
            return false;
        }
```

```java
public class SortedArrayStringMap implements IndexedStringMap {
    @Override
    public boolean equals(final Object obj) {
        if (obj == this) {
            return true;
        }
        if (!(obj instanceof SortedArrayStringMap)) {
            return false;
        }
```

This is a bummer for unit testing, because otherwise identical maps can be considered unequal depending upon how they were constructed and whether they were copied or just wrapped. The recent efficiency changes in the context map implementation seem to have brought this problem to the fore; evidently there are cases now where immutable maps are wrapped whereas previously they would have been copied.

rschmitt:
This change rewrites the `equals` and `hashCode` implementations in `SortedArrayStringMap` and `JdkMapAdapterStringMap` to support value-based equality between the two implementations.

Fixes #3669.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
@@ -181,20 +181,24 @@ public class DefaultThreadContextMap implements ThreadContextMap, ReadOnlyString
 
     @Override
     public int hashCode() {
-        final int prime = 31;
-        int result = 1;
-        final Object[] state = localState.get();
-        result = prime * result + ((state == null) ? 0 : getMap(state).hashCode());
-        return result;
+        return toMap().hashCode();
     }
 
     @Override
-    public boolean equals(final Object obj) {
+    public boolean equals(Object obj) {
         if (this == obj) {
             return true;
         }
         if (obj == null) {
             return false;
+        }
+        if (obj instanceof ReadOnlyStringMap) {
+            if (size() != ((ReadOnlyStringMap) obj).size()) {
+                return false;
+            }
+
+            // Convert to maps and compare
+            obj = ((ReadOnlyStringMap) obj).toMap();
         }
         if (!(obj instanceof ThreadContextMap)) {
             return false;
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LogEventWrapper.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/bridge/LogEventWrapper.java
@@ -213,5 +213,24 @@ public class LogEventWrapper implements LogEvent {
         public <V> V getValue(final String key) {
             return (V) super.get(key);
         }
+
+        @Override
+        public boolean equals(final Object obj) {
+            if (obj == this) {
+                return true;
+            }
+            if (obj instanceof ReadOnlyStringMap) {
+                // Convert to maps and compare
+                final Map<String, String> thisMap = toMap();
+                final Map<String, String> otherMap = ((ReadOnlyStringMap) obj).toMap();
+                return thisMap.equals(otherMap);
+            }
+            return super.equals(obj);
+        }
+
+        @Override
+        public int hashCode() {
+            return toMap().hashCode();
+        }
     }
 }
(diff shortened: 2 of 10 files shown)


vy:
Wouldn't it be better to

1. `ReadOnlyStringMapUtil` *internal* class such that
   * It contains a static `equals(ReadOnlyStringMap,ReadOnlyStringMap)` method
   * It contains a `hashCode(ReadOnlyStringMap)` method
2. Use `ReadOnlyStringMapUtil` to implement `equals` and `hashCode` in *all* `ReadOnlyStringMap` implementations

rschmitt:
@vy OK

rschmitt:
What should I do about these baseline failures?

vy:
@rschmitt, bump the `@Version` tag to `2.25.0` in `package-info.java` files located where you touched files. (Just answering your question, next week I will review your PR in detail.)

log4j-api/src/main/java/org/apache/logging/log4j/util/ReadOnlyStringMapUtil.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Map;
+
+/**
+ * Utility class for ReadOnlyStringMap implementations. Provides methods for equals and hashCode calculations.
+ *
+ * @since 2.25.0
+ */
+@InternalApi
+public final class ReadOnlyStringMapUtil {
```
ppkarwasz, line 27: I'm a bit concerned about introducing a new public class in `log4j-api`, especially since it's primary purpose is to solve a testing problem. While standardizing `ReadOnlyStringMap` behavior has value, doing so in the Log4j API introduces a version dependency: `log4j-core` 2.25.x would no longer be compatible with `log4j-api` 2.24.x.

To avoid this incompatibility, what do you think about implementing `equals()` and `hashCode()` directly like this instead?

```java
@Override
public boolean equals(final Object obj) {
    if (obj == this) {
        return true;
    }
    if (obj instanceof ReadOnlyStringMap) {
        return toMap().equals(((ReadOnlyStringMap) other).toMap());
    }
    return false;
}

@Override
public int hashCode() {
    return toMap().hashCode();
}
```

This would preserve compatibility while still ensuring consistent behavior in tests.

log4j-api/src/main/java/org/apache/logging/log4j/spi/DefaultThreadContextMap.java
```diff
@@ -196,6 +193,9 @@ public boolean equals(final Object obj) {
         if (obj == null) {
             return false;
         }
+        if (obj instanceof ReadOnlyStringMap) {
+            return ReadOnlyStringMapUtil.equals(this, (ReadOnlyStringMap) obj);
+        }
```
ppkarwasz, line 198: `DefaultThreadContextMap` is primarily an implementation of `ThreadContextMap`. I am not sure if the fact it implements `ReadOnlyStringMap` is ever used.

ppkarwasz (changes requested):
Thanks everyone for tackling this pull request!

Just to add my perspective: while `ReadOnlyStringMapUtil` initially seemed like a solid solution, introducing it creates a binary incompatibility between `log4j-core` 2.25.x and `log4j-api` 2.24.x. Since the main goal of #3669 is to support unit testing, it might be preferable to avoid that incompatibility and take a simpler—albeit slightly less efficient—approach by using `ReadOnlyStringMap.toMap()` within the `equals()` and `hashCode()` implementations.

log4j-api/src/main/java/org/apache/logging/log4j/util/ReadOnlyStringMapUtil.java
rschmitt: I'm relieved to hear you say this

PR #3676, opened 2025-05-19
Fix mvnw on macOS Sequoia

rschmitt:
This change pulls in a fix from a newer version of the maven-wrapper-plugin that fixes a bug in verifying the distributionSha256Sum using the version of `sha256sum` that ships with macOS Sequoia.

See:

* https://issues.apache.org/jira/browse/MWRAPPER-150
* https://github.com/apache/maven-wrapper/commit/b2ff9cdfa2d4a1f2bc0d4f039c0efc297b0c5875

### DIFF
--- a/mvnw
+++ b/mvnw
@@ -227,7 +227,7 @@ if [ -n "${distributionSha256Sum-}" ]; then
     echo "Please disable validation by removing 'distributionSha256Sum' from your maven-wrapper.properties." >&2
     exit 1
   elif command -v sha256sum >/dev/null; then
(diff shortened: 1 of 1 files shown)


vy (approved):
Let's not piss off contributors using macOS Sequoia, but let's also keep an eye on the upcoming Maven Wrapper `3.3.3` release.

ppkarwasz:
I would rather update the whole `mvnw` file to a new version.

rschmitt:
@ppkarwasz I tried that first. The fix isn't available yet in any released version of the wrapper plugin.

PR #3681, opened 2025-05-21
Expunge stale entries in InternalLoggerRegistry

### LINKED ISSUE #3430: Expunge stale entries in `InternalLoggerRegistry`
`InternalLoggerRegistry` is essentially a `Map<MessageFactory, Map<String, WeakReference<Logger>>>`. [As reported](/apache/logging-log4j2/issues/3399#issuecomment-2626946587) by @tristantarrant, stale entries are not expunged in its current form. That is, the inner map does not get compacted for entries whose associated `WeakReference`s are reclaimed by the garbage collector. This might cause high memory usage if an excessive amount of ephemeral loggers get registered.

Reclaimed references can be tracked by a `ReferenceQueue` passed to the `WeakReference::new` and processed as follows:

```java
private final ReferenceQueue<Logger> staleLoggerRefs = new ReferenceQueue<>();

private void expungeStaleEntries() {
    Reference<? extends Logger> loggerRef;
    while ((loggerRef = staleLoggerRefs.poll()) != null) {
        Logger logger = loggerRef.get();
        if (logger != null) {
            removeLogger(logger);
        }
    }
}
```

When to run `expungeStaleEntries()` should be decided with care, since `removeLogger()` mutates the registry map.

### On method invocation

All `ILR` methods (`getLogger()`, `computeIfAbsent()`, etc.) can first try to expunge stale entries.
This approach is exercised by `WeakHashMap` itself and in #3427.

**Pros:**

- Implies no changes besides `ILR`

**Cons:**

- Read-only operations, which are accessed most, might try to acquire the write-lock.
- If `ILR` is not accessed, stale entries will still be kept. Consider an application where all `Logger`s are assigned, and a temporary operation creates excessive amount of loggers and completes. All created temporary `Logger` instances will be reclaimed by GC, yet will continue taking space for the associated `Map.Entry<String, WeakReference<Logger>>`.

### By a janitor thread

`ILR` can start a janitor thread listening on the `ReferenceQueue<Logger>`.

**Cons:**

- `ILR` must have a life cycle (to start/stop the janitor thread) managed by `LoggerContext`.

jhl221123:
This pull request addresses issue #3430 by implementing a mechanism to expunge stale entries in `InternalLoggerRegistry`. 

It builds upon PR #3474 and incorporates feedback from @vy and @ppkarwasz.

Key changes include:
*   **Enabled stale entry detection via `ReferenceQueue`**: When a `Logger` is GC'd, its `WeakReference` is enqueued. Subsequent calls to methods like `getLogger()` or `computeIfAbsent()` trigger `expungeStaleEntries()` to process it.
*   **Implemented `expungeStaleEntries` logic**: If the `ReferenceQueue` is not empty, a write lock is acquired, the queue is drained, and map entries are scanned once to remove `WeakReferences` with null referents and clean up any empty maps. No action or locking occurs if the queue is initially empty.
*   **Added unit tests**: Verified the expunging mechanism under various conditions, including GC scenarios and concurrent access.

@vy, `ReferenceQueue` has no `peek()`/`isEmpty()`, so I couldn’t do a proper double-check and used an initial `poll()` instead. Feedback is welcome if there's a better way or if I missed something. Thanks!

Fixes #3430

### DIFF
--- a/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
+++ b/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
@@ -16,7 +16,12 @@
  */
 package org.apache.logging.log4j.core.test.util;
 
-import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.lang.ref.WeakReference;
 import java.util.concurrent.CountDownLatch;
--- a/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
+++ b/log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.test.util;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.lang.ref.WeakReference;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import org.apache.logging.log4j.core.Logger;
+import org.apache.logging.log4j.core.LoggerContext;
+import org.apache.logging.log4j.core.util.internal.InternalLoggerRegistry;
+import org.apache.logging.log4j.message.MessageFactory;
+import org.apache.logging.log4j.message.SimpleMessageFactory;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class InternalLoggerRegistryTest {
+    private InternalLoggerRegistry registry;
+    private MessageFactory messageFactory;
+
+    @BeforeEach
+    void setUp() {
+        registry = new InternalLoggerRegistry();
+        messageFactory = new SimpleMessageFactory();
+    }
+
+    @Test
+    void testGetLoggerReturnsNullForNonExistentLogger() {
+        assertNull(registry.getLogger("nonExistent", messageFactory));
+    }
+
+    @Test
+    void testComputeIfAbsentCreatesLogger() {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+        assertNotNull(logger);
+        assertEquals("testLogger", logger.getName());
+    }
+
+    @Test
+    void testGetLoggerRetrievesExistingLogger() {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+        assertSame(logger, registry.getLogger("testLogger", messageFactory));
+    }
+
+    @Test
+    void testHasLoggerReturnsCorrectStatus() {
+        assertFalse(registry.hasLogger("testLogger", messageFactory));
+        registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                .getLogger(name, factory));
+        assertTrue(registry.hasLogger("testLogger", messageFactory));
+    }
+
+    @Test
+    void testExpungeStaleEntriesRemovesGarbageCollectedLoggers() throws InterruptedException {
+        Logger logger =
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+
+        WeakReference<Logger> weakRef = new WeakReference<>(logger);
+        logger = null; // Dereference to allow GC
+
+        // Retry loop to give GC time to collect
+        for (int i = 0; i < 10; i++) {
+            System.gc();
+            Thread.sleep(100);
+            if (weakRef.get() == null) {
+                break;
+            }
+        }
+
+        // Access the registry to potentially trigger cleanup
+        registry.computeIfAbsent("tempLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                .getLogger(name, factory));
+
+        assertNull(weakRef.get(), "Logger should have been garbage collected");
+        assertNull(
+                registry.getLogger("testLogger", messageFactory), "Stale logger should be removed from the registry");
+    }
+
+    @Test
+    void testConcurrentAccess() throws InterruptedException {
+        int threadCount = 10;
+        ExecutorService executor = Executors.newFixedThreadPool(threadCount);
+        CountDownLatch latch = new CountDownLatch(threadCount);
+
+        for (int i = 0; i < threadCount; i++) {
+            executor.submit(() -> {
+                registry.computeIfAbsent("testLogger", messageFactory, (name, factory) -> LoggerContext.getContext()
+                        .getLogger(name, factory));
+                latch.countDown();
+            });
+        }
+
+        latch.await();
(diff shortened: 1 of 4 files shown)


log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Test
+    void testGetLoggerReturnsNullForNonExistentLogger() {
+        assertNull(registry.getLogger("nonExistent", messageFactory));
+    }
+
+    @Test
+    void testComputeIfAbsentCreatesLogger() {
+        Logger logger = registry.computeIfAbsent(
+                "testLogger", messageFactory, (name, factory) -> loggerContext.getLogger(name, factory));
```
ppkarwasz, line 74: `loggerContext.getLogger` calls back into `InternalLoggerRegistry`, which means we're triggering nested `computeIfAbsent` calls here. Is this behavior intentional? If not, it might be safer to use a mocked `Logger` to avoid unintended side effects or recursion.

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
```diff
@@ -0,0 +1,180 @@
...
+            assertTrue(
+                    isExpungeStaleEntries,
+                    "Stale WeakReference entries were not removed from the inner map for MessageFactory");
+        });
+    }
+
+    @Test
+    void testExpungeStaleMessageFactoryEntry() {
+        Logger logger = registry.computeIfAbsent(
+                "testLogger", messageFactory, (name, factory) -> loggerContext.getLogger(name, factory));
```
ppkarwasz, line 131: The `messageFactory` parameter is defined as `SimpleMessageFactory.INSTANCE` and therefore can not be garbage collected. Should we use a different `messageFactory` for this test case?

ppkarwasz (changes requested):
@jhl221123, thank you for picking up this pull request!

The main code looks good to me overall. I do have a few suggestions regarding the unit tests:

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
jhl221123: Not intended! The logger supplier lambdas now use `new MockLogger(...)` directly.

log4j-core-test/src/main/java/org/apache/logging/log4j/core/test/util/InternalLoggerRegistryGCTest.java
jhl221123: Good point! That test case now uses a locally created `new SimpleMessageFactory()` instance.

ppkarwasz (approved):
The changes look solid—nice work! There's just one final task before we can wrap this up: could you please add a changelog entry under `src/changelog/.2.x.x`, as outlined in the [changelog documentation](https://logging.apache.org/log4j/tools/log4j-changelog.html#changelog-entry-file)?

Also, just a quick note on `loggerRefByNameByMessageFactory`: since it's a `WeakHashMap`, it doesn’t follow all the typical `Map` invariants. From what I can tell, you're using it safely, and I didn’t spot any issues that would lead to unexpected behavior. That said, given the sensitivity of this part of the codebase, I’d like a second pair of eyes on it.

@vy, could you take a moment to review this as well?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistry.java
```diff
@@ -53,13 +58,43 @@ public final class InternalLoggerRegistry {
...
+    /**
+     * Expunges stale entries for logger references and message factories.
+     */
+    private void expungeStaleEntries() {
+        Reference<? extends Logger> loggerRef = staleLoggerRefs.poll();
+
+        if (loggerRef != null) {
+            writeLock.lock();
+            try {
+                while (staleLoggerRefs.poll() != null) {
```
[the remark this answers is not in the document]
ppkarwasz: False positive, the comment is already there.

vy:
I would like to and I will. Please allow me some time.

Op vr 23 mei 2025 om 21:27 schreef Piotr P. Karwasz <
***@***.***>

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+            throws NoSuchFieldException, IllegalAccessException {
+        Field loggerMapField = registry.getClass().getDeclaredField("loggerRefByNameByMessageFactory");
+        loggerMapField.setAccessible(true);
+        @SuppressWarnings("unchecked")
+        Map<MessageFactory, Map<String, WeakReference<Logger>>> loggerMap =
+                (Map<MessageFactory, Map<String, WeakReference<Logger>>>) loggerMapField.get(registry);
+        return loggerMap;
+    }
+
+    private class MockLogger extends Logger {
```
vy, line 183: I don't think you really need this definition. All your `new MockLogger(...)` calls can be replaced with `new Logger(...) {}` instead.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
vy, line 1: Can you use `final` modifier for variables wherever applicable, please?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+            Logger logger = registry.computeIfAbsent(
+                    loggerNamePrefix + i,
+                    messageFactory,
+                    (name, factory) -> new MockLogger(loggerContext, name, factory));
+            logger.info("Using logger {}", logger.getName());
+        }
+
+        await().atMost(10, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
+            System.gc();
+            System.runFinalization();
```
vy, line 108: Don't use this deprecated-for-removal method, please.
```suggestion
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+            boolean isExpungeStaleEntries = true;
+            for (int i = 0; i < numberOfLoggers; i++) {
+                if (loggerRefByName.containsKey(loggerNamePrefix + i)) {
+                    isExpungeStaleEntries = false;
+                    break;
+                }
+            }
+            assertTrue(
+                    isExpungeStaleEntries,
+                    "Stale WeakReference entries were not removed from the inner map for MessageFactory");
```
vy, line 125: Instead of this, could you _count_ the number of unexpected logger entries and assert that they are equal to zero, and if not, report how many are found, please?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+    void testExpungeStaleMessageFactoryEntry() {
+        SimpleMessageFactory mockMessageFactory = new SimpleMessageFactory();
+        Logger logger = registry.computeIfAbsent(
+                "testLogger", mockMessageFactory, (name, factory) -> new MockLogger(loggerContext, name, factory));
+        logger.info("Using logger {}", logger.getName());
+        logger = null;
+
+        await().atMost(10, SECONDS).pollInterval(100, MILLISECONDS).untilAsserted(() -> {
+            System.gc();
+            System.runFinalization();
```
vy, line 139:
```suggestion
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
```diff
@@ -0,0 +1,189 @@
...
+            Map<MessageFactory, Map<String, WeakReference<Logger>>> loggerRefByNameByMessageFactory =
+                    reflectAndGetLoggerMapFromRegistry();
+            assertNull(
+                    loggerRefByNameByMessageFactory.get(mockMessageFactory),
+                    "Stale MessageFactory entry was not removed from the outer map");
+        });
+    }
+
+    @Test
+    void testConcurrentAccess() throws InterruptedException {
```
vy, line 151: @jhl221123, @ppkarwasz, do we have a inspiration source (scenario, bug report, etc.) for this particular test?

vy (changes requested):

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/internal/InternalLoggerRegistryTest.java
jhl221123: @vy, Thanks for the review!
I couldn't think of a specific scenario, so I've removed it for now. Please let me know if anything changes!

ppkarwasz: We've previously seen issues with `InternalLoggerRegistry` under concurrent usage—see #3252 and #3399—caused by recursive calls from third-party code. I believe this was addressed in #3418, though I didn’t add a regression test due to the difficulty of reproducing static initialization edge cases.

That said, since `InternalLoggerRegistry` replaced `ConcurrentHashMap` (multi-lock) with a single lock, it might be worth adding a benchmark to assess any performance regression.

vy (approved):

Issue #3176, opened 2024-11-05
Validate scripts in global `<Scripts>` container


ppkarwasz:
The [`ScriptsPlugin`](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ScriptsPlugin.java) should perform a validation of the script nested configuration elements and:

- Remove the scripts that do not have a name.
- Issue a status logger warning, when this condition occurs.

JWT007:
Hi @ppkarwasz,

So I was going to fix this but in testing realized that no script is every really created with no name.

The super-constructor for Script/ScriptRef/ScriptFile when passed a `null` name evaluates a default from `toString()`.

```
public AbstractScript(final String name, final String language, final String scriptText) {
    this.language = language;
    this.scriptText = scriptText;
    this.name = name == null ? this.toString() : name;
}
```

Example: `org.apache.logging.log4j.core.script.Script@7857fe2`.

So by the time the array of scripts shows up in the `ScriptsPlugin` there are no scripts without names.

JWT007:
@ppkarwasz what should be done with this ticket? see last comment.

Of course, the Script builders and AbstractScript constructor can be modified...

ppkarwasz:
> @ppkarwasz what should be done with this ticket? see last comment.
> 
> Of course, the Script builders and AbstractScript constructor can be modified...

Since `AbstractScript.getName()` is used extensively accross the code, I believe that we need an additional property (`getConfiguredName()`?) to allow the `ScriptsPlugin` to remove the scripts without an explicit name.

ppkarwasz:
> If adding `configuredName` to `AbstractScript` and filtering in `ScriptsPlugin` is still the plan, I'd like to work on it. Is this up for grabs?

Hi @jhl221123,
Thanks for offering to take this on — I’ve assigned the issue to you.

There’s no finalized solution yet, so you have flexibility in how you want to approach it. Since this issue is part of a broader problem involving script referencing in configuration files, I’ve outlined the overall context and a couple of potential solution paths below. These are just suggestions to help guide your thinking — feel free to build on them or propose your own approach.

## Problem Statement

We aim to ensure consistent and reliable handling of script references within configuration files. This involves two related sub-problems:

* **Part 1 (this issue):**
  Enforce that all children of the global `<Scripts>` container define a user-visible `name`, enabling them to be referenced from other parts of the configuration.

* **Part 2 (issue #3336):**
[...]

PR #3691, opened 2025-05-26
Enforce explicit names for global scripts and refine script id handling

jhl221123:
This PR addresses issue #3176 by ensuring scripts defined in the global `<Scripts>` container have explicit, non-blank names. It also introduces a clearer distinction between user-defined names and internal script identifiers.

**Key Changes:**
* `AbstractScript` now provides `getName()` strictly for user-configured names, which can be null or blank. It also introduces `getId()` to consistently return a non-blank internal identifier, falling back to `toString()` if no explicit name is set.
* `ScriptsPlugin` now throws a `ConfigurationException` if any script defined directly under the global `<Scripts>` element lacks an explicit name.
* `ScriptManager` and various script-using components were updated. They now utilize the `script.getId()` for internal referencing.
* Added unit tests for `AbstractScript` and `ScriptsPlugin`.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/script/package-info.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/script/package-info.java
@@ -18,7 +18,7 @@
  * Log4j 2 Script support.
  */
 @Export
-@Version("2.20.2")
+@Version("2.25.0")
 package org.apache.logging.log4j.core.script;
 
 import org.osgi.annotation.bundle.Export;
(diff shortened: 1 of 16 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/script/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Log4j 2 Script support.
  */
 @Export
-@Version("2.20.2")
+@Version("2.25.0")
```
ppkarwasz, line 21:
```suggestion
@Version("2.26.0")
```
Tonight I'll release `2.25.0`, so the next minor version will become `2.26.0`.

ppkarwasz (approved):
Hi @jhl221123,

Thanks for the PR, and apologies for the delay—we're a bit short on bandwidth at the moment.

You've done a great job here. The PR includes all the necessary elements, and I appreciate that you followed our custom policy for `@Version` annotations as well. Nicely done!

PR #3703, opened 2025-05-31
Add option to provide URIs to monitor in addition to the config file

vy:
This PR is an identical clone of #3501, though additionally adds commit signatures. Created this PR instead of force-pushing signatures to that PR to avoid truncating the git and GitHub conversations history.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/MonitorResourcesTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/MonitorResourcesTest.java
@@ -1,0 +1,123 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.waitAtMost;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import org.apache.logging.log4j.core.config.Configuration;
+import org.apache.logging.log4j.core.config.ConfigurationSource;
+import org.apache.logging.log4j.core.config.Configurator;
+import org.apache.logging.log4j.core.config.builder.api.ComponentBuilder;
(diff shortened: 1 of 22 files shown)


log4j-core-test/src/test/java/org/apache/logging/log4j/core/MonitorResourcesTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.CleanupMode;
+import org.junit.jupiter.api.io.TempDir;
+
+public class MonitorResourcesTest {
+
+    @Test
+    void test_reconfiguration(@TempDir(cleanup = CleanupMode.ON_SUCCESS) final Path tempDir) throws IOException {
+        final ConfigurationBuilder<PropertiesConfiguration> configBuilder =
+                ConfigurationBuilderFactory.newConfigurationBuilder(PropertiesConfiguration.class);
```
ppkarwasz, line 49: _Nit_: Why don't we just use the default `BuiltConfiguration`?

```suggestion
        final ConfigurationBuilder<PropertiesConfiguration> configBuilder =
                ConfigurationBuilderFactory.newConfigurationBuilder();
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/api/ConfigurationBuilder.java
```diff
@@ -61,6 +61,15 @@ public interface ConfigurationBuilder<T extends Configuration> extends Builder<T
...
     ConfigurationBuilder<T> add(CustomLevelComponentBuilder builder);
 
+    /**
+     * Adds a top level component.
+     * @param builder The ComponentBuilder with all of its attributes and sub components set.
+     * @return this builder instance.
+     */
+    default ConfigurationBuilder<T> addComponent(ComponentBuilder<?> builder) {
+        throw new UnsupportedOperationException();
+    }
```
ppkarwasz, line 71: Can you add a `@since` tag?

Instead of adding a default method that throws an exception, I would prefer introducing an abstract method and annotating the class with [@ProviderType](https://docs.osgi.org/javadoc/osgi.annotation/7.0.0/org/osgi/annotation/versioning/ProviderType.html).

From the perspective of consumers of `ConfigurationBuilder`, the behavior remains effectively the same: if an implementation does not define the method, it will fail at runtime (via either `AbstractMethodError` or `UnsupportedOperationException`). However, for implementors of `ConfigurationBuilder`, declaring it as an abstract method is more explicit and has the advantage of failing at compile time, prompting them to implement the method directly.

ppkarwasz (approved):
Looks good to me, thanks!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/MonitorResourcesTest.java
vy: Fixed in 2fea3657a6507069d997ecaefc94e8e013d314d5.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/builder/api/ConfigurationBuilder.java
vy: Added the `@since` tag in 2fea3657a6507069d997ecaefc94e8e013d314d5.

> I would prefer introducing an abstract method

This is an `interface`, not an `abstract class`, do you imply not providing any default methods at all?

ppkarwasz: Yes, that would be my preferred option.

If someone implemented `ConfigurationBuilder`, he can expect compilation errors after a minor version bump.

vy: It causes `bnd-baseline:baseline` to fail, i.e., that'd be a backward incompatible change.

PR #3719, opened 2025-06-06
Refactor StatusLogger to use precompiled Pattern constants

kamilkrzywanski:
We should improve the implementation of StatusLogger by extracting all regular expressions used for property normalization and matching into clearly named, precompiled Pattern constants. This change would:

Avoid repeated calls to Pattern.compile(...), improving performance.
Increase readability and maintainability by clearly naming intent of regex patterns.


vy:
> We should improve the implementation of StatusLogger by extracting all regular expressions used for property normalization and matching into clearly named, precompiled Pattern constants.

Should we? Aren't these only used once during the `Config` singleton initialization? Would you mind sharing the performance problem you're having?

PR #3785, opened 2025-06-25
fix: Define version of `error_prone_annotations` in parent POM

### LINKED ISSUE #3779: Missing property in the log4j-parent pom file in version 2.25.0
## Description

`error.prone.version` property is being used in log4j-parent pom.xml file but this property in not defined  and because of this log files are not generated for the spring boot application. 

## Configuration

**Version:** 
Log4j 2.25.0
springboot : 2.7.18

**Operating system:** Ubuntu24.04

**JDK:** JDK 21.0.3

## Logs

```
'dependencies.dependency.version' for com.google.errorprone:error_prone_annotations:jar must be a valid version but is '${error-prone.version}'. in org.apache.logging.log4j:log4j-web:2.25.0
```

ppkarwasz:
This change explicitly defines the version of the `error_prone_annotations` dependency in the `log4j` parent POM.

Previously, the version was inherited via the `error-prone.version` property from `logging-parent`. However, because the `log4j-bom` POM is flattened during publication, this property no longer resolves in consumer projects.

Fixes #3779

### DIFF
--- a/log4j-parent/pom.xml
+++ b/log4j-parent/pom.xml
@@ -80,6 +80,7 @@
     <conversant.disruptor.version>1.2.15</conversant.disruptor.version>
     <disruptor.version>3.4.4</disruptor.version>
     <embedded-ldap.version>0.9.0</embedded-ldap.version>
+    <error-prone-annotations.version>2.38.0</error-prone-annotations.version>
     <felix.version>7.0.5</felix.version>
     <groovy.version>4.0.27</groovy.version>
     <guava.version>33.4.8-jre</guava.version>
@@ -1109,7 +1110,7 @@
             <dependency>
               <groupId>com.google.errorprone</groupId>
               <artifactId>error_prone_annotations</artifactId>
-              <version>${error-prone.version}</version>
+              <version>${error-prone-annotations.version}</version>
             </dependency>
             <dependency>
               <groupId>org.osgi</groupId>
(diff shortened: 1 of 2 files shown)


vy (approved):

ppkarwasz:
Unfortunately, this PR did **not** resolve #3779. I mistakenly applied the new `error-prone-annotations.version` property only to the Gradle Module Metadata plugin configuration:

https://github.com/apache/logging-log4j2/blob/65d30ee65e110e0c241c33b34907874b6271f356/log4j-parent/pom.xml#L1110-L1114

The Maven dependency configuration still relies on the inherited `error-prone.version`, which is later trimmed by the Flatten Maven Plugin:

https://github.com/apache/logging-log4j2/blob/65d30ee65e110e0c241c33b34907874b6271f356/log4j-parent/pom.xml#L859-L864

The intention behind introducing a separate `error-prone-annotations.version` property—rather than overriding `error-prone.version`—was to avoid *false positives*, since `error-prone.version` **is** defined during the build but stripped afterward. However, this workaround seems to have backfired.

This highlights a deeper issue: the problem is difficult to reliably reproduce or detect:

* Maven ignores the `provided` scope for dependencies in many contexts.
* Gradle uses its own `.module` metadata and doesn't seem to be affected, as confirmed by the success of our Gradle integration test in [release build 16091508381](https://github.com/apache/logging-log4j2/actions/runs/16091508381).

I'm open to suggestions on how we might better catch issues like this in the future—especially around metadata mismatches and flattened POM artifacts.

PR #3789, opened 2025-06-28
Create `NamedInstantPattern` to make date & time patterns supported by Pattern Layout programmatically accessible

ashr123:
I've created an enum for named patterns, that way I think it will be easier to maintain.
In addition the enum modifier is `public` because it lets users to reuse those patterns in their application

## Checklist

### ✅ Required checks

- [x] **Commit signatures**: All commits are signed and verifiable. (See [GitHub Docs on Commit Signature Verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)).
  <details>

  </details>
  <details>

    Run: `./mvnw verify`

  </details>

I got the following error:

<details>
  <summary>bnd-baseline-maven-plugin failure</summary>
> ```
> [ERROR] Failed to execute goal biz.aQute.bnd:bnd-baseline-maven-plugin:7.1.0:baseline (check-api-compat) on project log4j-core: An error occurred while calculating the baseline: Baseline problems detected. See the report in /Users/royash/IdeaProjects/logging-log4j2/log4j-core/target/baseline/log4j-core-2.26.0-SNAPSHOT.txt.
> [ERROR] ===============================================================
> [ERROR]   Name                                               Type       Delta      New        Old        Suggest    
> [ERROR]   org.apache.logging.log4j.core                      BUNDLE     MAJOR      2.26.0.SNAPSHOT 2.25.0     -          
> [ERROR] ===============================================================
> [ERROR]   Name                                               Type       Delta      New        Old        Suggest    If Prov.
> [ERROR]   org.apache.logging.log4j.core                      PACKAGE    UNCHANGED  2.24.2     2.24.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender             PACKAGE    MINOR      2.26.0     2.20.3     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.db          PACKAGE    UNCHANGED  2.21.0     2.21.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.db.jdbc     PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.mom         PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.mom.jeromq  PACKAGE    UNCHANGED  2.21.0     2.21.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.mom.kafka   PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.nosql       PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.rewrite     PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.rolling     PACKAGE    MINOR      2.26.0     2.21.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.rolling.action PACKAGE    MINOR      2.26.0     2.24.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.appender.routing     PACKAGE    MINOR      2.26.0     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.async                PACKAGE    MINOR      2.26.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config               PACKAGE    MINOR      2.26.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.arbiters      PACKAGE    UNCHANGED  2.21.0     2.21.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.builder.api   PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.builder.impl  PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.composite     PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.json          PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins       PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.convert PACKAGE    UNCHANGED  2.24.0     2.24.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.processor PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.util  PACKAGE    UNCHANGED  2.20.2     2.20.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.validation PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.validation.constraints PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.validation.validators PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.plugins.visitors PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.properties    PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.status        PACKAGE    UNCHANGED  2.20.2     2.20.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.xml           PACKAGE    UNCHANGED  2.20.2     2.20.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.config.yaml          PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.filter               PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.filter.mutable       PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.impl                 PACKAGE    UNCHANGED  2.24.1     2.24.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.jackson              PACKAGE    UNCHANGED  2.24.1     2.24.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.jmx                  PACKAGE    UNCHANGED  2.23.0     2.23.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.layout               PACKAGE    MINOR      2.26.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.lookup               PACKAGE    UNCHANGED  2.24.1     2.24.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.net                  PACKAGE    UNCHANGED  2.20.3     2.20.3     ok         -
> [ERROR]   org.apache.logging.log4j.core.net.ssl              PACKAGE    UNCHANGED  2.20.3     2.20.3     ok         -
> [ERROR]   org.apache.logging.log4j.core.osgi                 PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.parser               PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR] * org.apache.logging.log4j.core.pattern              PACKAGE    MINOR      2.24.1     2.24.1     2.25.0     -
> [ERROR]   MINOR                PACKAGE    org.apache.logging.log4j.core.pattern
> [ERROR]     ADDED              ENUM       org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            ACCESS     final
> [ERROR]       ADDED            EXTENDS    java.lang.Enum
> [ERROR]       ADDED            IMPLEMENTS java.io.Serializable
> [ERROR]       ADDED            IMPLEMENTS java.lang.Comparable
> [ERROR]       ADDED            IMPLEMENTS java.lang.constant.Constable
> [ERROR]       ADDED            FIELD      ABSOLUTE
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ABSOLUTE_MICROS
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ABSOLUTE_NANOS
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ABSOLUTE_PERIOD
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      COMPACT
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DATE
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DATE_PERIOD
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DEFAULT
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DEFAULT_MICROS
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DEFAULT_NANOS
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      DEFAULT_PERIOD
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_BASIC
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_BASIC_PERIOD
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_OFFSET_DATE_TIME_HH
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_OFFSET_DATE_TIME_HHCMM
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_OFFSET_DATE_TIME_HHMM
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_PERIOD
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      ISO8601_PERIOD_MICROS
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      US_MONTH_DAY_YEAR2_TIME
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            FIELD      US_MONTH_DAY_YEAR4_TIME
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            METHOD     clone()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     protected
> [ERROR]         ADDED          RETURN     java.lang.Object
> [ERROR]       ADDED            METHOD     compareTo(java.lang.Enum)
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     int
> [ERROR]       ADDED            METHOD     compareTo(java.lang.Object)
> [ERROR]         ADDED          ACCESS     abstract
> [ERROR]         ADDED          RETURN     int
> [ERROR]       ADDED            METHOD     describeConstable()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     java.util.Optional
> [ERROR]       ADDED            METHOD     equals(java.lang.Object)
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     boolean
> [ERROR]       ADDED            METHOD     finalize()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          ACCESS     protected
> [ERROR]         ADDED          RETURN     void
> [ERROR]       ADDED            METHOD     getDeclaringClass()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     java.lang.Class
> [ERROR]       ADDED            METHOD     getLegacyPattern()
> [ERROR]         ADDED          RETURN     java.lang.String
> [ERROR]       ADDED            METHOD     getNonLegacyPattern()
> [ERROR]         ADDED          RETURN     java.lang.String
> [ERROR]       ADDED            METHOD     hashCode()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     int
> [ERROR]       ADDED            METHOD     name()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     java.lang.String
> [ERROR]       ADDED            METHOD     ordinal()
> [ERROR]         ADDED          ACCESS     final
> [ERROR]         ADDED          RETURN     int
> [ERROR]       ADDED            METHOD     toString()
> [ERROR]         ADDED          RETURN     java.lang.String
> [ERROR]       ADDED            METHOD     valueOf(java.lang.String)
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern
> [ERROR]       ADDED            METHOD     values()
> [ERROR]         ADDED          ACCESS     static
> [ERROR]         ADDED          RETURN     org.apache.logging.log4j.core.pattern.NamedPattern[]
> [ERROR]   org.apache.logging.log4j.core.script               PACKAGE    MINOR      2.26.0     2.20.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.selector             PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.time                 PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.tools                PACKAGE    UNCHANGED  2.20.1     2.20.1     ok         -
> [ERROR]   org.apache.logging.log4j.core.util                 PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR]   org.apache.logging.log4j.core.util.datetime        PACKAGE    UNCHANGED  2.21.2     2.21.2     ok         -
> [ERROR]   org.apache.logging.log4j.core.util.internal.instant PACKAGE    UNCHANGED  2.25.0     2.25.0     ok         -
> [ERROR] -> [Help 1]
> [ERROR] 
> [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
> [ERROR] Re-run Maven using the -X switch to enable full debug logging.
> [ERROR] 
> [ERROR] For more information about the errors and possible solutions, please read the following articles:
> [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
> [ERROR] 
> [ERROR] After correcting the problems, you can resume the build with the command
> [ERROR]   mvn <args> -rf :log4j-core
> ```
</details>

### 🧪 Tests (select one)

- [x] I have added or updated tests to cover my changes.
- [ ] No additional tests are needed for this change.

### 📝 Changelog (select one)


ppkarwasz (changes requested):
Hi @ashr123,

Thanks for the PR!

From my perspective, this refactoring does improve the readability of the `decodeNamedPattern` method—cleaner logic and better structure are always welcome. That said, readability can be subjective, so I'd also like to hear what @vy (the original author) thinks about the change.

Regarding the visibility of the new `NamedPattern` enum: as I mentioned in https://github.com/apache/logging-log4j2/discussions/3788#discussioncomment-13601110:

> I'm not sure there's a strong case for making such an `enum` public. Unless there's a clear user-facing need, we prefer to keep our API surface area minimal.

Would it be possible to make the `NamedPattern` enum package-private? Or would that conflict with the purpose of your change?

Currently, the PR triggers an API compatibility check failure:

```
[ERROR]   Name                                               Type       Delta      New        Old        Suggest    If Prov.
[ERROR] * org.apache.logging.log4j.core.pattern              PACKAGE    MINOR      2.24.1     2.24.1     2.25.0     -
[ERROR]   MINOR                PACKAGE    org.apache.logging.log4j.core.pattern
[ERROR]     ADDED              ENUM       org.apache.logging.log4j.core.pattern.NamedPattern
```

This is caused by our [BND Baseline](https://github.com/apache/logging-log4j2/blob/2.x/BUILDING.adoc#fixing-api-compatibility-check-failures) mechanism, which flags any additions to the public API. If you do intend to make `NamedPattern` public, you'll need to update the `@Version` annotation in the `package-info.java` file for the `org.apache.logging.log4j.core.pattern` package. If not, making the enum package-private would avoid triggering this check.

Looking forward to your thoughts!

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedPattern.java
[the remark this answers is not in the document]
ashr123: @vy:

1. Java convention is to distinguish between date, time and date-time entities, if I’ll call it `DatePattern` it may be a bit confusing, don’t you thing, and indeed it concerns itself only about what you called named-patterns
3. regarding point 3, don’t you prefer using `InstantPatternFormatter.LEGACY_FORMATTERS_ENABLED` (which is constant) directly instead of creating another constant in `DatePattern`?
5. regarding 5, the hole point is to use the built-in mechanism of mapping `String` into an enum.. why should we iterate over the literals (sort of inventing the wheel)?

src/changelog/.2.x.x/exported_named_patterns_into_public_enum.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="added">
+  <issue id="3789" link="https://github.com/apache/logging-log4j2/pull/3789"/>
+  <description format="asciidoc">Add and export `org.apache.logging.log4j.core.pattern.NamedDatePattern` class where users can programmatically access named date & time patterns supported by Pattern Layout</description>
```
vy, line 7:
```suggestion
  <description format="asciidoc">Add and export `org.apache.logging.log4j.core.pattern.NamedDatePattern` class where users can programmatically access named date &amp; time patterns supported by Pattern Layout</description>
```

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedDatePattern.java
```diff
@@ -0,0 +1,114 @@
...
+    //   Yet legacy formatters use repeated `n` to denote sub-milliseconds precision of certain length.
+    //   This doesn't work for `DateTimeFormatter`, which needs
+    //
+    //   - `SSSSSS` for 6-digit microsecond precision
+    //   - `SSSSSSSSS` for 9-digit nanosecond precision
+    //
+    // - Legacy formatters use `X`, `XX,` and `XXX` to choose between `+00`, `+0000`, or `+00:00`.
+    //   This is the correct behaviour for `SimpleDateFormat`.
+    //   Though `X` in `DateTimeFormatter` produces `Z` for zero-offset.
+    //   To avoid the `Z` output, one needs to use `x` with `DateTimeFormatter`.
```
ppkarwasz, line 61: Of course this comment can be removed once we remove legacy patterns from this class.

ppkarwasz (changes requested):
Hi @ashr123,

Thanks for the PR!

Overall, the changes look good. However, I have some concerns regarding the values returned by the public method `NamedDatePattern.getPattern()`. Specifically, I think we should:

* Clearly define that this method **must always** return a pattern compatible with [`DateTimeFormatter`](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).
* Document this contract explicitly in the method’s Javadoc.
* If `LEGACY_FORMATTERS_ENABLED` is `true`, fall back to using `FixedDateFormat.FixedFormat` instead of `NamedDatePattern`.

### Rationale

I don’t think we should expose legacy `FixedFormat` patterns via the new enum, for the following reasons:

* These patterns are not compatible with any standard Java datetime formatters: not [`SimpleDateFormat`](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html), [`DateTimeFormatter`](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html), nor Commons Lang’s [`FastDateFormat`](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/time/FastDateFormat.html).
* They are also incompatible with our own forked `o.a.l.l.core.util.datetime.FastDateFormat`.
* In practice, they only work with `FixedDateFormat`.
* We want to discourage third-party usage of `FixedDateFormat`, as Log4j is not intended to serve as a general-purpose date formatting library.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedDatePattern.java
```diff
@@ -0,0 +1,114 @@
...
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.pattern;
+
+import static org.apache.logging.log4j.core.util.internal.instant.InstantPatternFormatter.LEGACY_FORMATTERS_ENABLED;
+
+import org.apache.logging.log4j.core.util.internal.instant.InstantPatternFormatter;
+
+/**
+ * Represents named date & time patterns for formatting log timestamps.
```
ppkarwasz, line 24:
```suggestion
 * Represents named date &amp; time patterns for formatting log timestamps.
```

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
```diff
@@ -109,84 +109,11 @@ private static String readPattern(@Nullable final String[] options) {
...
-                return "yyyy-MM-dd'T'HH:mm:ss." + (compat ? "nnnnnn" : "SSSSSS");
-            case "US_MONTH_DAY_YEAR2_TIME":
-                return "dd/MM/yy HH:mm:ss.SSS";
-            case "US_MONTH_DAY_YEAR4_TIME":
-                return "dd/MM/yyyy HH:mm:ss.SSS";
+        try {
+            return NamedDatePattern.valueOf(pattern).getPattern();
+        } catch (IllegalArgumentException ignored) {
+            return pattern;
         }
```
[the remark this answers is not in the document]
ashr123: @ppkarwasz do you think that finding the the literal by iterating all the values has better performance than catching the exception thrown?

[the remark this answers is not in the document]
ppkarwasz: Yes — even today, exception handling remains relatively expensive compared to standard control flow.

To compare both approaches, I wrote a small JMH benchmark. In the more common case (based on most user questions I’ve seen), where predefined patterns are **not** used, the exception-based approach incurs a significant performance penalty due to the cost of creating and handling the exception:

| Benchmark                                 | Mode | Cnt |  Score   |  Error  | Units |
|------------------------------------------|------|-----|----------|---------|--------|
| NamedDatePatternBenchmark.withException  | avgt |  25 | 643.222  | ±16.663 | ns/op  |
| NamedDatePatternBenchmark.withoutException | avgt |  25 |  42.912  | ± 0.191 | ns/op  |

For the smaller group of users who **do** use predefined patterns, `NameDateFormat.valueOf()` is of course faster, as it uses a map lookup rather than a linear search:

| Benchmark                                 | Mode | Cnt |  Score |  Error  | Units |
|------------------------------------------|------|-----|--------|---------|--------|
| NamedDatePatternBenchmark.withException  | avgt |  25 |  2.202 | ± 0.299 | ns/op  |
| NamedDatePatternBenchmark.withoutException | avgt |  25 | 46.423 | ±11.093 | ns/op  |

It is worth noting, however, that this logic runs only during logging configuration — typically at startup — so performance here is not critical. Still, I believe it's good practice to avoid exceptions for control flow when cleaner alternatives exist.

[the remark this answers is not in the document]
ppkarwasz: > I still prefer the try-catch option, especially if it done only at startup, I still think it more readable than iterating over the values. if this is that prevents from this PR to be merged, I'll give up and we will use the iteration version… 🫤

I think the iterative version better reflects the semantics of the enum — using enum values is the exception, not the norm. That said, I’m not strongly opposed to the `try-catch` version if you prefer it. The main goal is readability, and both approaches seem clear to me.

vy:
> > 2. Shall we rename the class from `NamedDatePattern` to `NamedInstantPattern` in line with our recent wordsmithing efforts in _"instant formatting"_?
> 
> That could be a sensible change. I don't have a strong preference either way, but aligning with our recent naming conventions might improve clarity.

@ashr123, would you mind renaming `NamedDatePattern` to `NamedInstantPattern`, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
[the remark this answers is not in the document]
ppkarwasz: > So if there is no objection, can you approve this PR? is something missing?

No objection to using `try-catch`, but my earlier comment (https://github.com/apache/logging-log4j2/pull/3789#pullrequestreview-2998821838) still stands: I believe it's incorrect for `NamedDatePattern.getPattern()` to return different values based on `LEGACY_FORMATTERS_ENABLED`.

How is a user expected to reliably use these patterns if their meaning changes depending on that flag?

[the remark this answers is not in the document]
ashr123: @ppkarwasz I agree with you completely, I'll in case  of `InstantPatternFormatter.LEGACY_FORMATTERS_ENABLED` is `true`, I'll use `FixedDateFormat.FixedFormat#getPattern`, is that ok?
Of course, `NamedInstantPattern` will contain only the non-legacy patterns

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedInstantPattern.java
```diff
@@ -0,0 +1,78 @@
...
+
+    US_MONTH_DAY_YEAR4_TIME("dd/MM/yyyy HH:mm:ss.SSS");
+
+    private final String pattern;
+
+    NamedInstantPattern(String pattern) {
+        this.pattern = pattern;
+    }
+
+    public String getPattern() {
```
ppkarwasz, line 75: Could you please add Javadoc to this public method stating that the returned pattern is compatible with `DateTimeFormatter`?

ppkarwasz (approved):
LGTM, although I would add some Javadoc to the `getPattern()` method.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedInstantPattern.java
ashr123: @ppkarwasz of course

ashr123:
@ppkarwasz can you add the JavaDoc yourself? I'm using another computer and I have a bit of trouble regarding the signature

ppkarwasz:
> > @vy should I use `FixedDateFormat.FixedFormat#getPattern` in case of `InstantPatternFormatter.LEGACY_FORMATTERS_ENABLED` is `true`?
> 
> @ashr123, no, absolutely not. They should not be used anywhere in Log4j. In `2.25.0` we put great effort into deprecating them – see [the release notes](https://logging.apache.org/log4j/2.x/release-notes.html#release-notes-2-25-0-instant-format).

This ties back to my earlier comment: https://github.com/apache/logging-log4j2/pull/3789#pullrequestreview-2998821838

I don’t think `NamedInstantFormat` should replicate the legacy pattern values — they’re not useful to external users. If needed, the `decodeNamedPattern` method can still retrieve them internally, as it does in the current version of the PR:

https://github.com/apache/logging-log4j2/blob/6b311cd8e81704643a6805e8703ed7325a8e3ef5/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java#L142-L148

vy:
> > @ashr123, no, absolutely not. They should not be used anywhere in Log4j. In `2.25.0` we put great effort into deprecating them – see [the release notes](https://logging.apache.org/log4j/2.x/release-notes.html#release-notes-2-25-0-instant-format).
> 
> This ties back to my earlier comment: [#3789 (review)](https://github.com/apache/logging-log4j2/pull/3789#pullrequestreview-2998821838)

It doesn't. In its current state, deprecated formatters are an island – no programmatic connections from anywhere. We can safely delete them, yet keep the legacy formatters configuration knob functional. Not saying we should, but the current design allows this. Supporting a legacy behavior and internal classes associated with them (of which should not have been exported, but now it is too late) are two different things.

In conclusion, deprecated formatters are an island, please keep it that way.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/DatePatternConverter.java
```diff
@@ -109,40 +109,19 @@ private static String readPattern(@Nullable final String[] options) {
      * @since 2.25.0
      */
     static String decodeNamedPattern(final String pattern) {
-        // If legacy formatters are enabled, we need to produce output aimed for `FixedDateFormat` and `FastDateFormat`.
```
vy, line 113: @ppkarwasz, would you mind restoring this comment block to its state in 0ef914b2a18c48675979c20094a8ceb9a1b5a579, please? I find it crucial to understand what we have done and why.

ppkarwasz: Can we simplify this comment a bit? It currently takes many lines to explain two main points:

* `FixedDateFormat` uses `n` to represent various sub-second fractions, while `DateTimeFormatter` uses `n` strictly for nanoseconds.
* `X` in `DateTimeFormatter` returns `"Z"` for zero offsets, whereas `SimpleDateFormat` returns `"00:00"`.

At the same time, it omits some helpful context — for example, why do we convert `ABSOLUTE_PERIOD` to `"HH:mm:ss.SSS"` instead of passing it directly to the legacy formatter?

ppkarwasz: > > At the same time, it omits some helpful context — for example, why do we convert `ABSOLUTE_PERIOD` to `"HH:mm:ss.SSS"` instead of passing it directly to the legacy formatter?
> 
> The legacy formatter is, as its name implies, legacy, and it shall not be used. It has several bugs, still.

What I wanted to explain is why we have to use:

```java
return NamedInstantPattern.valueOf(pattern).getLegacyPattern();
```

instead of a simpler:

```java
return NamedInstantPattern.valueOf(pattern).name();
```

> But, can we keep the existing details, without omission, please?

Sure, although here it looks a little bit out of context. What do you think about adding it to `NamedInstantPattern#getLegacyPattern()`?

ppkarwasz (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedInstantPattern.java
```diff
@@ -0,0 +1,163 @@
...
+     *       we use <code>x</code>, <code>xx</code>, or <code>xxx</code> instead.
+     *     </p>
+     *   </li>
+     * </ul>
+     *
+     * @return the legacy pattern string as used in
+     * {@link org.apache.logging.log4j.core.util.datetime.FixedDateFormat.FixedFormat}
+     */
+    String getLegacyPattern() {
+        String legacyPattern = pattern.replace('x', 'X');
```
vy, line 156: @ppkarwasz, does this extra logic fix anything in the previous design, where legacy pattern was hardcoded and clearly disclosing the difference?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternLegacyFormatter.java
```diff
@@ -46,7 +46,10 @@ final class InstantPatternLegacyFormatter implements InstantPatternFormatter {
     private final BiConsumer<StringBuilder, Instant> formatter;
 
     InstantPatternLegacyFormatter(final String pattern, final Locale locale, final TimeZone timeZone) {
-        this.precision = new InstantPatternDynamicFormatter(pattern, locale, timeZone).getPrecision();
+        // Replaces 'n' used in legacy patterns with 'S' to obtain the right precision.
+        // In legacy patterns, the precision of 'n' depends on the pattern length, but
+        // in `DateTimeFormatter`, it is always nanoseconds.
+        this.precision = new InstantPatternDynamicFormatter(pattern.replace('n', 'S'), locale, timeZone).getPrecision();
```
vy, line 52: Doesn't `IPLF::new`  always get passed the legacy pattern? What does this fix? Is this already an issue in `2.x`, i.e., excluding changes of this PR?

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/util/internal/instant/InstantPatternLegacyFormatter.java
vy: Got it. Created #3816 to address this. I will revert this change.

log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/NamedInstantPattern.java
vy: It all began years ago when we tried to make sense of the scattered `pattern.replace('x', 'X')` calls across the codebase, particularly in tests, to resolve time zone–related issues. It took a massive effort to dig through commit histories and untangle the subtle differences between competing date and time formatters. I vividly remember the hours-long video calls we had in 2024 just to understand and clarify all this.

Personally, I strongly prefer the more explicit and elaborate `LEGACY ? "FDF-pattern" : "DTF-pattern"` style. You’re remarkably good at context switching and diving deep into complex areas – like the date and time discrepancies at the heart of this PR. You also happen to have the time and flexibility to focus on the project extensively. I, on the other hand, don’t possess that context-switching superpower, nor do I have spare capacity. I believe that reflects the situation of most other Log4j maintainers and contributors as well.

So while I agree that this is, in part, a matter of style, there’s more to it than that. The `pattern.replace('x', 'X')` approach has historically proven to be hard to understand, and most of us are in desperate need of code that is simple, self-explanatory, and well-documented – in that order.

ppkarwasz: > Personally, I strongly prefer the more explicit and elaborate `LEGACY ? "FDF-pattern" : "DTF-pattern"` style.

I hardcoded the patterns again in https://github.com/apache/logging-log4j2/pull/3789/commits/bb8b41f8bd48414c11422098284c3d365361f1c2

vy (approved):

vy (approved):

PR #3839, opened 2025-07-21
Add API to list supported configuration file extensions

yybmion:
### Summary

Add ConfigurationFactory.getActiveFileExtensions() and getSupportedExtensions() methods to allow external frameworks like Spring Boot to determine which configuration file formats are currently supported at runtime.

- Add abstract getSupportedExtensions() method to ConfigurationFactory
- Add static getActiveFileExtensions() method to aggregate results
- Implement getSupportedExtensions() in all ConfigurationFactory subclasses
- Add test coverage for new functionality

### Example Usage

```java
List<String> supportedExtensions = ConfigurationFactory.getActiveFileExtensions();
// Returns: ["xml", "json", "jsn", "properties"] (if Jackson available)
// Returns: ["xml", "properties"] (if Jackson missing)
```

## Checklist

### ✅ Required checks

- [x] **Commit signatures**: All commits are signed and verifiable. (See [GitHub Docs on Commit Signature Verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)).
  <details>

  </details>
  <details>

    Run: `./mvnw verify`

  </details>

### 🧪 Tests (select one)

- [x] I have added or updated tests to cover my changes.
- [ ] No additional tests are needed for this change.

### 📝 Changelog (select one)

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
@@ -27,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.core.LoggerContext;
@@ -262,6 +263,14 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
     }
 
     protected abstract String[] getSupportedTypes();
+
+    /**
+     * Returns the file extensions supported by this configuration factory.
+     *
+     * @return list of supported file extensions (e.g., ["xml", "json"])
+     * @since 2.25.0
+     */
+    public abstract List<String> getSupportedExtensions();
 
     protected String getTestPrefix() {
         return TEST_PREFIX;
@@ -554,6 +563,16 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
         }
 
         @Override
+        public List<String> getSupportedExtensions() {
+            return getFactories().stream()
+                    .filter(factory -> factory != this)
+                    .filter(ConfigurationFactory::isActive)
+                    .flatMap(factory -> factory.getSupportedExtensions().stream())
+                    .distinct()
+                    .collect(Collectors.toList());
+        }
+
+        @Override
         public Configuration getConfiguration(final LoggerContext loggerContext, final ConfigurationSource source) {
             if (source != null) {
                 final String config = source.getLocation();
@@ -616,5 +635,20 @@ public abstract class ConfigurationFactory extends ConfigurationBuilderFactory {
 
     static List<ConfigurationFactory> getFactories() {
         return factories;
+    }
+
+    /**
+     * Returns all configuration file extensions currently supported at runtime.
+     * This aggregates extensions from all active configuration factories.
+     *
+     * @return list of supported file extensions
+     * @since 2.25.0
+     */
+    public static List<String> getActiveFileExtensions() {
+        return getFactories().stream()
+                .filter(ConfigurationFactory::isActive)
+                .flatMap(factory -> factory.getSupportedExtensions().stream())
+                .distinct()
+                .collect(Collectors.toList());
     }
 }
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationFactory.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/config/Log4j1ConfigurationFactory.java
@@ -18,6 +18,8 @@ package org.apache.log4j.config;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collections;
+import java.util.List;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.ConfigurationException;
@@ -52,5 +54,10 @@ public class Log4j1ConfigurationFactory extends ConfigurationFactory {
     @Override
     protected String[] getSupportedTypes() {
         return SUFFIXES;
+    }
+
+    @Override
+    public List<String> getSupportedExtensions() {
+        return Collections.emptyList();
     }
 }
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfigurationFactory.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/config/PropertiesConfigurationFactory.java
@@ -16,6 +16,8 @@
  */
 package org.apache.log4j.config;
 
+import java.util.Collections;
+import java.util.List;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.Configuration;
 import org.apache.logging.log4j.core.config.ConfigurationFactory;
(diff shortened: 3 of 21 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -617,4 +636,19 @@ private String[] parseConfigLocations(final String configLocations) {
...
     }
+
+    /**
+     * Returns all configuration file extensions currently supported at runtime.
+     * This aggregates extensions from all active configuration factories.
+     *
+     * @return list of supported file extensions
+     * @since 2.25.0
+     */
+    public static List<String> getActiveFileExtensions() {
```
vy, line 647: What is the use case for this? I have the impression the instance method would just do fine.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -263,6 +264,14 @@ public static void removeConfigurationFactory(final ConfigurationFactory factory
 
     protected abstract String[] getSupportedTypes();
 
+    /**
+     * Returns the file extensions supported by this configuration factory.
+     *
+     * @return list of supported file extensions (e.g., ["xml", "json"])
+     * @since 2.25.0
+     */
+    public abstract List<String> getSupportedExtensions();
```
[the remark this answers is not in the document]
ppkarwasz: `getSupportedFileExtensions` sounds good, although I would keep this a list to convey the order in which those extensions are checked.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
yybmion: Thanks for the review @vy ! 

You're absolutely right about the redundancy. The static method duplicates the functionality already provided by the instance method. 

Therefore, I think it would be better to remove the static `getActiveFileExtensions()` method and use the instance method approach instead.

vy:
> > 1. It duplicates quite some logic (e.g., in `JsonConfigurationFactory`, extensions and suffixes are semantically identical).
> 
> Totally agree. Let’s put a **default** implementation in `ConfigurationFactory` so most factories don’t have to repeat themselves.

Please don't – you will still end up having two methods doing the same thing.

> I prefer the dedicated `getSupportedExtensions()` API for a few reasons:
> 
> * `isActive()` and `getSupportedTypes()` are **protected**, not public, so they’re not available to consumers; exposing them (or a raw `getFactories()`) would leak internal details.

Agreed. Let's make `isActive()` public.

> * Implementations are **inconsistent** today: some factories always report active but conditionally restrict types (e.g., Log4j 1 XML returns `xml` only when a property is set), while others return a constant type set but flip `isActive()` based on optional deps. A single public method lets us **normalize** that behavior and present a stable contract.

What you're describing is not a problem, but a behavior and it is all good, AFAICT. It is up to the `CF` what to return from `isActive()`. It makes sense `CF.getFactories()` filtered on `isActive()` yields a different outcome when a user has enabled/disabled Log4j 1 support. As a matter of fact, this is a powerful feature that allows `CF`s to dynamically determine their active status.

> * In Log4j 3, `isActive()` isn’t needed anymore: optional Maven **modules** provide support for additional formats, not per-factory optional dependencies, so I would propose to remove `isActive()` from Log4j 3.

That is a good idea.

> * Conceptually, `ConfigurationFactory` should look like a **singleton service**; how it multiplexes concrete factories is an implementation detail. Exposing `getFactories()` would couple users to that detail, whereas `getSupportedExtensions()` gives them exactly what they need without exposing internals.

`CF::getSupportedFileExtensions` _only_ hoists `CF::getSupportedTypes` for a handful of users. OTOH, `CF::instances` would enable several more use cases with less API impact.

vy:
Another concern I have regarding adding a new, as a matter of fact, 2nd after `getSupportedTypes()`, _"supported file extensions"_ concept to `CF` is that it reinforces the idea of a `CF` must be associated with a _file_, whereas it actually is not necessarily.

On Tuesday, in a meeting where @ppkarwasz and I were present, after hearing that one of the biggest Log4j users using programmatic configuration extensively, I am very reluctant to marry the concept of files with configurations.

vy:
> * **`isActive()`**: aside from its inconsistent semantics today (which we could technically fix)

Created #3896 for this.

> **TL;DR:** Spring Boot does not actually need to know what extensions we support. With the existing `ConfigurationFactory.getConfiguration` API (plus potentially a small addition for composite configs), they can cover all their use cases. The only real gap is an ergonomic way to build composite configurations; everything else can stay encapsulated inside Log4j Core.

I'm happy to see that the nudge by @yybmion helped us to avoid yet another [peak of complexity](https://www.youtube.com/watch?v=Yiye8lqh0Ig). @ppkarwasz, thanks so much for your patience along this conversation and helping with evaluating alternatives. I liked your proposal on changing `CF::gC` as you suggested in the paragraph I quoted. Would you mind sharing more details such that @yybmion can flesh it out further in a PR<sup>1</sup>?

<sup>1</sup> I'm inclined to leave this PR as is for records, and implement the new `CF::gC` change in a separate PR.

ppkarwasz:
My idea for the proposed method

```java
public Configuration getConfiguration(LoggerContext context, String name, List<URI> uris)
```

is to encapsulate the logic currently in `ConfigurationFactory.Factory`:

https://github.com/apache/logging-log4j2/blob/769b9249176b69ad4474fc4bd831dec67775c953/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java#L381-L402

This would allow each `ConfigurationFactory` to support building a composite configuration from multiple URIs and will allow Spring Boot to specify the URI of the overrides, without any knowledge of Log4j's composite configuration.

One open question is how to define the nullability/return contract. The existing single-URI overload,

```java
public Configuration getConfiguration(LoggerContext context, String name, URI uri)
```

has inconsistent behavior across implementations:

* Individual `ConfigurationFactory` implementations typically return `null` if the URI itself is `null`.
* `ConfigurationFactory.Factory`, however, falls back to a `DefaultConfiguration`.

Most (but not all) callers explicitly check for `null`, so the contract is somewhat unclear. For the new multi-URI variant, we should decide whether the method is allowed to return `null`, or whether it must always return a valid `Configuration` (possibly default).

vy:
> One open question is how to define the nullability/return contract. The existing single-URI overload,
> 
> ```java
> public Configuration getConfiguration(LoggerContext context, String name, URI uri)
> ```
> 
> has inconsistent behavior across implementations:
> 
> * Individual `ConfigurationFactory` implementations typically return `null` if the URI itself is `null`.
> * `ConfigurationFactory.Factory`, however, falls back to a `DefaultConfiguration`.
> 
> Most (but not all) callers explicitly check for `null`, so the contract is somewhat unclear. For the new multi-URI variant, we should decide whether the method is allowed to return `null`, or whether it must always return a valid `Configuration` (possibly default).

@ppkarwasz, since there is no established default that one can claim as a stable API, I'm inclined to throw `IAE`, or `NPE` (when arg is null) on failures. WDYT?

Will this new API necessitate an amendment/improvement to the manual? (I've skimmed through combined/composite configuration sections, but couldn't see any.)

> Should I start working on this right away, or would you prefer to clarify a few implementation details first (like the nullability contract and specific behavior expectations)?

@yybmion, please go ahead. I strongly advise you to work out the essentials first, and leave out the tests, docs, etc. Once we get it reviewed, you can continue with polishing the PR.

yybmion:
@vy @ppkarwasz Thanks for the detailed guidance.

I created a new PR based on this discussion: #3921

Implemented the suggested approach
- New method: `getConfiguration(LoggerContext, String, List<URI>)` 
- Encapsulates composite configuration creation
- Throws exceptions instead of returning null
- Work out the essentials first

vy:
Closing in favor of #3921.

Issue #2117, opened 2023-12-21
Consider replacing `FileOutputStream` with `Files.newOutputStream`


ppkarwasz:
Since Java 7 two methods are available to open new files:

- [`FileOutputStream#new`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/FileOutputStream.html),
- [`Files#newOutputStream`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html#newOutputStream(java.nio.file.Path,java.nio.file.OpenOption...)).

While they mostly perform the same OS operations, there is a significant difference between the two on Windows: the IO method opens the file **without** the `FILE_SHARE_DELETE` option, whereas the NIO method adds the `FILE_SHARE_DELETE` option (cf. [this SO question](https://stackoverflow.com/q/63584311/11748454)).

Since log files on UNIX can already be deleted without stopping the application, this change would allow an alignment between the behavior of Log4j on Windows and UNIX.

I discussed this with @garydgregory on Slack.

vy:
There are [more (platform-independent!) hazards surrounding `FileOutputStream::new`](https://www.cloudbees.com/blog/fileinputstream-fileoutputstream-considered-harmful). Please indeed replace them with `Files::newOutputStream`.

jvz:
I like this idea because it also supports use of other file system implementations based on what the user has installed. That could also be useful in tests if we used something like https://github.com/google/jimfs for in-memory file system testing.

ppkarwasz:
This is related to #2537.

PR #3855, opened 2025-07-23
Add org.apache.logging.log4j.core.appender.FileManager.getPath()

garydgregory:
Add `org.apache.logging.log4j.core.appender.FileManager.getPath()`

> [!IMPORTANT]  
> Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise.

## Checklist

### ✅ Required checks

- [ ] **Commit signatures**: All commits are signed and verifiable. (See [GitHub Docs on Commit Signature Verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)).
  <details>

  </details>
  <details>

    Run: `./mvnw verify`

  </details>

Note that there are still problems on Windows:
```
[ERROR] Failures:
[ERROR]   SyslogAppenderCustomLayoutTest>SyslogAppenderTest.testTCPAppender:56->SyslogAppenderTestBase.sendAndCheckLegacyBsdMessage:74->SyslogAppenderTestBase.checkTheNumberOfSentAndReceivedMessages:112 The number of received messages should be equal with the number of sent messages ==> expected: <1> but was: <0>
[ERROR]   RollingAppenderDeleteScriptTest.testAppender:74 target\rolling-with-delete-script\test\test-2.log should have odd index expected:<1> but was:<0>
[ERROR]   AsyncLoggersWithAsyncLoggerConfigTest.testLoggingWorks:46 Incorrect number of events. Expected 2, got 0 expected:<2> but was:<0>
[INFO]
[ERROR] Tests run: 8447, Failures: 3, Errors: 0, Skipped: 36
```

### 🧪 Tests (select one)

- [x] No additional tests are needed for this change.

### 📝 Changelog (select one)

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/FileManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/FileManager.java
@@ -246,19 +246,20 @@ public class FileManager extends OutputStreamManager {
     protected OutputStream createOutputStream() throws IOException {
         final String filename = getFileName();
         LOGGER.debug("Now writing to {} at {}", filename, new Date());
-        final File file = new File(filename);
+        final Path path = getPath();
+        final File file = path.toFile();
         createParentDir(file);
         final FileOutputStream fos = new FileOutputStream(file, isAppend);
         if (file.exists() && file.length() == 0) {
             try {
                 final FileTime now = FileTime.fromMillis(System.currentTimeMillis());
-                Files.setAttribute(file.toPath(), "creationTime", now);
+                Files.setAttribute(path, "creationTime", now);
             } catch (Exception ex) {
                 LOGGER.warn("Unable to set current file time for {}", filename);
             }
             writeHeader(fos);
         }
-        defineAttributeView(Paths.get(filename));
+        defineAttributeView(path);
         return fos;
     }
 
@@ -343,6 +344,16 @@ public class FileManager extends OutputStreamManager {
     public String getFileName() {
         return getName();
     }
+
+    /**
+     * Returns the Path of the file being managed.
+     * @return The name of the file being managed.
+     * @since 2.26.0
+     */
+    public Path getPath() {
+        return Paths.get(getFileName());
+    }
+
     /**
      * Returns the append status.
      * @return true if the file will be appended to, false if it is overwritten.
--- a/src/changelog/.2.x.x/add_FileManager_getPath_.xml
+++ b/src/changelog/.2.x.x/add_FileManager_getPath_.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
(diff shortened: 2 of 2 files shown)


vy:
@garydgregory, would you mind sharing the use case, please? What problem is this change set addressing?

garydgregory:
Hi all,

TLDR: Let's bring this PR in and then...

Let's not confuse file Appenders and file Managers ;-)

I took a glance at the rolling file manager and it's a much larger code base. This PR is _small and focused_, which is a good thing. I am willing to volunteer to migrate the other manager code ((rolling) from IO to NIO _in a separate PR_ over the weekend probably, or at the very least initially from presenting `Path` instead/in addition to  `String in the API. I can do this in a binary compatible way of course, but view is that these are internal classes up for breakage. So we'd need to agree on what kind of compatibility we want for `AbstractRolloverStrategy` and friends. This work would include https://github.com/apache/logging-log4j2/issues/2117.

I want to think about the migration from IO to NIO in two steps:
- PR 1: Migrate the API: Use Path where possible instead of String and File
- PR 2: Migrate the implementation, there might be behavioral differences using `Files` that we only want to give users in 3.0 for example.

WDYT?


For 3.0, all bets are off and breaking BC is OK IMO in this area.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/FileManager.java
```diff
@@ -343,6 +344,16 @@ protected synchronized void writeToDestination(final byte[] bytes, final int off
     public String getFileName() {
         return getName();
     }
+
+    /**
+     * Returns the Path of the file being managed.
+     * @return The name of the file being managed.
+     * @since 2.26.0
+     */
+    public Path getPath() {
```
ppkarwasz, line 353: A small naming suggestion: since we already have a `getFileName()` method, it might be clearer to name this new method `getFilePath()` instead of `getPath()`.

It would also help to document the expected characteristics of the returned `Path`, for example:

* Whether the returned path is **absolute** and **normalized**. Since multiple appenders (even across different logger contexts) may share a single `FileManager`, using an absolute, normalized `Path` avoids ambiguity and makes comparisons safer.
* Whether the value may **change over time**. In particular, `RollingFileManager` can update its active file during rollover (especially with the *DirectWriteRolloverStrategy*), so it would be useful to document that the returned path may not be stable across invocations.

garydgregory: > A small naming suggestion: since we already have a getFileName() method, it might be clearer to name this new method getFilePath() instead of getPath().

Names are important and I'm glad we're discussing them.

IMO, "getFilePath()" is bad because it implies this is a Path to a file:// only, and since  a Path is an interface for locating a file in a file system, we should not set expectations that the implementation of the Path is only for file-schemed Paths.

It doesn't matter that there's already an API called "getFileName()", a file name is very different from a Path, where the expectation (from a NIO Path POV) is that the file name is the last segment, the "bar.txt" in "/foo/bar.txt". Having a common prefix between these two methods has no benefit.

ppkarwasz: I was interpreting `filePath` as “path of the file” as opposed to the path to its parent directory or similar.

I agree that “path with `file:` scheme” would be the wrong interpretation, even if currently that is the only scheme we support.

PR #3858, opened 2025-07-29
Document API compatibility checks (#3175)

jvz:
This documents the existence of the BND Baseline API compatibility checks and how to fix the generated errors.

---

Forward-port of #3175.


vy (approved):
It might be better to update version numbers used in examples from `2.x` to `3.x`, but anyhow... LGTM.

PR #3868, opened 2025-08-01
Fix sizing of Maps

kilink:
Fix several instances where Maps were sized incorrectly when the number of elements to insert is known; specifying the capacity without taking into account the load factor meant it would always result in a rehash / resize operation.

To this end, add a simple Maps utility class that handles the capacity calculation, much like the static HashMap.newHashMap helper that was added in JDK19.

## Checklist

### ✅ Required checks

- [ X ] **License**: I confirm that my changes are submitted under the [Apache License, Version 2.0](https://apache.org/licenses/LICENSE-2.0).
- [ X ] **Commit signatures**: All commits are signed and verifiable. (See [GitHub Docs on Commit Signature Verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)).
- [ X ] **Code formatting**: The code is formatted according to the project’s style guide.
- [ X ] **Build & Test**: I verified that the project builds and all unit tests pass.
  <details>

    Run: `./mvnw verify`

  </details>

### 🧪 Tests (select one)

- [ X ] No additional tests are needed for this change.

### 📝 Changelog (select one)

- [ ] I added a changelog entry in `src/changelog/.2.x.x`. (See [Changelog Entry File Guide](https://logging.apache.org/log4j/tools/log4j-changelog.html#changelog-entry-file)).
- [ X ] This is a trivial change and does not require a changelog entry.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.logging.log4j.util.Maps;
 
 /**
  * Adds entries to the {@link ThreadContext stack or map} and them removes them when the object is closed, e.g. as part
@@ -206,7 +207,7 @@ public class CloseableThreadContext {
         }
 
         private void closeMap() {
-            final Map<String, String> valuesToReplace = new HashMap<>(originalValues.size());
+            final Map<String, String> valuesToReplace = Maps.newHashMap(originalValues.size());
             final List<String> keysToRemove = new ArrayList<>(originalValues.size());
             for (final Map.Entry<String, String> entry : originalValues.entrySet()) {
                 final String key = entry.getKey();
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/ThreadContextTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/ThreadContextTest.java
@@ -22,10 +22,10 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.Map;
 import org.apache.logging.log4j.test.ThreadContextUtilityClass;
 import org.apache.logging.log4j.test.junit.UsingAnyThreadContext;
+import org.apache.logging.log4j.util.Maps;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 
@@ -115,7 +115,7 @@ class ThreadContextTest {
         assertTrue(ThreadContext.isEmpty());
         assertFalse(ThreadContext.containsKey("key"));
         final int mapSize = 10;
-        final Map<String, String> newMap = new HashMap<>(mapSize);
+        final Map<String, String> newMap = Maps.newHashMap(mapSize);
         for (int i = 1; i <= mapSize; i++) {
             newMap.put("key" + i, "value" + i);
         }
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/spi/DefaultThreadContextMapTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/spi/DefaultThreadContextMapTest.java
@@ -20,11 +20,11 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import org.apache.logging.log4j.test.junit.UsingThreadContextMap;
 import org.apache.logging.log4j.test.spi.ThreadContextMapSuite;
+import org.apache.logging.log4j.util.Maps;
 import org.apache.logging.log4j.util.PropertiesUtil;
 import org.junit.jupiter.api.Test;
 
@@ -56,7 +56,7 @@ class DefaultThreadContextMapTest extends ThreadContextMapSuite {
         assertTrue(map.isEmpty());
         assertFalse(map.containsKey("key"));
         final int mapSize = 10;
-        final Map<String, String> newMap = new HashMap<>(mapSize);
+        final Map<String, String> newMap = Maps.newHashMap(mapSize);
         for (int i = 1; i <= mapSize; i++) {
             newMap.put("key" + i, "value" + i);
         }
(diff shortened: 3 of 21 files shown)


vy:
Thanks so much for the patch. Would you mind sharing some context, please?
Did you observe a particular performance issue?

kilink:
> Did you observe a particular performance issue?

I don't recall how I stumbled upon it or if it showed up in any performance profiles (it definitely has for other libraries). It's just a common thing I've noticed across various projects (e.g., spring and grpc-java) since the HashMap constructor is a foot gun and it's not easy to do correctly until Java 19. Since the intent here was to presize the map, and we know how many elements to insert, it just seemed to make sense to size it so we always avoid the resize / rehash operations.

ppkarwasz:
Hi @kilink,

Adding a new utility class to `log4j-api` and then using it in `log4j-core` reinforces an unnatural dependency between the two modules, and would introduce a hard incompatibility between Log4j Core `2.26.0` and Log4j API `2.25.x`. The Core module should implement the API, not depend on it for internal utilities. I realize this kind of coupling already exists in some places, but I’d prefer not to increase it.

Could you instead:

* Move `Maps` to `o.a.l.l.util.internal` to make it clear it’s not intended for external use. While `o.a.l.l.util` is *theoretically* internal, it remains exported due to past API leaks (e.g., `o.a.l.l.util.Supplier`).
* Create a separate `o.a.l.l.core.util.internal.Maps` helper for `log4j-core`, so it can be used independently without introducing additional cross-module dependencies.

I agree with @vy that, without clear benchmark results, the value of this enhancement is questionable. We avoid using `HashMap` in performance-critical paths (such as the logging context map) because its instantiation and operation costs are high for small maps. Instead, we use `o.a.l.l.util.StringMap`, which offers enhanced features like mutability control (`freeze()`) and a `foreach` method (introduced for `Map` only in JDK 8). However, `StringMap` represents a significant technical debt we would like to eliminate. If you are interested in map performance, could you explore the feasibility of replacing `StringMap` with `Map` in version 3.x?

vy:
I'm against enhancements without convincing figures obtained from benchmarks mapping to real-world use cases. If this optimization yields a tangible benefit on the hot path, provide a reproducer please.

kilink:
I personally don't think it's worthwhile to spend time writing a benchmark for what amounts to a bug fix / correctness issue. If you'd rather not accept the change, feel free to close the issue. I'll leave my rationale for fixing it however:

- Without a benchmark, we can see that the incorrect supplied capacity will cause a resize / rehash, and result in a map that is over-sized for the number of items inserted. It's easy to avoid with a trivial calculation which takes into account the load factor.
- It's true that not all instances of this I changed in the PR are likely to be on the hot path, but fixing it everywhere is more of a consistency thing; if the helper is there, why not, otherwise it would just be more confusing.
- There are likely callers in the wild that at least would be hitting the `toMap` implementation of the `ReadOnlyStringMap`, e.g. likely through `ThreadContextMap` (I can confirm such instances internally, as well as in OSS projects such as Sentry).
- Whoever wrote those lines of code was attempting to size the HashMaps precisely by passing in a capacity, it just so happens that the number of items to insert != the map capacity, so seems like an easy mistake to have made. I assume care was taken here though since log4j generally tries to be optimal in terms of allocations and performance.

I assume the main point of contention is the addition of a utility class, and I agree it is ugly, but it's a stopgap until (?) Log4j can use the static `HashMap.newHashMap` added in Java 19, I presume when the baseline JDK version log4j requires allows us to use that, we can just switch to it and remove any such shim utility class. Or another way to put it, I presume you would readily accept this change if it were simply delegating to the JDK helpers (assuming we could use it)?

log4j-api/src/main/java/org/apache/logging/log4j/CloseableThreadContext.java
```diff
@@ -206,7 +207,7 @@ public void close() {
         }
 
         private void closeMap() {
-            final Map<String, String> valuesToReplace = new HashMap<>(originalValues.size());
+            final Map<String, String> valuesToReplace = Maps.newHashMap(originalValues.size());
```
vy, line 210: I've reviewed all `log4j-api` and `log4j-core` changes, and found this to be only change that one _might_ claim to be of importance. I prefer a PR that

1. Replaces all `HashMap#new(int)` calls with a `HashMap#new()` (Since they're all incorrectly implemented, it is better to just stick to defaults.)
2. Replaces only this line in `CloseableThreadContext` with `new HashMap((int) Math.ceil(originalValues.size() / 0.75))`

@ppkarwasz, WDYT?

ppkarwasz: Can you point out an example, where the usage of `HashMap#new(int)` is inappropriate?

I am mostly +1 on the proposed changes, as long as they don't introduce a new API. We have already dealt with methods and classes backported from newer JDK versions.

ppkarwasz: > All changes this PR highlights show `HashMap#new(int)` usages where the author of the associated piece of code _incorrectly_ assumed the provided capacity would avoid allocations, but it does not.

I must be tired, but could you point me to some examples? I checked a couple of usages and, if the capacity of the map is computed correctly, they don't seem to cause the map to be resized.

vy: For one, all `*Test` suggestions are redundant. The rest of the usages are incorrect, see `ThreadDumpMessage`:

```
Map<ThreadInformation, StackTraceElement[]> threads = new HashMap<>(map.size());
for (Map.Entry<Thread, StackTraceElement[]> entry : map.entrySet()) {
    threads.put(...);
}
```

Author had the intention of creating an initial `HashMap` of capacity `map.size()` to avoid resizing.
1. This is _incorrect_, since it _will be_ resized at the .75 mark
2. This code is hardly on a hot path, if used at all

I'm not saying what this PR implements is _incorrect_, instead, what is already there in the existing code base is incorrect. Plus, there is nothing to be corrected if just the defaults are used. Being extra smart does not buy anything but complicate the code.

vy: > * Keep the current code **unchanged**, accepting that it may trigger a resize for some values of n (but not all, since the threshold is always 0.75 × the smallest power of 2 ≥ n).

That is a fair path forward. This effectively translates to only fixing the `CloseableThreadContext` with an inline math operation, @ppkarwasz, do you agree?

ppkarwasz: The `internal` packages are not for users, but for us: BND Baseline will not complain if we remove the class and we'll not address any users complains, since they have been sufficiently warned.

I am for merging this one, if my comment above is addressed.

kilink: I've made the changes suggested by @ppkarwasz, thanks for the discussion / review.

PR #3874, opened 2025-08-08
Remove `jvmrunargs` lookup

### LINKED ISSUE #2726: The `jvmrunargs` lookup is always empty
The [`JmxRuntimeInputArgumentsLookup`](org.apache.logging.log4j.core.lookup.JmxRuntimeInputArgumentsLookup) lookup created with the standard constructor always returns `null`.

The lookup has a singleton instance `JMX_SINGLETON`, which is not used anywhere in the code.

ramanathan1504:
`jvmrunargs` was added in 2014-08-20 (1f64e68ed1), and never worked until it got accidentally fixed in `2.25.0` (see #3071 and f7c26cd7), which was released in 2025-06-13. We decided to remove it<sup>1</sup>. This fixes #2726.

<sup>1</sup> Link to [the internal Slack conversation](https://the-asf.slack.com/archives/C4TQW0M5L/p1754949663412429?thread_ts=1754895587.905569&cid=C4TQW0M5L). (Only accessible by certain ASF committers and some limited set of guests.)


log4j-core-test/src/test/java/org/apache/logging/log4j/core/lookup/JvmRunArgsLookupTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.lookup;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import org.apache.logging.log4j.core.LogEvent;
+import org.junit.Test;
```
ppkarwasz, line 23: While JUnit 4 is still present on the classpath, we use mainly JUnit5.
Could you convert this class to JUnit 5?

ppkarwasz (changes requested):
Hi @ramanathan1504,

Thanks for the PR! :100:

From what I can tell, the `jvmrunargs` lookup was already fixed in version `2.25.0`, so the main code changes here may not be needed. However, the unit tests you’ve added are very valuable. They’ll help ensure the lookup keeps working. The absence of such tests is likely why a feature that never fully worked managed to ship in so many releases.

Could you:

* Remove the main code changes but keep `JvmRunArgsLookupTest`.
* Consider adding documentation for `jvmrunargs` (either in this PR or in a follow-up). See #3876 for details.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/lookup/JvmRunArgsLookupTest.java
ramanathan1504: Updated the test cases with JUnit 5

vy (changes requested):
@ramanathan1504, would you mind implementing the following changes, please?

1. Remove `log4j2-test.xml` that you added
2. Remove `o.a.l.l.core.lookup.Interpolator::LOOKUP_KEY_JVMRUNARGS` field
3. Remove `o.a.l.l.core.lookup.JmxRuntimeInputArgumentsLookup`
5. Move the changelog entry file to `src/changelog/.2.x.x/3874_remove_jvmrunargs_lookup.xml`
6. Fix the changelog entry file syntax (see `3176_validate_scripts_in_ScriptsPlugin.xml` for an example)
7. Set the `entry.type` to `removed` in the changelog entry file

ramanathan1504 (commented):
Fixes #3874

- Removed `log4j2-test.xml`
- Removed `LOOKUP_KEY_JVMRUNARGS` field from `Interpolator`
- Removed `JmxRuntimeInputArgumentsLookup` class
- Moved and updated the changelog entry to `src/changelog/.2.x.x/3874_remove_jvmrunargs_lookup.xml` with the correct format and type

Please review and let me know if any further changes are needed.

src/changelog/.2.x.x/3874_remove_jvmrunargs_lookup.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="removed">
+    <issue id="3874" link="https://github.com/apache/logging-log4j2/issues/3874"/>
+    <description format="asciidoc">
+      Removed the `jvmrunargs` lookup and related code due to lack of clear use case and potential for confusion.
```
vy, line 10: We don't need to elaborate on our reasoning in the changelog, interested people can check the linked issue:

```suggestion
      Remove the `jvmrunargs` lookup
```

src/changelog/.2.x.x/3874_remove_jvmrunargs_lookup.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="removed">
+    <issue id="3874" link="https://github.com/apache/logging-log4j2/issues/3874"/>
```
vy, line 8:
```suggestion
    <issue id="3874" link="https://github.com/apache/logging-log4j2/pull/3874"/>
```

vy (changes requested):
@ramanathan1504, could you remove `JmxRuntimeInputArgumentsLookup`, please? (Please review the `Files changed` tab in GitHub after pushing your changes.)

ramanathan1504:
Understood, I’ll remove it. I’ll be more careful in future PRs to ensure such issues don’t occur again.

vy (approved):

vy:
@ramanathan1504, you need to sign your commits. The _"In a GitHub PR, how can I sign (not sign-off!) existing commits?"_ prompt creates [a good enough ChatGPT response](https://chatgpt.com/share/689f0f3a-22b8-8002-90d7-6b7c1d702ffa).

Note that since you'll be truncating the history anyway, you can as well squash all your changes into a single commit, sign it, and force-push it.

> I’ll be more careful in future PRs to ensure such issues don’t occur again.

No worries. Shit happens.

ppkarwasz (changes requested):
Hi @ramanathan1504,

Thanks for the contribution!

Removing a class is *technically* a breaking change, which is why the [BND Baseline check](https://github.com/apache/logging-log4j2/blob/2.x/BUILDING.adoc#fixing-api-compatibility-check-failures) fails.

However, since we’ve decided this particular removal should **not** be treated as a breaking change, you’ll need to override the check by doing the following:

1. Identify the element that **contains** the removed class. For classes, this means the containing package:
    https://github.com/apache/logging-log4j2/blob/2f79c39696a998ec1530aaf12de67a3c51ee1f13/log4j-core/src/main/java/org/apache/logging/log4j/core/lookup/package-info.java#L23-L28

2. Add a `@BaselineIgnore("2.26.0")` annotation to that package. This tells BND to ignore the semver violation up to version `2.26.0` of `log4j-core`.

3. Even though we’re not treating this as a `MAJOR` change, it **is** a `MINOR` change for the package. Per [OSGi semantic versioning](https://bnd.bndtools.org/chapters/170-versioning.html), you must bump the package version (`2.24.1`) by at least one minor version.
   Our convention is to align the package version with the library version, so update it to `@Version("2.26.0")`. This way, the package version clearly answers the question: *“In which Log4j version did the last API change in this package occur?”*

After these annotation changes, please run the build to identify and remove the tests that rely on the removed lookup.

ramanathan1504:
Hi @ppkarwasz,

Thanks for the detailed instructions! I’ll make the changes as suggested:
	•	Add @BaselineIgnore("2.26.0") to the package.
	•	Bump the package version to @Version("2.26.0").
	•	Update/remove tests that rely on the removed class.

I’ll push the updates shortly.

ramanathan1504:
Hi, I have applied the changes as you requested. The BND baseline check is now overridden.

I also had to fix an unrelated test failure in NamedInstantPatternTest that was causing the build to fail.

The full mvn clean install now passes successfully. This should be ready for another look. Thanks!

ramanathan1504:
@ppkarwasz 
Thanks for the clear guidance! You're right, I should have separated the concerns.
I have reverted the change to NamedInstantPatternTest.java and updated the PR. It now only contains the BND baseline changes as originally requested.
I've also opened a new Jira issue for the test failure here: issues #3885
I'll be sure to keep future PRs focused on a single issue. Thanks again for the review!

vy (approved):

ppkarwasz (approved):

PR #3880, opened 2025-08-13
Raise default async discard threshold to ERROR

rschmitt:
There are two basic scenarios where the ring buffer fills up. One is that an application is simply logging too much and the flushing process can't keep up, and in this case a discard threshold of WARN or INFO is probably sufficient to mitigate the problem. However, if flushing has stopped making progress altogether, e.g. due to a full or failed disk, then logging calls will block indefinitely. This can result in a production outage.

This change sets the default discard threshold to ERROR, in order to better mitigate the scenario where the disk fills up, fails, or is in the process of failing. With this threshold, logging should only block at the FATAL level, which would typically mean that the operation is already failing anyway.


rschmitt:
More controversially, I think that `asyncQueueFullPolicy` should default to `Discard` for basically the same reasons. One weird thing about this is that the blocking queue-full policy is literally named `DefaultAsyncQueueFullPolicy` and is denoted by the property value `Default`. =\

vy:
> if flushing has stopped making progress altogether, e.g. due to a full or failed disk, then logging calls will block indefinitely. This can result in a production outage.

If logging is a vital component of your application, and it doesn't work, I think it is reasonable to signal that the application is down. Consider this scenario in a cluster, say, Kubernetes, environment: container signals down in liveliness probes due to logging buffer failures, pod will be taken down, and re-spawned in a new environment with sufficient logging capacity. This is what you'd want, instead of losing all logging for an indefinite amount of time. I think this is a good default. If you indeed want the other way around, it makes sense that you need to opt-in for extra configuration, which is `log4j2.discardThreshold` in this case.

@remkop, @ppkarwasz, WDYT?

> More controversially, I think that `asyncQueueFullPolicy` should default to `Discard` for basically the same reasons. One weird thing about this is that the blocking queue-full policy is literally named `DefaultAsyncQueueFullPolicy` and is denoted by the property value `Default`. =\

This is a valid remark. I'd support a PR

1. renaming the default from `Default` to `Discard`, and
2. translating `Default` usages to `Discard` with a `WARN`'ing logged

rschmitt:
> If logging is a vital component of your application, and it doesn't work, I think it is reasonable to signal that the application is down.

See, this is what I thought too, but over the years we had so many outages caused by blocking log statements that we adopted these overridden defaults years ago, and I'm not aware of any problems that they've caused. Additionally, the default logging behavior is synchronous, right? If you opt in to asynchronous logging, you already run the risk of a crash causing all of the buffered log messages to be lost (and the log statements leading up to the crash are probably the most important ones!). The current defaults seem like the worst of both worlds: you're exposed to the risk of data loss _and_ the risk of an outage caused by indefinite blocking. The principled reason to discard instead of blocking is that that's the only way to guarantee asynchronous behavior (which is probably what you _think_ you're opting in to by using async logging in the first place).

rschmitt:
> That said, I’m not opposed to [...] adding a `Block` synonym for `Default`, so users can more clearly express their intent without having to remember the historical naming.

Yeah, I can do that.

remkop:
I agree with @ppkarwasz that changing the default behavior would be against the original intent of Log4j2 to not drop events. It would be surprising to users to change the default behavior. 

I would not oppose changing the name of the value used to configure the default behavior to a label that expresses the intention/behavior better.

PR #3884, opened 2025-08-15
Port GraalVM reachability to 3.x

### LINKED ISSUE #3055: Add GraalVM reachability metadata to `main`
We need to port #3053 to the `main` branch.

jvz:
This resolves #3055 by porting #3053. I've ported the annotation processor from 2.x and did some refactoring of the other annotation processor to work more like this new one.


log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/model/PluginRegistry.java
```diff
@@ -101,24 +103,44 @@ private void loadPlugins(final ClassLoader classLoader, final Namespaces namespa
...
+                                        .setClassName(in.readUTF())
+                                        .setName(in.readUTF())
+                                        .setPrintable(in.readBoolean())
+                                        .setDeferChildren(in.readBoolean())
+                                        .get();
+                                index.add(entry);
+                            }
+                        }
+                    }
+                }
```
ppkarwasz, line 132: I don't see how the changes to `log4j-plugins` are relevant to this PR. Could you revert them?

ppkarwasz (approved):
Hi @jvz,

Thanks for the contribution, and sorry for the delay. The PR looks good overall, but I noticed it includes some unrelated changes. Could you please separate those into a different PR? It helps keep the history cleaner and makes reviews easier.

log4j-plugins/src/main/java/org/apache/logging/log4j/plugins/model/PluginRegistry.java
jvz: This is an example of the type of trivial change that would have been simple to directly apply to the main branch but are now unable to do so without the formality of a PR and peer review.

ppkarwasz: Since they are “trivial” they will be reviewed fast and they give you the opportunity to explain why the change is needed.

PR #3901, opened 2025-08-27
Bump elasticsearch-java to version 9.x

### LINKED ISSUE #3663: Bump `elasticsearch-java` to version 9.x
## Description

Since `elasticsearch-java` version 9.x has breaking changes, it can not be upgraded automatically.

ramanathan1504:
Fixes #3663.

This pull request updates the dependencies in the `log4j-layout-template-json-test/pom.xml` file to keep the project up to date with the latest Elastic Stack versions and adds a new test dependency. The most important changes are:

**Dependency updates:**

* Bumped the `elastic.version` property from `8.17.3` to `9.1.3` to align with the latest Elastic Stack release.

**New test dependency:**

* Added the `elasticsearch-rest-client` library (version `8.13.4`) as a test-scoped dependency to support testing with the Elasticsearch REST client.

### DIFF
--- a/log4j-layout-template-json-test/pom.xml
+++ b/log4j-layout-template-json-test/pom.xml
@@ -84,15 +84,15 @@
     </dependency>
 
     <dependency>
-      <groupId>org.elasticsearch.client</groupId>
-      <artifactId>elasticsearch-rest-client</artifactId>
-      <version>${elastic.version}</version>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
       <scope>test</scope>
     </dependency>
 
     <dependency>
-      <groupId>org.awaitility</groupId>
-      <artifactId>awaitility</artifactId>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-client</artifactId>
+      <version>${elastic.version}</version>
       <scope>test</scope>
     </dependency>
 
@@ -360,9 +360,6 @@
                 <log4j.logstash.gelf.port>${logstash.gelf.port}</log4j.logstash.gelf.port>
                 <log4j.logstash.tcp.port>${logstash.tcp.port}</log4j.logstash.tcp.port>
               </systemPropertyVariables>
-              <jdkToolchain>
-                <version>17</version>
-              </jdkToolchain>
 
             </configuration>
             <executions>
--- a/log4j-layout-template-json-test/pom.xml
+++ b/log4j-layout-template-json-test/pom.xml
@@ -48,7 +48,7 @@
          2. The Docker image version of the ELK-stack
 
          As of 2024-09-16, these all (Maven artifacts and Elastic products) get released with the same version. -->
-    <elastic.version>8.17.3</elastic.version>
+    <elastic.version>9.1.3</elastic.version>
 
   </properties>
 
@@ -129,6 +129,13 @@
     <dependency>
       <groupId>co.elastic.logging</groupId>
       <artifactId>log4j2-ecs-layout</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-client</artifactId>
+      <version>8.13.4</version>
       <scope>test</scope>
     </dependency>
 
--- a/log4j-layout-template-json-test/pom.xml
+++ b/log4j-layout-template-json-test/pom.xml
@@ -48,7 +48,7 @@
          2. The Docker image version of the ELK-stack
 
          As of 2024-09-16, these all (Maven artifacts and Elastic products) get released with the same version. -->
-    <elastic.version>9.1.3</elastic.version>
+    <elastic.version>8.17.3</elastic.version>
 
   </properties>
 
@@ -195,6 +195,26 @@
 
       <build>
         <plugins>
+
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-toolchains-plugin</artifactId>
+            <version>3.1.0</version>
+            <configuration>
+              <toolchains>
+                <jdk>
+                  <version>17</version>
+                </jdk>
+              </toolchains>
+            </configuration>
+            <executions>
+              <execution>
+                <goals>
+                  <goal>toolchain</goal>
+                </goals>
+              </execution>
+            </executions>
+          </plugin>
 
           <plugin>
             <groupId>io.fabric8</groupId>
--- a/log4j-layout-template-json-test/pom.xml
+++ b/log4j-layout-template-json-test/pom.xml
@@ -84,15 +84,15 @@
     </dependency>
 
     <dependency>
-      <groupId>org.awaitility</groupId>
-      <artifactId>awaitility</artifactId>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-client</artifactId>
+      <version>${elastic.version}</version>
       <scope>test</scope>
     </dependency>
 
     <dependency>
-      <groupId>co.elastic.clients</groupId>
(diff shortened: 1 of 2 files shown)


ramanathan1504:
@ppkarwasz 
Thank you for the feedback
I will close this PR and create new ones and follow your proposal.

vy:
@ramanathan1504, please don't close PRs and start a new one each time – it makes it impossible to track the evolution of the eventual PR.

ppkarwasz:
I agree: the changes to `.github/dependabot.yaml` must be done on the **default** branch, which is `2.x`, so they can be done in this PR.

The `@FailAt` annotation also needs to be done in `2.x`.

Only the upgrade to `Rest5Client` needs a separate PR.

vy:
I think

1. **we're comparing apples with oranges**. We use ES client to test a functionality, the fact that it depends on Java 17 should not stop us from using it, in particular, given that we already use Java 17 to compile the sources. That is, we already have Java 17 at our disposal while running tests – saying _"Don't use Java 17 because our CI tests run in Java 8"_ does not make sense to me.
2. There is only one problematic test: `LogstashIT`. It is executed via `maven-failsafe-plugin` configured in the `docker` profile defined in `log4j-layout-template-json-test/pom.xml`, **it has nothing to do with `java8-tests` profile** in `log4j-parent/pom.xml`, and the latter forces Java 8.
3. Using different major versions across `2.x` and `main` will obstruct **porting process**.

@ppkarwasz, what about the following plan instead?

1. Use the latest and greatest ES client on `2.x`
2. Update `maven-failsafe-plugin` in the `docker` profile in `log4j-layout-template-json-test/pom.xml` to force Java 17

ppkarwasz:
Sounds good!

BTW, I was also considering dropping tests on Java 8, since the difference between JRE 8 and 9 is very painful.

ramanathan1504:
@ppkarwasz @vy 
I am reopened the pr, also little confused if anyone of you describe what's needed to do?

log4j-layout-template-json-test/pom.xml
```diff
@@ -132,6 +132,13 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-client</artifactId>
```
vy, line 137: Can you implement the following changes, please?

1. Remove the `elasticsearch-java` dependency (We don't need this anymore, right?)
2. Place the `elasticsearch-rest-client` dependency adhering to the lexicographic order of artifact IDs
3. Replace hardcoded `<version>8...` with `<version>${elastic.version}</version>`

log4j-layout-template-json-test/pom.xml
```diff
@@ -189,6 +196,26 @@
       <build>
         <plugins>
 
+          <plugin>
```
vy, line 199: You don't need this plugin. You just need to add a [jdkToolchain] entry to the `maven-failsafe-plugin` configuration.

vy (changes requested):

log4j-layout-template-json-test/pom.xml
```diff
@@ -84,15 +84,15 @@
     </dependency>
 
     <dependency>
-      <groupId>org.awaitility</groupId>
-      <artifactId>awaitility</artifactId>
+      <groupId>org.elasticsearch.client</groupId>
+      <artifactId>elasticsearch-rest-client</artifactId>
+      <version>${elastic.version}</version>
```
ppkarwasz, line 102: Could you keep the test dependencies in alphabetical order by `artifactId`?

log4j-layout-template-json-test/pom.xml
```diff
@@ -360,6 +360,10 @@
                 <log4j.logstash.gelf.port>${logstash.gelf.port}</log4j.logstash.gelf.port>
                 <log4j.logstash.tcp.port>${logstash.tcp.port}</log4j.logstash.tcp.port>
               </systemPropertyVariables>
+              <jdkToolchain>
+                <version>17</version>
+              </jdkToolchain>
```
ppkarwasz: I would drop this, since it requires all users to have Toolchains configured.

Since the default is to use the JVM running Maven and that one is already a JDK 17, no toolchains are necessary.

ppkarwasz (approved):
Hi @ramanathan1504,

Looks good to me, except the two remarks below.

log4j-layout-template-json-test/pom.xml
ramanathan1504: updated

log4j-layout-template-json-test/pom.xml
ramanathan1504: removed the toolchain

ppkarwasz:
@ramanathan1504,

You lost some signatures, can you rebase with `git rebase -S`?

ramanathan1504:
@ppkarwasz 
ok i will do that

ramanathan1504:
@ppkarwasz 
commits all are verified

ppkarwasz (approved):
LGTM, thanks! :100:

log4j-layout-template-json-test/src/test/java/org/apache/logging/log4j/layout/template/json/LogstashIT.java
vy, line 1: @ramanathan1504, why did you replace the strongly-typed client calls with raw JSON parsing? AFAIK, typed, say, `SearchResponse` is still supported[1]. If we'll perform raw HTTP-querying and deal with JSON marshalling ourselves, what is the point of using the ES client?

[1]: https://www.elastic.co/docs/reference/elasticsearch/clients/java/usage/searching

ramanathan1504: @vy 
I thought upgrading to the newer version might cause some issues, so I added a static JSON-based approach as a safer option. 

I will rework the PR to use SearchResponse instead.

ramanathan1504:
@vy 
Is this right?

recent 8.x version of the modern client library (co.elastic.clients:elasticsearch-java) for the Java code.

the new 9.x version for the Docker images of Elasticsearch and Logstash.

vy:
> @vy Is this right?
> 
> recent 8.x version of the modern client library (co.elastic.clients:elasticsearch-java) for the Java code.
> 
> the new 9.x version for the Docker images of Elasticsearch and Logstash.

No, use ES 9 everywhere. We use ES only for integration tests and, as @ppkarwasz stated earlier (I cannot find the commit associated with his message, since you force-pushed changes, but the conversation is still visible), ITs run using Java 17 anyway.

ramanathan1504:
> > @vy Is this right?
> > recent 8.x version of the modern client library (co.elastic.clients:elasticsearch-java) for the Java code.
> > the new 9.x version for the Docker images of Elasticsearch and Logstash.
> 
> No, use ES 9 everywhere. We use ES only for integration tests and, as @ppkarwasz stated earlier (I cannot find the commit associated with his message, since you force-pushed changes, but the conversation is still visible), ITs run using Java 17 anyway.
@vy 
am reverted back the LogstashIT to old and bumped the version only

vy (approved):

vy:
@ramanathan1504, after several review rounds, I could not succeed in making you remove redundant dependencies and unrelated changes. I eventually carried them out myself in 561107e6e. Please be more attentive to these details next time – it will make life easier for both of us.

ramanathan1504:
Really sorry for this,i will take care these things in future and i will retrospect this pr and comments..next time will do the good one thanks @vy

vy:
@ramanathan1504, would you submit a PR porting this to `main`, please? (Bump the version to `9.1.5` instead.)

ramanathan1504:
@vy sure i will do
thanks

ramanathan1504:
@vy can you review the pr #3951

Issue #2792, opened 2024-08-02
Uniformize TLS configuration


ppkarwasz:
Many Log4j components have a configuration attribute to enable the verification of the TLS server certificate:

- Network appenders (`Socket`, `SMTP`) use the [`verifyHostname`](https://logging.apache.org/log4j/2.x/manual/appenders.html#ssl-configuration) attribute of the `SSL` nested component to provide the same feature. Its default value is `false`.

- The [HTTP Appender](https://logging.apache.org/log4j/2.x/manual/appenders.html#HttpAppender) has a configuration attribute `verifyHostname`. The value defaults to `true`. Note that the HTTP Appender can also have a nested `SSL` component, but the value of `SSL.verifyHostname` is ignored.

- We also have a `log4j2.sslVerifyHostName` configuration property that is used if the `SSL` component is absent.

I understand that in the past only HTTP servers had a X509 certificate issued by a public CA. However nowadays most SMTP servers have also publicly verifiable X509 certificates, so we can switch both defaults to `true`.

Besides that a public X509 certificate was never required by our appenders: they only connect to a single host.

## Proposed changes

- Let us deprecate `HTTP.verifyHostname` in `2.x`. In `3.x` we can still keep it, but set its default value to `SSL.verifyHostname`.
- We can switch the default value of `SSL.verifyHostname` to `true`. This might require some additional work: the SMTP and HTTP appenders only connect to the host, when there is a log event to send. In the case of SMTP this happens only for `ERROR` log events by default, so users might realize that they have a configuration problem much later.
- Currently if the user configures an `SSL` element, all the `log4j2.ssl*` configuration properties are ignored. I think those properties should still be used to provide default values for `SSL`.

vy:
@ppkarwasz, good examination. Agreed with the proposal. Some remarks:

> Let us deprecate `HTTP.verifyHostname` in `2.x`. In `3.x` we can still keep it, but set its default value to `SSL.verifyHostname`.

I'd be in favor of removing `HTTP.verifyHostname` in `3.x`.

> Currently if the user configures an `SSL` element, all the `log4j2.ssl*` configuration properties are ignored. I think those properties should still be used to provide default values for `SSL`.

This is a good improvement anyway. I would rather implement this in a separate PR and avoid unnecessarily involving it in the discussions we will have for other proposed changes in this PR.

ppkarwasz:
[...]

PR #3902, opened 2025-08-28
Unify TLS configuration

jhl221123:
- Introduced `TlsConfiguration` interface
- Defaulted `verifyHostname` to `true`
- Repurposed `Http.verifyHostname` as a toggle for backward compatibility

Related to #2792

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/HttpAppender.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/HttpAppender.java
@@ -82,7 +82,8 @@ public final class HttpAppender extends AbstractAppender {
             }
 
             if (verifyHostname != null) {
-                LOGGER.warn("Http#verifyHostname is deprecated. please configure this on the TLS instead.");
+                LOGGER.warn(
+                        "`verifyHostname` attribute of `HttpAppender` is deprecated and ignored. Use a `TlsConfiguration` element to configure this attribute.");
             } else if (sslConfiguration != null) {
                 verifyHostname = sslConfiguration.isVerifyHostName();
             } else {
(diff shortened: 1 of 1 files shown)


jhl221123:
@vy Thanks for the feedback!

> @jhl221123, do you plan to update this PR to reflect this too?

Absolutely! But I noticed in #2792 you mentioned that it might be better to handle this in a separate PR. Which approach would you prefer: should I continue updating this PR, or would it be better to open a separate one for it?

ppkarwasz:
> I'm not able to see the value of a new `TlsConfiguration` interface here. The fact that it contains a `getSslContext` method (note the `Ssl` in the name!) makes it even more clumsy. I know that SSL keyword is superseded by TLS, but from an API point of view in the Java ecosystem, is it really? What will be the benefit of introducing `TlsFoo` as long as we will have `SslFoo` and `SslBar` around for functionality, backward compatibility, etc. reasons?

I don't see the point in calling it `TlsConfiguration` instead of `SslConfiguration` either, except the fact that the `SslConfiguration` name is already used.

Where I *do* see value, though, is in introducing an **interface** rather than tying everything to the concrete `SslConfiguration` class. That opens the door for third-party Log4j Core plugins to provide their own TLS material without having to subclass Log4j internals.

Most popular Java HTTP clients (`HttpsURLConnection`, Java 11+ `HttpClient`, OkHttp, Apache HttpClient) can all be configured with essentially three knobs:

```java
public interface TlsConfiguration {
    SSLContext getSslContext();
    HostnameVerifier getHostnameVerifier();
    /** Optional **/
    SSLParameters getSslParameters();
}
```

Jakarta Mail is the only outlier that still needs an `isVerifyHostName()` flag.

With such an abstraction, you could, for example, plug in Spring Boot’s [`SslBundle`](https://docs.spring.io/spring-boot/reference/features/ssl.html) to supply the key/trust material. That would make it straightforward to integrate Log4j into environments where TLS credentials are managed externally (e.g. centralized config via Spring Cloud).

Each host would only need to declare which `SslBundle` to use, and Log4j could obtain everything else through the interface.

@rgoers: does this kind of abstraction line up with what you’d find useful in practice?

vy:
@ppkarwasz, thanks for clarifying the use case in mind. That certainly helps with setting the feature's intended context.

> Where I _do_ see value, though, is in introducing an **interface** rather than tying everything to the concrete `SslConfiguration` class.

I understand and share your preference for an interface, yet given the current state of the code, wouldn't it be less intrusive to provide a similar adapter pattern support by adding a sufficient `SslConfiguration.create(SSLContext, HostnameVerifier, SSLParameters)` factory method?

PR #3919, opened 2025-09-07
fix: make `%ex` behavior context-independent

### LINKED ISSUE #3873: Add support for old-style exception rendering in Pattern Layout
Due to https://github.com/apache/logging-log4j2/pull/3045, all exception messages are prefixed with a newline now:

```text
[SIGTSTP] [1] ~/src/logging-log4j2 # rg -txml pattern
src/main/resources/log4j2.xml
6:                <pattern>%date{EEE MMM dd HH:mm:ss yyyy zzz}{UTC} [%p] (%t) %c: %m %ex%n</pattern>
```

```text
[0] [1] ~/src/logging-log4j2 # ./run.sh
Darwin 842f57b044d8 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:53:27 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6041 arm64
Log4j 2.24.3:
Thu Aug 07 17:22:01 2025 UTC [ERROR] (main) org.example.App: Going to call a method now
Thu Aug 07 17:22:01 2025 UTC [ERROR] (main) org.example.App: Got exception while calling throwException, how could this have happened java.lang.RuntimeException: This is the message passed to the RuntimeException ctor
	at org.example.App.throwException(App.java:20)
	at org.example.App.main(App.java:12)

Thu Aug 07 17:22:01 2025 UTC [ERROR] (main) org.example.App: Done calling method

Log4j 2.25.0:
Thu Aug 07 17:22:02 2025 UTC [ERROR] (main) org.example.App: Going to call a method now
Thu Aug 07 17:22:02 2025 UTC [ERROR] (main) org.example.App: Got exception while calling throwException, how could this have happened
java.lang.RuntimeException: This is the message passed to the RuntimeException ctor
	at org.example.App.throwException(App.java:20)
	at org.example.App.main(App.java:12)

Thu Aug 07 17:22:02 2025 UTC [ERROR] (main) org.example.App: Done calling method
```

I don't see a way to reproduce the 2.24.3 style of log messages with the [pattern layout](https://logging.apache.org/log4j/2.x/manual/pattern-layout.html#converter-exception).

1. Using `%ex{0}` to get the first line doesn't work, because a line separator is prepended.
2. I can't find anything in `ThrowableFormatOptions` that provides the type of the exception thrown. A pattern like `%m %ex{short.message}%n` doesn't fully reproduce the top line of the stack trace.
3. Using `%n%ex` to print the rest of the stack trace repeats the first line. In other words, there's no way to just print the _frames_ of the stack trace.

Since this layout change affects log scans, I'd like it if there were a way to specify a pattern that reproduces the pre-2.25 layout for anyone who depends on it.

ppkarwasz:
Historically, throwable pattern converters (`%ex`, `%xEx`, etc.) behaved in a context-sensitive way:

* If the **preceding formatter’s expansion** did not end with whitespace, the converter automatically inserted a space before rendering the exception.

In version `2.25.0`, this was changed to insert a **newline** instead of a space, but the behavior was still dependent on surrounding context.

#### What this change does

This PR removes the context-dependent behavior altogether and makes `%ex` expansion fully predictable, while remaining backward-compatible:

* When `%ex` is **added implicitly** because `alwaysWriteExceptions=true`:

  * If the pattern already ends with `%n`, a plain `%ex` is appended.
  * Otherwise, `%notEmpty{%n%ex}` is appended. This ensures exceptions are always clearly separated from the main log message by a newline, without adding extra characters when no exception is present.
* When `%ex` is **explicitly included** in the pattern by the user:

  * Its expansion is rendered exactly as written in the pattern.
  * It will **never** prepend a newline on its own.

#### Why

* Eliminates confusing context-sensitive behavior.
* Makes output consistent and predictable.
* Preserves legacy expectations by only modifying implicitly added `%ex`.

Closes #3873

### DIFF
--- a/src/changelog/.2.x.x/3873_throwable_converter_new_line.xml
+++ b/src/changelog/.2.x.x/3873_throwable_converter_new_line.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+    <issue id="3873" link="https://github.com/apache/logging-log4j2/issues/3873"/>
+    <description format="asciidoc">
+        Throwable converters no longer prepend newlines based on context; output follows the pattern exactly.
+    </description>
+</entry>
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/PatternParserTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/PatternParserTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.logging.log4j.core.pattern;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
@@ -23,6 +24,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Calendar;
 import java.util.List;
+import java.util.stream.Stream;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.MarkerManager;
 import org.apache.logging.log4j.core.LogEvent;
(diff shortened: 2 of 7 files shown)


src/changelog/.2.x.x/3873_throwable_converter_new_line.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+    <issue id="3873" link="https://github.com/apache/logging-log4j2/issues/3873"/>
+    <description format="asciidoc">
+        Throwable converters no longer prepend newlines based on context; output follows the pattern exactly.
```
vy:
```suggestion
    <issue id="3873" link="https://github.com/apache/logging-log4j2/issues/3873"/>
    <issue id="3919" link="https://github.com/apache/logging-log4j2/pull/3919"/>
    <description format="asciidoc">
        Fix Pattern Layout exception stack trace converters to no longer prepend newlines based on context
```

vy (approved):

PR #3921, opened 2025-09-08
Add new `ConfigurationFactory::getConfiguration` accepting multiple URIs

### LINKED ISSUE #3775: Add API to List Supported Configuration File Locations
**Summary:**

Introduce a method in Log4j Core to list the configuration file locations currently supported by the deployment.

### 🔍 Motivation

The logic for determining which configuration file formats are supported differs between Log4j Core **2.x** and **3.x**:

* **In 2.x**:
  All `ConfigurationFactory` implementations reside in `log4j-core`, but some depend on **optional** dependencies (JSON and YAML), meaning not all formats are necessarily available at runtime.

* **In 3.x**:
  Each `ConfigurationFactory` is moved to its own module/artifact. These factories are activated as soon as their respective artifact is on the classpath.

  * One exception is `XmlConfigurationFactory`, which remains in `log4j-core`, but can only be used if:

    * The `java.xml` module is present (on the JVM), or
    * An external `DocumentBuilderFactory` is available (e.g., on Android where the default factory lacks required options)

This evolving behavior makes it difficult for external systems to know **which config files Log4j Core will actually process**.

### 📦 Use Case

**Spring Boot** needs a reliable way to determine which configuration file formats are supported by the current Log4j Core deployment, in order to correctly customize its configuration file search logic (e.g., to support files like `log4j2-spring.xml`).

Currently, the [`getStandardConfigLocations()`](https://docs.spring.io/spring-boot/api/java/org/springframework/boot/logging/AbstractLoggingSystem.html#getStandardConfigLocations%28%29) and `getSpringConfigLocations()` methods in `Log4j2LoggingSystem` rely on **hardcoded assumptions** about which formats are supported. This approach is fragile and tightly coupled to specific versions of Log4j Core.

This hardcoding becomes particularly problematic with Log4j Core 3:

* For example, Spring Boot will **not** look for a `log4j2-spring.json` configuration file if the **Jackson** library is missing, despite JSON being a supported format in Log4j Core 3.
* Conversely, it may attempt to load a `log4j2-spring.properties` file even when the `log4j-config-properties` module is not present, resulting in configuration failure.

### ✅ Proposed Solution

Introduce a public API in Log4j Core that returns the set of **actively supported default configuration file locations**, based on the currently available `ConfigurationFactory` implementations.

One option would be to add a method like `getDefaultConfigLocations()` to the `ConfigurationFactory` class. This method would return the list of standard file names (e.g., `log4j2.xml`, `log4j2.json`, etc.) that the factory is capable of processing in the current runtime environment.

```java
public abstract class ConfigurationFactory {

    /**
     * Returns the default configuration file names that this factory supports,
     * given the specified context name. For example:
     * ["log4j2.xml", "log4j2" + contextName + ".xml"]
     */
    public abstract String[] getDefaultConfigLocations(String contextName);
}
```

Each factory would return only the file names it can currently handle — based on available modules, libraries, or runtime conditions (e.g., presence of `java.xml` for XML parsing) and the default `ConfigurationFactory` would just aggregate the results of all available configuration factories.

**Note:** As discussed in #1501, it may be worth adding a warning mechanism when a configuration file matching the pattern `log4j2<contextName>.<extension>` is present on the classpath, but no active `ConfigurationFactory` is available to handle that file type.

yybmion:
Adds a new `ConfigurationFactory::getConfiguration` overload accepting multiple URIs.

This work is a spin-off from #3839 and fixes #3775.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
@@ -140,8 +140,7 @@ class ConfigurationFactoryTest {
     void testGetConfigurationWithNullUris() {
         final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
-            factory.getConfiguration(context, "test", (List<URI>) null);
-            Mockito.verify(factory).getConfiguration(Mockito.same(context), Mockito.eq("test"), (URI) Mockito.isNull());
+            assertThrows(NullPointerException.class, () -> factory.getConfiguration(context, "test", (List<URI>) null));
         }
     }
 
@@ -158,9 +157,8 @@ class ConfigurationFactoryTest {
     void testGetConfigurationWithNullInList() {
         final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
-            final List<URI> listWithNull = Collections.singletonList(null);
-            factory.getConfiguration(context, "test", listWithNull);
-            Mockito.verify(factory).getConfiguration(Mockito.same(context), Mockito.eq("test"), (URI) Mockito.isNull());
+            final List<URI> listWithNull = Arrays.asList(URI.create("path:://to/nowhere"), null);
+            assertThrows(NullPointerException.class, () -> factory.getConfiguration(context, "test", listWithNull));
         }
     }
 
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
@@ -19,7 +19,6 @@ package org.apache.logging.log4j.core.config;
 import static org.apache.logging.log4j.util.Unbox.box;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -27,7 +26,6 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
 import java.net.URI;
-import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
@@ -48,6 +46,7 @@ import org.apache.logging.log4j.test.junit.TempLoggingDir;
 import org.apache.logging.log4j.util.Strings;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
 
 class ConfigurationFactoryTest {
 
@@ -139,43 +138,41 @@ class ConfigurationFactoryTest {
 
     @Test
     void testGetConfigurationWithNullUris() {
-        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
-            assertThrows(NullPointerException.class, () -> factory.getConfiguration(context, "test", (List<URI>) null));
+            factory.getConfiguration(context, "test", (List<URI>) null);
+            Mockito.verify(factory).getConfiguration(Mockito.same(context), Mockito.eq("test"), (URI) Mockito.isNull());
         }
     }
 
     @Test
     void testGetConfigurationWithEmptyUris() {
-        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
-            assertThrows(
-                    IllegalArgumentException.class,
-                    () -> factory.getConfiguration(context, "test", Collections.emptyList()));
+            factory.getConfiguration(context, "test", Collections.emptyList());
+            Mockito.verify(factory).getConfiguration(Mockito.same(context), Mockito.eq("test"), (URI) Mockito.isNull());
         }
     }
 
     @Test
     void testGetConfigurationWithNullInList() {
-        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
             final List<URI> listWithNull = Collections.singletonList(null);
-            assertThrows(ConfigurationException.class, () -> factory.getConfiguration(context, "test", listWithNull));
+            factory.getConfiguration(context, "test", listWithNull);
+            Mockito.verify(factory).getConfiguration(Mockito.same(context), Mockito.eq("test"), (URI) Mockito.isNull());
         }
     }
 
     @Test
     void testGetConfigurationWithSingleUri() throws Exception {
-        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final ConfigurationFactory factory = Mockito.spy(ConfigurationFactory.getInstance());
         try (final LoggerContext context = new LoggerContext("test")) {
-            final URL resource = getClass().getResource("/log4j-test1.xml");
-            assertNotNull(resource);
-
-            final List<URI> singleUri = Collections.singletonList(resource.toURI());
-            final Configuration config = factory.getConfiguration(context, "test", singleUri);
-
-            assertNotNull(config);
-            assertFalse(config instanceof CompositeConfiguration);
+            final URI configLocation =
+                    getClass().getResource("/log4j-test1.xml").toURI();
+            factory.getConfiguration(context, "test", Collections.singletonList(configLocation));
+            Mockito.verify(factory)
+                    .getConfiguration(Mockito.same(context), Mockito.eq("test"), Mockito.eq(configLocation));
         }
     }
 
@@ -183,16 +180,13 @@ class ConfigurationFactoryTest {
     void testGetConfigurationWithMultipleUris() throws Exception {
         final ConfigurationFactory factory = ConfigurationFactory.getInstance();
         try (final LoggerContext context = new LoggerContext("test")) {
-            final URL resource1 = getClass().getResource("/log4j-test1.xml");
-            final URL resource2 = getClass().getResource("/log4j-xinclude.xml");
-            assertNotNull(resource1);
-            assertNotNull(resource2);
-
-            final List<URI> multipleUris = Arrays.asList(resource1.toURI(), resource2.toURI());
-            final Configuration config = factory.getConfiguration(context, "test", multipleUris);
-
-            assertNotNull(config);
-            assertTrue(config instanceof CompositeConfiguration);
+            final URI configLocation1 =
+                    getClass().getResource("/log4j-test1.xml").toURI();
+            final URI configLocation2 =
+                    getClass().getResource("/log4j-xinclude.xml").toURI();
+            final List<URI> configLocations = Arrays.asList(configLocation1, configLocation2);
+            final Configuration config = factory.getConfiguration(context, "test", configLocations);
+            assertInstanceOf(CompositeConfiguration.class, config);
         }
     }
 }
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
@@ -19,14 +19,19 @@ package org.apache.logging.log4j.core.config;
 import static org.apache.logging.log4j.util.Unbox.box;
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -36,6 +41,7 @@ import org.apache.logging.log4j.core.Filter;
 import org.apache.logging.log4j.core.Logger;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.appender.ConsoleAppender;
+import org.apache.logging.log4j.core.config.composite.CompositeConfiguration;
 import org.apache.logging.log4j.core.filter.ThreadContextMapFilter;
 import org.apache.logging.log4j.core.test.junit.LoggerContextSource;
 import org.apache.logging.log4j.test.junit.TempLoggingDir;
@@ -129,5 +135,64 @@ class ConfigurationFactoryTest {
         checkConfiguration(context);
         final Path logFile = loggingPath.resolve("test-properties.log");
         checkFileLogger(context, logFile);
+    }
+
+    @Test
+    void testGetConfigurationWithNullUris() {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            assertThrows(NullPointerException.class, () -> factory.getConfiguration(context, "test", (List<URI>) null));
+        }
+    }
+
+    @Test
+    void testGetConfigurationWithEmptyUris() {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            assertThrows(
+                    IllegalArgumentException.class,
+                    () -> factory.getConfiguration(context, "test", Collections.emptyList()));
+        }
+    }
+
+    @Test
+    void testGetConfigurationWithNullInList() {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            final List<URI> listWithNull = Collections.singletonList(null);
+            assertThrows(ConfigurationException.class, () -> factory.getConfiguration(context, "test", listWithNull));
+        }
+    }
+
+    @Test
+    void testGetConfigurationWithSingleUri() throws Exception {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            final URL resource = getClass().getResource("/log4j-test1.xml");
+            assertNotNull(resource);
+
+            final List<URI> singleUri = Collections.singletonList(resource.toURI());
+            final Configuration config = factory.getConfiguration(context, "test", singleUri);
+
+            assertNotNull(config);
+            assertFalse(config instanceof CompositeConfiguration);
+        }
+    }
+
+    @Test
+    void testGetConfigurationWithMultipleUris() throws Exception {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            final URL resource1 = getClass().getResource("/log4j-test1.xml");
+            final URL resource2 = getClass().getResource("/log4j-xinclude.xml");
+            assertNotNull(resource1);
+            assertNotNull(resource2);
+
+            final List<URI> multipleUris = Arrays.asList(resource1.toURI(), resource2.toURI());
+            final Configuration config = factory.getConfiguration(context, "test", multipleUris);
+
+            assertNotNull(config);
+            assertTrue(config instanceof CompositeConfiguration);
+        }
     }
 }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import org.apache.logging.log4j.Level;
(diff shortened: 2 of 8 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,73 @@ public Configuration getConfiguration(
...
+        for (final URI uri : uris) {
+            try {
+                final Configuration config = getConfiguration(loggerContext, name, uri);
+
+                if (config != null) {
+                    if (config instanceof AbstractConfiguration) {
+                        configurations.add((AbstractConfiguration) config);
+                    } else {
+                        LOGGER.error("Configuration at {} is not an AbstractConfiguration", uri);
+                        failedUris.add(uri);
```
vy: I'm in favor of keeping things simple: if `config instanceof AC` fails, throw a `ConfigurationException`. This will render `failedUris` booking, `config != null` checks, and `catch` at line 385 redundant.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,73 @@ public Configuration getConfiguration(
...
+            throw new IllegalArgumentException("URI list cannot be empty");
+        }
+
+        if (uris.size() == 1) {
+            final Configuration config = getConfiguration(loggerContext, name, uris.get(0));
+            if (config == null) {
+                throw new ConfigurationException("Failed to create configuration from: " + uris.get(0));
+            }
+            return config;
+        }
```
vy: You don't need this `if` block, it is already handled by
```
return configurations.size() == 1 ? configurations.get(0)
```

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
yybmion: Changes applied as requested.

yybmion:
Thanks for the review, @vy. I've updated the PR.  

If this looks good, I'll proceed with adding tests and changelog.

vy:
> If this looks good, I'll proceed with adding tests and changelog.

@yybmion, yes, please proceed with tests and a changelog.

yybmion:
> @yybmion, yes, please proceed with tests and a changelog.

@vy , Added tests and changelog as requested.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
vy, line 1: Would you mind putting all `final LoggerContext context = new LoggerContext("test")` lines in try-with-resources blocks, please? That is,

```
try (final LoggerContext context = new LoggerContext("test")) {
    // ...
}
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
```diff
@@ -130,4 +136,54 @@ void properties(final LoggerContext context) throws IOException {
...
+    }
+
+    @Test
+    void testGetConfigurationWithEmptyUris() {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final LoggerContext context = new LoggerContext("test");
+
+        assertThrows(
+                IllegalArgumentException.class,
+                () -> factory.getConfiguration(context, "test", Collections.emptyList()));
```
vy: Could you also add a test for a list containing `null`, please? That is, `Collections.emptyList(null)`.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
```diff
@@ -130,4 +136,54 @@ void properties(final LoggerContext context) throws IOException {
...
+        assertFalse(config instanceof CompositeConfiguration);
+    }
+
+    @Test
+    void testGetConfigurationWithMultipleUris() throws Exception {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        final LoggerContext context = new LoggerContext("test");
+
+        final URL resource1 = getClass().getResource("/log4j-test1.xml");
+        final URL resource2 = getClass().getResource("/log4j-test1.properties");
```
vy: If I'm not mistaken, both files are semantically identical. Could you use a totally different configuration for the `resource2`, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/json/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Classes and interfaces supporting configuration of Log4j 2 with JSON.
  */
 @Export
-@Version("2.20.1")
+@Version("2.21.0")
```
vy:
```suggestion
@Version("2.26.0")
```

See [Fixing API compatibility check failures](https://github.com/apache/logging-log4j2/blob/2.x/BUILDING.adoc#fixing-api-compatibility-check-failures).

log4j-core/src/main/java/org/apache/logging/log4j/core/config/properties/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Configuration using Properties files.
  */
 @Export
-@Version("2.20.1")
+@Version("2.21.0")
```
vy:
```suggestion
@Version("2.26.0")
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/xml/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Classes and interfaces supporting configuration of Log4j 2 with XML.
  */
 @Export
-@Version("2.20.2")
+@Version("2.21.0")
```
vy:
```suggestion
@Version("2.26.0")
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/yaml/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Classes and interfaces supporting configuration of Log4j 2 with YAML.
  */
 @Export
-@Version("2.20.1")
+@Version("2.21.0")
```
vy:
```suggestion
@Version("2.26.0")
```

src/changelog/.2.x.x/add_getConfiguration_method_for_mulitiple_URIs.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+  type="added">
+  <issue id="3775" link="https://github.com/apache/logging-log4j2/issues/3775"/>
+  <description format="asciidoc">
+    Adds a new public API to ConfigurationFactory for creating configurations from multiple URIs, automatically combining them into a CompositeConfiguration when needed.
```
vy: Add a new `ConfigurationFactory::getConfiguration` method accepting multiple ``URI``s

src/changelog/.2.x.x/add_getConfiguration_method_for_mulitiple_URIs.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+  type="added">
+  <issue id="3775" link="https://github.com/apache/logging-log4j2/issues/3775"/>
```
vy, line 8:
```suggestion
  <issue id="3775" link="https://github.com/apache/logging-log4j2/issues/3775"/>
  <issue id="3921" link="https://github.com/apache/logging-log4j2/pull/3921"/>
```

vy (changes requested):

yybmion:
Thanks for the review @vy, Changes applied as requested.

Added a **null element check in the URI list** to throw a ConfigurationException when any element is null. This ensures that `testGetConfigurationWithNullInList` passes.

```java
for (final URI uri : uris) {

     if (uri == null) {
          throw new ConfigurationException("URI list contains null element");
   }
...
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
yybmion: Yes, you're right — that was my mistake. I've updated the resources to use two different configurations: `/log4j-test1.xml` and `/log4j-xinclude.xml`.

vy (approved):
@yybmion, thanks so much for your patience. I am happy with the last state of the PR.

@ppkarwasz, do you any remarks?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/json/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Classes and interfaces supporting configuration of Log4j 2 with JSON.
  */
 @Export
-@Version("2.20.1")
+@Version("2.26.0")
```
vy, line 21: @yybmion, just a sec, why do we need these version bumps? We did not touch to these packages.

vy (changes requested):

yybmion:
@vy, I'm sorry about the confusion. Previously, I bumped the versions of those packages in the last PR (#3839) after making some changes, but I forgot to revert them. I've now restored them to their original values.

vy:
@yybmion, CI builds are failing. Would you mind reading [the Log4j Development Guide](https://github.com/apache/logging-log4j2/blob/2.x/BUILDING.adoc#development) and resolving all build failures, please?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,52 @@ public Configuration getConfiguration(
...
+     * Creates a Configuration from multiple configuration URIs.
+     * If multiple URIs are successfully loaded, they will be combined into a CompositeConfiguration.
+     *
+     * @param loggerContext the logger context (may be null)
+     * @param name the configuration name (may be null)
+     * @param uris the list of configuration URIs (must not be null or empty)
+     * @return a Configuration created from the provided URIs
+     * @throws NullPointerException if uris is null
+     * @throws IllegalArgumentException if uris is empty
+     * @throws ConfigurationException if no valid configuration could be created
```
ppkarwasz: This change introduces an inconsistency between two similarly named methods:

* `getConfiguration(LoggerContext, String, URI)`: returns `null` on error
* `getConfiguration(LoggerContext, String, List<URI>)`: throws `ConfigurationException` on error

Personally, I prefer the **exception-throwing** approach since it forces the caller to handle the error explicitly.
That said, we could instead align with the existing “`StatusLogger` + return null” pattern here.

If we keep the latter approach, then `LoggerContext.start(Configuration)` and `LoggerContext.reconfigure(Configuration)` should defensively null-check the `Configuration` argument to ensure that the call ignores `null` as argument.

yybmion:
> @yybmion, just a sec, why do we need these version bumps? We did not touch to these packages.

@vy I think the version bumps are actually needed because the new public method in ConfigurationFactory is inherited by all its subclasses (JsonConfigurationFactory, XmlConfigurationFactory, etc.), which adds the method to those packages' public API.

This is causing the build failure. I've updated the versions back to 2.26.0.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,52 @@ public Configuration getConfiguration(
...
+     * from any of the provided URIs
+     * @since 2.26.0
+     */
+    public Configuration getConfiguration(final LoggerContext loggerContext, final String name, final List<URI> uris) {
+
+        Objects.requireNonNull(uris, "uris parameter cannot be null");
+
+        if (uris.isEmpty()) {
+            throw new IllegalArgumentException("URI list cannot be empty");
+        }
```
ppkarwasz: If the list is empty, I’d delegate to `getConfiguration(loggerContext, name, (URI) null)`. This allows the factory to either auto-detect a configuration or return `null`.

Also note: the `name` parameter only matters in the empty-list case. When a non-`null` `URI` is provided, `getConfiguration(LoggerContext, String, URI)` ignores `name` entirely.
So, if we decide to forbid empty lists, the `name` parameter becomes unnecessary.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,52 @@ public Configuration getConfiguration(
...
+
+        final List<AbstractConfiguration> configurations = new ArrayList<>();
+
+        for (final URI uri : uris) {
+
+            if (uri == null) {
+                throw new ConfigurationException("URI list contains null element");
+            }
```
ppkarwasz: *Nit*: `IllegalArgumentException` seems more appropriate here. `ConfigurationException` suggests an unexpected runtime failure, whereas a null-check is a precondition the caller should do himself.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +334,52 @@ public Configuration getConfiguration(
...
+
+            final Configuration config = getConfiguration(loggerContext, name, uri);
+
+            if (config == null) {
+                throw new ConfigurationException("Failed to create configuration from: " + uri);
+            }
+
+            if (!(config instanceof AbstractConfiguration)) {
+                throw new ConfigurationException("Configuration at " + uri + " is not an AbstractConfiguration");
+            }
```
ppkarwasz: This check is only necessary when `uris` has more than one element and we need to build a `CompositeConfiguration`.

If `uris` is a singleton, we can skip the check and just return the configuration directly.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
```diff
@@ -130,4 +136,63 @@ void properties(final LoggerContext context) throws IOException {
...
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
+        try (final LoggerContext context = new LoggerContext("test")) {
+            final List<URI> listWithNull = Collections.singletonList(null);
+            assertThrows(ConfigurationException.class, () -> factory.getConfiguration(context, "test", listWithNull));
+        }
+    }
+
+    @Test
+    void testGetConfigurationWithSingleUri() throws Exception {
+        final ConfigurationFactory factory = ConfigurationFactory.getInstance();
```
ppkarwasz: Since the goal is to test the basic implementation, we could use a mock that delegates to the real `getConfiguration(LoggerContext, String, List<URI>)` method, while mocking only `getConfiguration(LoggerContext, String, URI)`.

ppkarwasz (changes requested):
Hi @yybmion,

Thanks for the PR, and apologies for the delayed feedback: I haven’t had as much time to spend on Log4j recently as I’d like.

I like the direction of your change, but there are still a few edge cases we need to address.

In particular, I expected the PR to integrate the newly introduced method into the existing code.

For example, `ConfigurationFactory.Factory#getConfiguration(LoggerContext, String, URI)` could benefit from using it: we currently have a couple of legacy mechanisms for passing multiple URIs through a single `URI` parameter:

* If `uri == null`, the `log4j2.configurationFile` property is treated as a comma-separated list of file paths or URIs.
* If `uri != null`, there’s a legacy (and admittedly hacky) method of cramming two URIs into one string (see `parseConfigLocations`), which unfortunately still needs to be supported.

These mechanisms are handled in `getConfiguration(LoggerContext, String, URI)`, but I think we should delegate them to the newly introduced method.

ppkarwasz:
I realize my comments go a bit beyond the immediate scope of this PR, but the section of code below has become a real maintenance headache. It would be great if we could take the opportunity to clean it up once and for all:

https://github.com/apache/logging-log4j2/blob/8a3fb534f999632d0688a9a204f11b3dd7f59f72/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java#L373-L478

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: I'm fine with delegating to `getConfiguration(loggerContext, name, (URI) null)`, but don't return `null` please.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: We have discussed this earlier, and agreed that, since this is a new API, we will go with the exception-throwing approach. @ppkarwasz you can see your thumbs up in the discussion: https://github.com/apache/logging-log4j2/pull/3839#issuecomment-3265754355

ppkarwasz: You're right that we discussed on [`dev@logging`](https://lists.apache.org/thread/h2oydyk6xld47ljttqvflbt4530o73vw) that we should **not** use partially constructed configurations.

However, my concern here is slightly different: we now have two **overloads** with fundamentally different error-handling conventions. Java deprecated the `File` API partly because it behaved inconsistently across methods and this proposal would introduce a similar inconsistency between these overloads.

I see three possible approaches:

1. **Align all methods to throw on error.**
   This approach would make error handling explicit and predictable. Currently, `getConfiguration` can already throw when an invalid properties configuration is encountered. However, most other factories behave differently: they either return `null` when the file is missing or the factory is inactive, or return a partially constructed (broken) configuration when a parsing error occurs.
   Unifying everything under the exception-throwing model would be cleaner, but we should proceed carefully, as it could introduce backward-compatibility issues for existing users.

2. **Align all methods to return `null` on error.**
   This is more consistent with current behavior, though not ideal: some callers expect a non-null configuration even when `URI` is `null`.

3. **Introduce a new method**, e.g. `getRequiredConfiguration`, which explicitly throws on failure.
   This would let us offer both behaviors without breaking existing expectations.

*P.S.:* A “thumbs up” on a PR announcement doesn’t necessarily mean approval: it just means I’ve seen it and plan to review when I can.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationFactoryTest.java
vy: Fixed in a738f0da5.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: Implemented as suggested in a738f0da5.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: Discarded null or blank URIs in a738f0da5.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: Implemented in a738f0da5.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: > 1. **Align all methods to throw on error.**
>    This approach would make error handling explicit and predictable. Currently, `getConfiguration` can already throw when an invalid properties configuration is encountered. However, most other factories behave differently: they either return `null` when the file is missing or the factory is inactive, or return a partially constructed (broken) configuration when a parsing error occurs.
>    Unifying everything under the exception-throwing model would be cleaner, but we should proceed carefully, as it could introduce backward-compatibility issues for existing users.

Agreed with this approach. Since this is a new API, there are no backward compatibility concerns for this particular method. In this PR we can throw on configuration failures. In a follow-up PR, we can align the rest of the `getConfiguration()` methods to do the same. Combining both tasks in a single PR will not only violate the scope of this PR, but make it difficult to review.

@ppkarwasz, do you agree?

@yybmion, would you be interested in implementing changes necessary to make other `ConfigurationFactory::getConfiguration` methods throw on failure too?

ppkarwasz: I am fine with exceptions in the new one, but we should add an issue to revamp the old one, so we don't forget.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +335,71 @@ public Configuration getConfiguration(
         return getConfiguration(loggerContext, name, configLocation);
     }
 
+    /**
+     * {@return a {@link Configuration} created using provided configuration location {@link URI}s}
+     * If the provided list of {@code URI}s is null or empty, {@code getConfiguration(loggerContext, name, (URI) null)} will be returned.
```
ppkarwasz: Should we explain more explicitly the behavior of an empty list?

```suggestion
     * <p>If the provided list of {@code URI}s is empty, the configuration factory
     * attempts to load an implementation-dependent set of default locations.
     * If no configuration can be found, a {@link ConfigurationException} is thrown.</p>
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -333,6 +335,71 @@ public Configuration getConfiguration(
...
+     *
+     * @param loggerContext a logger context, may be null
+     * @param name a configuration name, may be null
+     * @param configLocations configuration location {@code URI}s, may be null or empty
```
ppkarwasz: *Nit*: I’d prefer not to allow `null` for a new API and instead require callers to pass an empty list.

yybmion:
Hi @vy and @ppkarwasz,

Thanks for the feedback! Before I finalize this PR, I'd like to clarify the scope of follow-up work.

Based on the discussion, I see two distinct improvement areas:

**1. Factory refactoring (removing duplicate code)**
> I realize my comments go a bit beyond the immediate scope of this PR, but the section of code below has become a real maintenance headache. It would be great if we could take the opportunity to clean it up once and for all:
> 
> https://github.com/apache/logging-log4j2/blob/8a3fb534f999632d0688a9a204f11b3dd7f59f72/log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java#L373-L478

Delegate composite configuration logic in `Factory.getConfiguration(ctx, name, URI)` to the new `getConfiguration(ctx, name, List<URI>)` method.

**2. Exception-throwing consistency**
>@yybmion, would you be interested in implementing changes necessary to make other ConfigurationFactory::getConfiguration methods throw on failure too?

Align all `getConfiguration` methods to throw exceptions instead of returning null

___

I'm currently working on two changes. Should I handle them in separate PRs?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: Changed as requested in 6a8987f83.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: Improved as requested in 6a8987f83.

vy (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
vy: @ppkarwasz, created #3958 to align the behavior of existing `ConfigurationFactory::getConfiguration` methods.

vy:
@ppkarwasz, I've addressed all your remarks. Unless you have more, would you mind approving the PR, please?

@yybmion, you've done a great job in this PR, nothing else is needed from your side. :100: When this PR gets merged, please submit another one addressing the newly created #3958.

ppkarwasz (approved):
LGTM

PR #4004, opened 2025-12-16
Merge `rel/2.25.3` to `2.x`

vy:


log4j-1.2-api/src/main/java/org/apache/log4j/helpers/package-info.java
```diff
@@ -18,7 +18,7 @@
  * Log4j 1.x compatibility layer.
  */
 @Export
-@Version("2.26.0")
+@Version("2.25.3")
```
ppkarwasz, line 21: These versions (updated in commit 8d43a991c3068c6ff4c8fc38ee2c73ba5d24299a and e2898a0ec334cd4527b6618b3190edd1d47535f4) are wrong:

- In the release `2.25.2` of `log4j-1.2-api`, the version of this package was `2.20.3`,
- Since version `2.25.3` only made **patch** level changes to the public interface, the version should have been update to `2.20.4`. The API of the package is identical to what we released in `2.20.0`, with some micro changes to annotations.

Anyway, since Log4j `2.25.3` is already out and package versions need to be **monotonic**, there is nothing we can do right now.

ppkarwasz (approved):
LGTM

The **minor** version number of many OSGi packages have been unnecessarily, but there is nothing we can do now.

**Note**: under semantic versioning we are allowed to even bump the **major** version of a package without changing anything in the public API.

ppkarwasz (approved):

log4j-1.2-api/src/main/java/org/apache/log4j/helpers/package-info.java
vy: Fixed in dde4d4a.

PR #4008, opened 2025-12-21
Fix writeHeader computation in RollingRandomAccessFileManager

NeatGuyCoding:
Fix writeHeader computation in RollingRandomAccessFileManager

The writeHeader boolean was computed after creating the RandomAccessFile, which creates the file if it doesn't exist. This made file.exists() always return true, causing incorrect header write decisions.

This change captures the file's pre-existing state before opening the RandomAccessFile by:
- Declaring a fileExistedBefore boolean initialized to false
- Setting it by checking fileName != null && new File(fileName).exists() before calling new RandomAccessFile(...)
- Using !fileExistedBefore instead of !file.exists() in the writeHeader calculation

When fileName is null, fileExistedBefore remains false as expected.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManagerTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManagerTest.java
@@ -339,10 +339,10 @@ class RollingRandomAccessFileManagerTest {
 
     @ParameterizedTest
     @CsvSource({
-            "true,true",
-            "true,false",
-            "false,true",
-            "false,false",
+        "true,true",
+        "true,false",
+        "false,true",
+        "false,false",
     })
     void testWriteHeaderWhetherAppendOrExists(final boolean append, final boolean fileExists) throws Exception {
         final File file = File.createTempFile("log4j2", "test");
@@ -357,9 +357,8 @@ class RollingRandomAccessFileManagerTest {
         file.deleteOnExit();
 
         final String header = "HEADER";
-        final PatternLayout layout = PatternLayout.newBuilder()
-                .setHeader(header)
-                .build();
+        final PatternLayout layout =
+                PatternLayout.newBuilder().setHeader(header).build();
 
         final RollingRandomAccessFileManager manager = RollingRandomAccessFileManager.getRollingRandomAccessFileManager(
                 file.getAbsolutePath(),
@@ -385,11 +384,14 @@ class RollingRandomAccessFileManagerTest {
         final String content = new String(fileContent);
         final boolean expectedHeaderWritten = !append || !fileExists;
         if (expectedHeaderWritten) {
-            assertTrue(content.startsWith(header), "File should start with header when append=" + append
-                    + ", fileExists=" + fileExists + ", content: " + content);
+            assertTrue(
+                    content.startsWith(header),
+                    "File should start with header when append=" + append + ", fileExists=" + fileExists + ", content: "
+                            + content);
         } else {
             // When append=true and fileExists=true, file has existing content, so header should not be written
-            assertTrue(!content.startsWith(header),
+            assertTrue(
+                    !content.startsWith(header),
                     "File should not start with header when append=" + append + ", fileExists=" + fileExists
                             + ", content: " + content);
         }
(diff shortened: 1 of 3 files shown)


vy:
@NeatGuyCoding, the report and the fix look legitimate, thanks. Would you mind providing also a changelog entry and a test, please?

I'd appreciate it if you can keep the test as minimal as possible.

NeatGuyCoding:
@vy done

src/changelog/.2.x.x/fix_RollingRandomAccessFileManager_writeHeader.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <description format="asciidoc">
+    Fix writeHeader computation in RollingRandomAccessFileManager. The writeHeader boolean was computed after creating the RandomAccessFile, which creates the file if it doesn't exist. This made file.exists() always return true, causing incorrect header write decisions when append=true and the file didn't exist before.
```
vy:
```suggestion
    Fix header write in `RollingRandomAccessFileManager` that was being incorrectly skipped if `append=true` and the file didn't exist before.
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
```diff
@@ -174,8 +174,10 @@ public static RollingRandomAccessFileManager getRollingRandomAccessFileManager(
                             long size = 0;
                             long time = System.currentTimeMillis();
                             RandomAccessFile raf = null;
+                            boolean fileExistedBefore = false;
                             if (fileName != null) {
                                 file = new File(name);
+                                fileExistedBefore = new File(name).exists();
```
vy: Why don't we reuse the `file` here?

```suggestion
                                fileExistedBefore = file.exists();
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
```diff
@@ -174,8 +174,10 @@ public static RollingRandomAccessFileManager getRollingRandomAccessFileManager(
                             long size = 0;
                             long time = System.currentTimeMillis();
                             RandomAccessFile raf = null;
+                            boolean fileExistedBefore = false;
```
vy, line 177: Can we replace the `file.exists()` on line 184 with `fileExistedBefore` too?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManagerTest.java
```diff
@@ -333,4 +334,40 @@ void testRolloverRetainsFileAttributes() throws Exception {
                 .permissions();
         assertEquals(filePermissions, actualFilePermissions);
     }
+
+    @Test
+    void testWriteHeaderWhenFileDoesNotExistBefore() throws IOException {
```
vy: I see that header write is not tested at all. Would you mind extending this to adapt the following footprint, please?

    @ParameterizedTest
    @CsvSource({
            "true,true",
            "true,false",
            "false,true",
            "false,false",
    })
    void testWriteHeaderWhenFileDoesNotExistBefore(final boolean append, final boolean fileExists) throws Exception {
        // ...
    }

If you think there are more cases (of header write) to be tested, please share.

vy (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManager.java
NeatGuyCoding: no, it should be re-evaluate to make sure time is not based on the one deleted

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/RollingRandomAccessFileManagerTest.java
NeatGuyCoding: fix

vy (approved):

PR #4011, opened 2025-12-23
feature:support random delayed compression

### LINKED ISSUE #4012: Proactively deferring compression can reduce disk I/O pressure.
At 0 o 'clock every day, a large number of machines begin to compress the log files of the previous day, resulting in the sharp rise of the disk IO of the container, which brings risk to the stability of the cloud disk. Now, the compression random delay x seconds is introduced to reduce the concentration of the compression trigger time and reduce the pressure on the disk.

https://github.com/apache/logging-log4j2/pull/4011

OpenZYK:
Fixes #4012 

At 0 o 'clock every day, a large number of machines begin to compress the log files of the previous day, resulting in the sharp rise of the disk IO of the container, which brings risk to the stability of the cloud disk. Now, the compression random delay x seconds is introduced to reduce the concentration of the compression trigger time and reduce the pressure on the disk.

 ```
         <RollingFile name="RollingFile" 
                     fileName="logs/app.log" 
                     filePattern="logs/app-%d{MM-dd-yyyy}-%i.log.gz">
            <PatternLayout>
                <Pattern>%d %p %c{1.} [%t] %m%n</Pattern>
            </PatternLayout>
            <Policies>
                <TimeBasedTriggeringPolicy interval="1" modulate="true"/>
                <SizeBasedTriggeringPolicy size="10 MB"/>
            </Policies>
            <!-- deferred compression 1s~30s -->
            <DefaultRolloverStrategy max="10" delayedCompressionSeconds="30"/>
        </RollingFile>
```

### CODE REVIEW

**Comment by Mrkell76** [comment]
@

File: log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java [hunk 1]
```diff
@@ -0,0 +1,200 @@
...
+
+import java.io.IOException;
+import java.util.concurrent.ThreadLocalRandom;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Wrapper action that schedules compression for defferred execution.
+ * This action wraps another action and schedules it to be executed
+ * after a random defer within a specified time window.
```
vy:
```suggestion
 * Decorates an {@link Action} to execute it after a certain amount of delay.
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private static final Logger LOGGER = StatusLogger.getLogger();
+
+    private final Action originalAction;
+    private final int maxDeferSeconds;
+    private boolean complete = false;
+    private boolean interrupted = false;
+
+    /**
+     * Creates a new DelayedCompressionAction.
```
vy:
```suggestion
     * Creates a new instance.
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+     *
+     * @param originalAction the action to be executed with defer
```
vy:
```suggestion
     * @param originalAction the action to be executed with delay
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Wrapper action that schedules compression for defferred execution.
+ * This action wraps another action and schedules it to be executed
+ * after a random defer within a specified time window.
+ */
+public class DeferredCompressionAction implements Action {
```
vy: I think your earlier attempt, `DelayedCompressionAction`, fits the bill better. AFAICT, `deferred` waits for a condition, which does not necessarily needs to be a time, whereas `delay` makes it clear the involvement of time.

Once you implement this change, please update all the usages of `[Dd]efer` in this PR.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+        this.originalAction = requireNonNull(originalAction, "originalAction");
+        this.maxDeferSeconds = maxDeferSeconds;
+    }
+
+    /**
+     * Executes the action with a defer using sleep.
+     *
+     * @return true if the action was successfully executed
+     * @throws IOException if an error occurs during execution
+     */
```
vy:
```suggestion
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+    @Override
+    public boolean execute() throws IOException {
+        if (interrupted) {
```
vy: What about `complete == true`?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/DeferredCompressionAction.java
```diff
@@ -0,0 +1,200 @@
...
+    private boolean complete = false;
+    private boolean interrupted = false;
+
+    /**
+     * Creates a new DelayedCompressionAction.
+     *
+     * @param originalAction the action to be executed with defer
+     * @param maxDeferSeconds the maximum defer in seconds (0 means immediate execution)
+     */
+    public DeferredCompressionAction(Action originalAction, int maxDeferSeconds) {
```
vy: _Question:_ Shouldn't we be receiving a range instead of an upper bound?

vy (changes requested):
@OpenZYK, I think the suggested feature sounds legitimate. I've skimmed through the changes, and dropped some remarks. Would you also mind providing a test and a changelog entry, please?

vy:
@OpenZYK, [TimeBasedTriggeringPolicy](https://logging.apache.org/log4j/2.x/manual/appenders/rolling-file.html#TimeBasedTriggeringPolicy) receives a `maxRandomDelay` configuration attribute — doesn't that precisely address the problem you're describing?

OpenZYK:
[TimeBasedTriggeringPolicy](https://logging.apache.org/log4j/2.x/manual/appenders/rolling-file.html#TimeBasedTriggeringPolicy) receives a maxRandomDelay configuration attribute —  This is a parameter for scrolling files, which may cause the logs of the first day to be written to the files of the next day. Splitting files itself is not very IO expensive, but what I'm addressing here is the impact of compression on disk IO. I will add some tests based on your suggestion

vy:
Can't we change the default behavior of the `TimeBasedTriggeringPolicy.maxRandomDelay` configuration attribute such that it will not affect the cut off time? That is, the roll over will happen precisely at the right time, but only the compression will honor the delay? @ppkarwasz, WDYT?

ppkarwasz:
@vy,

I think delaying **when a rollover is triggered** and delaying **when the asynchronous rollover actions execute** should have separate configuration knobs.

The execution of asynchronous actions lives in `RollingFileManager`, so tying that behavior to an attribute on a specific `TriggeringPolicy` (or rollover strategy) feels like a layering violation.

For this PR specifically, I believe there’s a simpler and cleaner approach: instead of wrapping actions and sleeping, we could replace the `ExecutorService` in `RollingFileManager` with a `ScheduledExecutorService` and apply the delay at scheduling time.

OpenZYK:
@ppkarwasz Your suggestion is great! I will follow this idea and add some test cases

ppkarwasz:
@OpenZYK,

To be clear what I suggested was to start the `AsyncAction` from `RollingFileManager` with a delay, not to start delayed actions from within `DelayedCompressAction`: the rollover process has many inter-dependent actions and it is much simpler to execute them sequentially. If you execute them in parallel, you will encounter many subtle bugs.

vy:
@OpenZYK, we were expecting you to only touch `RollingFileManager`, and there schedule the action with an arbitrary delay. But we still see `DelayedCompressionAction`. Are we still on the same page?

OpenZYK:
Hi @vy @ppkarwasz, removed `DelayedCompressionAction` entirely, replaced executor with `ScheduledThreadPoolExecutor`, split delay into `minDelaySeconds`/`maxDelaySeconds`, added 10 test cases and changelog entry. Would you mind taking another look?

OpenZYK:
Hi @vy @ppkarwasz, we noticed that #4071 has been merged into 2.26.0, which adds compression delay support at the `Action` level (`GzCompressAction`/`ZipCompressAction`).

Our approach in this PR is different: we apply the delay at the `RollingFileManager` scheduling layer using `ScheduledThreadPoolExecutor`, which aligns with @ppkarwasz's earlier suggestion to avoid executing actions in parallel and prevent subtle bugs.

Could you let us know whether this PR is still worth pursuing, or if the feature is considered covered by #4071? Happy to close it if it's no longer needed.

vy:
> Hi @vy @ppkarwasz, we noticed that #4071 has been merged into 2.26.0,

No, **fortunately** it doesn't! :sweat_smile: It was merged to `2.x` and planned for `2.27.0`.

> which adds compression delay support at the `Action` level (`GzCompressAction`/`ZipCompressAction`).
> 
> Our approach in this PR is different: we apply the delay at the `RollingFileManager` scheduling layer using `ScheduledThreadPoolExecutor`, which aligns with @ppkarwasz's earlier suggestion to avoid executing actions in parallel and prevent subtle bugs.
> 
> Could you let us know whether this PR is still worth pursuing, or if the feature is considered covered by #4071? Happy to close it if it's no longer needed.

@OpenZYK, it is super awesome that you still follow this feature request. :star_struck: I will take care of submitting a new PR steering the incorrectly implemented #4071 in the direction @ppkarwasz suggested, and ask for your review.

ramanathan1504:
fixes are already merged in `2.x` with 4165

PR #4014, opened 2025-12-25
Fix warning when last argument is null (#3975)

### LINKED ISSUE #3975: Logging with last implicit param a null Throwable generates framework warning
## Description

[Issue #2380](https://github.com/apache/logging-log4j2/issues/2380) introduces generation of framework warning when last (implicit) parameter is a null `Throwable`.

## Configuration

**Version:** 2.25.1

**Operating system:** Windows 11

**JDK:** Java HotSpot(TM) 64-Bit Server VM (build 25.461-b11, mixed mode)

## Logs

```
2025-10-30T13:17:14,249 ERROR [main] MyTest {} - Execution completed with result invalid
java.lang.RuntimeException: failed
	at ...

2025-10-30T11:17:14.252438500Z main WARN found 1 argument placeholders, but provided 2 for pattern `Execution completed with result {}`
2025-10-30T13:17:14,251 ERROR [main] MyTest  {} - Execution completed with result valid

```

## Reproduction
Some libraries provide execution results in an 'Either' style of structure (e.g. `dev.failsafe.event.ExecutionCompletedEvent`) --result or an exception. Such cases could conveniently be logged like
```
logger.error("Execution completed with result {}", outcome.result, outcome.failed);
```

That is, if we have failure `Throwable` -- log its stack trace. But if it is `null` and the error is described in `.result` -- just log what we have. This used to work properly until a fix for #2380 introduces a warning (WARN found 1 argument placeholders, but provided 2 for pattern), see sample test below and its output above.

```java
  record Result(
          String result,
          Throwable failed
  ) {}

  @Test
  void testLog4j() {
    var logger = org.apache.logging.log4j.LogManager.getLogger(MyTest.class);
    var failedResult = new Result("invalid", new RuntimeException("failed"));
    logger.error("Execution completed with result {}", failedResult.result, failedResult.failed);

    var validResult = new Result("valid", null);
    logger.error("Execution completed with result {}", validResult.result, validResult.failed);
  }
```

lrozenblyum:
Allow extra null argument to ParameterFormatter without issuing a warning.

Fixes #3975

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
@@ -247,7 +247,7 @@ final class ParameterFormatter {
         // #2380: check if the count of placeholder is not equal to the count of arguments
         if (analysis.placeholderCount != argCount) {
             final int noThrowableArgCount =
-                    argCount < 1 ? 0 : argCount - ((args[argCount - 1] instanceof Throwable) ? 1 : 0);
+                    argCount < 1 ? 0 : argCount - (isLastArgumentThrowable(args, argCount) ? 1 : 0);
             if (analysis.placeholderCount != noThrowableArgCount) {
                 STATUS_LOGGER.warn(
                         "found {} argument placeholders, but provided {} for pattern `{}`",
@@ -266,6 +266,12 @@ final class ParameterFormatter {
         else {
             formatMessageContainingNoEscapes(buffer, pattern, args, argCount, analysis);
         }
+    }
+
+    private static boolean isLastArgumentThrowable(final Object[] args, final int argCount) {
+        final Object lastArgument = args[argCount - 1];
+        // #3975: tolerate null in the last argument since it could have been Throwable parameter
+        return (lastArgument == null) || (lastArgument instanceof Throwable);
     }
 
     private static void formatMessageContainingNoEscapes(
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
```diff
@@ -268,6 +267,12 @@ static void formatMessage(
         }
     }
 
+    private static boolean ignoreLastArgument(final Object[] args, final int argCount) {
```
ppkarwasz: _Nit_: naming is hard, but I think `isLastArgumentThrowable` would better reflect its purpose.

Alternatively the following logic could be refactored into a `getArgCountWithoutException` method:

```java
argCount < 1 ? 0 : argCount - (ignoreLastArgument(args, argCount) ? 1 : 0)
```

ppkarwasz (changes requested):

log4j-api/src/main/java/org/apache/logging/log4j/message/ParameterFormatter.java
lrozenblyum: I've applied the first suggestion (`isLastArgumentThrowable`).

Extracting a method to get arg count - haven't done that because its naming itself is a big question :) Even the name `getArgCountWithoutLastThrowable` would be ambiguous. Technically it should sound `getArgCountIgnoringLastArgumentIfItIsThrowable` - yet it's too clumsy...

ppkarwasz (approved):
LGTM, thanks!

lrozenblyum:
Thank you for fast review @ppkarwasz!

Issue #4019, opened 2026-01-02
Throwable pattern conversion loses Exception details


aber4mx:
## Description

Log4j 2's Pattern Layout allows to log Throwables. Several configuration elements allow to configure how the exception stack trace should be logged (max number of lines, package filtering..).
However the default behavior without advanced configuration should be to print the stacktrace "as is". It is mentionned in the documentation:
`If this mode is employed without any configuration, the output will be identical to the one obtained from Throwable#printStackTrace().`
and it was working as of 2.24.3, but the behavior changed in 2.25.3.

On a standard Exception, the logged stacktrace is the same. However Exceptions can override the `Throwable#toString()` method which is used by `Throwable#printStackTrace()` or they could also override `Throwable#printStackTrace()` directly.
Since 2.25.x, log4j emulates `Throwable#printStackTrace()`, even if there is no advanced paramater that justifies it. By upgrading from 2.24.3 the exceptions logged in our application lose some information and meaning.

Could we stick to use `Throwable#toString()` in the default case ? And in the advanced case where that requires cusomization on the StackTraceElements, rather use `Throwable#toString()` for the message part instead of a hard-coded `"throwable.getClass().getName()" + ": " + throwable.getLocalizedMessage()` ?

I attach a patch with a unit test that is green at 2.24.3 but is red at 2.25.3

## Configuration

**Version:** 2.25.3

**Operating system:** windows 11 / RedHat 9

**JDK:** OpenJDK 17.0.2

## Logs

See reproduction scenario

## Reproduction

The below patch adds a unit test red on 2.25.3, but green on 2.24.3.
The unit test logs a custom exception whose class overrides the toString method

[exception_rendering_test.patch](https://github.com/user-attachments/files/24410476/exception_rendering_test.patch)

Kimbumsoo99:
Hi @aber4mx. 

Thanks for the test case, It clearly demonstrates the issue.
I'd like to take a look at this and provide a fix. May I be assigned to this?

vy:
@Kimbumsoo99, please feel free to submit a PR — you don't need to be assigned or whatsoever.

vy:
Some random thoughts on the matter.

1. `ThrowableStackTraceRenderer` should *ideally* produce an identical output to `Throwable::printStackTrace` (and render `toString` just like the latter)
2. There are users in the wild trying to parse and/or locate a stack trace in a log file generated using Pattern Layout, which is explicitly documented to be *not* used as a structured layout
[...]

Issue #3623, opened 2025-04-17
Uniformise `%ex` pattern and `exception` JTL resolver


ppkarwasz:
## Description

Currently there is a difference in the way exceptions are formatted by the [`%ex` pattern converter](https://logging.apache.org/log4j/2.x/manual/pattern-layout.html#converter-exception) and the following [JSON Template Layout snippet](https://logging.apache.org/log4j/2.x/manual/json-template-layout.html#event-template-resolver-exception):

```json
{
  "$resolver": "exception",
  "field": "stackTrace",
  "stackTrace": {
    "stringified": true
  }
}
```

- JTL simply uses `Throwable.printStackTrace()` and the first line of the output is obtained by calling `Object.toString()`.
- Pattern Layout uses `getClass().getName() + ": " + getLocalisedMessage()`.

Those lines agree, unless an exception overrides `Object#toString()`.

PR #4033, opened 2026-01-23
Take `Throwable#toString()` into account while rendering stack traces in Pattern Layout

JongminChung:
Align renderThrowableMessage(...) to use Throwable.toString(), restoring the default printStackTrace() behavior documented for the %ex converter.

- This resolves both #4019  and #3623.
- I added test case that issue reporter gives. ([TestCase](https://github.com/apache/logging-log4j2/issues/4019#:~:text=exception_rendering_test.patch))

---

## Checklist

* Tests are provided

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
@@ -79,6 +79,23 @@ public class ThrowablePatternConverterTest {
         }
     }
 
+    /**
+     * Test exception whose {@link #toString()} is intentionally overridden to return a fixed value.
+     */
+    private static final class ToStringOverridingException extends RuntimeException {
+
+        private static final ToStringOverridingException INSTANCE = new ToStringOverridingException();
+
+        private ToStringOverridingException() {
+            super(EXCEPTION);
+        }
+
+        @Override
+        public String toString() {
+            return "foo";
+        }
+    }
+
     static Stream<SeparatorTestCase> separatorTestCases() {
         final String level = LEVEL.toString();
         return Stream.of(
@@ -214,6 +231,14 @@ public class ThrowablePatternConverterTest {
             assertThat(actualStackTrace).as("pattern=`%s`", effectivePattern).isEqualTo(expectedStackTrace);
         }
 
+        @Test
+        void full_output_should_use_custom_toString() {
+            final Throwable exception = ToStringOverridingException.INSTANCE;
+            final String expectedStackTrace = renderStackTraceUsingJava(exception);
+            final String actualStackTrace = convert(patternPrefix, exception);
+            assertThat(actualStackTrace).isEqualTo(expectedStackTrace);
+        }
+
         // This test does not provide `separator` and `suffix` options, since the reference output will be obtained from
         // `Throwable#printStackTrace()`, which doesn't take these into account.
         @ParameterizedTest
@@ -252,10 +277,14 @@ public class ThrowablePatternConverterTest {
         }
 
         private String renderStackTraceUsingJava() {
+            return renderStackTraceUsingJava(EXCEPTION);
+        }
+
+        private String renderStackTraceUsingJava(final Throwable throwable) {
             final Charset charset = StandardCharsets.UTF_8;
             try (final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
                     final PrintStream printStream = new PrintStream(outputStream, false, charset.name())) {
-                EXCEPTION.printStackTrace(printStream);
+                throwable.printStackTrace(printStream);
                 printStream.flush();
                 return new String(outputStream.toByteArray(), charset);
             } catch (final Exception error) {
@@ -545,9 +574,13 @@ public class ThrowablePatternConverterTest {
     }
 
     static String convert(final String pattern) {
+        return convert(pattern, EXCEPTION);
+    }
+
+    static String convert(final String pattern, final Throwable throwable) {
         final List<PatternFormatter> patternFormatters = PATTERN_PARSER.parse(pattern, false, true, true);
         final LogEvent logEvent =
-                Log4jLogEvent.newBuilder().setThrown(EXCEPTION).setLevel(LEVEL).build();
+                Log4jLogEvent.newBuilder().setThrown(throwable).setLevel(LEVEL).build();
         final StringBuilder buffer = new StringBuilder();
         for (final PatternFormatter patternFormatter : patternFormatters) {
             patternFormatter.format(logEvent, buffer);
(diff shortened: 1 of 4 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/ThrowableStackTraceRenderer.java
```diff
@@ -138,12 +138,7 @@ private void renderCause(
     }
 
     static void renderThrowableMessage(final StringBuilder buffer, final Throwable throwable) {
-        final String message = throwable.getLocalizedMessage();
-        buffer.append(throwable.getClass().getName());
-        if (message != null) {
-            buffer.append(": ");
-            buffer.append(message);
-        }
+        buffer.append(throwable);
```
JongminChung, line 141: Link [Throwable.java#L697](https://github.com/openjdk/jdk/blob/ca37dba4d40bf3f71c5489829c893346faec1c56/src/java.base/share/classes/java/lang/Throwable.java#L697)

JongminChung:
Hi, @vy 

Would love to get your eyes on this when you get a moment.
Let me know if any changes are needed. Thanks!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
```diff
@@ -545,9 +572,13 @@ private static List<Exception> createExceptionsOfDifferentDepths() {
     }
 
     static String convert(final String pattern) {
+        return convert(pattern, EXCEPTION);
+    }
+
+    static String convert(final String pattern, final Throwable throwable) {
```
vy: This can be `private`.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
```diff
@@ -79,6 +79,21 @@ public String toString() {
         }
     }
 
+    private static final class CustomException extends Exception {
```
vy: Could you implement the following changes, please?

1. Rename this to `ToStringOverridingException`
2. Use `super(EXCEPTION)` in ctor, and receive no arguments
3. Create `ToStringOverridingException.INSTANCE` field and use it (similar to `TestFriendlyException.INSTANCE`)
4. Return a constant in `toString()`, e.g., `return "foo"`

vy (changes requested):
Could you add a changelog entry and make sure `./mvnw verify -pl :log4j-core,:log4j-core-test` passes, please?

vy:
Could you replace the entire ~AI slop~ description with a short meaningful summary, please?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
JongminChung: This can't be private since it's accessed from other classes within the same package.

([Link](https://github.com/apache/logging-log4j2/blob/2.x/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/RootThrowablePatternConverterTest.java#L22))

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
JongminChung: Thanks for the review, @vy! I've applied all your suggestions:

1. Renamed to ToStringOverridingException
2. Changed constructor to use super(EXCEPTION) with no arguments
3. Added ToStringOverridingException.INSTANCE static field (following TestFriendlyException.INSTANCE pattern)
4. toString() now returns a constant ("foo")

---

image

JongminChung:
> Could you replace the entire ~AI slop~ description with a short meaningful summary, please?

As this is my first PR, I wanted to provide more background context. I've shortened it now. 
Thanks for the thorough review!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
vy: It should be, since the new method is only used by this class:

```
$ cat p/A.java 
package p;
public class A {
    public static void main(String[] args) { System.out.println(f(1,2)); }
    static int f(int i) { return i; }
    private static int f(int i, int j) { return i + j; }
}

$ cat p/B.java 
package p;
import static p.A.f;
public class B {
    public static void main(String[] args) { System.out.println(f(1)); }
}

$ javac p/*.java
$ java p.A
3
$ java p.B
1
```

Am I missing something?

vy (changes requested):
@JongminChung, thanks for the update.

Please, don't ever force-push. It completely nukes all the conversation history.

Would you also mind adding a changelog entry file, please?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/ThrowablePatternConverterTest.java
JongminChung: I am sorry. I made a mistake. I will correct it to private.

JongminChung:
> @JongminChung, thanks for the update.
> 
> Please, don't ever force-push. It completely nukes all the conversation history.
> 
> Would you also mind adding a changelog entry file, please?

Sorry, I deleted them thinking they were unnecessary commitments through force-push, but I overlooked that the context of the previous conversation disappeared.

Regarding the changelog you requested, I will read the document and write it.

Thank you for reviewing it in detail every time.

src/changelog/.2.x.x/4033_fix_custom_throwable_to_sting.xml
```diff
@@ -0,0 +1,13 @@
...
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="changed">
+    <issue id="3623" link="https://github.com/apache/logging-log4j2/pull/3623"/>
+    <issue id="4033" link="https://github.com/apache/logging-log4j2/pull/4033"/>
+    <description format="asciidoc">
+        Take Throwable#toString() into account while rendering stack traces in Pattern Layout.
```
vy:
```suggestion
    <issue id="3623" link="https://github.com/apache/logging-log4j2/issues/3623"/>
    <issue id="4033" link="https://github.com/apache/logging-log4j2/pull/4033"/>
    <description format="asciidoc">
        Take `Throwable#toString()` into account while rendering stack traces in Pattern Layout.
```

vy (approved):

aber4mx:
Thank you very much @JongminChung  @vy !

PR #4041, opened 2026-01-29
Avoid referring to `PluginBuilderAttribute.class` in `PluginProcessor`

cushon:
This makes it easier to build PluginProcessor by itself separately from the rest of log4j-core, and then include that processor when building the rest of core, instead of having to compile all of core twice.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessor.java
@@ -53,7 +53,6 @@ import javax.tools.FileObject;
 import javax.tools.StandardLocation;
 import org.apache.logging.log4j.core.config.plugins.Plugin;
 import org.apache.logging.log4j.core.config.plugins.PluginAliases;
-import org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute;
 import org.apache.logging.log4j.util.Strings;
 
 /**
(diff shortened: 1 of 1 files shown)


vy:
@cushon, I'm sorry for the late reply. All of my Log4j time was sucked by AI slop — Piotr mentioned about it in #4052.

> This makes it easier to build PluginProcessor by itself separately from the rest of log4j-core, and then include that processor when building the rest of core, instead of having to compile all of core twice.

If so, we should be able to simplify the `maven-compiler-plugin` configuration in the `plugin-processing` profile in `/log4j-parent/pom.xml` too, right? Can we carry out this change too, please?

cushon:
@vy thanks for taking a look!

> If so, we should be able to simplify the `maven-compiler-plugin` configuration in the `plugin-processing` profile in `/log4j-parent/pom.xml` too, right? Can we carry out this change too, please?

I'm open to trying that, I'm not sure exactly how it would work yet. I noticed this issue trying to build log4j2 with Bazel, not using the maven build, I am less familiar with maven.

This change makes it possible to build the plugins by building just the following sources:

```
org/apache/logging/log4j/core/config/plugins/processor/**/*.java
org/apache/logging/log4j/core/config/plugins/Plugin.java
org/apache/logging/log4j/core/config/plugins/PluginAliases.java
org/apache/logging/log4j/core/util/JsonUtils.java
```

So it would still be necessary to compile those parts of log4j-core twice, but it might be possible to compile just the above sources in the first phase.

Another option I was wondering about is whether it might be possible to move these processors to a separate maven submodule, to compile them separately before `log4j-core`, and avoid the need for the `maven-compiler-plugin` configuration that compiles in two phases. Do you think that's worth considering?

vy (approved):

vy:
> I noticed this issue trying to build log4j2 with Bazel

Would you mind elaborating on this a bit, please? Do you have an in-house Log4j build using Bazel?

> Another option I was wondering about is whether it might be possible to move these processors to a separate maven submodule, to compile them separately before `log4j-core`, and avoid the need for the `maven-compiler-plugin` configuration that compiles in two phases. Do you think that's worth considering?

That'd be a disruptive change. `PluginProcessor` users will need to adapt their class paths. Log4j 3 will have that modularity, but its release date is still uncertain.

cushon:
> > I noticed this issue trying to build log4j2 with Bazel
> 
> Would you mind elaborating on this a bit, please? Do you have an in-house Log4j build using Bazel?

Yes, we have in-house Bazel builds for some dependencies as an alternative to maven/gradle builds or published prebult jars.

PR #4060, opened 2026-03-06
Add debug logs for successful resource loading in `Loader`

### LINKED ISSUE #4058: loader should log where it found config files
In several places in Loader, such as here:
https://github.com/apache/logging-log4j2/blob/2.x/log4j-core/src/main/java/org/apache/logging/log4j/core/util/Loader.java#L94
There’s a message at trace level saying basically “Trying to find [FILENAME] using THINGY”.
This is very useful, because it lets you know all the places that log4j is looking for config files.
However, there’s no good way to find which of these attempts succeeded.
Earlier today I had to put a debug breakpoint inside this class to find out which log4j2.xml was getting loaded.
It would be nice if successful attempts were logged, perhaps even at a level higher than trace.  (maybe debug??)

ban-xiu:
Close https://github.com/apache/logging-log4j2/issues/4058

### DIFF
--- a/src/changelog/.2.x.x/add_debug_logs_for_successful_resource_loading.xml
+++ b/src/changelog/.2.x.x/add_debug_logs_for_successful_resource_loading.xml
@@ -6,6 +6,7 @@
            https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
        type="added">
     <issue id="4058" link="https://github.com/apache/logging-log4j2/issues/4058"/>
+    <issue id="4060" link="https://github.com/apache/logging-log4j2/pull/4060"/>
     <description format="asciidoc">
         Added debug level logs for successful resource loading in `Loader` class to show where resources are found
     </description>
--- a/src/changelog/.2.x.x/add_debug_logs_for_successful_resource_loading.xml
+++ b/src/changelog/.2.x.x/add_debug_logs_for_successful_resource_loading.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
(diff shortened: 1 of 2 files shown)


markyag-spgdji:
We could rewrite getResourceAsStream in terms of getResource, and then get the stream via url.openStream().  Note that this is what java.lang.ClassLoader itself does.  The benefit is that we log locations for BOTH methods.

vy:
> We could rewrite getResourceAsStream in terms of getResource, and then get the stream via url.openStream(). Note that this is what java.lang.ClassLoader itself does. The benefit is that we log locations for BOTH methods.

@markyag-spgdji, correct, but this will not work for custom `ClassLoader` implementations which implement `getResourceAsStream` differently. Am I mistaken?

vy (changes requested):
@ban-xiu, would you mind adding a changelog entry file too, please?

ban-xiu:
> @ban-xiu, would you mind adding a changelog entry file too, please?

done.

src/changelog/.2.x.x/add_debug_logs_for_successful_resource_loading.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="added">
+    <issue id="4058" link="https://github.com/apache/logging-log4j2/issues/4058"/>
```
vy, line 8:
```suggestion
    <issue id="4058" link="https://github.com/apache/logging-log4j2/issues/4058"/>
    <issue id="4060" link="https://github.com/apache/logging-log4j2/pull/4060"/>
```

vy (approved):

markyag-spgdji:
thanks, both of you!

PR #4065, opened 2026-03-16
Remove redundant warning from `feature_request.md`

### LINKED ISSUE #4057: new issue template is out of date and misleading
I was about to create file a new issue.  I chose the "feature request" template, and it was pre-filled with the following:
> **Warning!**
> It is highly recommended to discuss feature requests in [the mailing lists](https://logging.apache.org/log4j/2.x/support.html) first.
>
> [A clear and concise description of the feature requested.]

There are at least two problems here.
### The URL is old and leads to a Not Found page.
The analogous URL these days is perhaps https://logging.apache.org/support.html
Or if the goal is to direct the user to log4j-specific lists, then perhaps this page is better: https://logging.apache.org/log4j/2.12.x/mail-lists.html
But that page embeds a version in the URL which is awkward: if the template is not constantly updated then if the mailing list info ever changes then once again the template will be directing users to an old page.
### There is no guidance as to WHICH mailing list to use for discussion.
There are at least 3 possibilities.
1. Log4j user list: I actually tried suggesting my feature there almost a week ago and got no replies.
The list seems to be very low traffic.  There are basically two kinds of posts.  One is new release announcements.
The other is a random user asking a question and getting no replies, neither from devs nor fellow users.
The last time a dev replied to a question on this list was around June, I think.
2. Log4j dev list: this list seems more or less dead since 2017.  I did not even try this one.
3. The "general" dev list.  This one seems active.
4. I thought that by symmetry there should be a "general" user list, but can't find one.

I don't think it's reasonable to expect people to
1. pick a list at random
2. send an email there
3. wait a few days with no response
4. pick a different list at random
5. repeat until no more lists

vy:
Fixes #4057.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -3,7 +3,4 @@ name: Feature request
 about: Submit a feature request
 ---
 
-**Warning!**
-It is highly recommended to discuss feature requests in [the mailing lists](https://logging.apache.org/log4j/2.x/support.html) first.
-
 [A clear and concise description of the feature requested.]


FreeAndNil (approved):
LGTM

PR #4070, opened 2026-03-19
Deprecate externalContext and migrate to Map-backed getObject/putObject

### LINKED ISSUE #2351: Missing servlet context in web lookup when using composite configuration
## Description

Web lookup doest not work when configuring a web application using a list of configuration files.

While using the log4j-web module and setting log4jConfiguration as a list of files the web initializer calls Configurator.initialize passing a Map.Entry<String, Object> as the last argument.

https://github.com/apache/logging-log4j2/blob/32075af9323dce1bbb8f66922aa2cccbb46cd4fd/log4j-web/src/main/java/org/apache/logging/log4j/web/Log4jWebInitializerImpl.java#L171-L183

The configurator does not implement a method which accepts a list of locations and an entry (as it does for the case of a single location) and instead matches this method:
https://github.com/apache/logging-log4j2/blob/32075af9323dce1bbb8f66922aa2cccbb46cd4fd/log4j-core/src/main/java/org/apache/logging/log4j/core/config/Configurator.java#L184-L188

Which in turn calls this which does not correctly set the context.
https://github.com/apache/logging-log4j2/blob/32075af9323dce1bbb8f66922aa2cccbb46cd4fd/log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jContextFactory.java#L290-L301

## Configuration

**Version:** 2.3.0

**Operating system:** windows 11

**JDK:** JDK 17

## Reproduction

- Add log4j-web module to web applciation.
- Set the init param log4jConfiguration as a list of properties (`<Parameter name="log4jConfiguration" value="log4j2-foo.properties,log4j2-bar.properties"/>`)
- Lookup the servletContextName or any other ServletContxt variable in one of the properties files (${web:servletContextName})

ramanathan1504:
Separate web `ServletContext` storage from the legacy `LoggerContext` external context, avoiding collisions with integrations such as Spring Boot. Also deprecate the legacy external-context APIs in favor of `getObject`/`putObject`, while retaining backwards compatibility.

Regression tests cover `${web:*}` lookups with comma-separated composite configurations for both `log4j-web` and `log4j-jakarta-web`.

Fixes #2351.


log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jContextFactory.java
```diff
@@ -300,10 +300,16 @@ public LoggerContext getContext(
             final boolean currentContext,
             final List<URI> configLocations,
             final String name) {
-        final LoggerContext ctx =
-                selector.getContext(fqcn, loader, currentContext, null /*this probably needs to change*/);
-        if (externalContext != null && ctx.getExternalContext() == null) {
-            ctx.setExternalContext(externalContext);
+        final LoggerContext ctx;
+        if (externalContext instanceof Map.Entry) {
```
vy, line 304: @ramanathan1504, would you mind pointing me the code calling this method with an `externalContext` of type `Map.Entry`, please?

ramanathan1504: @vy 

The externalContext is passed as a Map.Entry<String, Object> from the log4j-web module.
In Log4jWebInitializerImpl.java (lines 177 and 183), it calls:
WebLoggerContextUtils.createExternalEntry(this.servletContext)

This returns the Map.Entry which is then passed through Configurator.initialize(...) down to the Log4jContextFactory.

I used instanceof Map.Entry in the Factory to ensure that when composite configurations are used, the externalContext is correctly unwrapped and set so that WebLookup can find the ServletContext. However, I'm open to moving this logic to Configurator or elsewhere if you feel the Factory should remain agnostic of this wrapper.

vy: @ppkarwasz, this entire `Map.Entry` family existing in `LC` method footprints appear like an ugly hack to me, yet it is public. That is, we cannot change it. @ramanathan1504's proposal in this PR looks legitimate to me. I'd appreciate a 2nd pair of eyes on this change. Would you mind quickly skimming through it and see if you can spot anything suspicious, please?

log4j-jakarta-web/src/test/java/org/apache/logging/log4j/web/WebLookupTest.java
```diff
@@ -94,5 +112,63 @@ public class WebLookupTest {
...
+            assertNotNull(
+                    WebLoggerContextUtils.getServletContext(),
+                    "ServletContext is null in composite configuration - "
+                            + "${web:*} lookups will not resolve (issue #2351)");
+
+            final Configuration config = ctx.getConfiguration();
+            assertNotNull(config, "No Configuration");
+
+            final StrSubstitutor substitutor = config.getStrSubstitutor();
+            assertNotNull(substitutor, "No StrSubstitutor");
```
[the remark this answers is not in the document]
ramanathan1504:
```java 
assertInstanceOf(CompositeConfiguration.class,config,"Expected CompositeConfiguration for comma-separated log4jConfiguration");
```
Updated

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jContextFactory.java
rgoers: We should stop using the externalContext field for anything internal and use the externalContextMap instead. Spring used to store its Environment in the externalContext field which would presumably collide with this. That was subsequently changed to use the external Context Map but other things could easily use this field.

ramanathan1504: @rgoers 

Agreed, but there's a catch:
The externalContext parameter is in the public API. If we stop auto-setting it, we break existing callers who expect ctx.getExternalContext() to work.
log4j-jakarta-web already uses the right pattern (Map.Entry with distinct keys), but other integrations might not.
How do you want to handle backward compatibility?
•Breaking change + migration docs?
•Deprecation path?
•Document-only fix?

Happy to implement whichever you prefer.

rgoers: We cannot break backward compatibility in 2.x.  I would suggest removing the externalContext field and mapping the set/get methods to a specific key in the map. The methods should be marked as deprecated and removed in 3.x but the key needs to be documented. Just my opinion though.

ramanathan1504: @rgoers, following discussions with @vy and @ppkarwasz, we have aligned on a Map-backed refactor to resolve the collision issue you raised. I have updated the PR title and description accordingly. This approach maintains backward compatibility for Spring Boot while providing a clean fix for the composite configuration issue. Resolving this conversation.

vy:
@ramanathan1504, I don't see any deprecations or whatsoever. What is the status of this PR?

ramanathan1504:
> @ramanathan1504, I don't see any deprecations or whatsoever. What is the status of this PR?

@vy, an update on my progress: I have completed the refactor to move to Map-based storage (putObject/getObject) and have applied the requested @Deprecated annotations across log4j-api, log4j-core, and log4j-web.

The full build mvn install is now successful. I am currently debugging a few remaining test failures in mvn verify to ensure all lookups and backward compatibility fallbacks are 100% correct. I will push the final fix shortly to resolve #2351 as discussed.

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jContextFactory.java
```diff
@@ -148,8 +148,11 @@ private void initializeShutdownCallbackRegistry() {
      * for the caller if a more appropriate Context can be determined.
      * @param externalContext An external context (such as a ServletContext) to be associated with the LoggerContext.
      * @return The LoggerContext.
+     * @deprecated Use {@link org.apache.logging.log4j.spi.LoggerContext#putObject(String, Object)} instead.
```
vy: This should point to `getObject`, not `putObject`.

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
```diff
@@ -163,11 +166,14 @@ public LoggerContext(final String name, final Object externalContext) {
      * @param name The context name.
      * @param externalContext The external context.
      * @param configLocn The location of the configuration as a URI.
+     * @deprecated Use {@link #LoggerContext(String)} and {@link #putObject(String, Object)} instead.
```
vy:
```suggestion
     * @deprecated Use {@link #LoggerContext(String)} to create a new context, and then use {@link #putObject(String, Object)} to attach the {@code externalContext}.
```

vy (changes requested):
@ramanathan1504, could you throw this to your LLM-of-choice and tell it to check

- No (binary and/or source) incompatibilities are introduced
- Check if Spring is also using these touched parts, and if so, these changes don't break Spring's integration

log4j-core/src/main/java/org/apache/logging/log4j/core/LoggerContext.java
ramanathan1504: @vy fix [75d9e2f](https://github.com/apache/logging-log4j2/pull/4070/commits/75d9e2f3cb8e27d2fac04f36a51b78671d670970)

log4j-core/src/main/java/org/apache/logging/log4j/core/impl/Log4jContextFactory.java
ramanathan1504: @vy updated [ac7173b](https://github.com/apache/logging-log4j2/pull/4070/commits/ac7173b3f5c9f2eb2cc51c0792e4dff7d34f387f)

ramanathan1504:
@vy, can I mirror the `externalContext` fix to `log4j-jakarta-web` (main + regression test), deprecate `getExternalContext` implements in `api/api-test/to-slf4j/to-jul/tagli` in this `PR` or create a separate `PR`

ramanathan1504:
> @ramanathan1504, could you throw this to your LLM-of-choice and tell it to check
> 
>     * No (binary and/or source) incompatibilities are introduced
> 
>     * Check if Spring is also using these touched parts, and if so, these changes don't break Spring's integration
@vy 
The diff only adds`@Deprecated` no signatures changed, so binary and source compat hold. Spring Boot's `Log4J2LoggingSystem` does use `get/setExternalContext  `

vy:
> can I mirror the `externalContext` fix to `log4j-jakarta-web` (main + regression test), deprecate `getExternalContext` implements in `api/api-test/to-slf4j/to-jul/tagli` in this `PR` or create a separate `PR`

@ramanathan1504, good call. Implemented and pushed it to this PR.

vy (approved):

PR #4083, opened 2026-03-26
Add initial `AGENTS.md` for AI-assisted vulnerability research

ppkarwasz:
This introduces an `AGENTS.md` file, which AI coding agents (such as Claude Code, Cursor, and GitHub Copilot) read automatically when entering a repository. It lets the project guide agent behavior without relying on researchers to configure their tools manually.

This first version targets vulnerability research workflows. Before an agent drafts or files a report, it is instructed to read:

- **Threat model**: to establish what is and isn't in scope
- **VDR**: to catch duplicates before they are reported
- **Security FAQ**: to filter out known, intentional behaviors that are not vulnerabilities

All three documents are referenced by their raw source URLs rather than rendered pages. This reduces token consumption and avoids context window pressure when agents are processing large codebases alongside these files.

The expected outcome is fewer duplicate reports, fewer false positives, and less triage burden.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,58 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+# Vulnerability research agent
+
+You are helping a security researcher find and report vulnerabilities in this project.
+Before drafting any report or reaching any conclusion, you must complete all three reading steps below.
+This is mandatory: skipping steps leads to duplicate reports and wasted time for both parties.
+
+## Before reporting anything
+
+### Step 1: Read the security model
+
+Fetch and read the project's security model before evaluating any finding:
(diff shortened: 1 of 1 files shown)


oscerd (approved):

FreeAndNil (approved):
LGTM

garydgregory:
Why?

ppkarwasz:
> Why?

Why do I want to add an `AGENTS.md`? We are receiving a lot of invalid security reports, often duplicates, certainly at least AI-aided. With such a file we control the instructions that AI agents follow.

garydgregory:
> > Why?
> 
> Why do I want to add an `AGENTS.md`? We are receiving a lot of invalid security reports, often duplicates, certainly at least AI-aided. With such a file we control the instructions that AI agents follow.

You're assuming a lot here. I hope to be wrong that this will be treated like robot.txt files. IOW, ignored by so many companies as to become useless and just clutter in our repository.

PR #4095, opened 2026-04-14
Switches from `protected_branches` to `rulesets`

ppkarwasz:
Similarly to apache/logging-parent#456 this PR switches from GitHub Branch Protection to Rulesets and:

- Keeps the same rules for `main`, while protection `2.x` is **temporarily** disabled in case we need to update the `.asf.yaml` file.
- Adds tag protection for the `rel/*` tags.


ppkarwasz:
After successfully enabling rulesets in `logging-parent`, I extended this PR to cover the `2.x` branch in cccea50. One small catch: until apache/infrastructure-asfyaml#93 is merged, these rulesets cannot be modified.

@puerco: are these controls sufficient to classify Log4j as **SLSA Source Level 4** compliant? As you know, we enforce quite a "Byzantine bureaucracy" here and would love something to show for it:

[SLSA 4](https://slsa.dev)

...alongside the in-toto attestations we will produce.

A couple of related questions:
- Should we also protect `release/*` branches to demonstrate that every commit up to a release has been reviewed? The `rel/*` tags sit on short-lived side branches that originate from the two protected branches (`2.x` and `main`).
- For tags, I opened #4096 to discuss cleaning up some historical inconsistencies before locking down `rel/*` tags.

> [!NOTE]
> For bystanders wondering: SLSA would not have prevented Log4Shell. It only protects against **malicious** actors, not honest mistakes that pass the full review process.

ppkarwasz:
Since `2.25.x` and `2.26.x` are also release branches, https://github.com/apache/logging-log4j2/pull/4095/commits/8ddf8cb27d7246f17bf8cbfbd843e6d06d2b31af adds them to the protected list.

FreeAndNil (approved):
lgtm

PR #4107, opened 2026-04-30
Switch reusable builds to `gha/v0`

ppkarwasz:
By switching to the `gha/v0` branch of the reusable builds we make sure that the GitHub Action dependency versions are always authorized.


vy:
Discussed with @ppkarwasz and decided to close in favor of #4108.

PR #4119, opened 2026-05-09
Remove the `patternFlags` attribute from `RegexFilter`

### LINKED ISSUE #3086: Log4j 2.x Documentation - RegexFilter does not document the possible nested element PatternFlags

ramanathan1504:
Fixes #3086 

This pull request refactors the `RegexFilter` to use a plugin builder pattern, restores support for configuring regex `patternFlags` via a comma-separated attribute, and improves robustness around the `useRawMsg` option. It also adds tests and documentation for the new pattern flags feature.

**RegexFilter Refactor and Feature Restoration:**

* Refactored `RegexFilter` to use a plugin builder (`Builder` class), simplifying construction and improving maintainability. The builder now uses a primitive boolean for `useRawMsg` to avoid null-unboxing issues. [[1]](diffhunk://#diff-240d93e5d49d2cce79e7ba0c5f0c78a5048288d3a122556b5dae7c9d81907d2dR119-R123) [[2]](diffhunk://#diff-240d93e5d49d2cce79e7ba0c5f0c78a5048288d3a122556b5dae7c9d81907d2dL133-L175)
* Restored and improved support for configurable regex `patternFlags` via a comma-separated string attribute (e.g., `patternFlags="CASE_INSENSITIVE"`), replacing the previous, more cumbersome approach.

**Testing Improvements:**

* Added a new test (`testPatternFlagsFromConfiguration`) in `RegexFilterTest.java` to verify that pattern flags are correctly loaded from configuration files.
* Added a corresponding test configuration file `RegexFilterPatternFlagsTest.xml` to exercise the new pattern flags feature.

**Documentation and Changelog:**

* Updated the manual to document the `patternFlags` attribute, listing all supported flag values and usage.
* Added a changelog entry describing the refactor, restoration of `patternFlags`, and the builder improvements.

**Other:**

* Bumped the OSGi version for the filter package to `2.26.0` to reflect the new feature.


log4j-core/src/main/java/org/apache/logging/log4j/core/filter/RegexFilter.java
```diff
@@ -130,48 +135,100 @@ public String toString() {
...
-     * @throws IllegalAccessException  When there is no access to the definition of the specified member.
-     * @throws IllegalArgumentException When passed an illegal or inappropriate argument.
      */
-    // TODO Consider refactoring to use AbstractFilter.AbstractFilterBuilder
-    @PluginFactory
     public static RegexFilter createFilter(
             // @formatter:off
             @PluginAttribute("regex") final String regex,
-            @PluginElement("PatternFlags") final String[] patternFlags,
+            @PluginAttribute("patternFlags") final String[] patternFlags,
```
vy: This sounds like a breaking API change to me. That is, I guess the following was working, and is broken with this change:

```
<RegexFilter regex="error" onMatch="ACCEPT" onMismatch="DENY">
  <PatternFlags>
    <tag-name-here-does-not-matter>UNIX_LINES</tag-name-here-does-not-matter>
    <tag-name-here-does-not-matter>DOTALL</tag-name-here-does-not-matter>
  </PatternFlags>
</RegexFilter>
```

@ramanathan1504, would you mind verifying this, please?

ramanathan1504: @vy, Yes, I also verified that this introduces a breaking change.
What would be the preferred approach here? Should we add OSGi/bnd versioning support for backward compatibility, or should we hold this change for the next major version? Another option would be to restore backward compatibility by supporting both the old and new configuration formats.

vy (approved):

PR #4124, opened 2026-05-10
Improve logging for `LinkageError` scenarios involving the LMAX Disruptor library

### LINKED ISSUE #2250: Log disruptor initialization errors
As shown in [this user report](https://lists.apache.org/thread/ynxrn9vsmdqs31jrml71xb582o7m74qf), linkage errors with the LMAX Disruptor library never end up in the status log.

We should add more debugging there to allow users to easily diagnose such problems.

Since upgrades to LMAX Disruptor 4.x will not cause any more linkage errors (cf. #1829), this task is not a priority, but a nice to have.

ramanathan1504:
FIxes #2250

## Reference URLs
- https://logging.apache.org/log4j/2.x/manual/async.html
- https://github.com/LMAX-Exchange/disruptor
- https://lists.apache.org/thread/ynxrn9vsmdqs31jrml71xb582o7m74qf

## What was done
-Improves user visibility by updating the logging message while keeping the existing initialization error handling behavior unchanged.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
@@ -32,6 +32,9 @@ import org.apache.logging.log4j.util.PropertiesUtil;
  */
 final class DisruptorUtil {
     private static final Logger LOGGER = StatusLogger.getLogger();
+    private static final String LINKAGE_ERROR_MESSAGE =
+            "Log4j2 detected a linkage error while initializing LMAX Disruptor. "
+                    + "Please check your classpath and ensure Disruptor version is compatible.";
     private static final int RINGBUFFER_MIN_SIZE = 128;
     private static final int RINGBUFFER_DEFAULT_SIZE = 256 * 1024;
     private static final int RINGBUFFER_NO_GC_DEFAULT_SIZE = 4 * 1024;
@@ -52,14 +55,13 @@ final class DisruptorUtil {
 
     static final int DISRUPTOR_MAJOR_VERSION = detectDisruptorMajorVersion();
 
-    // TODO: replace with LoaderUtil.isClassAvailable() when TCCL is removed
-    // See: https://github.com/apache/logging-log4j2/issues/3706
     private static int detectDisruptorMajorVersion() {
+        // Avoid using `LoaderUtil`, which might choose an incorrect class loader (e.g. TCCL in OSGi) – see #2768.
         try {
             Class.forName(
-                    "com.lmax.disruptor.SequenceReportingEventHandler", true, DisruptorUtil.class.getClassLoader());
+                    "com.lmax.disruptor.SequenceReportingEventHandler", false, DisruptorUtil.class.getClassLoader());
             return 3;
-        } catch (final ClassNotFoundException e) {
+        } catch (final ClassNotFoundException | LinkageError ignored) {
             return 4;
         }
     }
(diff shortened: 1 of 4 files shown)


vy (approved):
Simplify logging

vy (changes requested):
> Added defensive LinkageError handling in DisruptorUtil when creating async logger exception handlers.

I don't see that.

> Guarded sequenceCallback.set(sequence) with a null check in RingBufferLogEventHandler4 to prevent NPE when callback is not initialized.

I don't see that either.

log4j-core/src/main/java/org/apache/logging/log4j/core/async/DisruptorUtil.java
```diff
@@ -55,13 +55,17 @@ final class DisruptorUtil {
...
             Class.forName(
-                    "com.lmax.disruptor.SequenceReportingEventHandler", true, DisruptorUtil.class.getClassLoader());
+                    "com.lmax.disruptor.SequenceReportingEventHandler", false, DisruptorUtil.class.getClassLoader());
+            version = 3;
             return 3;
-        } catch (final ClassNotFoundException e) {
-            return 4;
+        } catch (final ClassNotFoundException ignored) {
+            version = 4;
         }
```
vy:
```suggestion
        int version = 4;
try {
Class.forName(
                    "com.lmax.disruptor.SequenceReportingEventHandler", false, DisruptorUtil.class.getClassLoader());
            version = 3;
return 3;
        } catch (final ClassNotFoundException ignored) {
            // Do nothing
}
```

ramanathan1504:
> > Added defensive LinkageError handling in DisruptorUtil when creating async logger exception handlers.
> 
> I don't see that.
> 
> > Guarded sequenceCallback.set(sequence) with a null check in RingBufferLogEventHandler4 to prevent NPE when callback is not initialized.
> 
> I don't see that either.

Forgot to mention this earlier @vy — for this PR, @ppkarwasz suggested that we do not need additional LinkageError handling here, since Log4j is already designed to handle it during the initialization stage. The main goal is just to provide better visibility to the user about what is happening.

That’s why I reverted the extra LinkageError handling from the earlier commit and only kept the logging message improvement.

https://github.com/apache/logging-log4j2/pull/4124/changes/dc9f81f979dcca0dcc8bd41969d233844985b4b8

vy (changes requested):
@ramanathan1504, would you mind adding a brief changelog entry, please?

I'd appreciate it if you can update the description of the PR to reflect the goal.

ramanathan1504:
> @ramanathan1504, would you mind adding a brief changelog entry, please?
> 
> I'd appreciate it if you can update the description of the PR to reflect the goal.

@vy 
updated pr description and added changelog

src/changelog/.2.x.x/fix-log-disruptor-initialization-errors.xml
```diff
@@ -0,0 +1,13 @@
...
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="changed">
+    <issue id="3176" link="https://github.com/apache/logging-log4j2/issues/2250"/>
+    <issue id="3176" link="https://github.com/apache/logging-log4j2/pull/4124"/>
+    <description format="asciidoc">
+        Improved status logging for LinkageError scenarios involving the LMAX Disruptor library to provide better diagnostics and user visibility during initialization failures. This change helps users more easily identify dependency compatibility issues that were previously not surfaced in the status log.
```
vy:
```suggestion
        Improve logging for `LinkageError` scenarios involving the LMAX Disruptor library
```

vy (approved):

PR #315, opened 2019-11-21
added retry count if failed to send

joongs4:
In my system, there is Kafka only for logging but as Kafka is not yet completely installed, it often crashes. If Kafka is down, an application gets blocked. I believe an application never gets affected by logging. Therefore, I added a attribute for retry count. 

I changed only two java files below.
- org.apache.logging.log4j.kafka.appender.KafkaManger
- org.apache.logging.log4j.kafka.appender.KafkaAppender

Thanks.

PR #3989, opened 2025-11-28
Disable commit signatures

vy:
Per this `dev@logging` [lazy consensus](https://lists.apache.org/thread/k6cb4yw2dc9p186l64osy71sm7j8sjn4), disable commit signatures. While at it, simplify the PR checklist, since nobody reads it.

PR #4125, opened 2026-05-14
Fix KafkaAppender reporting error after successful retry

SebTardif:
Fix KafkaAppender retry logic that always reports an error to the error handler even when a retry succeeds.

When `retryCount` is configured and the initial `tryAppend()` fails, the retry loop uses `break` to exit on success. However, `break` only exits the `while` loop; execution always reaches the `error()` call afterward. This causes spurious error notifications for transient Kafka failures that were successfully recovered by a retry.

This change replaces `break` with `return` so that a successful retry exits `append()` without reporting an error. Retry exceptions are now logged at DEBUG level for diagnostics instead of being silently discarded.

Also removes dead code in `Builder.getRetryCount()` where `Integer.valueOf(int)` was wrapped in a `NumberFormatException` catch that can never fire.

The bug was introduced in #315.

## Checklist

* Tests are provided


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaAppender.java
```diff
@@ -216,16 +210,18 @@ public void append(final LogEvent event) {
             try {
                 tryAppend(event);
             } catch (final Exception e) {
-
-                if (this.retryCount != null) {
-                    int currentRetryAttempt = 0;
-                    while (currentRetryAttempt < this.retryCount) {
-                        currentRetryAttempt++;
+                if (this.retryCount != null && this.retryCount > 0) {
```
ramanathan1504:
```java
if (this.retryCount != null) {
                    int currentRetryAttempt = 0;
                    while (currentRetryAttempt < this.retryCount) {
                        currentRetryAttempt++;
                        try {
                            tryAppend(event);
                            return; // Fix: Exit early on success
                        } catch (final Exception retryException) {
                            LOGGER.debug(
                                    "Unable to write to Kafka in appender [{}], retry attempt [{}/{}].",
                                    getName(),
                                    currentRetryAttempt,
                                    this.retryCount,
                                    retryException);
                        }
                    }
                }

This still applies all three of your excellent fixes (early return, no swallowed exceptions, and preserving the stack trace in debug mode) but modifies fewer lines of code.

**Review by ramanathan1504 — COMMENTED** [review]
LGTM overall — great fix for the `break`/`return` retry behavior, and thanks for adding DEBUG logging plus cleaning up `getRetryCount()`.

One small suggestion commented to keep the diff minimal: retain the existing `while` loop shape and just apply the early `return` + DEBUG logging inside it.

File: log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaAppender.java (hunk 1 above)
Comment (SebTardif) [reply log4j-core/src/main/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaAppender.java]: Good call, adopted. The while-loop preserves the original structure with a smaller diff. Force-pushed.

**Comment by ramanathan1504** [comment]
**Hi @SebTardif ,**

Thank you so much for your patience and for contributing to this! We really appreciate the time and effort you've put into getting this `KafkaAppender` issue resolved. The code changes look good and the review is cleared! 

Before we can finally merge this, there are just a few quick administrative things we need to address:

*   **Avoid Force-Pushing during reviews:** I noticed that the recent updates were force-pushed. For future PRs, please just push new commits normally rather than force-pushing. Force-pushing overwrites the Git history, which makes it very difficult for reviewers to track the differences between the old and new changes. 
*   **Commit Signatures:** It looks like your commits are currently signed with a self-signed key. For security and verification purposes, we require commits to be signed off using a valid **GPG key**. Could you please set up a GPG key, re-sign your commits, and update the PR?
*   **Add a Changelog Entry:** Please add an entry to our changelog file for this fix. You can use this exact wording: 
    > *"Fixed an issue where `KafkaAppender` incorrectly reported an error to the error handler despite a successful retry."*

**Next Steps:**
Since you need to re-sign your previous commits and add the changelog, you will actually need to force-push **one last time** to update this PR. Once you've added the changelog and signed with a GPG key, go ahead and push it up, and we'll get this merged!

Thanks again for your great work on this!

File: log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/mom/kafka/KafkaAppenderTest.java [hunk 2]
```diff
@@ -206,6 +213,40 @@ public void testAppendWithRetryCount() {
...
     }
 
+    @Test
+    public void testRetrySuccessDoesNotReportError() {
+        retrySendCount.set(0);
+        final AtomicBoolean errorReported = new AtomicBoolean(false);
+        final Appender appender = ctx.getRequiredAppender("KafkaAppenderWithRetryCount");
+        final ErrorHandler originalHandler = ((KafkaAppender) appender).getHandler();
+        try {
+            ThreadContext.put("KafkaAppenderRetrySuccessTest", "true");
```
vy: You might consider using [CloseableThreadContext](https://logging.apache.org/log4j/2.x/manual/thread-context.html#CloseableThreadContext).

vy:
Great guidance @ramanathan1504! :100: 

> * **Commit Signatures:** It looks like your commits are currently signed with a self-signed key. For security and verification purposes, we require commits to be signed off using a valid **GPG key**.

@ramanathan1504, @SebTardif, we've dropped the requirement for commit signatures — see #3989. You can skip that chore.

SebTardif:
Thanks @ramanathan1504 for the review guidance and @vy for clarifying on signatures.

Adopted `CloseableThreadContext` as suggested. Pushed as a new commit this time (no force-push).

ramanathan1504 (approved):

ramanathan1504:
@SebTardif 

Thanks for the update and for addressing the review comments.

I noticed the build failed due to Spotless formatting checks, so I pushed a small formatting-only fix to get the workflows green again. For future PRs, please make sure to run the Spotless checks locally before pushing to avoid CI failures.

Everything looks good now.

PR #4127, opened 2026-05-21
Fix resource leaks in `ConfigurationSource` when loading via URL

SebTardif:
## What this PR does

In `log4j-core/.../config/ConfigurationSource.java`, method `getConfigurationSource(URL)`: `url.openConnection()` creates a URLConnection, but on both error paths (FileNotFoundException, IOException/URISyntaxException) the connection is never closed or disconnected.

For `HttpURLConnection` this leaks a socket. For `JarURLConnection` this leaks a file handle (relevant to locked JARs on Tomcat redeploy scenarios).

**Fix:**
- Added a `boolean success` flag with try-finally pattern to disconnect/close on error paths
- Reused the existing `file` variable instead of calling `FileUtils.fileFromUri(url.toURI())` a second time

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
@@ -178,23 +178,17 @@ public class ConfigurationSourceTest {
             openConnectionCalls.set(0);
             getInputStreamCalls.set(0);
 
-            final Method method =
-                    ConfigurationSource.class.getDeclaredMethod("getConfigurationSource", URL.class);
+            final Method method = ConfigurationSource.class.getDeclaredMethod("getConfigurationSource", URL.class);
             method.setAccessible(true);
             final Object result = method.invoke(null, url);
 
-            assertTrue(
-                    openConnectionCalls.get() > 0,
-                    "Custom URLStreamHandler was not used by UrlConnectionFactory");
+            assertTrue(openConnectionCalls.get() > 0, "Custom URLStreamHandler was not used by UrlConnectionFactory");
 
             assertNull(result, "Expected null return for a 404 response");
 
             assertTrue(disconnected.get(), "disconnect() must be called on the error path");
 
-            assertEquals(
-                    1,
-                    getInputStreamCalls.get(),
-                    "getInputStream() should be called exactly once");
+            assertEquals(1, getInputStreamCalls.get(), "getInputStream() should be called exactly once");
         } finally {
             if (previous != null) {
                 System.setProperty(propKey, previous);
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
@@ -134,11 +134,6 @@ public class ConfigurationSourceTest {
      * Verifies that the error-path cleanup in {@code getConfigurationSource}
      * properly disconnects an {@link HttpURLConnection} on failure without
      * calling {@link URLConnection#getInputStream()} a second time.
-     *
-     * <p>This test fails against the original code (no cleanup) because
-     * {@code disconnect()} is never called, and fails against an intermediate
-     * fix that added a {@code finally} block calling {@code getInputStream()}
-     * again to obtain a stream for closing.
      */
     @Test
     void getConfigurationSource_disconnectsHttpConnection_onError() throws Exception {
@@ -194,13 +189,8 @@ public class ConfigurationSourceTest {
 
             assertNull(result, "Expected null return for a 404 response");
 
-            // Negative‐test #1: the original code has no finally block, so
-            // disconnect() is never called on the HttpURLConnection.
             assertTrue(disconnected.get(), "disconnect() must be called on the error path");
 
-            // Negative‐test #2: an intermediate fix called getInputStream()
-            // a second time in the finally block instead of closing the
-            // already-tracked stream reference.
             assertEquals(
                     1,
                     getInputStreamCalls.get(),
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
@@ -25,15 +25,22 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import com.sun.management.UnixOperatingSystemMXBean;
 import java.io.ByteArrayInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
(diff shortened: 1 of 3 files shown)


ramanathan1504:
@SebTardif 

"Great work on this! Resource leaks in configuration loading can be incredibly difficult to debug, especially in containerized or high-availability environments. This is a solid improvement to Log4j's stability.

#### Suggestions for Completion:

To make this PR "merge-ready" for the standards, we should add a failure test case and a changelog entry:

**1. Add a Failure Test Case**
Since the fix specifically targets the failure path (`!success`), we should add a test that intentionally fails to find a file inside a JAR to ensure the handle is released.

Add this to `ConfigurationSourceTest.java`:
```java
@Test
void testJarFileLeakOnFailure() throws Exception {
    final Path jarFile = prepareJarConfigURL(tempDir);
    // Path inside JAR that does NOT exist
    final URL brokenJarConfigURL = new URL("jar:" + jarFile.toUri().toURL() + "!/missing/file.xml");
    
    final long expectedFdCount = getOpenFileDescriptorCount();
    
    // This triggers the FileNotFoundException internally
    ConfigurationSource configSource = ConfigurationSource.fromUri(brokenJarConfigURL.toURI());
    
    assertNull(configSource, "Source should be null on failure");

    // 1. Verify File Descriptors (Unix/Linux)
    assertEquals(expectedFdCount, getOpenFileDescriptorCount(), 
        "File descriptors leaked after failed configuration load!");

    // 2. Verify File Lock (Windows)
    try {
        Files.delete(jarFile);
    } catch (IOException e) {
        fail("Could not delete JAR file! A handle was leaked in the failure path.");
    }
}
```

**2. Add a Changelog Entry**
Log4j 2 uses an automated tool for release notes. Please create a new file: `src/changelog/.2.x.x/fix_ConfigurationSource_leak.xml` :

```xml
<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns="https://logging.apache.org/xml/ns"
       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
       type="fixed">
  <issue id="4127" link="https://github.com/apache/logging-log4j2/pull/4127"/>
  <description format="markdown">
     Fix resource leaks in ConfigurationSource when loading via URL
  </description>
</entry>
```

Thanks again for the contribution! don't forgot to apply spotless !!😊

SebTardif:
Thanks for the detailed suggestions @ramanathan1504! Added both:

- **Failure test**: `testNoJarFileLeakOnFailure` — creates a JAR, requests a missing entry (`!/missing/file.xml`), and asserts no FD leak and no file lock after the failed load. Follows the same pattern as the existing `testNoJarFileLeak` and `testLoadConfigurationSourceFromJarFile` tests.

- **Changelog entry**: `src/changelog/.2.x.x/fix_configuration_source_url_leak.xml`

Spotless applied. All 4 ConfigurationSourceTest tests pass.

ramanathan1504:
@SebTardif 

I pushed a follow-up commit to address this in the PR branch.

What I changed:
- Updated `ConfigurationSource` to close only the already-opened `InputStream` on failure paths, so we avoid potential URL/JAR resource leaks.
- Updated `ConfigurationSourceTest` to validate the failure case (missing JAR entry), confirm no file descriptor increase, and ensure the JAR is deletable afterward.

Please avoid force-pushing in future updates so review history and inline comments remain easy to track.

Thanks for the support!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/ConfigurationSourceTest.java
vy, line 1: @SebTardif, this test passes without your changes. Would you mind adding a test that fails without the changes, please?

SebTardif:
@vy Added a negative test in 4bfa56c. It injects a counting `HttpURLConnection` via a custom `URLStreamHandler` and asserts two things:

1. `disconnect()` is called on the error path (fails against the original code, which has no `finally` block)
2. `getInputStream()` is called exactly once (fails against the intermediate fix that re-invoked `getInputStream()` in the `finally` block instead of closing the tracked stream)

vy:
@SebTardif, thanks for the fix, LGTM.

I'm okay with the state of the things, but @ppkarwasz is not. He said he will review this.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
```diff
@@ -359,24 +360,44 @@ public String toString() {
...
+                    if (openedStream != null) {
+                        try {
+                            openedStream.close();
+                        } catch (final IOException ignored) {
+                            // Best-effort cleanup
+                        }
+                    }
+                    if (urlConnection instanceof HttpURLConnection) {
+                        ((HttpURLConnection) urlConnection).disconnect();
+                    }
```
ppkarwasz: Thank you for your PR!

This `finally` clause and `success` flag complicate the code more than necessary, and the `openedStream != null` branch looks like dead code: `openedStream` is only assigned by `getInputStream()`, so if that call throws an `IOException` the variable is still `null`, and if it succeeds we go straight to building the `ConfigurationSource` and returning (the constructors don't throw). There is no path where `openedStream` is non-null and `success` is `false`, so `openedStream.close()` is never actually reached.

Your real goal here is to call `disconnect()` on the `HttpURLConnection` on the error path, right? If so, I'd suggest splitting the nested `try...catch` into two sequential ones:

1. The first assigns `file` and `urlConnection`. Nothing has been opened yet, so it needs no cleanup (it just catches `URISyntaxException` and `IOException` and logs).
2. The second calls `getInputStream()` and builds the source, and calls `disconnect()` **only** in the `IOException` catch. No `success` flag, and no `openedStream.close()`.

A single `catch (IOException)` in the second block covers both the generic case and `FileNotFoundException` (which is what a missing file, or an HTTP 404/410, surfaces as), so the two can share one handler, but we can have different log messages.

One small thing to fold in: read `urlConnection.getLastModified()` *before* `getInputStream()` in the last branch, so the stream is only opened once everything that could fail has already succeeded.

ppkarwasz (changes requested):

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
SebTardif: Good catch, you are right that `openedStream.close()` was dead code. Simplified as you suggested in 66b96e2:

- Split into two sequential try-catch blocks (resolve file/connection first, then open stream)
- `disconnect()` only in the `IOException` catch, no `success` flag
- Moved `getLastModified()` before `getInputStream()` in the generic branch

ppkarwasz (approved):

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationSource.java
ppkarwasz: LGTM!

vy:
@SebTardif, would you mind fixing the build (Spotless) failures, please?

Since @ppkarwasz is on board as well, let's carry this PR to the finish line and get it integrated.

SebTardif:
Fixed Spotless formatting in ebab7a2. CI should be green now.

vy (approved):

PR #4131, opened 2026-05-28
Fix CI

vy:
Ports #4108 to `main`.


vy:
@ppkarwasz, would you mind reviewing this and checking if we need to do something about `process-dependabot.yaml` in `2.x`, please?

ppkarwasz (approved):

PR #4152, opened 2026-06-26
[main] Fix RegexFilter NPE when `useRawMsg` is null (#3265 port)

vpelikh:
Ports the fix from **#3265** from the `2.x` branch to `main`.

In the `RegexFilter.createFilter()` factory method, the `useRawMsg` parameter is declared as `Boolean` (boxed). When passed directly to the `RegexFilter(boolean raw, ...)` constructor, a `null` value causes an NPE during unboxing.

This change uses `Boolean.TRUE.equals(useRawMsg)` instead, which safely handles `null` by defaulting to `false`.

## Changes

- **`log4j-core/.../filter/RegexFilter.java`**: Changed `useRawMsg` → `Boolean.TRUE.equals(useRawMsg)` in constructor call
- **`log4j-core-test/.../filter/RegexFilterTest.java`**: Added test ensuring no exception when `useRawMsg` is `null`

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/RegexFilterTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/filter/RegexFilterTest.java
@@ -18,6 +18,7 @@ package org.apache.logging.log4j.core.filter;
 
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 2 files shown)


ramanathan1504:
@vpelikh, Looks good, the fix and tests work perfectly locally and adapt well to the Log4j 3 variables.

However, we don't port changelog entries for direct syncs from 2.x to main. Please delete the changelog file (src/changelog/.3.x.x/3239_npe_fix_regex_filter_creator.xml). I'll merge this as soon as it's removed!

vpelikh:
Thank you for the review, @ramanathan1504!

I've removed the changelog file from this PR. I also just went through all my other open backport PRs a few minutes ago and cleaned up the changelog entries there, so they should all be consistent now and won't cause any extra noise.

ramanathan1504:
Thanks for the update, @vpelikh!

One quick request for the future: please try to avoid force-pushing when updating a PR. While it's fine for this because here there in no PR review comments, force-pushing erases the commit history and makes it very difficult for us to review iterative changes.

Everything else looks perfect. Thanks

ramanathan1504:
@vpelikh 
For future PRs, please make sure to check our [Development Guidelines (BUILDING.adoc)](https://www.google.com/url?sa=E&q=https%3A%2F%2Fgithub.com%2Fapache%2Flogging-log4j2%2Fblob%2Fmain%2FBUILDING.adoc) and run ./mvnw verify (or at least ./mvnw spotless:apply) locally before committing. This will help catch these formatting errors before the CI does!

ramanathan1504 (approved):

PR #3195, opened 2024-11-09
Fix(#2769): Modify the annotation processor in 2.x to give a warning if a plugin builder attribute does not have a public setter.

jaykataria1111:
The following commit modifies the annotation processor in 2.x to shell out a warning if the plugin builder attribute does not have a public setter. 

A ` @SuppressWarnings("log4j.public.setter")` attribute can be used to ignore this compilation warning incase it is a known issue.

## Checklist

* Base your changes on `2.x` branch if you are targeting Log4j 2; use `main` otherwise
* `./mvnw verify` succeeds (if it fails due to code formatting issues reported by Spotless, simply run `./mvnw spotless:apply` and retry)
* Non-trivial changes contain an entry file in the `src/changelog/.2.x.x` directory
* Tests for the changes are provided
* [Commits are signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) (optional, but highly recommended)

PR #4156, opened 2026-06-26
[main] Warn when `@PluginBuilderAttribute` field lacks public setter (#3195 port)

vpelikh:
This ports the annotation processor validation from #3195 (2.x) to `main` (3.x).

### What it does
The `PluginProcessor` annotation processor now validates that every `@PluginBuilderAttribute` field in a plugin builder class has a public `setXxx()` or `withXxx()` method that:

1. Takes exactly 1 parameter matching the field type
2. Returns a type assignable to the enclosing builder class
3. Is declared `public`

If no such setter is found, a compilation **ERROR** is emitted:

```
The field `myField` does not have a public setter. Note that
@SuppressWarnings("log4j.public.setter") can be used on the field to
suppress the compilation error.
```

### Suppressing false positives
Use `@SuppressWarnings("log4j.public.setter")` on the field to suppress the warning for legitimate cases (e.g., fields set via constructor or reflection).

### Files changed
- **`PluginProcessor.java`** — Added setter validation logic in `processBuilderAttributeFields()`
- **`PluginProcessorPublicSetterTest.java`** — Test with 3 cases: public setter OK, missing setter → error, suppressed → no error
- **`FakePluginPublicSetter.java.source`** — Test plugin for the above test
- **`SocketPerformancePreferences.java`** — Added `@SuppressWarnings` to 3 fields (setters return `void`, not builder type)
- **`StringMatchFilter.java`** — Added `@SuppressWarnings` (`setMatchString` name doesn't match field `text`)
- **`Rfc5424Layout.java`** — Added `@SuppressWarnings` (field `enterpriseNumber` has no setter)

### Note
Unlike the 2.x version, the 3.x `PluginProcessor` uses `@SupportedAnnotationTypes({"org.apache.logging.log4j.plugins.Plugin", "org.apache.logging.log4j.plugins.PluginBuilderAttribute", "org.apache.logging.log4j.core.config.plugins.PluginBuilderAttribute"})` and iterates by FQN, because 3.x has two `@PluginBuilderAttribute` variants.


ramanathan1504:
@vpelikh

However, your new processor works so well that it actually caught a pre-existing violation in main! When building locally, the compile fails in log4j-jdbc because JdbcAppender is missing public setters:

```
[ERROR] .../log4j/jdbc/appender/JdbcAppender.java:[61,25] The field immediateFail does not have a public setter.
[ERROR] .../log4j/jdbc/appender/JdbcAppender.java:[81,22] The field reconnectIntervalMillis does not have a public setter.
```

Could you please update `JdbcAppender.java` in this PR (either by adding the missing setters or suppressing the warning) so the CI builds successfully? Once the build is green, I'll happily approve and merge!

ramanathan1504:
@vpelikh, Also, I noticed Spotless formatting on 3 PR as well. When you push your updates, please make sure to run `./mvnw spotless:apply` Follow the `BUILDING.adoc`

vpelikh:
Thank you for the review, @ramanathan1504 ! Great catch – you're absolutely right, the new validation uncovered real issues in `JdbcAppender`.

I've just pushed two commits that:
1. Fix the setter return types – changed `setImmediateFail()` and `setReconnectIntervalMillis()` from `void` to `B` (the builder type), so they now satisfy the annotation processor's requirements.
2. Apply Spotless formatting – ran `./mvnw spotless:apply` to fix all formatting violations across the affected files.

The build should be green now. Please take another look when you have a moment – happy to adjust anything else if needed.

vpelikh:
@ramanathan1504 gentle ping, PR ready for re-review. Thanks!

ramanathan1504:
> @ramanathan1504 gentle ping, PR ready for re-review. Thanks!

Noted!. I will Try to close within a week. Thanks

ramanathan1504:
@vpelikh  both my earlier points are fixed. Nice catch on the `withXxx` guard too; 2.x only checks `startsWith("set")`, so its `with` branch is dead code.

One blocker: the new test leaves `example/` and a `PluginService` file in the module directory. The `@AfterEach` deletes `Log4j2Plugins.dat` — that's the 2.x artifact name, it doesn't exist in 3.x — so it cleans up a file that's never created and misses the two that are. A `@TempDir` output path fixes it.

Also worth a look: adding `PluginBuilderAttribute` to `@SupportedAnnotationTypes` isn't needed — 2.x resolves it with `getTypeElement()` inside `process()`, and `internal/Annotations.java` already owns both FQNs. And the `StringMatchFilter` suppression collides with #4153, which renames that setter.

vpelikh:
@ramanathan1504, thanks for review! Addresed your comments.

ramanathan1504 (changes requested):
1. `log4j-core/src/main/java/org/apache/logging/log4j/core/filter/StringMatchFilter.java:239` — I pulled the branch and `./mvnw compile` fails in log4j-core: ``The field `text` does not have a public setter``. #4153 does add `setText`, but it's still open, so right now this PR breaks the build on main. Could we keep the suppression here and let #4153 remove it in the same hunk when it merges? Then neither PR depends on the other's merge order.

2. `log4j-core/src/main/java/org/apache/logging/log4j/core/layout/Rfc5424Layout.java:637` — on 2.x this one isn't suppressed, the builder has a real `setEnterpriseNumber(Integer)` (Rfc5424Layout.java:1144 on 2.x). 3.x only has `setEin(String)`, so the field can't be set from Java at all, which looks like exactly the gap the processor is meant to find. Would porting that setter over be better than suppressing it?

3. Small one on the tests — I restored `PluginProcessor.java` to origin/main and re-ran: only `warnWhenPluginBuilderAttributeLacksPublicSetter` fails, the other two pass with the processor gone. An `assertThat(errorDiagnostics).hasSize(1)` would make them real.

vpelikh:
Thanks for the review, @ramanathan1504! I’ve addressed all three points:

1. StringMatchFilter — Kept the @SuppressWarnings here with a // Remove once #4153 renames the setter to match the field name comment, so neither PR depends on the other’s merge order. #4153 can drop it in the same hunk when it merges.

2. Rfc5424Layout — Good catch, that was exactly the gap the processor is meant to find. I ported the fluent setEnterpriseNumber(Integer) setter over from 2.x and removed the suppression instead of hiding it.

3. Tests — Added assertThat(errorDiagnostics).hasSize(1) to all three cases so they fail if the processor is absent, not just when the specific message is absent.

Thanks again for the thorough review and for staying patient!

ramanathan1504 (commented):
@vpelikh 

1. `log4j-plugin-processor/src/test/java/.../PluginProcessorPublicSetterTest.java:60` — the `.java.source` copy and the `@AfterEach` delete are a 2.x thing, the fixture had to live under `src/test/java` there. `PluginProcessorTest` in this same module reads its fixtures straight out of `src/test/resources/example/*.java` with `getResource`. Renaming the fixture to `.java` would drop the copy, the delete and the commons-io use — want to follow that instead?

2. `log4j-plugin-processor/src/main/java/.../PluginProcessor.java:135` — `@SupportedAnnotationTypes` only declares `Plugin`, so `annotations` can't hold anything else and the loop plus `contentEquals` is doing no work. The old `if (!annotations.isEmpty())` with your `processBuilderAttributeFields(roundEnv)` call appended keeps it to two lines.

Happy either way on both

vpelikh:
Thanks for the review, @ramanathan1504! I've addressed both points.

**Test refactor** — Good call on following the existing `PluginProcessorTest` pattern. I renamed the fixture to `FakePluginPublicSetter.java` and now read it straight out of `src/test/resources` via `getResource`, which eliminated the copy and the commons-io `FileUtils` usage. The generated-file output directory is now a JUnit `@TempDir`, so there's no manual delete (and no commons-io in this test) either.

**`process()` simplification** — You're right that the annotation loop plus `contentEquals` was doing no work once `@SupportedAnnotationTypes` only declares `Plugin`. Reverted to the original `if (!annotations.isEmpty())` guard and appended the `processBuilderAttributeFields(roundEnv)` call, keeping it to two lines.

One question: would you prefer I squash/rebase the review commits before this is merged, or leave the history as is? Happy to tidy things up whichever way you prefer.

Let me know if there's anything else.

ramanathan1504 (commented):
@vpelikh 
One small thing to push before you merge, not worth another round. I deleted `@SuppressWarnings("log4j.public.setter")` from `FakePluginPublicSetter.java:34` and the three tests still passed, then I changed the suppression check in `PluginProcessor.java:199` to `if (false)` and they still passed. The reason is `FakePluginPublicSetter.java:45-50` — `attributeWithoutPublicSetterButWithSuppressAnnotation` has a public setter, so it was never going to error with or without the annotation. 2.x declares that field with no setter at all. Dropping those six lines is the whole fix — I tried it and the `hasSize(1)` assertions catch it on their own afterwards, 3 failures with `Expected size: 1 but was: 2`, no test change needed.

vpelikh:
Thanks @ramanathan1504 — good catch. Done in [26ce1ea](https://github.com/apache/logging-log4j2/pull/4156/commits/26ce1ea9f446e29d7c1d654ad66c268d43a9893a).

And I’m sorry this PR has been taking so long — thanks for your patience through the back-and-forth.

log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
```diff
@@ -124,6 +135,8 @@ public boolean process(final Set<? extends TypeElement> annotations, final Round
         if (!annotations.isEmpty()) {
             processPluginAnnotatedClasses(ElementFilter.typesIn(roundEnv.getElementsAnnotatedWith(Plugin.class)));
         }
+        // Validate @PluginBuilderAttribute fields in builder classes
```
ramanathan1504:
```suggestion
```
remove this its not any special case to add inline comments

log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
```diff
@@ -164,6 +177,76 @@ private void processPluginAnnotatedClasses(Set<TypeElement> pluginClasses) {
...
+                if (element instanceof VariableElement) {
+                    processBuilderAttributeField((VariableElement) element);
+                }
+            }
+        }
+    }
+
+    private void processBuilderAttributeField(final VariableElement element) {
+        final String fieldName = element.getSimpleName().toString();
+        // Check for @SuppressWarnings("log4j.public.setter")
```
ramanathan1504:
```suggestion
```
remove this its not any special case to add inline comments

log4j-core/src/main/java/org/apache/logging/log4j/core/filter/StringMatchFilter.java
```diff
@@ -237,6 +237,8 @@ public static StringMatchFilter.Builder newBuilder() {
     public static class Builder extends AbstractFilterBuilder<StringMatchFilter.Builder>
             implements org.apache.logging.log4j.core.util.Builder<StringMatchFilter> {
         @PluginBuilderAttribute
+        // Remove once #4153 renames the setter to match the field name
```
ramanathan1504:
```suggestion
```

Before give in pr Please check these comments are not need

log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
ramanathan1504, line 1: clean all inline comments

log4j-plugin-processor/src/test/java/org/apache/logging/log4j/plugin/processor/PluginProcessorPublicSetterTest.java
```diff
@@ -0,0 +1,107 @@
...
+        final URL fakePluginUrl = PluginProcessorTest.class.getResource(FAKE_PLUGIN_SOURCE);
+        assertThat(fakePluginUrl).isNotNull();
+        final Path fakePluginPath = Paths.get(fakePluginUrl.toURI());
+        diagnosticCollector = new DiagnosticCollector<>();
+        try {
+            final JavaCompiler compiler = ToolProvider.getSystemJavaCompiler();
+            final StandardJavaFileManager fileManager =
+                    compiler.getStandardFileManager(diagnosticCollector, Locale.ROOT, UTF_8);
+            try {
+                // Route generated files (Log4jPlugins.java, META-INF/services/...) to a temp directory
```
ramanathan1504:
```suggestion
```
here too

ramanathan1504 (changes requested):

log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
ramanathan1504: @vpelikh can you clear remaining in this file?

ramanathan1504:
@vpelikh Please review Before the commits all are good and aligned..Thanks

log4j-plugin-processor/src/main/java/org/apache/logging/log4j/plugin/processor/PluginProcessor.java
ramanathan1504: [375694d](https://github.com/apache/logging-log4j2/pull/4156/commits/375694dd2bd99f4db54b2e1f470ce7bb1053030e)

ramanathan1504 (approved):

PR #4158, opened 2026-06-26
Make XInclude processing in XML configurations opt-in

ppkarwasz:
This PR makes XInclude processing in XML configurations **opt-in**: it is disabled by default and enabled with the `log4j2.configurationEnableXInclude`property.

> [!NOTE]
> Stacked on #4155 (`fix/2.x/xinclude-fixup-attributes`), which is the base of this PR, so the diff shown here is only the opt-in change.
> Please review/merge that one first.

## Behavior

- `XmlConfiguration` reads `log4j2.configurationEnableXInclude` in its constructor and configures the `DocumentBuilder` accordingly. XInclude is no longer always on, and the previous LOG4J2-1127 "retry without XInclude" fallback is removed.
- When XInclude is disabled (the default), `xi:include` elements are left unexpanded and ignored, so a configuration that relied on them falls back to the default setup.

## Documentation

- The new property is documented alongside the other configuration-factory system properties
- The XInclude section of the manual notes that it is now opt-in.

## Tests

- `ConfigurationFactoryTest.xinclude` enables the property, and `testGetConfigurationWithMultipleUris` no longer incidentally loads an XInclude file.
- A new `ConfigurationFactoryUnitTest` loads the same logical configuration as XML, JSON, YAML, Java properties, an XInclude-composed file (with the property enabled) and a cross-format (XML + YAML) composite, asserting through the `Configuration` interface that each yields the same components. It also covers the XInclude-disabled fallback to the default configuration.

## Checklist

* [x] Based on `2.x` (stacked on #4155)
* [x] `./mvnw verify` succeeds (only the affected modules were tested locally)
* [x] Changelog entry in `src/changelog/.2.x.x`
* [x] Tests provided


ppkarwasz:
The `Simplify XInclude enablement` commit (b75050644662d12a7b6b2cf919e7dc3de6fdc909) drops the defensive workaround for [LOG4J2-3531](https://issues.apache.org/jira/browse/LOG4J2-3531) (set `setXIncludeAware(true)`, build a throwaway `DocumentBuilder` to probe support, and silently fall back to `setXIncludeAware(false)` when the parser rejects it).

Now that XInclude is opt-in, that dance is no longer needed: if a user's JAXP parser does not support XInclude, they simply don't enable `log4j2.configurationEnableXInclude`. So we set `setXIncludeAware(true)` directly and let an unsupported parser surface the error, instead of degrading to a silent no-op.

PR #4071, opened 2026-03-19
Add support for max compression delay in compression actions

ramanathan1504:
Fixes #4012 

### Changes
- Added `maxCompressionDelay` parameter to `GzCompressAction` and `ZipCompressAction`
- Extended `FileExtension` to support max compression delay configuration
- Updated `DefaultRolloverStrategy` to pass compression delay settings to compression actions
- Added backward-compatible constructors for existing code
- Comprehensive test coverage for compression delay functionality

### Files Modified
- **Core Implementation**
  - `log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/FileExtension.java`
  - `log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/GzCompressAction.java`
  - `log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZipCompressAction.java`
  - `log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/DefaultRolloverStrategy.java`

- **Tests**
  - `log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/FileExtensionCompressDelayTest.java` (new)
  - `log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/action/GzCompressActionTest.java` (new)
  - `log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/action/ZipCompressActionTest.java` (new)

PR #4159, opened 2026-06-26
feat(rolling): enhance compression delay with minCompressionDelaySeconds/maxCompressionDelaySeconds

### LINKED ISSUE #4012: Proactively deferring compression can reduce disk I/O pressure.

OpenZYK:
## Summary

Enhances the compression delay feature introduced in #4071 by replacing the single `maxCompressionDelaySeconds` parameter with a `minCompressionDelaySeconds` + `maxCompressionDelaySeconds` pair.

The actual delay is now chosen uniformly at random from the closed interval `[min, max]`, giving operators fine-grained control over the compression window (e.g. `min=30, max=120` to always wait between 30 s and 2 min before compressing).

## Changes

- **`AbstractAction`**: add `blockThread(int min, int max)` that picks a uniform random delay in `[min, max]` seconds; keep the deprecated single-arg overload as a backward-compatible delegate to `blockThread(0, max)`.
- **`GzCompressAction` / `ZipCompressAction` / `CommonsCompressAction`**: add `minDelaySeconds` field; new 6-arg constructors; deprecated 5-arg constructors delegate to the 6-arg form with `min=0`.
- **`FileExtension`**: add abstract `createCompressAction(…, min, max)` method and implement it in every enum constant; existing 5-arg overloads delegate to the new 6-arg form with `min=0` for backward compatibility.
- **`DefaultRolloverStrategy`**: add `@PluginBuilderAttribute minCompressionDelaySeconds` to Builder; add `setMinCompressionDelaySeconds` setter; add `getMinCompressionDelaySeconds` getter; new 10-arg constructor; deprecated 9-arg constructor delegates to the 10-arg form with `min=0`; `rollover()` passes both min and max to `createCompressAction`.
- **Tests**: extend `AbstractActionTest` with `blockThread(min,max)` edge-case tests; extend `FileExtensionCompressDelayTest` with GZ/ZIP min/max range tests and `DefaultRolloverStrategy` builder tests (21 tests pass).
- **Changelog**: add `src/changelog/.2.x.x/4012_enhance_compression_delay_min_max.xml`.

## Backward Compatibility

All existing configurations using only `maxCompressionDelaySeconds` continue to work unchanged (`min` defaults to `0`, producing the same `[0, max]` range as before).

## Configuration Example

```xml
<DefaultRolloverStrategy max="10"
    minCompressionDelaySeconds="30"
    maxCompressionDelaySeconds="120"/>
```

Closes #4012

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/FileExtensionCompressDelayTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/FileExtensionCompressDelayTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.logging.log4j.core.appender.rolling;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
 import static org.junit.jupiter.api.Assertions.assertThrows;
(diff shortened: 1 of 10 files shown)


ramanathan1504:
@OpenZYK 

  1.`Redundant for Load Leveling`: A simple `[0, max]` uniform distribution (PR
    #4071) already spreads the `I/O` spikes perfectly; forcing a minimum delay
    provides no extra infrastructure benefit.

  2.`API Maintenance Burden`: Introducing a second parameter (min) pollutes core
    public APIs with overloaded constructors and builder attributes that we must
    maintain indefinitely.

  3.`Thread & Memory Retention`: High minimum delays `(e.g., 20+ minutes)` will
    unnecessarily block execution threads and keep file/action states in the `JVM`
    heap, risking thread starvation and memory bloat.

  4.`Delayed Disk Cleanup`: Forcing a minimum wait time keeps large, uncompressed
    log files on the disk longer, defeating the goal of rapid disk-space
    recovery.

  5.`Configuration Overhead`: A single `maxCompressionDelaySeconds` parameter is
    simpler for users to configure and achieves the exact same operational
    outcome.

OpenZYK:
Thanks for the feedback. Closing this PR as the single maxCompressionDelaySeconds parameter is sufficient for the use case.

PR #4160, opened 2026-06-27
Consolidate string-to-URI conversion in NetUtils.toURI

ppkarwasz:
## Motivation

`NetUtils.toURI(String)` is the single entry point that turns a configuration location (given as a URI or a file system
path) into a `URI`.
Its previous implementation mixed several special cases and behaved inconsistently across platforms:

- an absolute UNIX path stayed a scheme-less URI while an absolute Windows path became a `file:` URI,
- a Windows drive letter (`C:`) could be mistaken for a URI scheme,
- and a URL repaired for blanks silently lost its user-info, port and query.

This PR rewrites the method around two rules and isolates the awkward cases in
one helper.

## Conversion logic

1. **Absolute URI** (a real scheme, not a single-letter Windows drive letter) is returned **as commanded**, unchanged.
   This covers `classpath:`, `file:`, `http(s):`, and Apache Commons VFS URLs such as `zip:file:///a.zip!/log4j2.xml`.

2. Otherwise, the value is a **file system path**:
    - an **absolute** path (as `File.isAbsolute()` judges it on the running OS: a UNIX `/...` path, a Windows drive
      `C:\...`/`C:/...`) becomes a `file:` URI holding the full path;
    - a **relative** path becomes a **scheme-less** URI, to be resolved later as a file or a class path resource.

> [!IMPORTANT]
>
> Log4j Core handles **relative** URIs specially: first it tries to resolve the URI as a local file and falls back to a classpath lookup.

## Behavioral changes

The conversion outcome changes in the following cases. The before/after values below were captured by running the new test suite against the previous implementation (`<cwd>` is the process working directory).

| Input                         | Before                              | After                               | Note                            |
|-------------------------------|-------------------------------------|-------------------------------------|---------------------------------|
| `http://h/a b/x?env=prod x`   | `http://h/a%20b/x`                  | `http://h/a%20b/x?env=prod%20x`     | query preserved                 |
| `https://user@h:8443/a b/x`   | `https://h/a%20b/x`                 | `https://user@h:8443/a%20b/x`       | user-info and port preserved    |
| `/path/without/spaces` (UNIX) | `/path/without/spaces`              | `file:/path/without/spaces`         | absolute path now a `file:` URI |
| `C:/dir/log4j2.xml`   | `C:/dir/log4j2.xml` (scheme `C`)    | `C%3A/dir/log4j2.xml` (scheme-less) | drive letter is not a scheme    |
| `C:\dir\log4j2.xml` (UNIX)    | `file:/<cwd>/C:%5Cdir%5Clog4j2.xml` | `C%3A%5Cdir%5Clog4j2.xml`           | relative path stays scheme-less |

Summarized:

1. **URL repair no longer drops components.** Blanks in an `http(s)`/`ftp`/VFS URL are encoded while user-info, port and query are kept.
2. **An absolute UNIX path resolves to a `file:` URI**, matching the existing Windows behavior, so absolute paths are symmetric across platforms.
3. **A Windows drive letter is never a URI scheme.** `C:/...` (and `C:\...`) resolve to a `file:` URI on Windows (the drive) and to a scheme-less relative URI on UNIX (the colon escaped), instead of a `scheme=C` URI or a working-directory-resolved `file:` URI.

These are corner cases (remote URLs with blanks, drive-letter paths used on the wrong OS, absolute UNIX config paths).
Common locations (`log4j2.xml`, `config/log4j2.xml`, `classpath:...`, `file:...`, a plain absolute Windows path) are unaffected.

## Tests

`NetUtilsTest` is reworked into parameterized cases.
Every input runs on every OS; only the expected `toString()` differs, selected through `SystemUtils.IS_OS_WINDOWS`.
The commits are ordered tests-first:

- the first commit adds the tests (failing against the old behavior, which is how the table above was produced),
- the second commit applies the fix.

## Follow-up

A `src/changelog/.2.x.x/*.xml` entry still needs to be added for these
behavioral changes before merge.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/NetUtilsTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/NetUtilsTest.java
@@ -18,67 +18,101 @@ package org.apache.logging.log4j.core.util;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assumptions.assumeThat;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.params.provider.Arguments.arguments;
 
-import java.io.File;
 import java.net.InetAddress;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.stream.Stream;
+import org.apache.commons.lang3.SystemUtils;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledOnOs;
-import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
 
 class NetUtilsTest {
 
-    @Test
-    void testToUriWithoutBackslashes() {
-        final String config = "file:///path/to/something/on/unix";
-        URI uri = NetUtils.toURI(config);
+    /**
+     * {@link NetUtils#toURI(String)} over a mix of URIs and file system paths.
+     *
+     * <p>Every input is exercised on every OS. Only the expected result differs.</p>
+     */
+    @ParameterizedTest
+    @MethodSource
+    void toUriConvertsLocationsToUris(final String input, final String expected) {
+        assertThat(NetUtils.toURI(input)).hasToString(expected);
+    }
 
-        assertNotNull(uri, "The URI should not be null.");
-        assertEquals("file:///path/to/something/on/unix", uri.toString(), "The URI is not correct.");
+    static Stream<Arguments> toUriConvertsLocationsToUris() {
+        return Stream.of(
+                // Genuine absolute URIs are returned verbatim on every platform.
+                arguments("file:///path/to/something/on/unix", "file:///path/to/something/on/unix"),
+                arguments("classpath:log4j2.xml", "classpath:log4j2.xml"),
+                arguments("classloader:log4j2.xml", "classloader:log4j2.xml"),
+                arguments("vfsfile:/some/path/log4j2.xml", "vfsfile:/some/path/log4j2.xml"),
+                // Well-formed layered Apache Commons VFS URLs are valid (opaque) URIs, returned verbatim.
+                arguments("zip:file:///path/a.zip!/dir/log4j2.xml", "zip:file:///path/a.zip!/dir/log4j2.xml"),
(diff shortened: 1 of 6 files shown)


ramanathan1504:
Hi @ppkarwasz,

I have tested these changes locally and verified that the behavior works as expected.

I do have a quick question regarding some of the inline comments (for example, // 1. A genuine absolute URI...) inside the utility functions. Since @vy has previously suggested in other pull requests to avoid inline comments when the code is self-documenting, should we remove or simplify them here, or is it preferred to keep them?

Thanks!

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/NetUtilsTest.java
vy, line 1: @ppkarwasz, shall we exercise more patterns taking following combinations into account?

- `//a/b/c` — Is this a path, or a URI where the authority component is `a`?
- `///a/b/c` — triple-slash
- `file:/C:/a/b` — single-slash & drive letter
- `http://example.com//a/b`
- `http://example.com///a/b`
- `http://example.com\c\a\b`
- `http://example.com/c\a\b`
- `http://example.com/c:\a\b`
- `http://example.com//c:\a\b`
- `http://example.com///c:\a\b`

These are inspirational, neither exhaustive, nor required. Some might even be invalid and throwing — maybe better to add a negative test case for those instead of discarding?

vy:
> I do have a quick question regarding some of the inline comments (for example, // 1. A genuine absolute URI...) inside the utility functions. Since @vy has previously suggested in other pull requests to avoid inline comments when the code is self-documenting, should we remove or simplify them here, or is it preferred to keep them?

@ramanathan1504, existing comments go beyond explaining self-explanatory code and do certainly add value:

```
// 2. Otherwise, the value is a file system path:
//
// - an absolute path becomes a `file:` URI holding the full path,
// - a relative one becomes a scheme-less URI, to be resolved later as a file or a class path resource.
```

@ppkarwasz, please keep 'em.

ramanathan1504:
> > I do have a quick question regarding some of the inline comments (for example, // 1. A genuine absolute URI...) inside the utility functions. Since @vy has previously suggested in other pull requests to avoid inline comments when the code is self-documenting, should we remove or simplify them here, or is it preferred to keep them?
> 
> @ramanathan1504, existing comments go beyond explaining self-explanatory code and do certainly add value:
> 
> ```
> // 2. Otherwise, the value is a file system path:
> //
> // - an absolute path becomes a `file:` URI holding the full path,
> // - a relative one becomes a scheme-less URI, to be resolved later as a file or a class path resource.
> ```
> 
> @ppkarwasz, please keep 'em.

understood 👍🏾

PR #4171, opened 2026-07-04
Add native tracing fields to LogEvent to eliminate async MDC overhead

### LINKED ISSUE #1976: Log events should have trace/span-related metadata
This idea is also related to a more general idea of supporting the [OpenTelemetry log data model](https://opentelemetry.io/docs/specs/otel/logs/data-model/), though the idea here would be to make standard log event attributes for [trace context metadata](https://www.w3.org/TR/trace-context/#trace-id) like the trace id, span id, and flags. This would make it easier for observability libraries to integrate with Log4j without having to store data in `ThreadContext` and all the performance overhead that implies.

ramanathan1504:
Fixes #1976 by introducing a TraceContextProvider SPI and adding
native W3C tracing fields (traceId, spanId, traceFlags) directly to LogEvent.

This allows frameworks like OpenTelemetry to bypass the ThreadContext map
entirely, eliminating the map-cloning memory overhead previously
required during asynchronous logging.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/spi/TraceContextProvider.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/spi/TraceContextProvider.java
@@ -17,8 +17,19 @@
 package org.apache.logging.log4j.spi;
 
 /**
- * Service Provider Interface (SPI) for retrieving standard W3C tracing metadata from the active context.
- * Placed in the API module to allow third-party providers to compile against it with minimal dependencies.
+ * Service Provider Interface (SPI) for retrieving distributed tracing metadata (such as W3C Trace Context)
+ * from the active execution context.
+ * <p>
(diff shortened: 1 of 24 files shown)


rgoers (approved):
I see no reason not to do this. The implemenation looks pretty simple and clean. However, I don't see an integration with JsonTemplateLayout. Is one needed?  Without that how would the data get injected into ElasticSearch?

ramanathan1504:
Hi @rgoers, that is a great point. I hadn't included the JsonTemplateLayout integration yet, but it is definitely needed to support structured logging properly without falling back to intermediate pattern allocations.

I will open a new, follow-up PR with the native JTL resolvers (TraceIdResolver, SpanIdResolver, TraceFlagsResolver) and link it back to this one.

ramanathan1504:
Hi @rgoers, the follow-up PR for the native JsonTemplateLayout resolvers is now open here: https://github.com/apache/logging-log4j2/pull/4175 Thanks again for the guidance!

PR #4172, opened 2026-07-04
fix: null-check parseAppender result in log4j-1.2-api XmlConfiguration

SebTardif:
## Summary

In log4j-1.2-api `XmlConfiguration`, `parseAppender()` can return null. The code called `appender.getName()` unconditionally, causing an NPE on bad appender configuration.

## Fix

Skip null appenders after `parseAppender`.

## Testing

Logic-only guard; existing suite in CI.

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
@@ -774,6 +774,9 @@ public class XmlConfiguration extends Log4j1Configuration {
                     break;
                 case APPENDER_TAG:
                     final Appender appender = parseAppender(currentElement);
+                    if (appender == null) {
+                        break;
+                    }
                     appenderMap.put(appender.getName(), appender);
                     addAppender(AppenderAdapter.adapt(appender));
                     break;
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
@@ -775,8 +775,6 @@ public class XmlConfiguration extends Log4j1Configuration {
                 case APPENDER_TAG:
                     final Appender appender = parseAppender(currentElement);
                     if (appender == null) {
-                        // parseAppender already logs the creation failure; warn at this
-                        // site so the configuration skip is visible to users.
                         LOGGER.warn(
                                 "Could not create appender named [{}] of class [{}]; ignoring.",
                                 subst(currentElement.getAttribute(NAME_ATTR)),
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
@@ -775,9 +775,22 @@ public class XmlConfiguration extends Log4j1Configuration {
                 case APPENDER_TAG:
                     final Appender appender = parseAppender(currentElement);
                     if (appender == null) {
+                        // parseAppender already logs the creation failure; warn at this
+                        // site so the configuration skip is visible to users.
+                        LOGGER.warn(
+                                "Could not create appender named [{}] of class [{}]; ignoring.",
+                                subst(currentElement.getAttribute(NAME_ATTR)),
+                                subst(currentElement.getAttribute(CLASS_ATTR)));
                         break;
                     }
-                    appenderMap.put(appender.getName(), appender);
+                    final String appenderName = appender.getName();
+                    if (appenderName == null) {
+                        LOGGER.warn(
+                                "Appender of class [{}] has a null name; ignoring.",
+                                subst(currentElement.getAttribute(CLASS_ATTR)));
+                        break;
+                    }
+                    appenderMap.put(appenderName, appender);
                     addAppender(AppenderAdapter.adapt(appender));
                     break;
                 default:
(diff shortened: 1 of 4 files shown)


ramanathan1504:
@SebTardif, thanks for the fix.  we'll need a unit test reproducing the NPE and changelog also.

Also, could you check if the other parsers in this switch statement (parseLogger, etc.) need the exact same null check? Let's fix them all at once. Lastly, let's make sure we log a warning when appender == null so it doesn't fail silently for the user.

1. Unit Test --> add a test case with an invalid XML configuration that triggers this scenario to prevent future regressions.
2. Audit sibling methods: Please check the other tags in this parse() switch block (e.g., parseCategory, parseLogger). If they can also return null on invalid configurations, they need the exact same null-guard.
3. Logging: Does parseAppender() emit a warning via the StatusLogger when it fails? If not, please add a LOGGER.warn(...) inside your if block so we don't silently skip configuration errors.
4. Name check: Can appender.getName() ever be null here? If so, we should guard against that as well so map insertion doesn't fail.

SebTardif:
@ramanathan1504

> we'll need a unit test reproducing the NPE and changelog also.
>
> Also, could you check if the other parsers in this switch statement (parseLogger, etc.) need the exact same null check? ... log a warning when appender == null ... Can appender.getName() ever be null here?

Thanks for the review. Addressed in the latest commit:

1. **Unit test** `XmlConfigurationTest#testInvalidAppenderDoesNotNpe` with `log4j1-invalid-appender.xml` (bogus appender class). Without the guard, parse hits NPE on `appender.getName()` and the valid `list` appender never registers; with the guard both the NPE is avoided and `list` loads.
2. **Changelog** `src/changelog/.2.x.x/fix_xml_parse_appender_npe.xml`.
3. **Sibling audit** of the same `parse()` switch:
   - `parseCategory` / `parseRoot` are `void` and do not use a nullable return (no equivalent NPE).
   - Renderer / factory tags already only log warnings.
   - Only `APPENDER_TAG` called `parseAppender(...).getName()` without a null check.
4. **`LOGGER.warn`** when `parseAppender` returns null (includes name + class from the element).
5. **Null name guard** before `appenderMap.put` / `addAppender`.

Spotless clean; local green:
`./mvnw -pl log4j-1.2-api test -Dtest=XmlConfigurationTest#testInvalidAppenderDoesNotNpe`

log4j-1.2-api/src/test/java/org/apache/log4j/config/XmlConfigurationTest.java
```diff
@@ -99,6 +101,23 @@ void testXML() throws Exception {
         assertTrue(file.length() > 0, "File A2 is empty");
     }
 
+    /**
```
ramanathan1504: here too and rest of the file also

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
```diff
@@ -774,7 +774,23 @@ private void parse(Element element) {
                     break;
                 case APPENDER_TAG:
                     final Appender appender = parseAppender(currentElement);
-                    appenderMap.put(appender.getName(), appender);
+                    if (appender == null) {
+                        // parseAppender already logs the creation failure; warn at this
```
ramanathan1504: Can you remove these comments? mostly its not needed in test case and same time its not critical part to mention here right? or its really needed here?

src/changelog/.2.x.x/fix_xml_parse_appender_npe.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4172" link="https://github.com/apache/logging-log4j2/pull/4172"/>
```
ramanathan1504, line 6: need to add your `PR` number too.please refer other change log, if you have doubt

log4j-1.2-api/src/test/java/org/apache/log4j/config/XmlConfigurationTest.java
SebTardif: @ramanathan1504

> here too and rest of the file also

Done. Removed the test javadoc and inline comments on `testInvalidAppenderDoesNotNpe`.

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
SebTardif: @ramanathan1504

> Can you remove these comments?

Done. Removed the explanatory comments around the `appender == null` guard in `XmlConfiguration`.

src/changelog/.2.x.x/fix_xml_parse_appender_npe.xml
SebTardif: @ramanathan1504

> need to add your `PR` number too. please refer other change log

The changelog entry already includes the PR:

```xml
<issue id="4172" link="https://github.com/apache/logging-log4j2/pull/4172"/>
```

That matches other single-PR entries (for example #4125 / #4127). There is no separate issue number for this change. Happy to adjust the format if you prefer a different layout.

ramanathan1504: oh yes i thought it have some linked issue

ramanathan1504 (approved):
Thanks

PR #4180, opened 2026-07-11
Include free-form message when encoding StructuredDataMessage to XML

### LINKED ISSUE #4141: Serializing `StructuredDataMessage` to XML omits the `message` field
`StructuredDataMessage` implements `MultiFormatStringBuilderFormattable` and returns `XML` as one of its supported encodings. When `StructuredDataMessage` is serialized to XML, it omits the `message` provided at construction:

```
new StructuredDataMessage("an id", "a message", "a type");
```

gets encoded to XML as:

```
<StructuredData>
<type>a type</type>
<id>an id</id>
<Map>
</Map>
</StructuredData>
```

whereas, the following was expected:

```
<StructuredData>
<type>a type</type>
<id>an id</id>
<Map>
  <Entry key="message">foo</Entry>
</Map>
</StructuredData>
```

There is another problem: What if the user _also_ explicitly provides a `message` entry:

```
var sdm = new StructuredDataMessage("an id", "a message", "a type");
sdm.put("message", "foo");
```

kamilkrzywanski:
## What

Fixes #4141: `StructuredDataMessage` XML encoding omitted the free-form message supplied at construction.

## Changes

- Encode the constructor `message` field as a distinct `<message>` element in `asXml`, with the same XML escaping used for `type` and `id`
- Keep it separate from map entries so a user-supplied `put("message", ...)` remains an independent SD-PARAM
- Omit the element when the constructor message is `null`
- Add regression tests for inclusion, escaping, and the map-key conflict case

### Example

```java
new StructuredDataMessage("an id", "a message", "a type");
```

now encodes as:

```xml
<StructuredData>
<type>a type</type>
<id>an id</id>
<message>a message</message>
<Map>
</Map>
</StructuredData>
```

When both the constructor message and a map entry keyed `"message"` are set, both appear:

```xml
<message>a message</message>
...
  <Entry key="message">foo</Entry>
```

## Testing

```bash
mvn -pl log4j-api-test -am test -Dtest=StructuredDataMessageTest -Dsurefire.failIfNoSpecifiedTests=false
```

All `StructuredDataMessageTest` cases pass.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/message/StructuredDataMessageTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/message/StructuredDataMessageTest.java
@@ -111,8 +111,7 @@ class StructuredDataMessageTest {
         final String id = "i<&d>" + XmlFixture.TEXT;
         final String type = "t>yp<e&" + XmlFixture.TEXT;
         // null message: keep this test focused on id/type escaping (and covers omission of <message>)
-        final String actualXml =
-                new StructuredDataMessage(id, null, type).getFormattedMessage(new String[] {"XML"});
+        final String actualXml = new StructuredDataMessage(id, null, type).getFormattedMessage(new String[] {"XML"});
         final String expectedXml = "<StructuredData>\n"
                 + "<type>t&gt;yp&lt;e&amp;" + XmlFixture.ENCODED_TEXT
                 + "</type>\n"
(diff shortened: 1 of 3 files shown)


ramanathan1504:
@kamilkrzywanski  Thanks for the `Pr`.  Give some time i will review this ASAP

ramanathan1504:
@kamilkrzywanski 

Tests look great and passed locally.
I noticed the Spotless formatting check failed, so I just pushed a quick commit to your branch to apply the formatting. For future `PRs`, just make sure to check out the `BUILDING.adoc` before committing so you can run the Spotless checks locally.

LGTM 👍🏾

ramanathan1504 (approved):

PR #4182, opened 2026-07-14
Add developer information in Pom

ramanathan1504:
Add developer information in Ramanathan


ppkarwasz (approved):
LGTM!

However look also at apache/logging-parent#490: we should probably remove the `developers` section entirely and rely on Git history.

PR #4184, opened 2026-07-20
Fix italic and underline ANSI escape codes

### LINKED ISSUE #4105: Missing ANSI Italic Style and Rendering Error in Underline Style
## Description

After upgrading to version >= 2.25.0, when rendering ansi, `italic` styling on log message no longer works (triggers a warning that the style italic is invalid) while `underline` styling on log message produces italic effect on rendered text. The warning message includes a stack trace that points to `AnsiEscape.createSequence` failing to create sequence for style name `italic` at the line below in `JAnsiTextRenderer`.

https://github.com/apache/logging-log4j2/blob/76c8bef88a5af15b09420445143e652f4696cab7/log4j-core/src/main/java/org/apache/logging/log4j/core/pattern/JAnsiTextRenderer.java#L251

This seems to boil down to https://github.com/apache/logging-log4j2/pull/3070, which rewrites `JAnsiTextRenderer` to use `AnsiEscape` instead of depending on external Jansi library to generate ansi escape sequences. However, the enum `AnsiEscape` differs from the enums in the old external Jansi library in that:
* `AnsiEscape` does not have the `ITALIC` entry.
* `AnsiEscape.UNDERLINE` has the code value of `3` which correspond to `ITALIC`.
* Code value `4` which corresponds to `underline` style seems to not be assigned to any enum value in `AnsiEscape`. 

(Style code referenced from https://en.wikipedia.org/wiki/ANSI_escape_code#Select_Graphic_Rendition_parameters)

I'm wondering if this was a bug that existed in `AnsiEscape` from a long time ago that was just never noticed. The enum values from this file seems to have been unchanged for a long time. Thank you for your assistant with this issue.

## Configuration

**Version:** >= 2.25.0

**Operating system:** Windows 11

**JDK:** 17.0.16

## Logs

```
// to be supplied
[Stacktraces, errors, etc. relevant applications logs.]
```

## Reproduction

Will supplement this at a later time.

kamilkrzywanski:
Fixes #4105.

`AnsiEscape` was missing `italic` (SGR 3) and had `underline` mapped to 3 instead of 4. After the Jansi removal, that made `italic` warn as invalid and `underline` render as italic.

## Changes

- Add `ITALIC("3")` and correct `UNDERLINE` to `"4"`
- Document `italic` in the pattern layout style keywords
- Add regression tests with hard-coded SGR sequences
- Changelog entry

## Checklist

* `./mvnw verify` passed
* Tests are provided

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/AnsiEscapeTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/AnsiEscapeTest.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.pattern;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression tests for ANSI SGR attribute codes (GitHub issue #4105).
+ * <p>
+ * Expected values are hard-coded SGR sequences so a wrong mapping cannot
+ * compare equal to itself.
+ * </p>
+ */
+class AnsiEscapeTest {
+
+    @Test
+    void italicIsSgr3() {
+        assertEquals("\u001B[3m", AnsiEscape.createSequence("italic"));
+    }
+
+    @Test
+    void underlineIsSgr4() {
+        assertEquals("\u001B[4m", AnsiEscape.createSequence("underline"));
+    }
+}
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/AnsiEscapeTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/AnsiEscapeTest.java
@@ -18,7 +18,8 @@ package org.apache.logging.log4j.core.pattern;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
 
 /**
  * Regression tests for ANSI SGR attribute codes (GitHub issue #4105).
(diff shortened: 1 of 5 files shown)


ramanathan1504:
@kamilkrzywanski, thanks! I'll try to get this cleared whenever I find the time. Thanks for your patience.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/AnsiEscapeTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression tests for ANSI SGR attribute codes (GitHub issue #4105).
+ * <p>
+ * Expected values are hard-coded SGR sequences so a wrong mapping cannot
+ * compare equal to itself.
+ * </p>
+ */
+class AnsiEscapeTest {
```
ramanathan1504, line 31: Can u extend the test case for remaining format also?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/JAnsiTextRendererTest.java
```diff
@@ -39,6 +39,11 @@ public static Stream<Arguments> testRendering() {
                         "",
                         "@|white key|@ = @|cyan,bold some value|@",
                         "\u001b[37mkey\u001b[m = \u001b[36;1msome value\u001b[m"),
+                // GitHub issue #4105: italic = SGR 3, underline = SGR 4
```
ramanathan1504: not needed remove the comments

log4j-core-test/src/test/java/org/apache/logging/log4j/core/pattern/JAnsiTextRendererTest.java
```diff
@@ -39,6 +39,11 @@ public static Stream<Arguments> testRendering() {
...
+                Arguments.of(
+                        "",
+                        "@|italic italic text|@ and @|underline underlined text|@",
+                        "\u001b[3mitalic text\u001b[m and \u001b[4munderlined text\u001b[m"),
                 // Return broken escapes as is
```
ramanathan1504, line 50: here also remove comments

ramanathan1504 (changes requested):
On the issue, @vy suspected `AnsiEscape` should also be "checked for other differences" against Jansi — that part is still open. Worth doing the full sweep: walk every constant in `org.fusesource.jansi.AnsiRenderer.Code` (the enum `JAnsiTextRenderer.toCode` used before #3070) and check it resolves in `AnsiEscape`. `italic` isn't the only casualty — try `faint` or `blink_slow` and you'll see the pattern. A quick way to check a candidate: `AnsiEscape.createSequence(name)` returning a bare `ESC[m` means the name didn't resolve, so the style silently does nothing and you get a `The style attribute X is incorrect.` warning on the `StatusLogger`.

kamilkrzywanski:
@ramanathan1504 thanks for the review.

Did the full Jansi `AnsiRenderer.Code` sweep and added the missing names (`faint`, `blink_slow`, etc.). Extended the tests and removed those comments.

Should be ready for another look.

ramanathan1504 (commented):
@kamilkrzywanski 
1. `pattern-layout.adoc` picked up only part of the new list — `intensity_bold`, `intensity_faint`, `negative_on`/`negative_off`, `conceal_on`/`conceal_off`, `underline_off`, `blink_off` and `bg_default` all work but aren't in the keyword list. add them, or say the list isn't exhaustive?
2. when `AnsiEscapeTest` fails it prints `expected: <> but was: <>` — the escape chars are invisible. could you pass the style name as the assertion message, the way `JAnsiTextRendererTest` shows them as hex? and while you're in there, the `@CsvSource` covers 22 of the 43 names, the rest are free.
3. `main` still has `UNDERLINE("3")` and no `ITALIC` — would you do the port too? no changelog entry there, the `2.x` one covers both.

PR #4186, opened 2026-07-21
Add literal Context Data (MDC) filtering to JSON Template Layout

### LINKED ISSUE #4166: `JsonTemplateLayout`: add MDC key filtering to `ThreadContextMapResolver` and harden `EcsLayout.json`
elastic/ecs-logging-java#404, released in v1.8.0, introduced a hardening measure: MDC entries whose keys match core ECS field names (`@timestamp`, `message`, `log.logger`, `log.level`, `event.dataset`, `process.thread.name`, `process.thread.id`, `ecs.version`) are silently dropped before serialization. This prevents user- or library-controlled MDC values from overriding the trusted fields written by the logger itself.

The `log4j-layout-template-json` module ships its own `EcsLayout.json` template. It currently provides no equivalent protection and is therefore not aligned with the upstream hardening.

### Proposed work

1. Add an optional `keyExcludes` configuration parameter to `ThreadContextMapResolver`. It should accept a list of literal key names to drop silently during serialization. A pattern-based extension (glob or regex) can be considered separately; a string list is sufficient for the initial use case.

   Example resolver snippet in a user-defined template:
   ```json
   {
     "$resolver": "mdc",
     "flatten": true,
     "keyExcludes": ["@timestamp", "message", "log.logger"]
   }
   ```

2. Update `EcsLayout.json` to use `keyExcludes` with the same reserved-key set introduced in `ecs-logging-java` v1.8.0, so the built-in ECS template provides the same field integrity guarantees out of the box.

**References**:
- `EcsJsonSerializer.RESERVED_KEYS` in `ecs-logging-core`

ramanathan1504:
Add `allowed`/`disallowed` literal key filters to JSON Template Layout's (Context Data) MDC and Map resolvers.

Update `EcsLayout.json` to prevent MDC entries from overwriting structural ECS fields.

Fixes #4166.


log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/ReadOnlyStringMapResolver.java
```diff
@@ -345,6 +356,9 @@ private enum LoopMethod implements TriConsumer<String, Object, LoopContext> {
 
         @Override
         public void accept(final String key, final Object value, final LoopContext loopContext) {
+            if (loopContext.keyExcludes != null && loopContext.keyExcludes.contains(key)) {
```
ppkarwasz: What about requiring `loopContext.keyExcludes` to be not null, so we don't need a null-check? The performance should be similar.

ppkarwasz (changes requested):
Hi @ramanathan1504,

Looks good to me, with two small comments:

log4j-layout-template-json/src/main/java/org/apache/logging/log4j/layout/template/json/resolver/ReadOnlyStringMapResolver.java
ramanathan1504: are you describing using Collections.emptySet() to avoid the != null check? I can make that change.

ppkarwasz: Yes, this should not impact performance and would be easier to understand by developers. You can even add a JSpecify `@NonNull` annotation on the field, so it is clear it can not be `null`.

ramanathan1504: @ppkarwasz, 
fixes 06399c8

log4j-perf-test/src/main/java/org/apache/logging/log4j/perf/jmh/MdcKeyFilterBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 1)
+@Fork(1)
+@State(Scope.Benchmark)
+public class MdcKeyFilterBenchmark {
```
ppkarwasz, line 42: We probably should compare the performance of two `ReadOnlyStringMapResolver` instances, instead of comparing `Pattern` vs `Map`. This might give us an idea if the changes are worth it.

`Map` is clearly faster than `Pattern`, but the relative performance gain can not be determined, since it lacks the time needed to format the map as JSON.

ramanathan1504:
```text
Benchmark                                       Mode  Cnt  Score   Error   Units
MdcKeyFilterBenchmark.testKeyExcludesResolver  thrpt    5  5.049 ± 0.175  ops/us
MdcKeyFilterBenchmark.testPatternResolver      thrpt    5  1.846 ± 0.023  ops/us
```
@ppkarwasz  5582862

ramanathan1504:
@vy done — `keyExcludes` is gone, it's `literal` with `allowed` and `disallowed` now, and the grammar block in the `ReadOnlyStringMapResolver` javadoc is updated to match the one u pasted.

Two things your example didn't pin down: `disallowed` currently wins over `allowed`, and when `pattern` is also set a key has to satisfy both — is that what u meant, or should `allowed` widen the pattern instead?

ramanathan1504:
@ppkarwasz reworked to the `literal` grammar @vy proposed on the issue, both your comments are still in — the sets are `Collections.emptySet()` when absent with `@NonNull` on the fields, and the benchmark compares two full layouts: `4.562` vs `1.833 ops/us`.

A mis-shaped `literal` now throws the way `flatten` does, but a typo'd sub-key (`disalowed`) still parses and silently filters nothing — worth rejecting unknown keys under `literal`, or leave it as is?

src/changelog/.2.x.x/4166_add_mdc_literal_key_filter.xml
```diff
@@ -0,0 +1,13 @@
...
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="added">
+    <issue id="4166" link="https://github.com/apache/logging-log4j2/issues/4166"/>
+    <issue id="4186" link="https://github.com/apache/logging-log4j2/pull/4186"/>
+    <description format="asciidoc">
+        Add `literal` key filtering to MDC and Map resolvers, and harden `EcsLayout.json` with it.
```
vy:
```suggestion
        Add `literal` key filtering to JSON Template Layout's Context Data (MDC) and Map resolvers.
        Harden `EcsLayout.json` using that.
```

vy (approved):

log4j-layout-template-json/src/main/resources/EcsLayout.json
```diff
@@ -26,7 +26,19 @@
...
+        "@timestamp",
+        "message",
+        "log.logger",
+        "log.level",
+        "event.dataset",
+        "process.thread.name",
+        "process.thread.id",
+        "ecs.version"
+      ]
+    }
```
ppkarwasz: Code looks good, but the naming for the configuration option seems odd. What about `filter` or `keyFilter`? The word `literal` doesn't tell me much, what this could do.

```json
{
  "filter": {
    "includes": [
      "@timestamp",
      ...
    ]
  }
}
```

ppkarwasz (approved):

log4j-layout-template-json/src/main/resources/EcsLayout.json
vy: I'd not use the word `filter`, it doesn't match what we already have: `key` and `pattern`, which both "filter".

@ramanathan1504, we had an internal conversation with @ppkarwasz, and [agreed] to overload `key` instead. That is, if `key` is of type string, use the existing behavior, if it is of type object, use the new semantics with `allowed`/`disallowed`. Would you mind updating the PR in that direction, please?

[agreed]: https://the-asf.slack.com/archives/D0394M6GWTX/p1788351977612879?thread_ts=1788344568.252929&cid=D0394M6GWTX

ramanathan1504: @vy [de4101c](https://github.com/apache/logging-log4j2/pull/4186/commits/de4101cddc6a34cace8e095600fe6efe273c764c)  `literal` is gone, `key` takes a string (unchanged) or an object with `allowed`/`disallowed`. The object form combines with `pattern`/`flatten` and `disallowed` wins over `allowed`. tell me if u meant `allowed` to widen the pattern.

ppkarwasz (approved):

PR #4198, opened 2026-07-23
resolve external entities through ConfigurationSource in log4j 1.x XmlConfiguration

jmestwa-coder:
The Log4j 1.x `XmlConfiguration` reader resolves external XML entities with the parser's default resolution.

- `Log4jEntityResolver` only handles `log4j.dtd` and returns `null` otherwise, so the parser falls back to default resolution and fetches external entities with no restrictions
- a legacy `log4j.xml` with a `SYSTEM` entity can read local files or reach any network URL

Fix, per review:

- DTD validation stays enabled as before
- `Log4jEntityResolver` still serves the bundled `log4j.dtd`; any other external resource is resolved through `ConfigurationSource.fromUri`, the same way the configuration file itself is, so it honors `log4j2.Configuration.allowedProtocols` (same route as #4161)
- resources that cannot be resolved that way get an empty source and a status warning, so existing configurations keep working and discouraged protocols need an explicit opt-in

Tests: a `file` entity is still resolved; an `http` entity is ignored with `file` as the only allowed protocol.

## Checklist

- [x] Base your changes on `2.x` branch
- [x] `./mvnw verify` succeeds
- [x] Non-trivial changes contain an entry file in the `src/changelog/.2.x.x` directory
- [x] Tests are provided

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
@@ -18,7 +18,10 @@ package org.apache.log4j.xml;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.core.config.ConfigurationSource;
 import org.apache.logging.log4j.status.StatusLogger;
 import org.apache.logging.log4j.util.Constants;
 import org.xml.sax.EntityResolver;
@@ -27,9 +30,11 @@ import org.xml.sax.InputSource;
 /**
  * An {@link EntityResolver} specifically designed to return
  * <code>log4j.dtd</code> which is embedded within the log4j jar
- * file. Any other external resource resolves to an empty source, so the
- * parser never fetches external entities from the file system or the
- * network.
+ * file. Any other external resource is resolved through
+ * {@link ConfigurationSource#fromUri(URI)}, the same way the configuration
+ * file itself is resolved, so it is subject to the same restrictions (e.g.
+ * the {@code log4j2.Configuration.allowedProtocols} property). Resources
+ * that cannot be resolved this way are replaced by an empty source.
  */
 public class Log4jEntityResolver implements EntityResolver {
     private static final Logger LOGGER = StatusLogger.getLogger();
@@ -37,7 +42,7 @@ public class Log4jEntityResolver implements EntityResolver {
 
     @Override
     public InputSource resolveEntity(final String publicId, final String systemId) {
-        if (systemId.endsWith("log4j.dtd") || PUBLIC_ID.equals(publicId)) {
+        if ((systemId != null && systemId.endsWith("log4j.dtd")) || PUBLIC_ID.equals(publicId)) {
             final Class<?> clazz = getClass();
             InputStream in = clazz.getResourceAsStream("/org/apache/log4j/xml/log4j.dtd");
             if (in == null) {
@@ -48,7 +53,21 @@ public class Log4jEntityResolver implements EntityResolver {
             }
             return new InputSource(in);
         }
-        LOGGER.warn("Ignoring external resource with public ID [{}] and system ID [{}].", publicId, systemId);
+        // Resolve using `ConfigurationSource`
+        if (systemId != null) {
+            try {
+                final ConfigurationSource source = ConfigurationSource.fromUri(new URI(systemId));
+                if (source != null) {
+                    final InputSource inputSource = new InputSource(source.getInputStream());
+                    inputSource.setSystemId(systemId);
+                    return inputSource;
+                }
+            } catch (final URISyntaxException e) {
+                LOGGER.warn("Resolution error: invalid URI {}", systemId, e);
+            }
+        }
+        // Fall back to empty resource
+        LOGGER.warn("Resolution error: unable to resolve {}", systemId);
         return new InputSource(new ByteArrayInputStream(Constants.EMPTY_BYTE_ARRAY));
     }
 }
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
@@ -27,7 +27,9 @@ import org.xml.sax.InputSource;
 /**
  * An {@link EntityResolver} specifically designed to return
  * <code>log4j.dtd</code> which is embedded within the log4j jar
- * file.
+ * file. Any other external resource resolves to an empty source, so the
+ * parser never fetches external entities from the file system or the
+ * network.
  */
 public class Log4jEntityResolver implements EntityResolver {
     private static final Logger LOGGER = StatusLogger.getLogger();
@@ -46,6 +48,7 @@ public class Log4jEntityResolver implements EntityResolver {
             }
             return new InputSource(in);
         }
-        return null;
+        LOGGER.warn("Ignoring external resource with public ID [{}] and system ID [{}].", publicId, systemId);
+        return new InputSource(new ByteArrayInputStream(Constants.EMPTY_BYTE_ARRAY));
     }
 }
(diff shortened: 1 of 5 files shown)


ramanathan1504:
@jmestwa-coder Thanks for the PR! I appreciate your patience. I'll try to get it closed ASAP.

ramanathan1504:
@vy @ppkarwasz could one of you take a look? I did rather hand this over than guess on a security related things.

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
```diff
@@ -180,6 +181,27 @@ private void doConfigure(final ParseAction action) throws FactoryConfigurationEr
         }
     }
 
+    private static void disableDtdProcessing(final DocumentBuilderFactory factory) {
+        factory.setValidating(false);
+        factory.setExpandEntityReferences(false);
+        setFeature(factory, "http://xml.org/sax/features/external-general-entities", false);
+        setFeature(factory, "http://xml.org/sax/features/external-parameter-entities", false);
+        setFeature(factory, "http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```
ppkarwasz: As the name suggests, this setting applies to **non-validating** parsers. Resolving the external subset is **required** in validating parsers and can not be turned off.

ppkarwasz:
Hi @jmestwa-coder,

Thank you for working on this.

I do not think the proposed parser features are the right approach for Log4j 1 XML configuration files.

1. These features are implementation-specific rather than required by JAXP. Xerces and the JDK’s Xerces fork support them, but other JAXP providers are not required to do so. Silently swallowing `ParserConfigurationException` would therefore produce provider-dependent, best-effort hardening: some parsers would be restricted, while others would continue resolving external resources.

2. As noted above, `http://apache.org/xml/features/nonvalidating/load-external-dtd` only applies to non-validating parsers. It does not prevent a validating parser from loading the external subset.

3. The parser already uses an `EntityResolver`. JAXP parsers are **required** to resolve external subsets and entities through that resolver, so we can enforce a provider-independent policy there: resolve the bundled `log4j.dtd` normally and return an empty source for every other external resource.

4. We also have the incubating [`commons-xml` project](https://github.com/apache/commons-xml), whose purpose is exactly to centralize this kind of XML parser hardening instead of requiring every project to reproduce, and often cargo-cult, the same collection of implementation-specific settings.

Log4j configuration files are considered [trusted, operator-controlled sources](https://logging.apache.org/security.html#threat-common-sources-configuration). Completely disabling external resources therefore provides limited additional protection, while potentially breaking existing configurations. The Log4j 1 compatibility layer needs to account for more than 25 years of configuration files deployed in real applications.

I therefore suggest that we:

* Keep DTD validation enabled, but constrain external resource resolution through `Log4jEntityResolver`. This would be similar to the approach proposed for `log4j-core` in #4161.
* Address the existing validation behavior separately. Validation is currently enabled, but validation errors are only written to the status logger and do **not** prevent the configuration from being used.

For the validation issue, we should consider one of the following approaches:

* Fix the bundled `log4j.dtd` and our example configurations so that all supported configurations validate without warnings.
* Make the bundled `log4j.dtd` stricter, so that it accepts only appenders and layouts that the Log4j 1 bridge can translate into Log4j 2 Core equivalents.

Custom Log4j 1 appenders have worked again since `2.17.1`, but their use is strongly discouraged, so the second option may be worth considering carefully.

garydgregory:
I generally agree with @ppkarwasz. In my view, the primary consideration should be backward compatibility. When I use the Log4j 1 compatibility layer, it's because I need it, not because I want to. As a result, compatibility needs to be as robust and faithful as possible.

Over time, both my transitive dependencies and my own applications will (or should) complete their migration to Log4j 2. Until then, compatibility takes precedence.

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
jmestwa-coder: right, missed that this feature only applies to non-validating parsers. dropped the feature-based approach entirely in favor of the resolver.

jmestwa-coder:
makes sense, reworked it along those lines:

- reverted the parser features and `setValidating(false)`; dtd validation stays enabled as before
- `Log4jEntityResolver` still serves the bundled `log4j.dtd`, but now returns an empty source instead of `null` for anything else, so the policy is enforced through the resolver and doesn't depend on the JAXP provider
- it logs a status warning with the public/system id when it ignores a resource, so configs that relied on external entities get a visible hint instead of a silent behavior change
- test unchanged, full log4j-1.2-api suite passes

agreed the validation behavior is a separate topic, so I left the `log4j.dtd` cleanup/strictening out of this PR. commons-xml looks like the right long-term home for the generic hardening once it's usable.

jmestwa-coder:
gentle ping

ramanathan1504:
> gentle ping

@jmestwa-coder sorry for the delay, we already noted give some time. Our team was engaged in other log4j stuff.I will try to close this within a week.Thanks for your patience 🙏🏿

log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
```diff
@@ -46,6 +48,7 @@ public InputSource resolveEntity(final String publicId, final String systemId) {
             }
             return new InputSource(in);
         }
-        return null;
+        LOGGER.warn("Ignoring external resource with public ID [{}] and system ID [{}].", publicId, systemId);
+        return new InputSource(new ByteArrayInputStream(Constants.EMPTY_BYTE_ARRAY));
```
ppkarwasz: This goes in the right direction in my opinion, but the main question remains: why are we making this change? What do users gain by upgrading?

As far as I can tell, nothing. Configuration files are trusted data. An attacker who can modify a Log4j configuration file will almost certainly:

- have write access to the rest of the classpath as well,
- be able to do considerably more damage than an SSRF, e.g. instantiate any class on the classpath and call its setters.

Meanwhile, by Hyrum's law, some deployment out there **does** rely on fetching entities from an external file, and we would be breaking it. To silence a SAST warning? To correct a `@SuppressFBWarnings` justification I wrote some years ago?

I would rather take the same route as #4161: don't forbid the usage, gate it behind an additional configuration toggle. `ConfigurationSource.fromUri`/`fromResource` are the methods we use to fetch the configuration file in the first place, so restricting external entity fetches to what those methods allow seems reasonable:

```java
// Resolve using `ConfigurationSource`
if (systemId != null) {
    try {
        final ConfigurationSource source = ConfigurationSource.fromUri(new URI(systemId));
        if (source != null) {
            final InputSource inputSource = new InputSource(source.getInputStream());
            inputSource.setSystemId(systemId);
            return inputSource;
        }
    } catch (final URISyntaxException e) {
        LOGGER.warn("Resolution error: invalid URI {}", systemId, e);
    }
}
// Fall back to empty resource
LOGGER.warn("Resolution error: unable to resolve {}", systemId);
return new InputSource(new ByteArrayInputStream(Constants.EMPTY_BYTE_ARRAY));
```

This doesn't prevent anyone from keeping their current configuration. It only asks them to explicitly **opt in** to discouraged practices, such as fetching resources over `http`.

Are we recommending external entities in 2026? Certainly not. But we don't recommend running a Log4j 1 configuration on Log4j 2 Core, or writing a `log4j2.properties` file either, and both are still supported. That choice isn't ours to make.

ppkarwasz (changes requested):

log4j-1.2-api/src/main/java/org/apache/log4j/xml/Log4jEntityResolver.java
jmestwa-coder: fair point, agreed there's little to gain from forbidding it outright. switched to the #4161 route:

- `log4j.dtd` is still served from the jar
- everything else goes through `ConfigurationSource.fromUri`, so `file`/`https`/`jar` keep working as before and anything else needs an opt-in via `log4j2.Configuration.allowedProtocols`
- unresolvable resources fall back to an empty source with a status warning

updated the test to cover both sides (a `file` entity still resolves, an `http` one is ignored with `file` as the only allowed protocol), and renamed it and the PR accordingly since it's no longer about disabling anything.

ppkarwasz (approved):
LGTM

PR #4199, opened 2026-07-23
Honor ConfigurationFactory active state

### LINKED ISSUE #3896: Fix `ConfigurationFactory::isActive` overrides
`ConfigurationFactory` implementations should override `isActive()` and throw `IllegalStateException` when they are used (e.g., via `getSupportedTypes()`) in inactive state. This change sufficiently be tested too.

See [this comment](https://github.com/apache/logging-log4j2/pull/3839#issuecomment-3211804379) on #3839 for details.

No `3.x` port is needed – `isActive()` does not exist there.

DragonFSKY:
Fixes #3896.

This change makes `ConfigurationFactory` active-state handling consistent:

* Factories expose their supported types regardless of active state.
* Inactive factories return `null`, and aggregate lookup does not invoke them to create configurations.
* Default resource discovery preserves the existing file-specific diagnostic for matching inactive factories and continues fallback lookup.
* JSON and YAML factories honor overridden `isActive()` implementations, while Log4j 1.x compatibility factories track the compatibility property dynamically.

Tests cover aggregate lookup paths, inactive-factory diagnostics and fallback, dynamic activation, overridden active state, and missing JSON/YAML dependencies. Existing Log4j 1.x configuration tests explicitly enable compatibility mode where required, and the affected OSGi package versions reflect the protected overrides.

## Checklist

* `./mvnw -pl log4j-core-test,log4j-1.2-api -am verify` succeeds ([the build instructions](https://logging.apache.org/log4j/2.x/development.html#building))
* Tests are provided


ramanathan1504:
@DragonFSKY I've just taken a quick look at your PR and noticed there are multiple new test cases. Since it touches multiple files, please give us a little time. We will review it ASAP!

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -616,12 +628,6 @@ private Configuration getConfiguration(
 
                     final ConfigurationSource source = ConfigurationSource.fromResource(configName, loader);
                     if (source != null) {
-                        if (!factory.isActive()) {
```
ramanathan1504, line 619: "Found configuration file {} for the inactive {} ... might be inactive due to a missing dependency" ERROR from ConfigurationFactory. That line is commit      
  5d47e931aa by @vy, which raised it from WARN to ERROR and added the missing-dependency wording. It only became unreachable because inactive factories are now skipped at the top  
  of the loop.

ramanathan1504 (changes requested):
Result: with `jackson-dataformat-yaml` absent, `getSupportedTypes()` still returns the YAML suffixes, so today a `log4j2.yaml` match logs that ERROR naming the file and the likely cause. After this change the factory is skipped before that point and nothing is logged. Please log before the `continue`.

DragonFSKY:
Thanks for catching this! I've fixed it. Please let me know if anything else needs adjusting.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
DragonFSKY: While reviewing the changes again, I noticed that my latest approach weakens the contract requested in #3896: an inactive factory is expected to reject `getSupportedTypes()`. The existing file-specific diagnostic can only be produced after querying the factory’s supported suffixes and finding a matching resource. Logging before the early `continue` would instead produce a factory-level message even when no matching configuration file exists. I have restored the strict inactive behavior for now. Could you please confirm whether this is the intended trade-off, or whether you would prefer another way to retain the file-specific diagnostic?

vy: AFAIU, what without this fix, when someone has a project with `log4j.json`, but without Jackson in the classpath, they were getting an ERROR-level log message stating that `log4j.json` could not be loaded due to a missing dependency. With this change, `log4j.json` will not be loaded *and* user will have no indication of this happened and why. I think this is not desirable from a user's pov. Shall we change the approach to the following:

- Correctly implement `isActive()` everywhere
- Make `getSupportedTypes()` work, always, regardless of the `isActive()` state
- Make `getConfiguration()` return null if inactive — this matches the existing behavior in `CF`
- Use `isActive` correctly & wisely in `CF` such that, if a `CF` is inactive, don't use it, but still warn the user if an inactive `CF` matches an existing configuration file

@ramanathan1504, @DragonFSKY, WDYT?

DragonFSKY: Thanks, this direction makes sense. I have updated the PR accordingly: supported types remain discoverable, inactive factories return `null`, and aggregate lookup does not invoke them to create configurations, while preserving the existing file-specific diagnostic and fallback behavior. Please let me know if anything else needs adjusting.

ramanathan1504:
@DragonFSKY don't use force push.it will completely clear the old review..Thanks

DragonFSKY:
Thanks for the reminder, and sorry about that. I'll avoid force-pushing and keep any further review updates as separate commits. Since I had already squashed the branch before seeing your note, would you prefer that I leave it as is, or split the current change into two commits for easier review? I understand that this would require one more force-push, so I won't change the history unless you prefer it.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java
```diff
@@ -621,8 +631,14 @@ private Configuration getConfiguration(
                                     "Found configuration file `{}` for the inactive `{}`. This `ConfigurationFactory` implementation might be inactive due to a missing dependency.",
```
ramanathan1504: With `log4j-1.2-api` on the classpath and `log4j1.compatibility` off, this fires for a plain `log4j.xml`, where nothing is missing — the message names a cause it cannot know.

````suggestion
                                    "Found configuration file `{}` for the inactive `{}`. This `ConfigurationFactory` implementation might be inactive due to a missing dependency or a disabled feature.",
````

File: log4j-core/src/main/java/org/apache/logging/log4j/core/config/ConfigurationFactory.java [hunk 3]
```diff
@@ -524,6 +528,9 @@ public Configuration getConfiguration(
                     if (types != null) {
                         for (final String type : types) {
                             if (type.equals(ALL_TYPES) || configLocationStr.endsWith(type)) {
+                                if (!factory.isActive()) {
```
ramanathan1504: Any reason this one sits inside the `for (type : types)` loop while the check at `:493` guards the factory loop? Here (and at `:594`, `:662`) `continue` skips a suffix rather than the factory, so `isActive()` runs once per supported suffix — for the 1.x factories that is a `PropertiesUtil` lookup each time.

src/changelog/.2.x.x/fix_configuration_factory_activation.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="changed">
+  <issue id="3896" link="https://github.com/apache/logging-log4j2/issues/3896"/>
```
ramanathan1504, line 6:
````suggestion
  <issue id="3896" link="https://github.com/apache/logging-log4j2/issues/3896"/>
  <issue id="4199" link="https://github.com/apache/logging-log4j2/pull/4199"/>
````

File: src/changelog/.2.x.x/fix_configuration_factory_activation.xml [hunk 5]
```diff
@@ -0,0 +1,11 @@
...
+  <description format="asciidoc">
+    Make `ConfigurationFactory` implementations consistently report active state and decline configuration creation
```
ramanathan1504:
````suggestion
    Skip inactive `ConfigurationFactory` implementations, while still reporting matching configuration files.
````

DragonFSKY:
Thanks for the review. I’ve moved the active-state checks to the factory loops and applied the suggested diagnostic and changelog updates.

ramanathan1504 (approved):
@DragonFSKY  Thanks for the contribution LGTM!

PR #4215, opened 2026-07-25
Pin snapshot reproducibility checks to the deployed artifacts

ppkarwasz:
Caller-side counterpart of apache/logging-parent#505: the snapshot reproducibility check now verifies against the repository artifact uploaded by the current run instead of the shared Nexus snapshot repository, which a concurrent run could overwrite. The release path keeps using the per-release staging repository URL.

> [!WARNING]
> Should be merged only after apache/logging-parent#505 lands and the `gha/v0` tag is updated.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -83,13 +83,21 @@ jobs:
         17
       project-id: log4j
 
-  verify-reproducibility:
-    needs: [ deploy-snapshot, deploy-release ]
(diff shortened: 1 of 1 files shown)


ramanathan1504 (approved):

ramanathan1504 (approved):
https://github.com/apache/logging-parent/pull/505

Successfully merged and all CI Pass

ppkarwasz:
Let's merge this and see if the check succeeds.

PR #4217, opened 2026-07-25
docs: Routing Appender security considerations for high-cardinality keys

### LINKED ISSUE #4181: Add security considerations to the RoutingAppender documentation
## Description
Currently `RoutingAppender` based on its configuration can materialize an unbounded number of routes. This coupled with a user controlled or high cardinality key can result in File descriptor exhaustion.

Example configuration which can be problematic:

```xml

<Configuration status="WARN">

    <Appenders>

        <Routing name="Routing">

            <Routes pattern="${ctx:userId}">

                <Route>

                    <File name="File-${ctx:userId}" fileName="logs/user-${ctx:userId}.log">

                        <PatternLayout pattern="%d %p %c{1.} [%t] %m%n"/>

                    </File>

                </Route>

            </Routes>

            <IdlePurgePolicy timeToLive="10" cleanupIntervalMillis="1000"/>

        </Routing>

            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>

        </Console>

    </Appenders>

    <Loggers>

        <Root level="info">

            <AppenderRef ref="Routing"/>

            <AppenderRef ref="Console"/>

        </Root>

    </Loggers>

</Configuration>
```

## Configuration

**Version:** 2.26.3

**Operating system:** Mac OS 15

**JDK:** openjdk64-25.0.3

## Logs
N/A

## Reproduction

log4j config:

```xml

<Configuration status="WARN">

    <Appenders>

        <Routing name="Routing">

            <Routes pattern="${ctx:userId}">

                <Route>

                    <File name="File-${ctx:userId}" fileName="logs/user-${ctx:userId}.log">

                        <PatternLayout pattern="%d %p %c{1.} [%t] %m%n"/>

                    </File>

                </Route>

            </Routes>

            <IdlePurgePolicy timeToLive="10" cleanupIntervalMillis="1000"/>

        </Routing>

            <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>

        </Console>

    </Appenders>

    <Loggers>

        <Root level="info">

            <AppenderRef ref="Routing"/>

            <AppenderRef ref="Console"/>

        </Root>

    </Loggers>

</Configuration>

```

Java Code to Trigger the bug:
```java

import org.apache.logging.log4j.ThreadContext;

import java.util.UUID;


public class Harness {

    private static final Logger logger = LogManager.getLogger(Harness.class);


        System.out.println("Starting DoS harness...");

        for (int i = 0; i < 1000000; i++) {

            String userId = UUID.randomUUID().toString();

            ThreadContext.put("userId", userId);

            [logger.info](http://logger.info/)("Log message for user: " + userId);

            if (i % 10000 == 0) {

                System.out.println("Logged " + i + " unique users...");

            }

        }

        System.out.println("Finished logging.");

    }

}
```
---
[Detected using Decypher](https://neuvem.io/decypher)

SebTardif:
## What Problem This Solves

When a default `Route` embeds an appender definition, the Routing Appender creates one subordinate appender per distinct routing key. If the key comes from untrusted or high-cardinality data (for example `${ctx:userId}`), that can exhaust file descriptors and memory.

## Evidence

Issue #4181 describes FD exhaustion with a user-controlled routing key and file-backed routes. This PR documents the risk and mitigations (`IdlePurgePolicy`, low-cardinality keys, prefer appender references).

## Summary

Adds a **Security considerations** section to the Routing Appender manual page, linked from the existing purge-policy discussion.

Fixes #4181

### Validation

Documentation-only change under `src/site/antora/.../appenders/delegating.adoc` plus changelog entry.


ramanathan1504:
The section only covers availability. The other half of what @ppkarwasz asked for
on #4181 is the threat model side: an untrusted key doesn't just create more appenders,
it ends up inside their config. I tried it on 2.26.1 and with
fileName="logs/${ctx:userId}.log" and a key of ../../../../tmp/x it writes /tmp/x.log
and doesn't report an error. One thing to watch: the key has to be a whole path segment
for that to happen. logs/user-${ctx:userId}.log just fails to open, so the example
matters. Could you also add the threat model link he mentioned
(https://logging.apache.org/security.html#threat-common-sources-configuration) and a
line on why Routing is different from appenders that get built at config time?


@ppkarwasz  can you look on this?

SebTardif:
Thanks @ramanathan1504. Pushed an update that:

- Splits the section into **Resource allocation** and **Threat model**
- Notes that default routes build subordinate appenders **at runtime**, so lookups can still carry untrusted data when attributes are resolved (unlike config-time appenders)
- Adds the path-segment `fileName="logs/${ctx:userId}.log"` example (and the contrast with `logs/user-${ctx:userId}.log`)
- Links the [configuration sources threat model](https://logging.apache.org/security.html#threat-common-sources-configuration) and the [path traversal FAQ](https://logging.apache.org/security/faq.html#path-traversal)
- Extends mitigations to path/URL/sink interpolation

Happy to adjust wording if @ppkarwasz wants a different emphasis.

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
```diff
@@ -945,6 +945,44 @@ If the `Route` element contains an appender definition, the appender will be ins
...
+[#RoutingAppender-security]
+=== Security considerations
+
+When a default <<Route,`Route`>> embeds an appender definition, the `Routing` Appender creates **one subordinate appender per distinct routing key value**.
+Unlike most appenders, which are fully built when the configuration is loaded, those subordinate appenders are created **at runtime** when a new key appears.
+Lookups used in the route (for example `${ctx:userId}`) can therefore still carry attacker-controlled data when appender attributes are resolved.
+
+Review two aspects of any dynamic routing configuration:
+
+==== Resource allocation
```
ramanathan1504, line 958:
```suggestion
[#RoutingAppender-security-resources]
==== Resource allocation

The `Routing` Appender is intentionally powerful: a separate log file per tenant or per long-running job, or round-robin routing between a fixed set of appenders, are all supported designs.
The responsibility that comes with that power is provisioning: the system must be able to allocate the resources required by every appender that can be created.
```

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
```diff
@@ -945,6 +945,44 @@ If the `Route` element contains an appender definition, the appender will be ins
...
+
+If the key is derived from untrusted or high-cardinality data (for example `${ctx:userId}`, a client IP, or a free-form request header), an attacker or a busy system can force creation of an unbounded number of appenders.
+
+That growth commonly leads to:
+
+* exhaustion of file descriptors (when each route opens a `File` or rolling file appender)
+* elevated memory use for appender state, buffers, and managers
+* difficulty shutting down or reconfiguring the application cleanly
+
+==== Threat model
```
ramanathan1504, line 972:
```suggestion
[#RoutingAppender-security-threat-model]
==== Threat model
```

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
```diff
@@ -945,6 +945,44 @@ If the `Route` element contains an appender definition, the appender will be ins
...
+
+An untrusted key is not only a resource problem: it is also substituted into the subordinate appender's configuration.
+Attributes such as `fileName` therefore inherit whatever the lookup returns.
+
+For example, with `fileName="logs/${ctx:userId}.log"`, a Thread Context value of `../../../../tmp/x` (as a whole path segment) can open `/tmp/x.log` without an error.
```
ramanathan1504, line 977:
```suggestion
Embedding the lookup inside a longer fixed segment, such as `logs/user-${ctx:userId}.log`, usually fails to open the file instead of escaping the directory, so the exact `fileName` pattern matters.
That failure is a side effect of how paths are resolved, not a mitigation: it is loud rather than safe, and it does not hold for every value a lookup can return.
```

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
```diff
@@ -945,6 +945,44 @@ If the `Route` element contains an appender definition, the appender will be ins
...
+Embedding the lookup inside a longer fixed segment, such as `logs/user-${ctx:userId}.log`, typically fails to open the file instead of escaping the directory, so the exact `fileName` pattern matters.
+
+This matches the project's {logging-services-url}/security.html#threat-common-sources-configuration[threat model for configuration sources]: operators are responsible for ensuring that appender configuration attributes come from trusted data.
+Only the application developer knows which Thread Context keys carry validated values and which are entirely attacker-controlled.
+See also {logging-services-url}/security/faq.html#path-traversal[path traversal in the security FAQ].
+
+==== Mitigations
```
ramanathan1504, line 989:
```suggestion
[#RoutingAppender-security-mitigations]
==== Mitigations

* Prefer the <<Route-attr-ref,`ref` attribute>> of a `Route`, pointing at a fixed, predeclared set of appenders, when the set of destinations is known.
* When dynamic routes are required, always configure a <<PurgePolicy,`PurgePolicy`>> (typically <<IdlePurgePolicy,`IdlePurgePolicy`>>) so idle route appenders are stopped and released.
Note that a purge policy bounds how long an unused appender survives, not the rate at which new ones are created.
* Constrain routing keys to a low-cardinality, validated domain (allow-lists, enums, hashed buckets) instead of raw user input.
* Avoid routing on pure user-controlled identifiers when each value would create a new file-backed appender, or when the key is interpolated into paths, URLs, or other sink configuration.
* Do not create an appender per web request. xref:manual/api.adoc#fish-tagging[Fish tagging] the events and filtering on the tag is a far better use of resources.
```

ramanathan1504 (changes requested):

ramanathan1504:
@ppkarwasz, Can you check the above. if its align with your suggestion. we will add this to next milestone.

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
SebTardif: @ramanathan1504

> [#RoutingAppender-security-resources]
> ==== Resource allocation
> The `Routing` Appender is intentionally powerful...

Applied in [`be7b167`](https://github.com/apache/logging-log4j2/pull/4217/commits/be7b16778fb6e159974347e0213bf6baf0b3d2a6).

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
SebTardif: @ramanathan1504

> [#RoutingAppender-security-threat-model]
> ==== Threat model

Applied in [`be7b167`](https://github.com/apache/logging-log4j2/pull/4217/commits/be7b16778fb6e159974347e0213bf6baf0b3d2a6).

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
SebTardif: @ramanathan1504

> usually fails to open the file instead of escaping the directory
> That failure is a side effect of how paths are resolved, not a mitigation

Applied in [`be7b167`](https://github.com/apache/logging-log4j2/pull/4217/commits/be7b16778fb6e159974347e0213bf6baf0b3d2a6).

src/site/antora/modules/ROOT/pages/manual/appenders/delegating.adoc
SebTardif: @ramanathan1504

> [#RoutingAppender-security-mitigations]
> Prefer the `ref` attribute of a `Route`
> Note that a purge policy bounds how long an unused appender survives
> Fish tagging

Applied in [`be7b167`](https://github.com/apache/logging-log4j2/pull/4217/commits/be7b16778fb6e159974347e0213bf6baf0b3d2a6). Also switched the first mitigation to `<<Route-attr-ref>>` (that is the existing `ref` attribute anchor).

ramanathan1504 (approved):

PR #4222, opened 2026-07-26
Add log4j-api-kotlin dependency to BOM

### LINKED ISSUE #4145: Add `log4j-api-kotlin` to the bill of materials (`log4j-bom`)
Currently, the latest version of `log4j-bom` does not include `log4j-api-kotlin`.
It would be convenient to add it.

My use case, as an example, is the following: I wanted to use it in a Spring Boot application and I am using their Gradle plugin, which automatically imports their BOM `spring-boot-dependencies`. This one, in turn, depends on `log4j-bom`. This allows me to add `log4j` modules to my Gradle dependencies without specifying a version, and it will resolve to the best matching version according to the BOMs. But it does not work because `log4j-bom` is missing the kotlin api as a dependency.

ramanathan1504:
This PR adds `log4j-api-kotlin` to the Log4j Bill of Materials (`log4j-bom` / root `pom.xml`). 

Fixes #4145.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -380,6 +380,8 @@
     <log4j.docgen.pluginDescriptorsDir.phase1>${maven.multiModuleProjectDirectory}/target/plugin-descriptors/phase1</log4j.docgen.pluginDescriptorsDir.phase1>
     <log4j.docgen.pluginDescriptorsDir.phase2>${maven.multiModuleProjectDirectory}/target/plugin-descriptors/phase2</log4j.docgen.pluginDescriptorsDir.phase2>
 
+    <!-- log4j-api-kotlin> -->
+    <log4j-api-kotlin.version>1.5.0</log4j-api-kotlin.version>
   </properties>
 
   <dependencyManagement>
@@ -495,6 +497,12 @@
         <groupId>org.apache.logging.log4j</groupId>
         <artifactId>log4j-jul</artifactId>
         <version>${project.version}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api-kotlin</artifactId>
+        <version>${log4j-api-kotlin.version}</version>
       </dependency>
 
       <dependency>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -497,6 +499,12 @@
         <version>${project.version}</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.apache.logging.log4j</groupId>
+        <artifactId>log4j-api-kotlin</artifactId>
```
vy: Note the comment beginning at the beginning of this `<dependencies>` block:

> List of managed dependencies (*in alphabetical order*) ...

Hence, `log4j-api-kotlin` should be placed after `log4j-api` and before `log4j-api-test`.

pom.xml
```diff
@@ -380,6 +380,8 @@
     <log4j.docgen.pluginDescriptorsDir.phase1>${maven.multiModuleProjectDirectory}/target/plugin-descriptors/phase1</log4j.docgen.pluginDescriptorsDir.phase1>
     <log4j.docgen.pluginDescriptorsDir.phase2>${maven.multiModuleProjectDirectory}/target/plugin-descriptors/phase2</log4j.docgen.pluginDescriptorsDir.phase2>
 
+    <!-- log4j-api-kotlin> -->
+    <log4j-api-kotlin.version>1.5.0</log4j-api-kotlin.version>
```
vy: This Maven property neither overrides an existing one, nor is used elsewhere but in one place. Hence, you better inline this value, no need for a property.

vy (changes requested):

vy:
@ramanathan1504, could you also do the following, please?

1. Make sure there is a short changelog entry
2. Make sure `/target/pom.xml` looks good<sup>1</sup> after `./mvnw install -N`

<sup>1</sup> This is the BOM. It should be free of any properties, dependencies, etc.

pom.xml
ramanathan1504: @vy fixes [787824e](https://github.com/apache/logging-log4j2/pull/4222/commits/787824e771573c5762209b9f8c7cdd92e27a620b)

ramanathan1504:
@vy Changelog entry added in the same commit [787824e](https://github.com/apache/logging-log4j2/pull/4222/commits/787824e771573c5762209b9f8c7cdd92e27a620b) and checked the flattened BOM: only coordinates,
  metadata and `<dependencyManagement>` — no properties, dependencies,
  modules, build or profiles, and no unresolved `${...}`.

vy (approved):

PR #4226, opened 2026-07-28
Fix startup delay and direct write file naming in CronTriggeringPolicy

tupelo-schneck:
Two defects in `CronTriggeringPolicy`, both hit by an appender configured without a `fileName`. Originally split across this PR and #4227; consolidated here at @ramanathan1504's request, and #4227 is closed in favour of this one.

Reproduced with:

```xml
<RollingFile name="errorLogAppender" filePattern="${LOG_DIR}/error.log-%d{yyyyMMdd}">
    <PatternLayout pattern="%m%n"/>
    <CronTriggeringPolicy schedule="0 0 0 ? * SUN" evaluateOnStartup="true"/>
</RollingFile>
```

---

## 1. Roughly three seconds of startup delay per appender

`initialize()` looks up the previous fire time relative to `RollingFileManager#getFileTime()`:

```java
final Date lastRollForFile = cronExpression.getPrevFireTime(new Date(this.manager.getFileTime()));
```

That value is 0 when there is no current file yet (`file == null || !file.exists() ? 0 : initialFileTime(file)`), which is the case on every startup of an appender that writes directly to the file its pattern resolves to. The lookup therefore runs against the epoch.

`CronExpression.getTimeBefore()` walks backwards one `findMinIncrement()` step at a time, calling `getTimeAfter()` until the result precedes the target. `getTimeAfter()` clamps its result to 1970, because an expression without a year field gets a year set starting there. For a target at the epoch the exit condition can therefore never be satisfied: every iteration returns the same 1970 fire time, which is neither `null`, nor before `MIN_DATE`, nor before the target. The search grinds back through several millennia of candidate dates until the calendar's era flips and `getTimeAfter()` bails out at its `YEAR > 2999` guard, returning `null`.

A stack sample, taken once a second, with the main thread pinned at 100% CPU throughout:

```
"main" #3 prio=5 os_prio=31 cpu=749.54ms elapsed=103.85s runnable
	at org.apache.logging.log4j.core.util.CronExpression.getTimeAfter(CronExpression.java:1201)
	at org.apache.logging.log4j.core.util.CronExpression.getTimeBefore(CronExpression.java:1584)
	at org.apache.logging.log4j.core.util.CronExpression.getPrevFireTime(CronExpression.java:1594)
	at org.apache.logging.log4j.core.appender.rolling.CronTriggeringPolicy.initialize(CronTriggeringPolicy.java:67)
	at org.apache.logging.log4j.core.appender.rolling.RollingFileManager.initialize(RollingFileManager.java:246)
	...
	at org.apache.logging.log4j.core.LoggerContext.start(LoggerContext.java:311)
```

Note that `evaluateOnStartup="false"` does not avoid the cost: the lookup runs unconditionally and only its result is guarded by that flag.

**Fix.** Bound the backward search at the epoch, and skip the lookup when there is no current file. Both paths already returned `null` here, so behaviour is unchanged and only the cost differs.

`getPrevFireTime(new Date(0))` for `0 0 0 ? * SUN`:

| | duration |
| --- | --- |
| before | 2854–3040 ms |
| after | 0.03–0.06 ms |

The two-appender configuration above goes from ~6.0 s to ~0.2 s.

The bound is the epoch rather than `MIN_DATE`, per @ramanathan1504's review. `MIN_CAL.set(1970, 0, 1)` leaves the time fields alone, so `MIN_DATE` carries the JVM's start time of day — two JVMs seconds apart print `09:40:09` and `09:40:17`. Bounding the candidate on it would make results depend on when the process started, and would lose real fire times just after the epoch: for `0 0 0 * * ?` at `1970-01-02 06:00`, `2.x` returns `1970-01-02 00:00` where a `MIN_DATE` bound returns `null`. A bound on the local 1970 instant in the expression's own zone was also tried, since `getTime() < 0` is a UTC-instant test while `getTimeAfter()` floors at local 1970; it gave identical results across `America/New_York`, `UTC` and `Australia/Sydney`, so the simpler form is used.

## 2. The direct write file is named after process start, not the rollover period

The file an appender writes to covers a whole rollover period, but it was named after the moment the appender started. Starting on Tuesday 2026-07-28 under the weekly schedule above writes to `error.log-20260728`, though the period began on Sunday 2026-07-26, so a restart on Wednesday opens `error.log-20260729` instead of continuing the same file. A week with daily restarts leaves seven files where the schedule implies one, and rollover only ever closes out whichever fragment is current.

`initialize()` records the period start correctly:

```java
final Date lastRegularRoll = cronExpression.getPrevFireTime(new Date());
aManager.getPatternProcessor().setCurrentFileTime(lastRegularRoll.getTime());
```

`DirectWriteRolloverStrategy.getCurrentFileName()` then discarded it:

```java
// LOG4J2-3339 - Always use the current time for new direct write files.
manager.getPatternProcessor().setCurrentFileTime(System.currentTimeMillis());
```

That override compensated for a different problem: `rollover()` passed `lastRollDate`, the *previous* period's start, as the new file's time, so each replacement file was named after the file just rolled. Substituting the current time hid that, and works only because at the instant of a rollover "now" is approximately the new period's start. At startup "now" is an arbitrary point inside the period, which is where it breaks.

**Fix.** Pass the roll time rather than the previous roll date, so the replacement file is named after the period it opens, and let `getCurrentFileName()` use the period start already recorded.

Blast radius is limited to policies that track a period. `PatternProcessor.formatFileName()` already falls back to the current time when the current file time is 0, and `updateTime()` resets it to 0, so `TimeBasedTriggeringPolicy`, which never sets it, is unchanged. Appenders configured *with* a `fileName` are unaffected, since `getCurrentFileName()` is only consulted on the direct write path.

**Compatibility.** This changes the names of direct write files for cron based appenders, from the process start time to the start of the rollover period — `error.log-20260726` rather than `error.log-20260728` above. Existing files are untouched, neither read nor renamed; a restart after upgrading simply begins writing to the period's file. The changelog entry calls this out.

`RollingAppenderDirectCronTest`, the regression test added with LOG4J2-3339, still passes: it asserts that a rolled file's name matches the timestamp of its first line, which this change preserves.

---

## Testing

Four tests, each confirmed to fail without its corresponding fix and pass with it:

* `CronExpressionTest#testPrevFireTimeAtEpochReturnsNullPromptly` — asserts `null` within a 1 s timeout, deliberately well below the ~2.9 s the unfixed path takes, since a generous timeout would pass with the defect present.
* `CronExpressionTest#testPrevFireTimeJustAfterEpochIsUnaffectedByBound` — pins the near-epoch behaviour above, in a fixed zone so it does not depend on the CI host.
* `CronTriggeringPolicyTest#testBuilderWithoutFileNameInitializesPromptly` — builds an appender without a `fileName` on a weekly schedule.
* `CronTriggeringPolicyTest#testDirectWriteFileNameUsesPeriodStart` — asserts the manager's file name matches the period start computed independently from the same expression. Without the fix: `expected: <target/testcmd5.log-20260726> but was: <target/testcmd5.log-20260728>`.

`./mvnw verify` passes across all 41 modules.

## Checklist

* [x] Based on `2.x`
* [x] `./mvnw verify` succeeds
* [x] Changelog entries in `src/changelog/.2.x.x`
* [x] Tests are provided


log4j-core/src/main/java/org/apache/logging/log4j/core/util/CronExpression.java
```diff
@@ -1581,6 +1581,13 @@ protected Date getTimeBefore(final Date targetDate) {
         Date prevFireTime;
         do {
             final Date prevCheckDate = new Date(start.getTime() - minIncrement);
+            // `getTimeAfter()` never returns a fire time before 1970, so once the candidate
+            // date precedes `MIN_DATE` the loop condition below can no longer be satisfied.
+            // Bail out here, otherwise the search walks back millennia one increment at a
+            // time before `getTimeAfter()` finally gives up at the calendar's upper bound.
+            if (prevCheckDate.before(MIN_DATE)) {
+                return null;
+            }
```
ramanathan1504: `MIN_CAL.set(1970, 0, 1)` never clears the time fields, so `MIN_DATE` carries the JVM's start time of day — I printed `12:01:08.151`, `12:03:05.302` and `12:01:39.611` across three runs.
That only mattered once this guard started bounding the *candidate*: with `0 0 0 * * ?`, `getPrevFireTime(1970-01-02 06:00)` gives `1970-01-02 00:00:00` on 2.x and `null` here. Bounding on the epoch is deterministic — verified green plus spotless.

````suggestion
            // `getTimeAfter()` never returns a fire time before 1970, so once the candidate
            // date precedes the epoch the loop condition below can no longer be satisfied.
            // Bail out here, otherwise the search walks back millennia one increment at a
            // time before `getTimeAfter()` finally gives up at the calendar's upper bound.
            if (prevCheckDate.getTime() < 0) {
                return null;
            }
````

File: log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/CronTriggeringPolicy.java [hunk 2]
```diff
@@ -64,7 +64,11 @@ private CronTriggeringPolicy(
     public void initialize(final RollingFileManager aManager) {
         this.manager = aManager;
         final Date now = new Date();
-        final Date lastRollForFile = cronExpression.getPrevFireTime(new Date(this.manager.getFileTime()));
+        // `RollingFileManager` reports a file time of 0 when there is no current file yet, which
+        // happens on every startup of an appender configured without a `fileName`. There is no
+        // previous roll to look up in that case, so skip the lookup entirely.
+        final long fileTime = this.manager.getFileTime();
```
ramanathan1504, line 67: `testBuilderWithoutFileNameInitializesPromptly` passes with either fix alone (0.134 s / 0.143 s) and fails only when both are reverted, so nothing pins this guard — is it load-bearing, or documentation like `TimeBasedTriggeringPolicy.initialize()`'s `getFileTime() == 0`?

ramanathan1504 (changes requested):
@tupelo-schneck  Thanks for the contribution. I seen the #4227 that's also relate to this, can You close the follow up `Pr` and work it out here?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/CronExpression.java
tupelo-schneck: Applied, thank you — this is a real defect in what I wrote, not a style preference.

I reproduced the nondeterminism: two JVMs started seconds apart print `MIN_DATE` as `09:40:09` and `09:40:17`, since `MIN_CAL.set(1970, 0, 1)` leaves the time fields alone. Your example reproduces exactly — for `0 0 0 * * ?` at `1970-01-02 06:00`, `2.x` returns `1970-01-02 00:00` where the `MIN_DATE` bound returned `null`. A correct answer silently replaced, and which answer you got depended on when the process started.

I also tried bounding on the local 1970 instant in the expression's own zone, on the theory that `getTime() < 0` is a UTC instant test while `getTimeAfter()` floors at local 1970. Across `America/New_York`, `UTC` and `Australia/Sydney` it produced results identical to the epoch bound on every case I tried, so the simpler form is the right one.

Added `testPrevFireTimeJustAfterEpochIsUnaffectedByBound` to pin this — it fails with the `MIN_DATE` bound and passes with the epoch, in a fixed zone so it does not depend on the CI host.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/CronTriggeringPolicy.java
tupelo-schneck: Documentation, not load-bearing — your second reading is right, and thank you for checking it rather than assuming.

I confirmed it: with the epoch bound in `CronExpression` in place and this guard reverted, `CronTriggeringPolicyTest` and `CronExpressionTest` pass in 0.18 s. Once the search is bounded, `getPrevFireTime(new Date(0))` returns `null` in microseconds for any expression, so nothing here depends on the guard for speed.

I have kept it, for the reason you name: it mirrors `TimeBasedTriggeringPolicy.initialize()`'s `getFileTime() == 0` check, and it makes `lastRollForFile` `null` by construction rather than by a lookup that happens to return `null` for a meaningless input. Happy to drop it if you would rather keep the diff minimal — no functional difference either way.

tupelo-schneck:
Thanks for reviewing both — consolidated as you asked.

#4227's change is now here as 79037d369, and I have closed #4227 pointing at this PR. The description is rewritten to cover both fixes, and the two changelog entries now both reference this one.

Briefly, on the second fix, since it arrived without its own discussion: a direct write file was being named after the moment the appender started rather than after the rollover period it covers, so a restart later in the period opened a second file for a period meant to have one. Under a weekly schedule with a `%d{yyyyMMdd}` pattern, a week of daily restarts leaves seven files instead of one. Fixing it partly unwinds LOG4J2-3339 and changes direct write file names for cron based appenders, so it is worth a look on its own terms — the reasoning and the compatibility note are in the description.

`./mvnw verify` passes on the combined branch across all 41 modules.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/DirectWriteRolloverStrategy.java
```diff
@@ -390,8 +390,10 @@ public String getCurrentFileName(final RollingFileManager manager) {
             final SortedMap<Integer, Path> eligibleFiles = getEligibleFiles(manager);
             final int fileIndex = eligibleFiles.size() > 0 ? (nextIndex > 0 ? nextIndex : eligibleFiles.lastKey()) : 1;
             final StringBuilder buf = new StringBuilder(255);
-            // LOG4J2-3339 - Always use the current time for new direct write files.
-            manager.getPatternProcessor().setCurrentFileTime(System.currentTimeMillis());
+            // Name the file after the start of the rollover period it belongs to, which the
+            // triggering policy records as the pattern processor's current file time. Policies
+            // that do not track a period leave that value at 0, and `formatFileName()` then
+            // falls back to the current time on its own.
```
ramanathan1504:
```suggestion
```
no need here

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/CronTriggeringPolicyTest.java
ramanathan1504, line 1: Remove Javadoc comments and inline comments

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/CronExpressionTest.java
ramanathan1504, line 1: Remove inline and Javadoc things

src/changelog/.2.x.x/fix_cron_triggering_policy_slow_startup.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
```
ramanathan1504:
```suggestion
       type="changed">
```

src/changelog/.2.x.x/fix_cron_triggering_policy_slow_startup.xml
```diff
@@ -0,0 +1,14 @@
...
+    <issue id="4226" link="https://github.com/apache/logging-log4j2/pull/4226"/>
+    <description format="asciidoc">
+        Fix a multi-second startup delay in `CronTriggeringPolicy` for appenders configured without a `fileName`.
+        `CronExpression.getPrevFireTime()` now bounds its backward search at 1970 instead of scanning millennia of
+        candidate dates before returning `null`.
```
ramanathan1504:
```suggestion
       Changes `CronTriggeringPolicy` startup delay and name its direct write files after the rollover period.
```

ramanathan1504 (commented):
@tupelo-schneck  can u remove the inline comments .

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/CronTriggeringPolicyTest.java
tupelo-schneck: Done

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/CronExpressionTest.java
tupelo-schneck: Done

ramanathan1504 (approved):
LGTM! Thanks @tupelo-schneck

PR #4229, opened 2026-07-29
do not apply POSIX file attributes through symbolic links

jmestwa-coder:
`PosixViewAttribute` applies permissions and ownership through symbolic links:

- `FileUtils.defineFilePosixAttributeView` looks the view up without `NOFOLLOW_LINKS`, so `setPermissions`/`setOwner`/`setGroup` land on the link target
- `walkFileTree` still hands symlinks to `visitFile`, so the `followLinks="false"` default documented on `AbstractPathAction` has no effect here
- a link planted in `basePath` whose name matches the `PathCondition` glob redirects the chmod/chown onto any file the process can reach
- `FileManager.defineAttributeView` goes through the same helper

Requesting the view with `NOFOLLOW_LINKS` makes the change fail on a link rather than hit its target, and the visitor now skips symlinks so a planted one does not abort the rollover action.

## Checklist

* Tests are provided


jmestwa-coder:
any update?

log4j-core/src/main/java/org/apache/logging/log4j/core/util/FileUtils.java
```diff
@@ -159,7 +164,8 @@ public static void defineFilePosixAttributeView(
             final String fileOwner,
             final String fileGroup)
             throws IOException {
-        final PosixFileAttributeView view = Files.getFileAttributeView(path, PosixFileAttributeView.class);
+        final PosixFileAttributeView view =
```
ramanathan1504:
```suggestion
        final PosixFileAttributeView view = Files.getFileAttributeView(path, PosixFileAttributeView.class);
```

ramanathan1504: This helper is also called by `FileManager:289` and, through `defineAttributeView`, by `RollingFileManager:409` and `RollingRandomAccessFileManager:270,333` — on the appender's own configured `fileName`, not on anything an attacker planted. If that name is a symlink I get `FileSystemException: Too many levels of symbolic links`, which `FileManager` swallows into `LOGGER.error("Could not define attribute view …")`, so permissions are silently not applied. Could this line stay as it was, and let the visitor guard do the work? I tried it locally and both `followLinks` modes then behave, with your `PosixViewAttributeActionTest` still passing. The `import java.nio.file.LinkOption;` added at the top can go with it.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/FileUtils.java
```diff
@@ -146,6 +147,10 @@ public static void makeParentDirs(final File file) throws IOException {
 
     /**
      * Define file POSIX attribute view on a path/file.
+     * <p>
+     * Symbolic links are never followed: if {@code path} is a link, the attributes of the link itself are
+     * modified and its target is left untouched.
+     * </p>
      *
      * @param path Target path
```
ramanathan1504, line 150: Permissions never land on the link — `setPermissions` throws, which is why the new `FileUtilsTest` has to swallow an `IOException`. Only owner and group do, via `lchown`. If the lookup goes back to following, this paragraph can go entirely.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeAction.java
```diff
@@ -369,6 +369,10 @@ protected FileVisitor<Path> createFileVisitor(final Path basePath, final List<Pa
         return new SimpleFileVisitor<Path>() {
             @Override
             public FileVisitResult visitFile(final Path file, final BasicFileAttributes attrs) throws IOException {
+                if (attrs.isSymbolicLink()) {
```
ramanathan1504, line 372: `followLinks="true"` is documented as supported — `manual/appenders/rolling-file.adoc:1033`, with its own security warning — but under `FOLLOW_LINKS` the attributes come from `stat`, so `attrs.isSymbolicLink()` is false and this guard never fires. Gating the skip on the action's own setting makes both modes work. `isFollowSymbolicLinks()` is already public on `AbstractPathAction:143`.


```suggestion
                if (!isFollowSymbolicLinks() && attrs.isSymbolicLink()) {
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/FileUtilsTest.java
```diff
@@ -87,6 +90,28 @@ void testFileFromUriWithSpacesAndPlusCharactersInName() throws Exception {
         assertTrue(file.exists(), "file exists");
     }
```
ramanathan1504, line 89:
```java
@Test
void testSymbolicLinksAreFollowedWhenConfigured(@TempDir final Path tempDir) throws Exception {
    final Path outsider = tempDir.resolve("outsider.txt");
    Files.write(outsider, "secret".getBytes(StandardCharsets.UTF_8));
    Files.setPosixFilePermissions(outsider, PosixFilePermissions.fromString("rw-------"));

    final Path baseDir = Files.createDirectory(tempDir.resolve("logs"));
    Files.createSymbolicLink(baseDir.resolve("app-2.log"), outsider);

    final Configuration config = new BasicConfigurationFactory().new BasicConfiguration();
    final PosixViewAttributeAction action = PosixViewAttributeAction.newBuilder()
            .setBasePath(baseDir.toString())
            .setFollowLinks(true)
            .setMaxDepth(1)
            .setPathConditions(PathCondition.EMPTY_ARRAY)
            .setConfiguration(config)
            .setFilePermissionsString("rw-rw-rw-")
            .build();

    action.execute();

    assertEquals(
            "rw-rw-rw-",
            PosixFilePermissions.toString(Files.getPosixFilePermissions(outsider)),
            "followLinks=\"true\" should still follow the link");
}
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/FileUtilsTest.java
```diff
@@ -87,6 +90,28 @@ void testFileFromUriWithSpacesAndPlusCharactersInName() throws Exception {
...
 
+    @Test
```
ramanathan1504: Deletion block for the old test:

```suggestion
```

ramanathan1504 (changes requested):

src/changelog/.2.x.x/fix_posix_view_attribute_symlink.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4229" link="https://github.com/apache/logging-log4j2/pull/4229"/>
+  <description format="asciidoc">Stop `PosixViewAttribute` from applying permissions and ownership through symbolic links</description>
```
ramanathan1504: Could the entry say which configuration changes? As written, a reader with links in their log directory cannot tell whether this affects them.

```suggestion
  <description format="asciidoc">Stop `PosixViewAttribute` from applying permissions and ownership through symbolic links found in `basePath` unless `followLinks` is set to `true`.</description>
```

log4j-core/src/main/java/org/apache/logging/log4j/core/util/FileUtils.java
jmestwa-coder: makes sense, I missed that the manager paths call this on the configured fileName where following is the intended behavior. reverted to the plain lookup and dropped the LinkOption import; the visitor guard carries the fix now.

log4j-core/src/main/java/org/apache/logging/log4j/core/util/FileUtils.java
jmestwa-coder: dropped the paragraph along with the revert.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeAction.java
jmestwa-coder: good catch on followLinks="true", applied the suggestion.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/FileUtilsTest.java
jmestwa-coder: added it to PosixViewAttributeActionTest next to the negative test, both pass.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/util/FileUtilsTest.java
jmestwa-coder: removed, along with the imports it pulled in.

src/changelog/.2.x.x/fix_posix_view_attribute_symlink.xml
jmestwa-coder: took the suggested wording.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeAction.java
```diff
@@ -369,6 +369,10 @@ protected FileVisitor<Path> createFileVisitor(final Path basePath, final List<Pa
         return new SimpleFileVisitor<Path>() {
             @Override
             public FileVisitResult visitFile(final Path file, final BasicFileAttributes attrs) throws IOException {
+                if (!isFollowSymbolicLinks() && attrs.isSymbolicLink()) {
```
ramanathan1504: Withdrawing my earlier suggestion here, sorry — under `FOLLOW_LINKS` a valid link is `stat`-resolved before `visitFile` sees it so the gate changes nothing, and on a *dangling* link it throws `NoSuchFileException` out of `walkFileTree` and ends the scan; your original line was right.

```suggestion
                if (attrs.isSymbolicLink()) {
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeActionTest.java
ramanathan1504: `testSymbolicLinksAreFollowedWhenConfigured` passes with and without the change to 372, so the dangling-link case is the one that actually guards F1 — sketch below rather than a suggestion block, since it is a whole method:

```java
@Test
void testBrokenSymbolicLinkDoesNotAbortTheScan(@TempDir final Path tempDir) throws Exception {
    final Path baseDir = Files.createDirectory(tempDir.resolve("logs"));
    final Path regularFile = baseDir.resolve("app-1.log");
    Files.write(regularFile, "log".getBytes(StandardCharsets.UTF_8));
    Files.setPosixFilePermissions(regularFile, PosixFilePermissions.fromString("rw-------"));
    Files.createSymbolicLink(baseDir.resolve("app-0-broken.log"), tempDir.resolve("gone.txt"));

    final Configuration config = new BasicConfigurationFactory().new BasicConfiguration();
    final PosixViewAttributeAction action = PosixViewAttributeAction.newBuilder()
            .setBasePath(baseDir.toString())
            .setFollowLinks(true)
            .setMaxDepth(1)
            .setPathConditions(PathCondition.EMPTY_ARRAY)
            .setConfiguration(config)
            .setFilePermissionsString("rw-rw-rw-")
            .build();

    action.execute();

    assertEquals(
            "rw-rw-rw-",
            PosixFilePermissions.toString(Files.getPosixFilePermissions(regularFile)),
            "a dangling link must not stop the scan");
}
```

src/changelog/.2.x.x/fix_posix_view_attribute_symlink.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xmlns="https://logging.apache.org/xml/ns"
+       xsi:schemaLocation="https://logging.apache.org/xml/ns https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4229" link="https://github.com/apache/logging-log4j2/pull/4229"/>
+  <description format="asciidoc">Stop `PosixViewAttribute` from applying permissions and ownership through symbolic links found in `basePath` unless `followLinks` is set to `true`.</description>
```
ramanathan1504:
```suggestion
  <description format="asciidoc">Stop `PosixViewAttribute` from applying permissions and ownership through symbolic links found in `basePath` unless `followLinks` is set to `true`</description>

**Review by ramanathan1504 — CHANGES_REQUESTED** [review]

File: log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeActionTest.java [hunk 9]
```diff
@@ -0,0 +1,104 @@
...
+                .setFilePermissionsString("rw-rw-rw-")
+                .build();
+
+        action.execute();
+
+        assertEquals(
+                "rw-rw-rw-",
+                PosixFilePermissions.toString(Files.getPosixFilePermissions(outsider)),
+                "followLinks=\"true\" should still follow the link");
+    }
```
ramanathan1504, line 103: `testSymbolicLinksAreFollowedWhenConfigured` passes with and without the change on line 372 — I reverted that line and it stayed green — so the dangling-link case is the one that actually guards it. Sketch rather than a suggestion block, since it is a whole method:

```java
@Test
void testBrokenSymbolicLinkDoesNotAbortTheScan(@TempDir final Path tempDir) throws Exception {
    final Path baseDir = Files.createDirectory(tempDir.resolve("logs"));
    final Path regularFile = baseDir.resolve("app-1.log");
    Files.write(regularFile, "log".getBytes(StandardCharsets.UTF_8));
    Files.setPosixFilePermissions(regularFile, PosixFilePermissions.fromString("rw-------"));
    Files.createSymbolicLink(baseDir.resolve("app-0-broken.log"), tempDir.resolve("gone.txt"));

    final Configuration config = new BasicConfigurationFactory().new BasicConfiguration();
    final PosixViewAttributeAction action = PosixViewAttributeAction.newBuilder()
            .setBasePath(baseDir.toString())
            .setFollowLinks(true)
            .setMaxDepth(1)
            .setPathConditions(PathCondition.EMPTY_ARRAY)
            .setConfiguration(config)
            .setFilePermissionsString("rw-rw-rw-")
            .build();

    action.execute();

    assertEquals(
            "rw-rw-rw-",
            PosixFilePermissions.toString(Files.getPosixFilePermissions(regularFile)),
            "a dangling link must not stop the scan");
}
```

ramanathan1504 (changes requested):

ramanathan1504:
@jmestwa-coder thanks — the `FileUtils` revert, the javadoc paragraph, the old test removal and the changelog wording all landed exactly right; `FileUtils.java` and `FileUtilsTest.java` are now byte-identical to `2.x`, `spotless:check` is green and `PosixViewAttributeActionTest` + `FileUtilsTest` pass 10/10 here.

Two things, and the first one is my fault.

1. **My suggestion on `PosixViewAttributeAction.java:372` was wrong — please back it out.** I claimed the guard never fires under `followLinks="true"`, and it does not, but that is precisely the behaviour you want: with `FOLLOW_LINKS` a valid symlink is already `stat`-resolved before `visitFile` sees it, so it gets followed either way. The only case `!isFollowSymbolicLinks() &&` actually changes is a *dangling* link, which the plain guard skipped harmlessly and the gated version now walks into. One broken link and one real `app-1.log` in `basePath` with `followLinks="true"`:

```
with    !isFollowSymbolicLinks() && :
  execute() threw java.nio.file.NoSuchFileException: .../app-0-broken.log
  app-1.log perms = rw-------      <- never updated

without (plain attrs.isSymbolicLink()):
  execute() returned normally
  app-1.log perms = rw-rw-rw-      <- updated
```

`visitFile` throws out of `walkFileTree`, so one stale link stops the whole scan. Suggestion on the line to restore it.

2. **`testSymbolicLinksAreFollowedWhenConfigured` passes with and without the fix** — I reverted line 372 and it stayed green, same reason as above. Worth keeping as documentation, but it does not guard anything. The broken-link case does: it fails on the gated version and passes on the plain one. Sketch left on the test file.

Sorry for the round trip on 372 — that one was mine to get right.

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeAction.java
jmestwa-coder: no worries, backed it out — plain attrs.isSymbolicLink() again.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/rolling/action/PosixViewAttributeActionTest.java
jmestwa-coder: added testBrokenSymbolicLinkDoesNotAbortTheScan from your sketch and kept the followed-when-configured one as documentation. fails with the gated guard, passes with the plain one here too.

jmestwa-coder:
both done: 372 is back to the plain guard and the dangling-link test is in. PosixViewAttributeActionTest + FileUtilsTest green, spotless:check clean on both modules.

ramanathan1504 (approved):
@jmestwa-coder Thanks!

ramanathan1504:
@ppkarwasz @vy this changes which files get new permissions after a rollover. Can one of you confirm before merge?

ramanathan1504:
> @ppkarwasz @vy this changes which files get new permissions after a rollover. Can one of you confirm before merge?

The change only skips symbolic links when `followLinks` is `false`, which is what the manual already says.                                          Both regression tests fail without the fix, and CI is green.

PR #4234, opened 2026-08-01
Add ZstdCompressAction to support configurable Zstandard compression levels

katstack:
This PR adds `ZstdCompressAction`, replacing the generic `CommonsCompressAction("zstd", ...)` path (added in #1514) with a dedicated action that supports a configurable `compressionLevel`, following the same incremental pattern already used by `GzCompressAction` (LOG4J2-2598) and `ZipCompressAction`.

Supported range is the standard positive Zstd levels `[1, ZstdConstants.ZSTD_CLEVEL_MAX]` (currently `[1, 22]`). `compressionLevel=-1` (unset) continues to resolve to the Zstd default level (`3`), consistent with the existing behavior documented in `FileExtension`/`rolling-file.adoc`.

### What this PR deliberately does not do

Negative ("fast compression") Zstd levels are not supported. The rolling-appender framework uses `Deflater.DEFAULT_COMPRESSION` (`-1`) as a format-agnostic sentinel for "unspecified compression level," so an explicit `compressionLevel=-1` would be indistinguishable from "not set." On `main`, `compressionLevel` has already been replaced by a more generic `compressionOptions` map (#2921); on `2.x` it remains the mechanism, so this PR stays scoped to extending it rather than adding another special case on top of the sentinel.

## Testing

`ZstdCompressActionTest` covers the new validation/mapping logic, including a test pinning the currently-assumed `ZstdConstants` bounds so a future zstd-jni/commons-compress upgrade that shifts them fails loudly instead of silently drifting from the documented range. Also ran the existing `GzCompressActionTest`/`ZipCompressActionTest`/`Bzip2CompressActionTest`/rolling-appender suites locally with no regressions, and `./mvnw verify` passes on both JDK 8 and JDK 17.

## References

- #2950 — ongoing discussion on generalizing compression configuration.

### DIFF
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/FileExtension.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/FileExtension.java
@@ -18,10 +18,12 @@ package org.apache.logging.log4j.core.appender.rolling;
 
 import java.io.File;
 import java.util.Objects;
+import org.apache.commons.compress.compressors.zstandard.ZstdConstants;
 import org.apache.logging.log4j.core.appender.rolling.action.Action;
 import org.apache.logging.log4j.core.appender.rolling.action.CommonsCompressAction;
 import org.apache.logging.log4j.core.appender.rolling.action.GzCompressAction;
 import org.apache.logging.log4j.core.appender.rolling.action.ZipCompressAction;
+import org.apache.logging.log4j.core.appender.rolling.action.ZstdCompressAction;
 import org.apache.logging.log4j.core.internal.annotation.SuppressFBWarnings;
 
 /**
@@ -99,8 +101,15 @@ public enum FileExtension {
                 final String compressedName,
                 final boolean deleteSource,
                 final int compressionLevel) {
-            // One of "gz", "bzip2", "xz", "zstd", "pack200", or "deflate".
-            return new CommonsCompressAction("zstd", source(renameTo), target(compressedName), deleteSource);
+            // -1 (Deflater.DEFAULT_COMPRESSION) is the framework-wide sentinel for 'unspecified compression level'.
+            // Unlike GZ/ZIP, where -1 has no meaning other than 'use Deflater's default' (java.util.zip.Deflater
+            // natively treats -1 as its own default-compression sentinel), Zstd defines -1 as a real, distinct
+            // fast-compression level. So the sentinel has to be mapped explicitly here to Zstd's own default level.
+            // Negative Zstd fast-compression levels are intentionally out of scope for now; see the
+            // discussion on generalizing compressionLevel at
+            // https://github.com/apache/logging-log4j2/discussions/2950.
+            final int level = compressionLevel == -1 ? ZstdConstants.ZSTD_CLEVEL_DEFAULT : compressionLevel;
+            return new ZstdCompressAction(source(renameTo), target(compressedName), deleteSource, level);
         }
     };
 
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZstdCompressAction.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZstdCompressAction.java
@@ -1,0 +1,208 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.logging.log4j.core.appender.rolling.action;
+
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Objects;
+import org.apache.commons.compress.compressors.zstandard.ZstdCompressorOutputStream;
+import org.apache.commons.compress.compressors.zstandard.ZstdConstants;
+
+/**
+ * Compresses a file using Zstandard compression.
+ * <p>
+ * Supports positive compression levels in the range [{@value #MIN_COMPRESSION_LEVEL}, {@link ZstdConstants#ZSTD_CLEVEL_MAX}].
+ * Negative (fast-compression) levels are not currently supported; this may change in a future release.
+ * </p>
+ *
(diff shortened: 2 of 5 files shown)


log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/FileExtension.java
```diff
@@ -99,8 +101,15 @@ public Action createCompressAction(
...
-            // One of "gz", "bzip2", "xz", "zstd", "pack200", or "deflate".
-            return new CommonsCompressAction("zstd", source(renameTo), target(compressedName), deleteSource);
+            // -1 (Deflater.DEFAULT_COMPRESSION) is the framework-wide sentinel for 'unspecified compression level'.
+            // Unlike GZ/ZIP, where -1 has no meaning other than 'use Deflater's default' (java.util.zip.Deflater
+            // natively treats -1 as its own default-compression sentinel), Zstd defines -1 as a real, distinct
+            // fast-compression level. So the sentinel has to be mapped explicitly here to Zstd's own default level.
+            // Negative Zstd fast-compression levels are intentionally out of scope for now; see the
+            // discussion on generalizing compressionLevel at
+            // https://github.com/apache/logging-log4j2/discussions/2950.
+            final int level = compressionLevel == -1 ? ZstdConstants.ZSTD_CLEVEL_DEFAULT : compressionLevel;
```
ramanathan1504, line 111: `ZstdConstants` reads its values from zstd-jni when the class loads. Here that happens during rollover, so without the zstd-jni jar this throws `NoClassDefFoundError` before the file is renamed. Can the `-1` mapping move into `ZstdCompressAction.execute`?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZstdCompressAction.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.commons.compress.compressors.zstandard.ZstdConstants;
+
+/**
+ * Compresses a file using Zstandard compression.
+ * <p>
+ * Supports positive compression levels in the range [{@value #MIN_COMPRESSION_LEVEL}, {@link ZstdConstants#ZSTD_CLEVEL_MAX}].
+ * Negative (fast-compression) levels are not currently supported; this may change in a future release.
+ * </p>
+ *
+ * @apiNote An explicitly configured level of -1 currently resolves to the Zstd default level (3).
```
ramanathan1504, line 36: The constructor throws for `-1`. Only `FileExtension.ZSTD` turns it into 3. Can this note go, since the manual already covers it?

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZstdCompressAction.java
```diff
@@ -0,0 +1,208 @@
...
+ * This is provisional behavior tied to the current lack of negative-level support and may change
+ * in a future release without a corresponding API signature change.
+ */
```
ramanathan1504, line 39: New public class on `2.x`, so it needs the version tag.

```suggestion
 * @since 2.27.0
 */
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/rolling/action/ZstdCompressAction.java
```diff
@@ -0,0 +1,208 @@
...
+     */
+    public ZstdCompressAction(
+            final File source, final File destination, final boolean deleteSource, final int compressionLevel) {
+        Objects.requireNonNull(source, "source");
+        Objects.requireNonNull(destination, "destination");
+
+        this.source = source;
+        this.destination = destination;
+        this.deleteSource = deleteSource;
+        this.compressionLevel = checkCompressionLevel(compressionLevel);
```
ramanathan1504, line 121: This throws `IllegalArgumentException` inside `DefaultRolloverStrategy.rollover`, where nothing catches it. With `compressionLevel="0"` on a `.zst` pattern the file stops rolling and the log events are lost. `0` was ignored for `.zst` before. Can the check run in `execute()` instead?

ramanathan1504 (commented):
Thanks @katstack, the write-up on why `-1` has to stay the zstd default is clear, and the `FileExtension` mapping has a test.
`FileExtension.ZSTD` now loads `ZstdConstants` and the constructor checks the range while the rollover is being built, so `compressionLevel="0"` or a missing zstd-jni jar stops the rollover and every later log event is dropped. Can the level check stay inside `execute()`, so those cases only fail the compression like on 2.x, and can you add a `src/changelog/.2.x.x` entry?

ramanathan1504:
- For @ppkarwasz / @vy: given #2950 proposed deprecating `compressionLevel`, is new 2.x-only public API (`ZstdCompressAction`) wanted, or should the manual just keep `.zst` as "not supported"?
- Should `testAssumedZstdLevelBoundsHaveNotChanged` stay, knowing a zstd-jni bump can turn it red?
- Separate issue: should `RollingFileManager` still write the event when `strategy.rollover` throws? Today a bad `.gz` level loses events the same way.

katstack:
Addressed in 892c9128df. `FileExtension.ZSTD` now passes the raw level without loading `ZstdConstants`; `ZstdCompressAction.execute()` resolves the unspecified `-1` level and validates it immediately before compression. Therefore an invalid level or missing zstd-jni only fails the asynchronous compression action, rather than preventing rollover. I also added the 2.x changelog entry, removed the `@apiNote`, and added `@since 2.27.0`.

ramanathan1504:
@ppkarwasz  `mvnw verify` is red on `log4j-osgi-test`: `ZstdCompressorOutputStream.builder()` pulls in `org.apache.commons.io.build` and bnd makes it a mandatory import, so the          
  `log4j-core` bundle stops resolving for everyone, zstd users or not. Pushed `org.apache.commons.io.*;resolution:=optional` into `log4j-core/pom.xml` next to the              
  commons-compress entry, and Equinox and Felix are 9/9 again.                                                                                                                  
  @ppkarwasz, one thing before this goes further: #2950 proposed deprecating `compressionLevel` and removing it in 3.x, so is a new 2.x-only `ZstdCompressAction` wanted at all?

ppkarwasz:
Hi @katstack,

Thank you for the PR.

As @ramanathan1504 mentioned, we are considering removing `compressionLevel` in `3.x` in favor of a more general mechanism (#2950): every compression algorithm has its own set of parameters, and adding a `RolloverStrategy` attribute for each of them does not scale. Could you have a look at #2950 and tell us about your use case? Your point of view would help us shape that design.

That said, the attribute exists today, so we might as well honor it wherever we can. ZStandard is not the only algorithm with a compression level: Deflate, GZip and BZip2 support one too. We prefer to change a whole class of implementations at once rather than a single one, so users get consistent behavior and the change can be properly described in the release notes. Would you be willing to extend the PR to the other three algorithms?

Two remarks on the implementation:

- Please move `ZstdCompressAction` to a new `actions.internal` package, so it is not exported via JPMS/OSGi. Everything we export becomes public API that we must keep compatible for years, so we only export components meant to be reused. These compression actions are Log4j-specific; reusable compression code belongs in Commons Compress.
- `CommonsCompressAction`, `GzCompressAction` and `ZipCompressAction` already duplicate a lot of logic. Extracting a common `AbstractCompressAction` (also in `actions.internal`) would let the new action share it instead of adding a fourth copy.

**TL;DR**: moving `ZstdCompressAction` to `actions.internal` is the one required change; covering the other `compressionLevel`-aware algorithms in this PR would be very welcome.

PR #4073, opened 2026-03-24
Rfc5424Layout: Fix sanitization of SD parameter names

ppkarwasz:
This change corrects the sanitization of `PARAM-NAME` in RFC 5424 structured data produced by `Rfc5424Layout`.

The issue was reported by @N0tre3l  on March 8th, 2026.

Previously, parameter names were sanitized using the same escaping mechanism as parameter values. However, RFC 5424 does not define an escape mechanism for `PARAM-NAME`; instead, names must follow the `SD-NAME` syntax (`1*32PRINTUSASCII` with additional character restrictions).

This change enforces these constraints when rendering structured data parameters:

* Invalid characters are replaced with `?`.
* Parameter names are truncated to a maximum of **32 characters**.
* If sanitization results in an empty name, `?` is used instead.

This ensures that generated structured data complies with the RFC 5424 grammar for `PARAM-NAME`.

PR #4235, opened 2026-08-03
sanitize SD-ID and MSGID in Rfc5424Layout

jmestwa-coder:
`Rfc5424Layout` sanitizes SD-PARAM-NAME and escapes SD-PARAM-VALUE, but two neighboring fields of the same record go out unmodified:

- `appendMessageId` appends `StructuredDataMessage.getType()` raw, so a newline in the type ends the record and what follows is read as a separate syslog message
- `formatStructuredElement` appends the SD-ID raw, so a `]` closes the element early and the remainder reads as a second, caller-controlled element
- neither `setType` nor the `StructuredDataId` constructors check characters, only length, and both fields carry application data

`new StructuredDataMessage("a] [forged@1 user=\"root", "login ok", "Audit\n<13>1 - - - - -")` through a `Syslog` appender, before:

```
<128>1 1970-01-01T05:30:00.001+05:30 host - 95177 Audit
<13>1 - - - - - [a] [forged@1 user="root] login ok
```

after:

```
<128>1 1970-01-01T05:30:00.001+05:30 host - 95119 Audit?<13>1?-?-?-?-?- [a??[forged@1?user??root] login ok
```

Same `?` replacement #4073 introduced for parameter names. SD-ID reuses the SD-NAME rule since RFC 5424 gives both fields the same production, MSGID keeps the wider printable US-ASCII range its own rule allows. Lengths are left alone so callers that raise `StructuredDataId`'s 32 character limit keep working.

## Checklist

* [x] Based on `2.x`
* [x] `./mvnw verify` succeeds (ran the build and `Rfc5424LayoutTest` for `log4j-api`, `log4j-core` and `log4j-core-test` locally)
* [x] Entry added under `src/changelog/.2.x.x`
* [x] Tests are provided

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/layout/Rfc5424LayoutTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/layout/Rfc5424LayoutTest.java
@@ -973,13 +973,67 @@ class Rfc5424LayoutTest {
         assertThat(actual).isEqualTo(expected);
     }
 
+    private static LogEvent createLogEventWithStructuredData(final String id, final String type) {
+        final MutableInstant instant = new MutableInstant();
+        instant.initFromEpochMilli(1L, 0);
+
+        return Log4jLogEvent.newBuilder()
+                .setInstant(instant)
+                .setMessage(new StructuredDataMessage(id, "MSG", type))
+                .build();
+    }
+
+    private static Stream<Arguments> testSdIdSanitization() {
+        return Stream.of(
+                Arguments.of("validId", "[validId@32473]"),
+                Arguments.of("event id", "[event?id@32473]"),
+                Arguments.of("event=id", "[event?id@32473]"),
+                Arguments.of("event]id", "[event?id@32473]"),
+                Arguments.of("event\"id", "[event?id@32473]"),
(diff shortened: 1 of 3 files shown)


vy:
The PR’s own premise conflicts with [the security model](https://logging.apache.org/security.html#threat-common-sources-structural), which states that MSGID and SD-ID are explicitly developer-controlled structural identifiers, so feeding attacker data into them is out of scope.

@raboof, @ppkarwasz, I'm closing this PR. Feel free to re-open it if you think otherwise.

raboof:
> states that MSGID and SD-ID are explicitly developer-controlled structural identifiers, so feeding attacker data into them is out of scope

the fact that we don't support untrusted input in those fields doesn't mean we necessarily need to forbid ourselves from sanitizing them: sanitizing them could be seen as a hardening improvement (reducing the impact when an application incorrectly does pass untrusted input to those fields) or a general improvement (reducing the chance of regular bugs when users pass data that would be misinterpreted here). I don't have a strong opinion if those motivations are worth making the change.

ppkarwasz:
My approach here would be to validate the values in the constructors of [`StructureDataId`](https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/message/StructuredDataId.html), which has the advantage of:

- Failing fast during static initialization, so the developer can fix it,
- Has no impact on runtime performance.

PR #4157, opened 2026-06-26
[main] Port `InternalLoggerRegistry` from `2.x` (#3418 and #3681 ports)

vpelikh:
Port `InternalLoggerRegistry` from `2.x` (#3418, #3681)

Minimizes lock usage by moving logger instantiation outside the write lock, preventing deadlocks that could occur when the logger constructor triggers property lookups or other pluggable operations (PR #3418, closes #3399).

Adds stale entry detection and removal via `ReferenceQueue` — when a `Logger` is garbage collected, its `WeakReference` is enqueued. Subsequent calls to `getLogger()`, `computeIfAbsent()`, or `hasLogger()` trigger expunging of stale entries, preventing memory leaks from accumulated dead entries (PR #3681).

**Changes:**
- `InternalLoggerRegistry` — new multi-level `WeakHashMap<MessageFactory, HashMap<String, WeakReference<Logger>>>` structure with `ReadWriteLock`
- `LoggerContext.getLogger()` — simplified to use `registry.computeIfAbsent()` (moves `newLogger()` outside the write lock)
- `InternalLoggerRegistryTest` — 6 tests covering creation, lookup, and stale entry expunging (WeakReference + MessageFactory GC scenarios)

PR #4240, opened 2026-08-04
Port `MessageFactory`-namespaced logger registry from `2.x`

### LINKED ISSUE #2962: Implement `MessageFactory`-namespaced logger registry (blocked by `2.24.1` release)
Port #2936 fixing #2933 in `2.x` to `main` when `2.24.1` is released.

vpelikh:
## Summary

Completes the port of #2962 by removing the now-dead `checkMessageFactory` machinery from `LoggerContext`.
The MessageFactory-namespaced `InternalLoggerRegistry` itself was already ported in #4157. This PR removes the leftover core code that only existed to warn about message-factory mismatches under the old name-keyed registry.

## Why `checkMessageFactory` was dead code

The `LoggerRegistry` now keys loggers by **(name, message factory)**. As a result:

- `computeIfAbsent(name, mf, newLogger)` always returns a logger whose message factory matches the requested one, so the warning branch `!loggerMessageFactory.equals(messageFactory)` could never fire.
- `getLogger` normalizes a `null` message factory to the default *before* reaching the registry, so the null-mismatch branch was unreachable as well.

The behavior this method guarded against — returning an existing logger created with a *different* message factory — no longer exists: a request with a different factory now returns a distinct, correctly-bound logger instead of the old logger plus a warning. `2.x` removed this dead method in PR #2936, the same change that introduced the namespaced registry there; this PR mirrors that change in the `3.x`, keeping it a faithful port.

**Note on registry placement:** as suggested in https://github.com/apache/logging-log4j2/issues/2962#issuecomment-2469688771, the namespaced registry was kept out of `log4j-api`; it lives in `log4j-core` (mirroring `2.x`) rather than `log4j-kit`, since kit sits below core and can't reference the core `Logger` type.

## Changes

1. **`log4j-core` — `LoggerContext`**
   - Removed private `checkMessageFactory(ExtendedLogger, MessageFactory)` and its call site in `getLogger(String, MessageFactory)`.
   - Removed the now-unused `ExtendedLogger` and `StatusLogger` imports.
   - Added `@Nullable` to the message-factory parameter of `getLogger(String, MessageFactory)` and `hasLogger(String, MessageFactory)`; refreshed the `getLogger` Javadoc to describe the message-factory-matching semantics.
2. **`log4j-core-test` — `LoggerTest`**
   - Added regression test `getLogger_String_MessageFactoryMismatchProducesNoWarning`: re-requests a logger with a different and with a `null` message factory, and asserts no "created with the message factory" status warning is emitted.
   - Kept the `builder` hardcoded source-line constant in sync with the file layout.

Closes #2962

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
@@ -52,6 +52,7 @@ import org.apache.logging.log4j.message.ReusableMessageFactory;
 import org.apache.logging.log4j.message.SimpleMessage;
 import org.apache.logging.log4j.message.StringFormatterMessageFactory;
 import org.apache.logging.log4j.message.StructuredDataMessage;
+import org.apache.logging.log4j.status.StatusLogger;
 import org.awaitility.Awaitility;
 import org.hamcrest.MatcherAssert;
 import org.junit.jupiter.api.Tag;
@@ -121,7 +122,7 @@ public class LoggerTest {
 
     @Test
     public void builder() {
-        final int currentLine = 124;
+        final int currentLine = 125;
         logger.atDebug().withLocation().log("Hello");
         final Marker marker = MarkerManager.getMarker("test");
         logger.atError().withMarker(marker).log("Hello {}", "John");
@@ -417,6 +418,17 @@ public class LoggerTest {
         assertEquals(
                 String.format("%,d", Integer.MAX_VALUE),
                 events.get(0).getMessage().getFormattedMessage());
+    }
+
+    @Test
+    public void getLogger_String_MessageFactoryMismatchProducesNoWarning(final TestInfo testInfo) {
+        final String name = testInfo.getTestMethod().map(Method::getName).orElseThrow(AssertionError::new);
+        testMessageFactoryMismatch(name, StringFormatterMessageFactory.INSTANCE, new ReusableMessageFactory());
+        testMessageFactoryMismatch(name + "Null", StringFormatterMessageFactory.INSTANCE, null);
+        final boolean mismatchWarning = StatusLogger.getLogger().getStatusData().stream()
+                .map(data -> data.getMessage().getFormattedMessage())
+                .anyMatch(message -> message.contains("created with the message factory"));
+        assertFalse(mismatchWarning, "The message factory mismatch warning should not be emitted");
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


vy:
@ramanathan1504, I'm dropping this one on you. Let me know if you disagree or need help.

ramanathan1504:
> @ramanathan1504, I'm dropping this one on you. Let me know if you disagree or need help.

@vy Got it. I'll take a look and let you know if I need any help.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
```diff
@@ -52,6 +52,7 @@
 import org.apache.logging.log4j.message.SimpleMessage;
 import org.apache.logging.log4j.message.StringFormatterMessageFactory;
 import org.apache.logging.log4j.message.StructuredDataMessage;
+import org.apache.logging.log4j.status.StatusLogger;
```
ramanathan1504: Only the new test uses this import, so it goes with the test.

```suggestion
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
```diff
@@ -121,7 +122,7 @@ public void basicFlow() {
 
     @Test
     public void builder() {
-        final int currentLine = 124;
+        final int currentLine = 125;
```
ramanathan1504: Back to `124` once the import above is removed.

```suggestion
        final int currentLine = 124;
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/LoggerTest.java
```diff
@@ -419,6 +420,17 @@ public void getLogger_String_MessageFactoryMismatchNull(final TestInfo testInfo)
...
+    @Test
+    public void getLogger_String_MessageFactoryMismatchProducesNoWarning(final TestInfo testInfo) {
+        final String name = testInfo.getTestMethod().map(Method::getName).orElseThrow(AssertionError::new);
+        testMessageFactoryMismatch(name, StringFormatterMessageFactory.INSTANCE, new ReusableMessageFactory());
+        testMessageFactoryMismatch(name + "Null", StringFormatterMessageFactory.INSTANCE, null);
+        final boolean mismatchWarning = StatusLogger.getLogger().getStatusData().stream()
+                .map(data -> data.getMessage().getFormattedMessage())
+                .anyMatch(message -> message.contains("created with the message factory"));
+        assertFalse(mismatchWarning, "The message factory mismatch warning should not be emitted");
+    }
```
ramanathan1504: This passes even when `checkMessageFactory` warns on every call, and the two `getLogger_String_MessageFactoryMismatch` tests above already check that each factory gets its own logger.

```suggestion
```

vpelikh:
@ramanathan1504 addressed you review in [b31b2c5](https://github.com/apache/logging-log4j2/pull/4240/commits/b31b2c53fb840eed5406fc3b74b5a23a8dcce58d).

ramanathan1504 (approved):
@vpelikh, thanks. `LoggerTest.java` is back to what `main` has, and the whole class passes.One ask from the last round never made it out of the review body: #3161 asks every port to carry a changelog entry under `src/changelog/.3.x.x/` that links #2936. Approving on the code, please add the entry before this merges.

vpelikh:
@ramanathan1504 done. Thanks for the review!

ramanathan1504 (approved):
LGTM!

PR #4245, opened 2026-08-06
fix: treat null parameters as empty in CsvParameterLayout

### LINKED ISSUE #4243: CsvParameterLayout throws NullPointerException on any event without parameters
## Description

`CsvParameterLayout.toSerializable` passes `Message.getParameters()` to `CSVFormat.printRecord` with no null check. It is `null` for `SimpleMessage`, so every plain `logger.info("some text")` throws.

Only `IOException` is caught, so the NPE escapes to the appender: one error per event, no output, JVM exits 0. Parameterised calls work, so it survives testing and appears in production.

```java
final Object[] parameters = message.getParameters();   // null for SimpleMessage
try {
    getFormat().printRecord(buffer, parameters);       // NPE
} catch (final IOException e) {                        // IOException only
```

**Present on both lines:**

- 2.x `04c93c1d33` — `log4j-core/.../core/layout/CsvParameterLayout.java:98`
- 3.x `main` — `log4j-csv/.../csv/layout/CsvParameterLayout.java:104`

Reproduced on 2.26.1, JDK 21.

## Logs

```
ERROR An exception occurred processing Appender CsvParams
java.lang.NullPointerException: Cannot read the array length because "values" is null
	at org.apache.commons.csv.CSVFormat.printRecord(CSVFormat.java:2265)
	at CsvParameterLayout.toSerializable(CsvParameterLayout.java:98)
```

## Reproduction

```xml
<File name="Csv" fileName="params.csv"><CsvParameterLayout format="Default"/></File>
```

```java
logger.info("order {} accepted", 4711);   // fine
logger.info("plain text");                // NPE, file gains nothing
```

## Suggested fix

```java
getFormat().printRecord(buffer, parameters == null ? EMPTY : parameters);
```

An empty row is well-formed and matches the present-but-zero-length case. Widening the `catch` would stop the throw but leave the column count wrong, which is worse for a machine-read format.

arimu1:
Fixes #4243

## Description

`CsvParameterLayout.toSerializable` passes `Message.getParameters()` to `CSVFormat.printRecord` without a null check. Parameter-less messages such as `SimpleMessage` return `null` from `getParameters()`, and Commons CSV throws `NullPointerException` when the values array is null. Only `IOException` is caught, so the NPE escapes to the appender and the event is lost.

This change treats a null parameter array as `Constants.EMPTY_OBJECT_ARRAY`, which produces a well-formed empty CSV record (record separator only) and matches the zero-length parameters case.

## Checklist

* Tests are provided

## Testing

- JDK 17 (branch enforcer requires `[17,18)`)
- `./mvnw test -pl :log4j-core-test -am -Dtest=CsvParameterLayoutTest#testNullParametersProduceEmptyRecord -Dsurefire.failIfNoSpecifiedTests=false` — **2/2 pass**
- Related existing layout tests in the same class (`testLayoutDefaultNormal`, `testLayoutDefaultObjectArrayMessage`, `testLayoutTab`, charset/content-type tests) also pass
- Spotless check clean on `:log4j-core` and `:log4j-core-test`

### DIFF
--- a/src/changelog/.2.x.x/4243_fix_CsvParameterLayout_null_parameters.xml
+++ b/src/changelog/.2.x.x/4243_fix_CsvParameterLayout_null_parameters.xml
@@ -6,6 +6,7 @@
            https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
        type="fixed">
   <issue id="4243" link="https://github.com/apache/logging-log4j2/issues/4243"/>
+  <issue id="4245" link="https://github.com/apache/logging-log4j2/pull/4245"/>
   <description format="asciidoc">
     Fix `NullPointerException` in `CsvParameterLayout` when a log event has no parameters (for example `SimpleMessage`).
   </description>
--- a/src/changelog/.2.x.x/4243_fix_CsvParameterLayout_null_parameters.xml
+++ b/src/changelog/.2.x.x/4243_fix_CsvParameterLayout_null_parameters.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4243" link="https://github.com/apache/logging-log4j2/issues/4243"/>
+  <description format="asciidoc">
+    Fix `NullPointerException` in `CsvParameterLayout` when a log event has no parameters (for example `SimpleMessage`).
+  </description>
+</entry>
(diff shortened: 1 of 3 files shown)


src/changelog/.2.x.x/4243_fix_CsvParameterLayout_null_parameters.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4243" link="https://github.com/apache/logging-log4j2/issues/4243"/>
```
vy, line 8:
```suggestion
  <issue id="4243" link="https://github.com/apache/logging-log4j2/issues/4243"/>
  <issue id="4245" link="https://github.com/apache/logging-log4j2/pull/4245"/>
```

vy (approved):

Issue #4242, opened 2026-08-05
log4j-cassandra: failed appender startup leaks the DataStax Cluster, so the JVM never exits


ramanathan1504:
## Description

`CassandraManager.startupInternal()` fails → the `Cluster` is never closed → the JVM never exits, because the DataStax 3.x driver's threads are non-daemon.

`cluster.close()` is only reachable from `shutdownInternal()`, and `AbstractDatabaseManager.shutdown()` skips that when `running == false` (see #4241).

**Version:** 2.x `04c93c1d33` (no 3.x release) · Cassandra 4.1.11, driver 3.11.x

## Logs

```
main DEBUG Shut down CassandraManager Cassandra, all resources released: true
```

Thread dump 8s later — five non-daemon threads, each carrying the configured `clusterName`:

```
"log4j-bench-nio-worker-0"  "log4j-bench-nio-worker-1"  "log4j-bench-timeouter-0"
"log4j-bench-connection-reaper-0"  "log4j-bench-scheduled-task-worker-0"
```

## Reproduction

Point a `Cassandra` appender at a live node with the keyspace **absent**, log one event, call `LogManager.shutdown()`, return from `main`. The JVM hangs.

## Suggested fix

Close the `Cluster` when `startupInternal()` fails — wrap its body in `try`/`catch` that calls `cluster.close()` and rethrows, or make `shutdownInternal()` null-tolerant and let it run irrespective of `running`, per #4241.

ramanathan1504:
https://github.com/apache/logging-log4j2/pull/4246 fixes in the same pr

PR #4246, opened 2026-08-06
fix: skip writes and still shut down after failed DB manager startup

### LINKED ISSUE #4241: AbstractDatabaseManager: a manager whose startup failed still receives writes, and is never shut down
## Description

`AbstractDatabaseManager` checks `isRunning()` on the shutdown path but not on the write path.

After `startupInternal()` throws, `startup()` logs the cause and leaves `running = false`. Then:

- **`write()` (line 297) does not check it** — keeps accepting events and dereferences state startup never assigned. NPE per event, naming an internal field, burying the one line that named the cause.
- **`shutdown()` (line 230) does check it** — so `shutdownInternal()` never runs and whatever startup built is leaked. Returns `true` anyway.

Affects every subclass whose startup fails: JDBC, JPA, NoSQL. #4242 is the case where the leak prevents JVM exit.

**Version:** 2.x `04c93c1d33`, and `main` — unchanged in this respect.

## Logs

```
main ERROR CassandraManager Could not perform database startup operations:
  InvalidQueryException: Keyspace 'log4j' does not exist
    at AbstractDatabaseManager.startup(AbstractDatabaseManager.java:259)

main ERROR Unable to write to database [Cassandra] for appender [Cassandra].
java.lang.NullPointerException: Cannot invoke "PreparedStatement.bind(Object[])"
  because "this.preparedStatement" is null        <- then this, per event
```

## Suggested fix

- `write()`: return early when `!isRunning()`, logging once rather than per event.
- `shutdown()`: run `shutdownInternal()` regardless of `running`, since startup may have acquired resources before throwing.

arimu1:
Fixes #4241 and #4242

## Description

`AbstractDatabaseManager` checks `isRunning()` on the shutdown path but not on the write path.

After `startupInternal()` throws, `startup()` logs the cause and leaves `running = false`. Then:

- **`write()` did not check it** — kept accepting events and dereferencing state startup never assigned (NPE per event, burying the original cause).
- **`shutdown()` only ran `shutdownInternal()` when `running`** — so resources acquired during a failed startup (or in the manager factory) were never released. Returns `true` anyway.

This change:

1. Makes `write()` return early when `!isRunning()`, logging a single status warning instead of one failure per event.
2. Makes `shutdown()` always invoke `shutdownInternal()` once (tracked independently of `running`), so partial startup state can be cleaned up.
3. Makes `CassandraManager.shutdownInternal()` null-safe when `session` was never created (motivating failure mode from the issue / related #4242).

## Checklist

* Tests are provided

## Testing

- JDK 17 (branch enforcer requires `[17,18)`) — JBR 17.0.14
- `./mvnw test -pl :log4j-core-test -am -Dtest=AbstractDatabaseManagerTest -Dsurefire.failIfNoSpecifiedTests=false` — **11/11 pass**
  - New: `testFailedStartupSkipsWrite`, `testFailedStartupStillShutsDown`, `testShutdownWithoutStartupStillRunsShutdownInternal`
- Spotless check clean on `:log4j-core`, `:log4j-core-test`, `:log4j-cassandra`

### DIFF
--- a/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
+++ b/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
@@ -6,6 +6,7 @@
            https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
        type="fixed">
   <issue id="4241" link="https://github.com/apache/logging-log4j2/issues/4241"/>
+  <issue id="4242" link="https://github.com/apache/logging-log4j2/issues/4242"/>
   <issue id="4246" link="https://github.com/apache/logging-log4j2/pull/4246"/>
   <description format="asciidoc">
     Fix `AbstractDatabaseManager` so a manager whose startup failed no longer accepts writes (which could NPE per event) and still runs `shutdownInternal()` to release resources acquired during startup.
--- a/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
+++ b/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
@@ -6,6 +6,7 @@
            https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
        type="fixed">
   <issue id="4241" link="https://github.com/apache/logging-log4j2/issues/4241"/>
+  <issue id="4246" link="https://github.com/apache/logging-log4j2/pull/4246"/>
   <description format="asciidoc">
     Fix `AbstractDatabaseManager` so a manager whose startup failed no longer accepts writes (which could NPE per event) and still runs `shutdownInternal()` to release resources acquired during startup.
   </description>
--- a/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
+++ b/src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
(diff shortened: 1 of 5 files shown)


src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4241" link="https://github.com/apache/logging-log4j2/issues/4241"/>
```
vy, line 8:
```suggestion
  <issue id="4241" link="https://github.com/apache/logging-log4j2/issues/4241"/>
  <issue id="4246" link="https://github.com/apache/logging-log4j2/pull/4246"/>
```

vy (approved):

src/changelog/.2.x.x/4241_fix_AbstractDatabaseManager_failed_startup.xml
```diff
@@ -0,0 +1,13 @@
...
```
vy, line 8:
```suggestion
  <issue id="4241" link="https://github.com/apache/logging-log4j2/issues/4241"/>
  <issue id="4242" link="https://github.com/apache/logging-log4j2/issues/4242"/>
```

vy (approved):

vy:
@arimu1, could you take care of the build failures, please?

arimu1:
Thanks for the review, @vy — CI should be green now.

`AbstractDatabaseAppenderTest.testAppend` was failing because the Mockito mock’s `isRunning()` is final and defaults to `false`, so `write()` returned early after the production guard. The test now uses a spy of a stub manager and calls `startup()` so the happy path still exercises `writeThrough()`.

Failed-startup behavior remains covered by `AbstractDatabaseManagerTest` (`testFailedStartupSkipsWrite`, `testFailedStartupStillShutsDown`, etc.).

PR #4247, opened 2026-08-08
Default `protocol` to UDP in the Log4j 1 `SyslogAppender` builder

### LINKED ISSUE #4238: Default `protocol` to UDP in the Log4j 1 `SyslogAppender` builder
`SyslogAppenderBuilder` currently defaults `protocol` to TCP. Log4j 1's `org.a.l.net.SyslogAppender` only ever supported UDP, and `protocol` has no counterpart in the Log4j 1 configuration format: it is a bridge extension.

This matters beyond fidelity. Over UDP the datagram boundary is the record boundary, so no character in the message requires escaping. Over TCP with RFC 6587 non transparent framing, LF is the record delimiter, and `Log4j1SyslogLayout` provides no mechanism to escape it. Defaulting to TCP therefore silently places legacy configurations on a transport the ported layout was never written for.

**Proposed changes:**

- Default `protocol` to UDP when the appender is built from a Log4j 1 configuration, matching Log4j 1 behaviour.
- Document `protocol` explicitly as a `log4j-1.2-api` extension, and note that non-UDP values are not recommended because `Log4j1SyslogLayout` does not escape newlines.
- Consider emitting a status logger warning when `protocol` is set to a stream transport.

All users should migrate away from a Log4j 1 configuration, but in particular users needing structured syslog over a framed transport should migrate to a Log4j 2 configuration and `Rfc5424Layout`, which implements newline escaping.

This was reported by @August829.

DragonFSKY:
Fixes #4238.

Defaults the Log4j 1 `SyslogAppender` bridge protocol to UDP for XML and properties configurations, matching Log4j 1 behavior. Explicit protocol values remain unchanged.

Also documents `protocol` as a `log4j-1.2-api` extension and why stream transports are not recommended with `Log4j1SyslogLayout`.

Verification:
- `./mvnw verify`

### DIFF
--- a/log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/SyslogAppenderBuilder.java
+++ b/log4j-1.2-api/src/main/java/org/apache/log4j/builders/appender/SyslogAppenderBuilder.java
@@ -77,7 +77,7 @@ public class SyslogAppenderBuilder extends AbstractBuilder implements AppenderBu
         final AtomicReference<String> facility = new AtomicReference<>();
         final AtomicReference<String> level = new AtomicReference<>();
         final AtomicReference<String> host = new AtomicReference<>();
-        final AtomicReference<Protocol> protocol = new AtomicReference<>(Protocol.TCP);
+        final AtomicReference<Protocol> protocol = new AtomicReference<>(Protocol.UDP);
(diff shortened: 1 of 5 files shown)


ppkarwasz (approved):
@DragonFSKY,

LGTM! :100: Thank you for your contribution.

PR #4249, opened 2026-08-12
Fix circular references in exceptions

vy:
Fix `Throwable` causal-chain handling for cyclic and identity-malfunctioning exceptions.

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/nosql/NoSqlDatabaseManagerTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/nosql/NoSqlDatabaseManagerTest.java
@@ -25,6 +25,7 @@ import static org.mockito.BDDMockito.given;
 import static org.mockito.BDDMockito.then;
 import static org.mockito.Mockito.mock;
 
+import foo.TestFriendlyException;
 import java.io.IOException;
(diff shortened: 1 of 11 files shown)


ramanathan1504 (commented):
@vy I reverted each production hunk locally and the cyclic cases are real: JPA converter → `StackOverflow`, `NoSqlDatabaseManagerTest` → `OutOfMemoryError` after 100s, `Throwables.getRootCause` → `r` instead of `r_c_c`.
`ThrowableExtendedStackTraceRenderer.java:115` — reverting only this hunk leaves all 1,643 pattern-converter tests green; `TestFriendlyException` is a single class so `classResourceInfoByName` comes out identical either way. Can one be made to fail without it, or is it defensive only?
`ThrowableProxy.java:119` — `causeVisited` never gets the root added, so `a.initCause(b); b.initCause(a)` prints `Caused by: java.lang.Exception: A` twice where `%xEx` prints `[CIRCULAR REFERENCE: ...]`. Out of scope here?
`src/changelog/.2.x.x/4249_fix-circular-exception.xml` — link #3933 too, the way the #4133 entry does?

vy:
@ramanathan1504, I've addressed your review remarks. See ceaff494c7 and eb75be4908.

ramanathan1504 (approved):
LGTM!

PR #4252, opened 2026-08-15
Honor spring.cloud.config.watch.enabled on Log4j2EventListener

### LINKED ISSUE #4244: Log4j2EventListener's @ConditionalOnProperty has no effect - the listener always runs
## Description

`Log4j2EventListener` carries `@ConditionalOnProperty("spring.cloud.config.watch.enabled")` but always runs.

It is registered twice:

- as `@Component` — the condition applies
- in `META-INF/spring.factories` (line 17) as an `ApplicationListener` — the condition **cannot** apply, because `SpringApplication` instantiates these directly, with no bean definition to suppress

So `spring.cloud.config.watch.enabled=false` disables nothing.

**Version:** 2.x `04c93c1d33` · JDK 21 · Spring Boot 3.4.3

## Logs

Property set to `false`, `monitorInterval` 300s:

```
RELOADED - the refresh event drove the reload, 298 seconds before
           the monitorInterval would have
```

298s ahead of the interval, so the event drove it, not the timer.

## Reproduction

Run with `-Dspring.cloud.config.watch.enabled=false` and publish an `EnvironmentChangeEvent` (POST `/actuator/refresh`). Log4j reconfigures anyway.

## Suggested fix

Check the property inside `onApplicationEvent` against the `Environment` — smallest change, keeps the documented switch meaningful. Or drop `@ConditionalOnProperty` so the source stops promising a switch that does not exist.

AzazelSensei:
`Log4j2EventListener` is also created from `spring.factories`, so `@ConditionalOnProperty` never applies and `spring.cloud.config.watch.enabled=false` does nothing.

I check that property in `onApplicationEvent` (via `EnvironmentAware` or the event source context). Explicit `false` skips the reload. Missing or true keeps the old behavior.

Fixes #4244

## Checklist

* Tests are provided

### DIFF
--- a/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2EventListener.java
+++ b/log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2EventListener.java
@@ -20,17 +20,48 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cloud.context.environment.EnvironmentChangeEvent;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationListener;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
 @Component
 @ConditionalOnProperty(value = "spring.cloud.config.watch.enabled")
-public class Log4j2EventListener implements ApplicationListener<EnvironmentChangeEvent> {
+public class Log4j2EventListener implements ApplicationListener<EnvironmentChangeEvent>, EnvironmentAware {
     private static Logger LOGGER = LogManager.getLogger(Log4j2EventListener.class);
+    private Environment environment;
+
+    @Override
+    public void setEnvironment(final Environment environment) {
+        this.environment = environment;
+    }
 
     @Override
     public void onApplicationEvent(final EnvironmentChangeEvent environmentChangeEvent) {
+        if (!isWatchEnabled(environmentChangeEvent)) {
+            LOGGER.debug("Ignoring environment change event; spring.cloud.config.watch.enabled is false");
+            return;
+        }
         LOGGER.debug("Application change event triggered");
         WatchEventManager.publishEvent();
+    }
+
+    /**
+     * {@code spring.factories} constructs this listener outside the bean factory, so
+     * {@code @ConditionalOnProperty} never applies. Honor the same property here.
+     */
+    private boolean isWatchEnabled(final EnvironmentChangeEvent event) {
+        Environment env = this.environment;
+        if (env == null) {
+            final Object source = event.getSource();
+            if (source instanceof ApplicationContext) {
+                env = ((ApplicationContext) source).getEnvironment();
+            }
+        }
+        if (env == null) {
+            return true;
+        }
+        return !Boolean.FALSE.equals(env.getProperty("spring.cloud.config.watch.enabled", Boolean.class));
     }
 }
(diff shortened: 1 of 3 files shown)


vy:
`log4j-spring-cloud-config-client` uses JUnit 4 and, sadly, `pom.xml` is missing the `org.junit.vintage:junit-vintage-engine` test dependency. @ramanathan1504, can you take care of this one, please?

Maybe in a follow-up PR, we can migrate `log4j-spring-cloud-config-client` to JUnit 5. Though I'd not be surprised if it turns out to be a tough job.

ramanathan1504:
This `PR` Should land after #4317 because the module has run no tests since removed `junit-vintage-engine` while `Log4j2EventListenerTest stayed` on JUnit 4. @vy I will create separate `PR` for `Junit 5` migration

log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2EventListener.java
```diff
@@ -20,17 +20,48 @@
...
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationListener;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Component;
 
 @Component
 @ConditionalOnProperty(value = "spring.cloud.config.watch.enabled")
-public class Log4j2EventListener implements ApplicationListener<EnvironmentChangeEvent> {
+public class Log4j2EventListener implements ApplicationListener<EnvironmentChangeEvent>, EnvironmentAware {
```
ramanathan1504: `setEnvironment` is never called on the instance from `spring.factories`, and nothing scans this package for `@Component`. Can `EnvironmentAware` and the field go, so only the event source check is left?

log4j-spring-cloud-config-client/src/main/java/org/apache/logging/log4j/spring/cloud/config/client/Log4j2EventListener.java
```diff
@@ -20,17 +20,48 @@
...
+        if (env == null) {
+            final Object source = event.getSource();
+            if (source instanceof ApplicationContext) {
+                env = ((ApplicationContext) source).getEnvironment();
+            }
+        }
+        if (env == null) {
+            return true;
+        }
+        return !Boolean.FALSE.equals(env.getProperty("spring.cloud.config.watch.enabled", Boolean.class));
```
ramanathan1504: `getProperty` with `Boolean.class` throws `ConversionFailedException` for a value like `maybe`, and that fails `/actuator/refresh`. This matches `@ConditionalOnProperty`, which never throws.

```suggestion
        return !"false".equalsIgnoreCase(env.getProperty("spring.cloud.config.watch.enabled"));
```

src/changelog/.2.x.x/4244_honor_watch_enabled_on_event_listener.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4244" link="https://github.com/apache/logging-log4j2/issues/4244"/>
```
ramanathan1504, line 8: Add pr details like issue tag

ramanathan1504 (commented):
@AzazelSensei  The event source check covers `/actuator/refresh`, and `doesNotReloadWhenWatchDisabled` fails without the listener change once it uses `watchManager()`.
With `2.x` merged the test no longer compiles, because `loggerContextRule` is gone. In `Log4j2EventListener.java`, `getProperty` with `Boolean.class` throws for a value like `maybe`, and `setEnvironment` is never called on the `spring.factories` instance. Can we compare the string with `"false"` and drop `EnvironmentAware`?

AzazelSensei:
Dropped EnvironmentAware and the field. The check only uses the event source now, and I compare the property as a string so a value like maybe does not throw on /actuator/refresh. Changelog has the PR tag. doesNotReloadWhenWatchDisabled uses watchManager() so it compiles after the 2.x merge.

ramanathan1504:
@AzazelSensei thanks, all four points are in and the test fails without the listener change.                                                                                 `spotless:check` still fails on the `new Log4j2EventListener()` call in `Log4j2EventListenerTest.java`. Can you run `./mvnw spotless:apply` and push?

Please check the `Building.adoc` its has all the steps.

ramanathan1504 (approved):
LGTM!

PR #4256, opened 2026-08-27
Switch the default minimum allowed plugin processor message severity from NOTE to ERROR

### LINKED ISSUE #4225: Consider making plugin builds less chatty
Our build has custom plugins and this is the output we have on a regular Gradle build:

```
Note: GraalVmProcessor: writing GraalVM metadata for 12 Java classes to `META-INF/native-image/log4j-generated/org.springframework.boot/spring-boot-log4j/reflect-config.json`.
Note: PluginProcessor: writing plugin descriptor for 13 Log4j Plugins to `META-INF/org/apache/logging/log4j/core/config/plugins/Log4j2Plugins.dat`.
```

One thing that always tricks me is that I often forgets it is about log4j, especially reading "Note: GraalVmProcessor: writing GraalVM metadata for 12 Java classes". 

Can you please consider this to be less chatty? Do we really need that log by default? If we really do, can't we do something to make it a bit more obvious that it is log4j doing something during the build? Thank you

vy:
Fixes #4225

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessorPublicSetterTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessorPublicSetterTest.java
@@ -113,7 +113,8 @@ public class PluginProcessorPublicSetterTest {
     }
 
     @Test
-    void noteEmittedByDefault() {
+    void noteEmittedWhenConfigured() {
+        setupWithOptions("-A" + PluginProcessor.MIN_ALLOWED_MESSAGE_KIND_OPTION + "=NOTE");
         final List<Diagnostic<? extends JavaFileObject>> noteDiagnostics = diagnosticCollector.getDiagnostics().stream()
                 .filter(d -> d.getKind() == Diagnostic.Kind.NOTE)
                 .collect(Collectors.toList());
@@ -141,7 +142,7 @@ public class PluginProcessorPublicSetterTest {
 
     @ParameterizedTest
     @ValueSource(strings = {"NOTE", "note"})
-    void explicitNoteKindBehavesLikeDefault(final String kindValue) {
+    void explicitNoteKindEmitsNotes(final String kindValue) {
         setupWithOptions("-A" + PluginProcessor.MIN_ALLOWED_MESSAGE_KIND_OPTION + "=" + kindValue);
 
         assertThat(errorDiagnostics).anyMatch(d -> d.getMessage(Locale.ROOT)
@@ -154,17 +155,13 @@ public class PluginProcessorPublicSetterTest {
     }
 
     @Test
-    void invalidKindValueEmitsWarning() {
+    void invalidKindValueDoesNotEmitWarningByDefault() {
         setupWithOptions("-A" + PluginProcessor.MIN_ALLOWED_MESSAGE_KIND_OPTION + "=INVALID");
 
         final List<Diagnostic<? extends JavaFileObject>> warningDiagnostics =
                 diagnosticCollector.getDiagnostics().stream()
                         .filter(d -> d.getKind() == Diagnostic.Kind.WARNING)
                         .collect(Collectors.toList());
-        assertThat(warningDiagnostics)
-                .anyMatch(d -> d.getMessage(Locale.ROOT)
-                                .startsWith(
-                                        "[Log4j] org.apache.logging.log4j.core.config.plugins.processor.PluginProcessor:")
-                        && d.getMessage(Locale.ROOT).contains("unrecognized value `INVALID`"));
+        assertThat(warningDiagnostics).isEmpty();
     }
 }
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
@@ -84,7 +84,7 @@ public class GraalVmProcessor extends AbstractProcessor {
     private final Map<String, ReachabilityMetadata.Type> reachableTypes = new HashMap<>();
     private final List<Element> processedElements = new ArrayList<>();
     private Annotations annotationUtil;
-    private Diagnostic.Kind minAllowedMessageKind = Diagnostic.Kind.NOTE;
+    private Diagnostic.Kind minAllowedMessageKind = Diagnostic.Kind.ERROR;
 
     @Override
     public synchronized void init(ProcessingEnvironment processingEnv) {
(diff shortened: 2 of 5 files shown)


log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessorPublicSetterTest.java
```diff
@@ -154,17 +155,13 @@ void explicitNoteKindBehavesLikeDefault(final String kindValue) {
     }
 
     @Test
-    void invalidKindValueEmitsWarning() {
+    void invalidKindValueDoesNotEmitWarningByDefault() {
```
ramanathan1504, line 158: Can this go back to asserting the warning is printed?

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
```diff
@@ -102,7 +102,7 @@ public synchronized void init(ProcessingEnvironment processingEnv) {
                                 GraalVmProcessor.class.getName(),
                                 kindValue,
                                 PluginProcessor.MIN_ALLOWED_MESSAGE_KIND_OPTION,
-                                Diagnostic.Kind.NOTE,
+                                Diagnostic.Kind.ERROR,
                                 Arrays.toString(Diagnostic.Kind.values())));
```
ramanathan1504, line 106: This warning goes through the filter it is reporting on

```suggestion
                processingEnv
                        .getMessager()
                        .printMessage(
                                Diagnostic.Kind.WARNING,
                                MESSAGE_PREFIX
                                        + String.format(
                                                "%s: unrecognized value `%s` for option `%s`, using default `%s`. Valid values: %s",
                                                GraalVmProcessor.class.getName(),
                                                kindValue,
                                                PluginProcessor.MIN_ALLOWED_MESSAGE_KIND_OPTION,
                                                Diagnostic.Kind.ERROR,
                                                Arrays.toString(Diagnostic.Kind.values())));
```

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessor.java
```diff
@@ -114,7 +114,7 @@ public void init(final ProcessingEnvironment processingEnv) {
                                 PluginProcessor.class.getName(),
                                 kindValue,
                                 MIN_ALLOWED_MESSAGE_KIND_OPTION,
-                                Diagnostic.Kind.NOTE,
+                                Diagnostic.Kind.ERROR,
                                 Arrays.toString(Diagnostic.Kind.values())));
```
ramanathan1504, line 118: Same here, a typo in the value prints nothing.

```suggestion
                processingEnv
                        .getMessager()
                        .printMessage(
                                Diagnostic.Kind.WARNING,
                                MESSAGE_PREFIX
                                        + String.format(
                                                "%s: unrecognized value `%s` for option `%s`, using default `%s`. Valid values: %s",
                                                PluginProcessor.class.getName(),
                                                kindValue,
                                                MIN_ALLOWED_MESSAGE_KIND_OPTION,
                                                Diagnostic.Kind.ERROR,
                                                Arrays.toString(Diagnostic.Kind.values())));
```

ramanathan1504 (commented):
@vy 
Does `plugins.adoc` need a change too? It still describes the notes as on by default.

Changing the default instead of removing the notes keeps the option useful for people who still want them, and `notesSuppressedByDefaultWithoutAffectingMetadataGeneration` shows the descriptor files are still written.

One thing has no line to comment on. The collapsible block in `plugins.adoc`, "Suppressing annotation processor notes in strict build environments", still says both processors emit `NOTE`-level diagnostics by default and that this option turns them off, and it lists `NOTE (default)` among the accepted values. The Maven and Gradle snippets under it pass `=WARNING`, which suppresses nothing now.

`CI` also failing because of spotless not applied correctly.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessorPublicSetterTest.java
vy: What do you mean? This test verifies the default behavior, which is to emit no warnings.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
vy: I'm confused with your suggestion. Mind elaborating?

vy:
> Does `plugins.adoc` need a change too?

@ramanathan1504, good call. Fixed in 5402ef3791.

vy:
> `CI` also failing because of spotless not applied correctly.

@ramanathan1504, right. Fixed in d1a559ac94.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
ramanathan1504: @vy If someone mistypes the option value, this warning is dropped by the `ERROR` default, so they get no hint. Sending it straight to `processingEnv.getMessager()` avoids that.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessorPublicSetterTest.java
ramanathan1504: @vy It passes `=INVALID`, so this is a mistyped value, not the default

vy: I'm not seeing what you say. The method name is _"invalid kind value does not emit warning by default"_. It is not saying _"the value is the default"_.

vy: Anyway, changed in 418976ab09.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/GraalVmProcessor.java
vy: Got it. Fixed in 418976ab09.

log4j-core/src/main/java/org/apache/logging/log4j/core/config/plugins/processor/PluginProcessor.java
vy: Fixed in 418976ab09.

src/site/antora/modules/ROOT/pages/manual/plugins.adoc
```diff
@@ -218,12 +218,11 @@ Provide these values to the processor using the `log4j.graalvm.groupId` and `log
 .Suppressing annotation processor notes in strict build environments
```
ramanathan1504, line 218: Title still says suppressing, but this now turns notes on

```suggestion
.Enabling annotation processor notes
```

ramanathan1504 (approved):
One small thing left inline on the `plugins.adoc` title. Thanks 🙏

PR #4264, opened 2026-08-30
Document XML parser features and clarify Log4j 1 bridge properties

### LINKED ISSUE #4263: Document XML parser features used for configuration files and clarify Log4j 1 bridge properties
## Context

While triaging a report received on our security mailing list (tracking reference: `https://lists.apache.org/thread/k38tg3hsbo7kl17psybp02nzl28lwloq`), we noticed that the features enabled on the XML parsers used to read configuration files are not documented, and that the relationship between the two Log4j 1 bridge configuration properties can be misread when their documentation entries are consulted out of context (e.g., through a deep link to a single anchor).

No behavior change is requested here; this issue only asks for documentation improvements.

## Document the XML parser features enabled for each configuration format

The documentation should state explicitly which XML features are enabled when parsing configuration files:

- **Log4j 1 XML configuration files** (handled by the `log4j-1.2-api` bridge, see the [Migrating from Log4j 1](https://logging.apache.org/log4j/2.x/migrate-from-log4j1.html) page): the files are parsed with **external entities enabled** for backward compatibility with the original Log4j 1 parser. **DTD validation is enabled** and parsing errors are printed as warnings to the status logger.
- **Log4j 2 XML configuration files** (see the [XML format](https://logging.apache.org/log4j/2.x/manual/configuration.html#xml-features) section of the Configuration page): **XInclude is enabled when available**, i.e., when the JAXP implementation supports it. The existing XInclude section describes how to use the feature, but the format description should state up front that the feature is active by default, since it affects which files a configuration file may pull in.

## Clarify the independence of the two Log4j 1 bridge properties

The [Migrating from Log4j 1](https://logging.apache.org/log4j/2.x/migrate-from-log4j1.html) page documents `log4j1.compatibility` and `log4j.configuration` under the sentence "set **one of the following** configuration properties", which correctly presents them as two independent, alternative ways to enable runtime conversion of Log4j 1 configuration files. However, each property's entry does not restate this, so a reader landing directly on the `log4j1.compatibility` anchor (for example via the link printed by the `DOMConfigurator`/`PropertyConfigurator` warning message) can incorrectly conclude that `log4j1.compatibility` is the sole opt-in gate for the Log4j 1 configuration file support.

ppkarwasz:
This is a documentation-only improvement prompted by a report on the security mailing list (see #4263 for context):

- Adds an "External entities and validation" section to the migration guide: external XML entities are enabled when parsing Log4j 1 configuration files (kept by the bridge for backward compatibility) but not Log4j 2 ones, with guidance to inline entities or convert them to XIncludes, and the DTD-vs-XML-Schema validation difference. The threat model is linked next to each capability statement, so the statements cannot be quoted out of context.
- States the Log4j 2 XML parser features (no external DTD/entity retrieval, XInclude enabled when available) at the top of the XML format documentation.
- Makes the `log4j.configuration` and `log4j1.compatibility` property entries self-contained, so neither can be misread as depending on the other when deep-linked.
- Corrects the documented environment variable of both properties: `log4j.configuration` maps to `LOG4J_CONFIGURATION` (not `LOG4J_CONFIGURATION_FILE`, which maps to `log4j.configurationFile`) and `log4j1.compatibility` maps to `LOG4J_1_COMPATIBILITY` (not `LOG4J_COMPATIBILITY`); both verified against `PropertiesUtil`.
- Improves the class Javadoc of both `XmlConfiguration` classes, corrects their `XXE_DOCUMENT` suppression justifications, and bumps the `org.apache.log4j.xml` and `org.apache.logging.log4j.core.config.xml` package versions by a micro version accordingly.

User-visible change: the `DOMConfigurator`/`PropertyConfigurator` warning messages now link to `logging.apache.org` instead of `logging.staged.apache.org` (staging link shipped since `2.24.0`).

Documentation for behavior that first ships in `2.27.0` (#4198) is deliberately excluded; it follows in a separate stacked PR (#4265).

Closes #4263

https://claude.ai/code/session_01LNsw2hhNuJ6tjsEU2tChnn


log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
```diff
@@ -131,7 +147,8 @@ public void doConfigure() throws FactoryConfigurationError {
             @Override
             @SuppressFBWarnings(
                     value = "XXE_DOCUMENT",
-                    justification = "The `DocumentBuilder` is configured to not resolve external entities.")
+                    justification =
+                            "External entities are resolved by `Log4jEntityResolver` through `ConfigurationSource`, which restricts their locations to the protocols allowed by the `log4j2.Configuration.allowedProtocols` property.")
```
ramanathan1504: `ConfigurationSource.fromUri` returns a `FileInputStream` for `file:` and schemeless URIs and a classloader resource for `classloader:`/`classpath:` ones, all before `UrlConnectionFactory` is reached — so the allowlist only covers the URL branch, and a `file:` entity still resolves with `log4j2.Configuration.allowedProtocols=_none`.

```suggestion
                            "External entities are resolved by `Log4jEntityResolver` through `ConfigurationSource`, the same way the configuration file itself is resolved. Configuration files must come from trusted sources.")
```

src/site/antora/modules/ROOT/partials/manual/configuration-xml-format.adoc
```diff
@@ -18,6 +18,17 @@
 [id=xml-features]
 = XML format
 
+Log4j 2 parses XML configuration files with the following parser features:
+
+* DTD validation and the retrieval of external DTDs and external entities are **disabled**.
```
ramanathan1504, line 23: The parser also sets `setExpandEntityReferences(false)`, and `XmlConfiguration` only walks `Element`/`Text` children — so an *internal* entity in element content is dropped without a warning, which is worth stating here given the migration guide in this same diff tells Log4j 1 users to inline their entities.

```suggestion
* DTD validation and the retrieval of external DTDs and external entities are **disabled**, and entity references are not expanded.
```

ramanathan1504 (changes requested):
log4j-1.2-api/src/main/java/org/apache/log4j/PropertyConfigurator.java:155 (question, no fence)
Cannot be a suggestion: the remaining occurrence is in a different file, outside every hunk.

`grep -rn logging.staged.apache.org` over the repo leaves one more after this PR — `src/site/antora/modules/ROOT/pages/faq.adoc:324` — worth sweeping it in the same commit?

src/site/antora/modules/ROOT/partials/manual/configuration-xml-format.adoc
ppkarwasz: I think that setting `setExpandEntityReferences(false)` is a bug we should correct and #4162 corrects it.

The only effect of the setting is as you describe: the text node is wrapped in an `EntityReference` and we drop that node. It has **no** impact on fetching external references, only on the internal ones and I would leave it as implementation details **without** documenting it.

ramanathan1504:
@ppkarwasz 
One thing left besides the `faq.adoc` sweep: the warning text in `PropertyConfigurator.java` and `DOMConfigurator.java` is user-visible, so does it want a `src/changelog/.2.x.x` entry alongside it?

ppkarwasz:
The warning is more of a debugging tool: both `java.util.logging` and Log4j 1 has methods that can **change** the logging configuration automatically, which were *active* by default before version `2.24.0`. We disabled them, because some old and deeply nested transitive dependencies were “helping” application developers in configuring logging (e.g. Nashorn).

Application developers usually don't need help with that. Those that were relying on it, will enable debug status logging and see the link to the **correct** documentation. I don't think a changelog entry is necessary for correcting a typo in a warning message.

ramanathan1504:
@ppkarwasz  `Changelog` entry missing

ppkarwasz:
Same as in #4265: the only changes in code are to some compile-time annotations, which are not used at runtime.

vy (approved):

ramanathan1504 (approved):

PR #4265, opened 2026-08-30
Document location restrictions on Log4j 1 external entities

ppkarwasz:
Stacked on #4264; only the last commit belongs to this PR.

Documents the behavior introduced by #4198, which first ships in `2.27.0`: the external subset and other external entities in Log4j 1 XML configuration files are resolved through `ConfigurationSource`, so they can only be retrieved from locations allowed by the `log4j2.configurationAllowedProtocols` property. Adds a note to the migration guide and a paragraph to the Javadoc of the bridge `XmlConfiguration`.

Kept separate from #4264 so that the documentation of unreleased behavior can be merged (or held back) independently. After #4264 merges, the base can be retargeted to `2.x`.

https://claude.ai/code/session_01LNsw2hhNuJ6tjsEU2tChnn


Yoyis8836 (approved):

log4j-1.2-api/src/main/java/org/apache/log4j/xml/XmlConfiguration.java
```diff
@@ -77,6 +78,12 @@
  *     Parsing and validation errors do not stop the configuration process; they are printed as warnings to the status
  *     logger.
  * </p>
+ * <p>
+ *     Since version <strong>2.27.0</strong>, external entities are resolved through
+ *     {@link ConfigurationSource#fromUri(URI)}, so they can only be retrieved from locations allowed by the
```
ramanathan1504: Same wording as the migration guide: `fromUri` resolves local files before reaching `UrlConnectionFactory`, so the property does not gate `file:` entities.

````suggestion
 *     {@link ConfigurationSource#fromUri(URI)}, so they can only be retrieved from local files or over the protocols allowed by the
````

File: src/site/antora/modules/ROOT/pages/migrate-from-log4j1.adoc [hunk 2]
```diff
@@ -360,6 +360,13 @@ and
 xref:manual/configuration.adoc#configuration-attribute-schema[`schema`]
 attributes of the XML configuration format.
 
+[NOTE]
+====
+Since version `2.27.0`, the external subset and other external entities can only be retrieved from locations allowed by the
```
ramanathan1504: `allowedProtocols` only gates the URL branch of `ConfigurationSource.fromUri` — local files are opened first (`ConfigurationSource.java:318-321`), so a `file:` entity still resolves even under `_none`, which the properties page also notes.

````suggestion
Since version `2.27.0`, the external subset and other external entities can only be retrieved from local files or over the protocols allowed by the
````

ramanathan1504 (commented):
@ppkarwasz Changelog is missing

ppkarwasz:
> Changelog is missing

This is documentation only, so I wouldn't add any changelog. For me a changelog is necessary, so users can rapidly find a change that might have broken their application. In this case, no code changes are present.

ramanathan1504:
> > Changelog is missing
> 
> 
> 
> This is documentation only, so I wouldn't add any changelog. For me a changelog is necessary, so users can rapidly find a change that might have broken their application. In this case, no code changes are present.

ok understood👍

ramanathan1504 (approved):

Issue #3704, opened 2025-06-01
Add Support for Custom Headers in SMTP Appender


ppkarwasz:
Currently, the [SMTP Appender](https://logging.apache.org/log4j/2.x/manual/appenders/network.html#SmtpAppender) does not support adding custom headers to outgoing email messages. In contrast, the [HTTP Appender](https://logging.apache.org/log4j/2.x/manual/appenders/network.html#HttpAppender) allows users to define custom headers using `<Property>` elements. Introducing similar support for the SMTP Appender would enhance feature parity across appenders and improve user experience and flexibility.

### Proposed Solution

Both the HTTP and SMTP Appenders extend `AbstractFilterable` and already support `<Property>` configuration elements. In the HTTP Appender, these `<Property>` elements are interpreted as HTTP headers. Similarly, the SMTP Appender could treat `<Property>` entries as custom email headers.

**Example:**

```xml
<Appenders>
  <SMTP name="Email" ...>
    <Property name="X-Custom-ID" value="12345"/>
    <Property name="Reply-To" value="noreply@example.com"/>
  </SMTP>
</Appenders>
```

This change would preserve existing behavior while enabling greater customization for advanced email use cases.

sidhantmourya:
Hello @ppkarwasz ,

I'd like to contribute to this enhancement.

Is this issue currently open for contribution? If so, could you please guide me on where to start in the codebase (such as which classes or modules are involved)? Also, are there any recommended steps for setting up the project locally, including running and debugging it? I'm particularly interested in any tips for working with a large codebase like this, as it's my first time contributing to a project of this scale.

Thank you.

ppkarwasz:
Hi @sidhantmourya,

Thanks for digging into this!

I must admit that when I introduced the Jakarta version of the SMTP appender, I only added a minimal abstraction for `MailManager`. In hindsight, I could have refactored more shared logic into the abstract base class. One reason I held back was that `SmtpManager` exposes Java EE Mail types in its public and protected methods, which we can’t remove without breaking backward compatibility.

That said, if you'd like to take on abstracting `MimeMessage`, feel free — it would be a useful improvement. Alternatively, you're also welcome to duplicate the logic across both `MailManager` implementations. The Java EE version will be removed in Log4j 3.x anyway, so code duplication is a short-term compromise.

A few quick notes on the code you shared:

```java
setHeader.invoke(message, header.getName(), substitutor.replace(header.getValue()));
```

[...]

PR #4267, opened 2026-08-30
Custom header for smtp appender

ramanathan1504:
#3704 Added support for custom `SmtpAppender` message headers, configured with nested `Property` elements.


vy (changes requested):
I'm not convinced this feature should exist in the first place. See [my comment](https://github.com/apache/logging-log4j2/issues/3704#issuecomment-5508475629).

PR #4269, opened 2026-08-31
Reject dynamic proxies in `FilteredObjectInputStream`

ppkarwasz:
> [!IMPORTANT]

On Java 8, `FilteredObjectInputStream` is the only deserialization protection available, but dynamic proxy class descriptors are resolved through `ObjectInputStream#resolveProxyClass`, not `resolveClass`, so they bypassed the allowlist entirely. A stream could thus instantiate a proxy of arbitrary interfaces on the Java 8 code path.

On Java 9 and later, the JEP 290 filter already rejects proxies, because their synthetic class names (`com.sun.proxy.$ProxyN`, `jdk.proxy1.$ProxyN`, …) never match the allowlist. This PR aligns the Java 8 path by overriding `resolveProxyClass` to reject proxies outright — no supported Log4j serialized form contains one.

The rejection also applies when extra allowed classes are supplied to the constructor; proxies are never legitimate in a Log4j stream, so there is no opt-out.

### DIFF
--- a/log4j-api/src/main/java/org/apache/logging/log4j/util/FilteredObjectInputStream.java
+++ b/log4j-api/src/main/java/org/apache/logging/log4j/util/FilteredObjectInputStream.java
@@ -71,6 +71,20 @@ public class FilteredObjectInputStream extends ObjectInputStream {
         return super.resolveClass(desc);
     }
 
+    /**
+     * Unconditionally rejects dynamic proxy classes.
+     * <p>
+     *     Proxy class descriptors do not pass through {@link #resolveClass(ObjectStreamClass)}, so they would
+     *     otherwise bypass the allowlist entirely. No supported Log4j serialized form contains a dynamic proxy, and
+     *     the JEP 290 filter used on Java 9 and later rejects proxy classes as well, since their synthetic class names
+     *     never match the allowlist.
+     * </p>
+     */
+    @Override
+    protected Class<?> resolveProxyClass(final String[] interfaces) throws IOException, ClassNotFoundException {
+        throw new InvalidObjectException("Proxy classes are not allowed for deserialization");
+    }
+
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/FilteredObjectInputStream.java
```diff
@@ -71,6 +71,20 @@ protected Class<?> resolveClass(final ObjectStreamClass desc) throws IOException
         return super.resolveClass(desc);
     }
 
+    /**
+     * Unconditionally rejects dynamic proxy classes.
+     * <p>
+     *     Proxy class descriptors do not pass through {@link #resolveClass(ObjectStreamClass)}, so they would
+     *     otherwise bypass the allowlist entirely. No supported Log4j serialized form contains a dynamic proxy, and
+     *     the JEP 290 filter used on Java 9 and later rejects proxy classes as well, since their synthetic class names
+     *     never match the allowlist.
```
ramanathan1504, line 80: This is true only with the default filter. `DefaultObjectInputFilter.checkInput` asks the process-wide `jdk.serialFilter` first and returns its answer. So a permissive one still lets a proxy through on Java 9+.

```suggestion
     *     {@code DefaultObjectInputFilter} rejects proxy classes on Java 9 and later, since their synthetic class
     *     names never match the allowlist, unless a process-wide serial filter allows them first.
```

ramanathan1504:
@ppkarwasz 
 I removed the `resolveProxyClass` override and ran the test. Both proxy tests fail with "nothing was thrown". So they test the new override, they do not trip over the handler
  class.                                                                                                                                                                        
                                                                                                                                                                                
  One correction for the Javadoc, left inline.

PR #4270, opened 2026-08-31
Apply the deserialization filter in test helpers on Java 9+

ppkarwasz:
> [!IMPORTANT]

The `SerialUtil` test helper returned a plain, unfiltered `ObjectInputStream` on Java 9 and later, so our test suite only exercised the deserialization allowlist when running on a Java 8 toolchain. This made it possible for a change to the serialized forms to pass CI while being rejected by the filter at runtime.

`SerialUtil` now installs `DefaultObjectInputFilter` on Java 9+ (reflectively, since the helper must still compile and run on Java 8), combined with an `ObjectInputFilter.Config.createFilter` delegate built from the caller-supplied extra allowed classes — reproducing `FilteredObjectInputStream`'s semantics on modern JDKs. When `DefaultObjectInputFilter` is unavailable — it only exists in the packaged multi-release `log4j-api` jar, not in classes-dir reactor builds — the helper falls back to `FilteredObjectInputStream`, which enforces the same allowlist on any Java version.

`SerializationTestHelper` in `log4j-1.2-api` now delegates to `SerialUtil`, so its `org.apache.log4j.*` extras are honored on all Java versions as well.

Test-only change; no user-visible behavior is affected.

### DIFF
--- a/log4j-1.2-api/src/test/java/org/apache/log4j/util/SerializationTestHelper.java
+++ b/log4j-1.2-api/src/test/java/org/apache/log4j/util/SerializationTestHelper.java
@@ -30,8 +30,7 @@ import java.io.ObjectOutputStream;
 import java.util.Arrays;
 import java.util.Collection;
 import org.apache.commons.io.FileUtils;
-import org.apache.logging.log4j.util.Constants;
-import org.apache.logging.log4j.util.FilteredObjectInputStream;
+import org.apache.logging.log4j.test.junit.SerialUtil;
 
 /**
  * Utiities for serialization tests.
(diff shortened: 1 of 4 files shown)


ramanathan1504:
I checked the Java 9+ branch with a packaged `log4j-api` jar. `Config.createFilter` returns UNDECIDED for names it does not list, so the default allowlist still runs after it. `java.io.File` is rejected with no extras and allowed when passed as one. Arrays of it too.

Nothing tests it though. On `2.x` today `SerialUtil.getObjectInputStream` returns a plain `ObjectInputStream` on JDK 17 and reads a `java.io.File` back fine. One `assertThrows(IOException.class, ...)` in `log4j-api-test` fails before this change and passes after, on Java 8 and Java 9+ both.

PR #4272, opened 2026-08-31
Bound array allocation in `SortedArrayStringMap` deserialization

ppkarwasz:
> [!IMPORTANT]

`SortedArrayStringMap.readObject` allocated its key and value arrays at the capacity declared in the stream before reading a single entry, so a forged capacity field caused an arbitrarily large allocation (`OutOfMemoryError`) — the classic serialization memory-amplification shape, where a few bytes of input commit the reader to megabytes of allocation.

The declared capacity is no longer trusted: deserialization pre-allocates at most 128 Ki entries (1 MiB per array on a typical 64-bit JVM) and grows the arrays as entries are actually read, so allocation is proportional to the data present in the stream. A `mappings` count exceeding the declared capacity is now rejected as an inconsistent stream.

The serialized form is unchanged — this only affects how a stream is read — so there is no compatibility impact in either direction.

Stacked on #4271.

### DIFF
--- a/log4j-api-test/src/test/java/org/apache/logging/log4j/util/SortedArrayStringMapTest.java
+++ b/log4j-api-test/src/test/java/org/apache/logging/log4j/util/SortedArrayStringMapTest.java
@@ -32,14 +32,18 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.InputStreamReader;
+import java.io.InvalidObjectException;
+import java.io.ObjectInputStream;
 import java.lang.reflect.Field;
 import java.net.URL;
 import java.net.URLDecoder;
+import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
 import java.util.Arrays;
 import java.util.ConcurrentModificationException;
 import java.util.HashMap;
 import java.util.Map;
+import org.apache.logging.log4j.test.junit.SerialUtil;
 import org.junit.jupiter.api.Test;
 
 /**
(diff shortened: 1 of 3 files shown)


log4j-api/src/main/java/org/apache/logging/log4j/util/SortedArrayStringMap.java
```diff
@@ -513,18 +521,30 @@ private void readObject(final java.io.ObjectInputStream s) throws IOException, C
...
+        if (mappings > capacity) {
+            throw new InvalidObjectException("Illegal mappings count: " + mappings + " for capacity: " + capacity);
+        }
 
-        // allocate the bucket array;
         if (mappings > 0) {
-            inflateTable(capacity);
+            // Do not trust the declared capacity: allocate a bounded amount up front
+            // and resize as entries are actually read.
+            int allocated = Math.min(capacity, MAX_DESERIALIZATION_CAPACITY);
```
FreeAndNil, line 531: **Security** — Would `mappings` work as the bound here instead of `capacity`?

`mappings` is validated as `<= capacity` just above and is the tighter of the two. As it stands, a stream declaring `capacity = Integer.MAX_VALUE, mappings = 1` still allocates two 131072-element arrays (~1 MiB) for a single key, and `initFrom0` carries `threshold` into every later copy of the map. The growth loop would still cover `mappings > 1 << 17`. Unless there is a reason to prefer `capacity` that I am not seeing.

```suggestion
            int allocated = Math.min(mappings, MAX_DESERIALIZATION_CAPACITY);
```

log4j-api/src/main/java/org/apache/logging/log4j/util/SortedArrayStringMap.java
```diff
@@ -513,18 +521,30 @@ private void readObject(final java.io.ObjectInputStream s) throws IOException, C
...
+            threshold = allocated;
         } else {
-            threshold = capacity;
-        }
-
-        // Read the keys and values, and put the mappings in the arrays
-        for (int i = 0; i < mappings; i++) {
-            keys[i] = (String) s.readObject();
-            values[i] = SerializationUtil.readWrappedObject(s);
+            threshold = Math.min(capacity, MAX_DESERIALIZATION_CAPACITY);
```
FreeAndNil, line 547: **Security** — The empty-map branch keeps the declared capacity, so the first `putValue` allocates ~1 MiB.

`threshold` survives as 131072 and `inflateTable` acts on it at the first insert, for a map holding one entry. An honestly written empty map declares `capacity = 4`, so clamping this to `DEFAULT_INITIAL_CAPACITY` looks like it would cost nothing.

log4j-api-test/src/test/java/org/apache/logging/log4j/util/SortedArrayStringMapTest.java
```diff
@@ -119,6 +123,78 @@ void testSerializationOfNonSerializableValue() {
...
+    void testDeserializationDoesNotPreallocateDeclaredCapacity() {
+        final SortedArrayStringMap original = new SortedArrayStringMap();
+        original.putValue("a", "avalue");
+        original.putValue("B", null);
+        original.putValue("3", "3value");
+
+        // A forged stream declaring a huge capacity must not force a huge allocation.
+        final byte[] forged = patchCapacity(serialize(original), 4, 3, Integer.MAX_VALUE);
+        final SortedArrayStringMap copy = deserialize(forged);
+        assertEquals(original, copy);
```
FreeAndNil, line 158: **Testability** — Both forged-capacity tests compare contents only, so they do not pin the new bound.

They fail on the old code because `new String[Integer.MAX_VALUE]` throws, so they would also pass with a much larger bound, or with a fix that over-allocated. Asserting the length of `values` would pin it down - the class already reads that field reflectively at lines 923 and 1012.

log4j-api-test/src/test/java/org/apache/logging/log4j/util/SortedArrayStringMapTest.java
```diff
@@ -119,6 +123,78 @@ void testSerializationOfNonSerializableValue() {
...
+        final byte[] forged = patchCapacity(serialize(original), 4, 3, 2);
+        final ObjectInputStream ois = SerialUtil.getObjectInputStream(forged);
+        assertThrows(InvalidObjectException.class, ois::readObject);
+    }
+
+    @Test
+    void testDeserializationResizesBeyondInitialAllocation() {
+        final SortedArrayStringMap original = new SortedArrayStringMap();
+        // One entry more than the bounded initial allocation of the deserialized arrays
+        final int count = (1 << 17) + 1;
```
FreeAndNil, line 189: **Maintainability** — `count` repeats the value of `MAX_DESERIALIZATION_CAPACITY` without being tied to it.

If the constant is raised later, `count` drops below it and this test quietly stops exercising the growth branch while still passing. Deriving it reflectively, as above, would keep the two in step.

log4j-api/src/main/java/org/apache/logging/log4j/util/SortedArrayStringMap.java
```diff
@@ -57,6 +57,14 @@ public class SortedArrayStringMap implements IndexedStringMap {
      */
     private static final int DEFAULT_INITIAL_CAPACITY = 4;
 
+    /**
+     * The maximum number of entries pre-allocated during deserialization: each array of this size occupies
+     * 1 MiB on a typical 64-bit JVM.
```
FreeAndNil, line 62: **Maintainability** — Nit: with compressed oops an array of this size is 512 KiB, not 1 MiB.

1 MiB is the pair of arrays, or a single one with compressed oops disabled. Only worth mentioning because the figure is what justifies the value of the constant.

FreeAndNil (commented):
Findings outside the diff, which cannot be anchored to a line:

log4j-api/src/main/java/org/apache/logging/log4j/util/SortedArrayStringMap.java:516 — **Bug** — Minor, and older than this PR: `capacity == 0` passes the check and leaves an unusable map.

With `capacity = 0, mappings = 0` the first `putValue` goes `inflateTable(0)` -> `insertAt` -> `ensureCapacity` (`0 >= 0`) -> `resize(0)`, and then `keys[0] = key` throws `ArrayIndexOutOfBoundsException`. `writeObject` never emits 0, so only a forged stream gets there. Since the field is being validated here anyway, would rejecting 0 be worth adding?

FreeAndNil (approved):
just some nits

PR #4280, opened 2026-09-01
Restore UNCACHED thread name updates in AsyncLogger

### LINKED ISSUE #4275: AsyncLogger ignores Thread.setName() changes despite the UNCACHED strategy since 2.25.0
## Description

Hello!

After upgrading Log4j Core from `2.24.3` to `2.25.0`, I noticed a change in how a fully asynchronous logger captures thread names.

My application executes multiple tasks on reused worker threads. Each task temporarily changes the name of the worker thread using `Thread.setName()` so that the currently executing task can be identified in the logs.

With Log4j `2.24.3`, the `%t` pattern reflects the name assigned by each task. With Log4j `2.25.0`, it keeps showing the name assigned by the first task executed on that worker thread.

This happens even when `log4j2.asyncLoggerThreadNameStrategy` is explicitly set to `UNCACHED`.

## Environment

- Operating system: Linux
- JDK: OpenJDK 26
- LMAX Disruptor: `4.0.0`
- Log4j versions compared:
  - `2.24.3`
  - `2.25.0`

## Reproduction

The example uses a fixed thread pool with exactly one worker. This guarantees that both tasks run on the same reused `Thread` instance while assigning it a different name for each task.

### `AsyncThreadNameExample.java`

```java
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;

public final class AsyncThreadNameExample {

    static {
        System.setProperty(
                "log4j2.contextSelector",
                "org.apache.logging.log4j.core.async.AsyncLoggerContextSelector"
        );
        System.setProperty(
                "log4j2.asyncLoggerThreadNameStrategy",
                "UNCACHED"
        );
        System.setProperty(
                "log4j2.enableThreadlocals",
                "true"
        );
    }

    public static void main(String[] args) throws InterruptedException {
        ExecutorService executorService = Executors.newFixedThreadPool(1);

        executorService.submit(new NamedTask("first-job"));
        executorService.submit(new NamedTask("second-job"));

        executorService.shutdown();

        if (!executorService.awaitTermination(10, TimeUnit.SECONDS)) {
            throw new IllegalStateException("Executor did not terminate in time");
        }

        LogManager.shutdown();
    }

    private static final class NamedTask implements Runnable {

        private static final Logger LOGGER = LogManager.getLogger(NamedTask.class);

        private final String taskName;

        private NamedTask(String taskName) {
            this.taskName = taskName;
        }

        @Override
        public void run() {
            Thread thread = Thread.currentThread();
            String originalThreadName = thread.getName();

            try {
                thread.setName(taskName);
                LOGGER.info("Executing {}", taskName);
            } finally {
                thread.setName(originalThreadName);
            }
        }
    }
}
```

The system properties are set before the logger in `NamedTask` is initialized.

### `log4j2.xml`

```xml
<Configuration status="WARN">
    <Appenders>
        <Console name="Console">
            <PatternLayout pattern="[%t] %m%n"/>
        </Console>
    </Appenders>

    <Loggers>
        <Root level="INFO">
            <AppenderRef ref="Console"/>
        </Root>
    </Loggers>
</Configuration>
```

The following JAR files are placed in the current directory:

```text
disruptor-4.0.0.jar
log4j-api-2.24.3.jar
log4j-core-2.24.3.jar
log4j-api-2.25.0.jar
log4j-core-2.25.0.jar
```

### Run with Log4j 2.24.3

```bash
java \
  -cp ".:log4j-api-2.24.3.jar:log4j-core-2.24.3.jar:disruptor-4.0.0.jar" \
  AsyncThreadNameExample.java
```

Output:

```text
[first-job] Executing first-job
[second-job] Executing second-job
```

### Run with Log4j 2.25.0

```bash
java \
  -cp ".:log4j-api-2.25.0.jar:log4j-core-2.25.0.jar:disruptor-4.0.0.jar" \
  AsyncThreadNameExample.java
```

Output:

```text
[first-job] Executing first-job
[first-job] Executing second-job
```

## Expected behavior

When `log4j2.asyncLoggerThreadNameStrategy` is set to `UNCACHED`, each log event should capture the current value returned by `Thread.getName()`:

```text
[first-job] Executing first-job
[second-job] Executing second-job
```

## Actual behavior

With Log4j `2.25.0`, the thread-local `RingBufferLogEventTranslator` retains the thread name captured during the first task:

```text
[first-job] Executing first-job
[first-job] Executing second-job
```

## Possible cause

The behavior appears to have changed in commit
[`dc6c53ab7bb195f3878d3ee92cd0ff20634326f1`](https://github.com/apache/logging-log4j2/commit/dc6c53ab7bb195f3878d3ee92cd0ff20634326f1),
merged through [PR #3171](https://github.com/apache/logging-log4j2/pull/3171).

That commit removed the following call from both `AsyncLogger.logWithThreadLocalTranslator(...)` overloads:

```java
initTranslatorThreadValues(translator);
```

It also removed the method that refreshed the thread values when the `UNCACHED` strategy was selected:

```java
private void initTranslatorThreadValues(
        final RingBufferLogEventTranslator translator
) {
    if (THREAD_NAME_CACHING_STRATEGY == ThreadNameCachingStrategy.UNCACHED) {
        translator.updateThreadValues();
    }
}
```

The PR discussion explains that a `RingBufferLogEventTranslator` is always used by the same thread that created it.

While the `Thread` instance remains the same in this example, its name changes between tasks through `Thread.setName()`. Therefore, retaining the values captured when the translator was created produces a stale thread name.

The current documentation for
[`log4j2.asyncLoggerThreadNameStrategy`](https://logging.apache.org/log4j/2.x/manual/systemproperties.html#log4j2.asyncLoggerThreadNameStrategy)
states that `UNCACHED` disables caching of the result of `Thread.getName()`.

## Questions

Is the behavior introduced in Log4j `2.25.0` intentional?

If changes made through `Thread.setName()` are no longer expected to be reflected by fully asynchronous loggers, should the `UNCACHED` strategy and its documentation be updated accordingly?

Otherwise, could this be considered a regression in the thread-local translator path?

AndrewBystrov:
Fixes #4275.

Restore the current thread name for fully asynchronous `AsyncLogger` events when `log4j2.asyncLoggerThreadNameStrategy` is set to `UNCACHED`.

The thread-local translator now refreshes its thread values for the `UNCACHED` strategy before publishing each event. The `CACHED` strategy and the vararg translator behavior remain unchanged.

The regression test submits two tasks to a single-thread executor, renames the reused worker before each log call, and verifies that both log events contain their corresponding thread names. The test fails against the base revision and passes with this fix.

The test class was renamed from `AsyncLoggerTestUncachedThreadName` to `AsyncLoggerUncachedThreadNameTest` because the old name did not match Maven Surefire's default test discovery patterns. As a result, the existing test was not included in regular test runs, despite detecting the regression when executed explicitly.

## Verification

* The regression test passes in both configured Disruptor test executions
* `AsyncLoggerThreadNameStrategyTest`: 6 tests pass in both executions
* AsyncLogger test group: 82 tests, 0 failures, 1 skipped in both executions
* `./mvnw verify -DskipTests -pl :log4j-core,:log4j-core-test`
* `./mvnw -N validate`
* `git diff --check`

## Checklist

* [x] Changes are based on the `2.x` branch
* [x] Relevant Maven tests and verification checks succeed
* [x] A changelog entry is included in `src/changelog/.2.x.x`
* [x] A regression test is provided


ramanathan1504:
@AndrewBystrov 

Noted Give some time I will review and close.

Thanks for the Contribution

AndrewBystrov:
Hi @ramanathan1504, just a friendly follow-up on this PR. I understand reviews can take some time. Please let me know if there’s anything I can clarify or change on my side. Thanks!

ramanathan1504:
> Hi @ramanathan1504, just a friendly follow-up on this PR. I understand reviews can take some time. Please let me know if there’s anything I can clarify or change on my side. Thanks!

Surely, I will try to close this weekend.👍

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLogger.java
```diff
@@ -356,6 +358,13 @@ private void initTranslator(
                 );
     }
 
+    private void initTranslatorThreadValues(final RingBufferLogEventTranslator translator) {
+        // constant check should be optimized out when using default (CACHED)
```
ramanathan1504, line 362: The default strategy is `UNCACHED` on current JDKs, so this comment is wrong. Can we drop it?

```suggestion
```

log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLogger.java
```diff
@@ -356,6 +358,13 @@ private void initTranslator(
...
+        if (THREAD_NAME_CACHING_STRATEGY == ThreadNameCachingStrategy.UNCACHED) {
+            translator.updateThreadValues();
```
ramanathan1504, line 364: `updateThreadValues()` is still `@Deprecated`, and its Javadoc says the translator is only used on the thread that created it, which is the reason this call was removed. Should the deprecation go, or should the Javadoc mention `Thread.setName()`?

log4j-core-test/src/test/java/org/apache/logging/log4j/core/async/AsyncLoggerUncachedThreadNameTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    currentThread.setName(expectedThreadName);
+                    log.info(msg);
+                } finally {
+                    currentThread.setName(originalThreadName);
+                }
+                return expectedThreadName;
+            });
+            final String firstTaskThreadName = (String) firstTask.get();
+            final String secondTaskThreadName = (String) secondTask.get();
+            assertNotNull(firstTaskThreadName, "firstTaskThreadName");
```
ramanathan1504, line 96: Each task returns a string literal, so these two `assertNotNull` calls cannot fail. Can we remove them?

ramanathan1504 (changes requested):
@AndrewBystrov thanks, this puts back the `UNCACHED` refresh that #3171 removed, and the renamed test covers the reused worker thread from the issue.
Two small points are inline on `AsyncLogger.java` and one on the test. @ppkarwasz can you confirm whether `RingBufferLogEventTranslator#updateThreadValues` should stay deprecated now that it is called again?

AndrewBystrov:
Thanks for the feedback! I’ll address the requested changes. Regarding the updateThreadValues() deprecation and Javadoc, let’s wait for @ppkarwasz’s guidance before making any changes there.

AndrewBystrov:
Thanks for the feedback! I’ve addressed the requested changes in the latest commit. For the deprecation point, I kept updateThreadValues() deprecated and updated its Javadoc to clarify its internal use after Thread.setName(String)

AndrewBystrov:
Hi @ppkarwasz, just a friendly follow-up on this PR. I’ve addressed the review comments and kept updateThreadValues() deprecated while clarifying its internal use in the Javadoc. Could you please take a look when you have a chance? Thanks!

ramanathan1504:
> Hi @ppkarwasz, just a friendly follow-up on this PR. I’ve addressed the review comments and kept updateThreadValues() deprecated while clarifying its internal use in the Javadoc. Could you please take a look when you have a chance? Thanks!

@AndrewBystrov we already noted most probably we will close within a week. if you are interested there is so many issue are listed.
you will pick anyone and help us. I will review that too.. Thanks👍

PR #4286, opened 2026-09-03
Remove workflows that are unused on `2.25.x`

ppkarwasz:
Cleans `2.25.x` of workflows that are either obsolete or never run on a non-default branch.

GitHub evaluates `schedule`, `issues`, `issue_comment` and `workflow_run` triggers only on the default branch, so these copies were dead code:

- `close-stale` (`schedule`)
- `labeler` (`issues`, `issue_comment`)
- `develocity-publish-build-scans` (`workflow_run`, already removed from `2.x`)
- `process-dependabot` (`workflow_run`, the `2.x` copy handles PRs against every branch)

`merge-dependabot` used `pull_request_target` and therefore did run from this branch, but it has been superseded by the `analyze-dependabot`/`process-dependabot` pair. `analyze-dependabot` stays, since `pull_request` runs the file from the PR merge commit.

`deploy-site` now only stages the website for `release/2.25.*` branches. The staging site for `2.x` and the production site (from `2.x-site-pro`) are deployed by the workflow on `2.x`, so the `2.x`-guarded jobs never ran here.

No changelog entry, since this only touches CI configuration.

https://claude.ai/code/session_014QXFCr2hQv3zoVf7wo5Lab


ramanathan1504:
@ppkarwasz 

The same rule this PR applies to `close-stale` and `labeler` also catches one file that
isn't in the diff: `.github/workflows/codeql-analysis.yaml` still declares

    schedule:
      - cron: '32 12 * * 5'

at lines 28-29. `schedule` is only evaluated on the default branch, so that stanza never
fires here either. The file itself must stay -- its `push` and `pull_request` triggers are
branch-local and working -- but the `schedule:` block is dead weight, same as the workflows
being removed.

Worth noting the consequence before dropping it: it means the weekly CodeQL sweep only ever
covers `2.x`. This branch gets CodeQL on push and PR only. If that's the intended posture for
a maintenance branch, then removing the stanza just makes it explicit; if it isn't, the fix
belongs on `2.x` instead. Happy either way -- flagging it since this PR is precisely an audit
of which triggers fire on which branch.

Minor while you're in here: the `concurrency` comment on line 31 still reads "pushes to 2.x
run to completion" -- copy-paste from `2.x`.

ramanathan1504 (approved):

ppkarwasz:
> `schedule` is only evaluated on the default branch, so that stanza never fires here either.

Good point! :100:

Unfortunately I enabled auto-merge that fired on your approval. Can you create a follow-up PR?

ramanathan1504:
> > `schedule` is only evaluated on the default branch, so that stanza never fires here either.
> 
> Good point! 💯
> 
> Unfortunately I enabled auto-merge that fired on your approval. Can you create a follow-up PR?

ok, i will do that

PR #4288, opened 2026-09-03
Remove workflows that are unused on `2.26.x`

ppkarwasz:
Same clean-up as #4286, applied to `2.26.x`.

GitHub evaluates `schedule`, `issues`, `issue_comment` and `workflow_run` triggers only on the default branch, so these copies were dead code:

- `close-stale` (`schedule`)
- `labeler` (`issues`, `issue_comment`)
- `process-dependabot` (`workflow_run`, the `2.x` copy handles PRs against every branch)

`analyze-dependabot` stays, since `pull_request` runs the file from the PR merge commit.

`deploy-site` now only stages the website for `release/2.26.*` branches. The staging site for `2.x` and the production site (from `2.x-site-pro`) are deployed by the workflow on `2.x`, so the `2.x`-guarded jobs never ran here.

No changelog entry, since this only touches CI configuration.

https://claude.ai/code/session_014QXFCr2hQv3zoVf7wo5Lab


ramanathan1504:
Same note as on #4286: `.github/workflows/codeql-analysis.yaml` (not in this diff) still has

    schedule:
      - cron: '32 12 * * 5'

at lines 28-29, which never fires off the default branch -- the same reason `close-stale` and
`labeler` are being removed here. The file stays, since `push`/`pull_request` are branch-local;
only the `schedule:` block is inert.

The consequence is that the weekly CodeQL sweep covers `2.x` alone, never this branch. Fine if
that's intended for a maintenance branch, but worth making explicit either way.

Also line 31's `concurrency` comment still says "pushes to 2.x run to completion".

ramanathan1504 (approved):

ppkarwasz:
> ```
> schedule:
>   - cron: '32 12 * * 5'
> ```

Sorry about this too. Please create a follow-up.

ramanathan1504:
> > ```
> > schedule:
> >   - cron: '32 12 * * 5'
> > ```
> 
> Sorry about this too. Please create a follow-up.

ok

PR #4307, opened 2026-09-07
  Use `TIMESTAMP` instead of the removed `DATETIME` alias in the H2 tests                                                                                                       

ramanathan1504:
Relates to #4292. Body carries the two-jar reproduction and separates the three sites: two block the build, one (JdbcAppenderBenchmark, src/main) doesn't.


ppkarwasz (approved):
LGTM! :100:

PR #4316, opened 2026-09-09
escape scope attribute names and values in DumpTag

jmestwa-coder:
`DumpTag` (`log:dump`) writes the names and values of every attribute in the selected scope straight into the page:

- request- and session-scope attributes are routinely attacker-influenced (a request parameter copied into a request attribute, a form bean, etc.), so any page using `log:dump` with those scopes emits reflected/stored XSS
- `doEndTag` concatenated the attribute `name` and `value` into the HTML output with no escaping
- both are now escaped with `StringBuilders.escapeXml` before writing, the same escaping `HtmlLayout` and `Log4j1XmlLayout` already apply to event data

## Checklist

- [x] Base your changes on the `2.x` branch
- [x] `./mvnw verify` succeeds for the affected module
- [x] Changelog entry added under `src/changelog/.2.x.x`
- [x] Tests are provided

### DIFF
--- a/log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
+++ b/log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
@@ -22,6 +22,7 @@ import javax.servlet.jsp.JspException;
 import javax.servlet.jsp.PageContext;
 import javax.servlet.jsp.tagext.Tag;
 import javax.servlet.jsp.tagext.TagSupport;
+import org.apache.logging.log4j.util.StringBuilders;
 
 /**
  * This class implements the {@code <log:dump>} tag.
@@ -60,8 +61,8 @@ public class DumpTag extends TagSupport {
                 final String name = names.nextElement();
                 final Object value = this.pageContext.getAttribute(name, this.scope);
 
-                this.pageContext.getOut().write("<dt><code>" + name + "</code></dt>");
-                this.pageContext.getOut().write("<dd><code>" + value + "</code></dd>");
+                this.pageContext.getOut().write("<dt><code>" + escapeHtml(name) + "</code></dt>");
+                this.pageContext.getOut().write("<dd><code>" + escapeHtml(String.valueOf(value)) + "</code></dd>");
             }
             this.pageContext.getOut().write("</dl>");
         } catch (final IOException e) {
@@ -69,5 +70,11 @@ public class DumpTag extends TagSupport {
         }
 
         return Tag.EVAL_PAGE;
+    }
+
+    private static String escapeHtml(final String value) {
+        final StringBuilder builder = new StringBuilder(value);
+        StringBuilders.escapeXml(builder, 0);
+        return builder.toString();
     }
 }
--- a/src/changelog/.2.x.x/fix_taglib_dump_tag_xss.xml
+++ b/src/changelog/.2.x.x/fix_taglib_dump_tag_xss.xml
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4316" link="https://github.com/apache/logging-log4j2/pull/4316"/>
+  <description format="asciidoc">
+    HTML-escape scope attribute names and values rendered by the `log:dump` JSP tag (`DumpTag`) to prevent cross-site scripting
+  </description>
+</entry>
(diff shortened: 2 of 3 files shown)


jmestwa-coder:
any update?

ramanathan1504:
@jmestwa-coder Thanks for yours patience. i will review ASAP. Thanks for the contribution

log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
```diff
@@ -60,8 +61,8 @@ public int doEndTag() throws JspException {
                 final String name = names.nextElement();
                 final Object value = this.pageContext.getAttribute(name, this.scope);
 
-                this.pageContext.getOut().write("<dt><code>" + name + "</code></dt>");
-                this.pageContext.getOut().write("<dd><code>" + value + "</code></dd>");
+                this.pageContext.getOut().write("<dt><code>" + escapeHtml(name) + "</code></dt>");
+                this.pageContext.getOut().write("<dd><code>" + escapeHtml(String.valueOf(value)) + "</code></dd>");
```
ramanathan1504:
```suggestion
                this.pageContext.getOut().write("<dd><code>" + escapeHtml(value) + "</code></dd>");
```

log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
```diff
@@ -70,4 +71,10 @@ public int doEndTag() throws JspException {
 
         return Tag.EVAL_PAGE;
     }
+
+    private static String escapeHtml(final String value) {
+        final StringBuilder builder = new StringBuilder(value);
+        StringBuilders.escapeXml(builder, 0);
+        return builder.toString();
+    }
```
ramanathan1504:
~~~suggestion
    private static String escapeHtml(final Object value) {
        final StringBuilder builder = new StringBuilder().append(value);
        StringBuilders.escapeXml(builder, 0);
        return builder.toString();
    }
~~~

ramanathan1504:
@jmestwa-coder 
Thanks. The escaping covers both the name and the value, and the new test fails without the fix.
One small thing in `DumpTag`: if an attribute's `toString()` returns null, `new StringBuilder(String.valueOf(value))` now throws an NPE where the old code printed `null`. Can `escapeHtml` take the `Object` and use `new StringBuilder().append(value)`?

jmestwa-coder:
Good catch. Pushed 3e104a4:

- `escapeHtml` now takes the `Object` and builds it with `new StringBuilder().append(value)`, so a null `toString()` renders as `null` like the old concatenation did instead of throwing
- the call site passes `value` directly, no more `String.valueOf`

`DumpTagTest` still passes and spotless is clean.

ramanathan1504 (changes requested):
One small thing in `DumpTag`: if an attribute's `toString()` returns null, `new StringBuilder(String.valueOf(value))` now throws an NPE where the old code printed `null`. Can `escapeHtml` take the `Object` and use `new StringBuilder().append(value)`?

log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
jmestwa-coder: Already applied in 3e104a4, the call site passes `value` directly now.

log4j-taglib/src/main/java/org/apache/logging/log4j/taglib/DumpTag.java
jmestwa-coder: Already in 3e104a4: `escapeHtml(final Object value)` with `new StringBuilder().append(value)`, same as the suggestion.

jmestwa-coder:
@ramanathan1504 looks like the changes-requested review was submitted after 3e104a4 went up, so both suggestions were already on the branch by then:

- `escapeHtml` takes `Object` and builds with `new StringBuilder().append(value)`, so a null `toString()` renders as `null` instead of throwing
- the call site passes `value` directly

Head matches both suggestions verbatim, `DumpTagTest` passes locally and CI is green. Could you take another look so the review can be cleared?

src/changelog/.2.x.x/fix_taglib_dump_tag_xss.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4316" link="https://github.com/apache/logging-log4j2/pull/4316"/>
+  <description format="asciidoc">
+    HTML-escape scope attribute names and values rendered by the `log:dump` JSP tag (`DumpTag`) to prevent cross-site scripting
```
ramanathan1504:
```suggestion
    Fix HTML-escaped `log:dump` JSP tag attributes to prevent XSS
```

ramanathan1504 (changes requested):

ramanathan1504 (approved):

ramanathan1504 (approved):

ramanathan1504 (approved):

PR #4319, opened 2026-09-11
Clear the database appender buffer when committing fails

### LINKED ISSUE #4318: Memory leak in JdbcAppender when a database failure occurs during the commit/close phase.
A memory leak occurs when using the JdbcAppender with a configured bufferSize. When the appender attempts to flush the buffer, it calls AbstractDatabaseManager.flush(), which in turn executes the SQL statements and calls commitAndClose().

If commitAndClose() throws an exception (e.g., due to a temporary database disconnection, transaction timeout, or constraint violation), the exception propagates out of the flush() method before buffer.clear() is executed. This results in:
1. Duplicate Events: The events remain in the buffer and will be processed again in the next flush attempt.
2. Memory Exhaustion (OOM): New events continue to be added to the internal buffer (or the buffer never gets a chance to reset its state), eventually leading to an OutOfMemoryError as the application continues to log.

Steps to Reproduce
1. Configure a JdbcAppender in log4j2.xml with a bufferSize (e.g., bufferSize="100").
2. Start the application and ensure it logs to the database correctly.
3. Induce a database failure (e.g., stop the database service or revoke the user's insert permissions).
4. Generate enough log events to trigger multiple buffer flushes.
5. Observe via a heap dump or memory profiler that the ArrayList (or equivalent buffer) inside the DatabaseManager continues to hold onto LogEvent objects and grows indefinitely if not cleared, or simply keeps the same failing events while the appender fails to recover.

Suggested Fix

The flush() method in AbstractDatabaseManager should ensure that the buffer is cleared even if an exception occurs during the database operations. A try-finally block should be used:
```
public synchronized void flush() {
    if (this.buffer.size() > 0) {
        try {
            this.connectAndStart();
            for (final LogEvent event : this.buffer) {
                this.writeInternal(event, this.layout.toSerializable(event));
            }
            this.commitAndClose();
        } finally {
            // Ensure the buffer is cleared to prevent OOM and duplicate processing attempts
            this.buffer.clear();
        }
    }
}
```

Alternatively, JdbcDatabaseManager.commitAndClose() should catch and log exceptions internally (as implemented in the local CustomJdbcDatabaseManager workaround) to prevent them from interrupting the manager's lifecycle.

dajiaohuang:
Fixes #4318.

`AbstractDatabaseManager#flush()` handed the buffered events to
`writeInternal()` and then ran `commitAndClose()` and `buffer.clear()` in
the same `finally` block:

```java
} finally {
    this.commitAndClose();
    // not sure if this should be done when writing the events failed
    this.buffer.clear();
}
```

When `commitAndClose()` throws (a database failure during the commit/close
phase), the `clear()` was skipped. The events therefore stayed in the
buffer and were written a second time by the following flush, and, because
the buffer keeps accumulating new events while the failure persists, it
grew without bound. `flush()` is also called from `shutdown()`, so a
failing commit at shutdown could also mask the original exception with a
secondary one from re-sending the same events.

The fix moves `buffer.clear()` into an inner `finally` block, so the events
are discarded no matter how `commitAndClose()` returns or throws. The
events were already passed to the database layer, so keeping them cannot
recover the transaction.

`testBufferedEventsAreDiscardedWhenCommitFails` makes `commitAndClose()`
fail once and then succeed, and asserts that `writeInternal()` is invoked
exactly twice in total (once per event), i.e. that the failed transaction
is not re-sent.

Note on scope: `connectAndStart()` is deliberately left outside the `try`,
so events are still retried when the *connection* cannot be established.
Only the commit/close phase changes behaviour here.

## Checklist

* Tests are provided

### DIFF
--- a/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManagerTest.java
+++ b/log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManagerTest.java
@@ -18,10 +18,13 @@ package org.apache.logging.log4j.core.appender.db;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.ArgumentMatchers.same;
 import static org.mockito.BDDMockito.then;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.reset;
 import static org.mockito.Mockito.spy;
@@ -225,6 +228,34 @@ class AbstractDatabaseManagerTest {
         then(manager).should().commitAndClose();
         then(manager).should().shutdownInternal();
         then(manager).shouldHaveNoMoreInteractions();
+    }
+
+    @Test
+    void testBufferedEventsAreDiscardedWhenCommitFails() throws Exception {
+        setUp("name", 10);
+
+        final LogEvent event1 = mock(LogEvent.class);
+        final LogEvent event2 = mock(LogEvent.class);
+
+        when(event1.toImmutable()).thenReturn(mock(LogEvent.class));
+        when(event2.toImmutable()).thenReturn(mock(LogEvent.class));
+
+        manager.startup();
+        manager.write(event1, null);
+        manager.write(event2, null);
+
+        // The first flush fails while committing the transaction, the next one succeeds.
+        doThrow(new DbAppenderLoggingException("Failed to commit the transaction"))
+                .doReturn(true)
+                .when(manager)
+                .commitAndClose();
+
+        assertThrows(DbAppenderLoggingException.class, manager::flush);
+
+        manager.flush();
+
+        // Events of a transaction that failed to commit must not be sent again.
+        verify(manager, times(2)).writeInternal(any(LogEvent.class), isNull());
     }
 
     @Test
--- a/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManager.java
+++ b/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManager.java
@@ -192,9 +192,14 @@ public abstract class AbstractDatabaseManager extends AbstractManager implements
                     this.writeInternal(event, layout != null ? layout.toSerializable(event) : null);
                 }
             } finally {
-                this.commitAndClose();
-                // not sure if this should be done when writing the events failed
-                this.buffer.clear();
+                try {
+                    this.commitAndClose();
+                } finally {
+                    // The events were already handed to the database layer, so they must not be kept
+                    // when committing fails: the next flush would send them again and the buffer
+                    // would grow without bound while the failure persists.
+                    this.buffer.clear();
+                }
             }
         }
     }
--- a/src/changelog/.2.x.x/4318_fix_database_appender_buffer_on_failed_commit.xml
+++ b/src/changelog/.2.x.x/4318_fix_database_appender_buffer_on_failed_commit.xml
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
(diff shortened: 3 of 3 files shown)


log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManagerTest.java
```diff
@@ -227,6 +230,34 @@ void testBuffering04() throws Exception {
...
+        final LogEvent event2 = mock(LogEvent.class);
+
+        when(event1.toImmutable()).thenReturn(mock(LogEvent.class));
+        when(event2.toImmutable()).thenReturn(mock(LogEvent.class));
+
+        manager.startup();
+        manager.write(event1, null);
+        manager.write(event2, null);
+
+        // The first flush fails while committing the transaction, the next one succeeds.
```
ramanathan1504, line 247:
```suggestion
```

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManagerTest.java
```diff
@@ -227,6 +230,34 @@ void testBuffering04() throws Exception {
...
+        doThrow(new DbAppenderLoggingException("Failed to commit the transaction"))
+                .doReturn(true)
+                .when(manager)
+                .commitAndClose();
+
+        assertThrows(DbAppenderLoggingException.class, manager::flush);
+
+        manager.flush();
+
+        // Events of a transaction that failed to commit must not be sent again.
```
ramanathan1504, line 257:
```suggestion
```

log4j-core/src/main/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManager.java
```diff
@@ -192,9 +192,14 @@ public final synchronized void flush() {
...
             } finally {
-                this.commitAndClose();
-                // not sure if this should be done when writing the events failed
-                this.buffer.clear();
+                try {
+                    this.commitAndClose();
+                } finally {
+                    // The events were already handed to the database layer, so they must not be kept
+                    // when committing fails: the next flush would send them again and the buffer
+                    // would grow without bound while the failure persists.
```
ramanathan1504:
```suggestion
```

src/changelog/.2.x.x/4318_fix_database_appender_buffer_on_failed_commit.xml
```diff
@@ -0,0 +1,14 @@
...
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+    <issue id="4318" link="https://github.com/apache/logging-log4j2/issues/4318"/>
+    <description format="asciidoc">
+        Clear the buffer of a buffered database appender when committing a transaction fails.
+        Previously the events were kept, so the next flush sent them again and the buffer grew
+        without bound while the database failure persisted.
```
ramanathan1504: One sentence is enough here. Use "when committing a transaction fails" if the retry on connect stays.

```suggestion
        Clear the buffer of a database appender when connecting or committing fails.
```

ramanathan1504 (changes requested):
@dajiaohuang Inline comments only for special cases.

The commit-failure path is fixed and the new test fails without it. The issue's repro stops the database, which fails in `connectAndStart()` before the `try` in`AbstractDatabaseManager`, so the buffer still grows by one event per log call. Can we clear it there too?

dajiaohuang:
Implemented the requested follow-up in fd234f9: flush now clears the buffer in an outer finally block, so connectAndStart() failures discard pending events as well as write/commit failures. Added a regression test for the failed-connection path and updated the changelog wording. Validation: git diff --check passed. The focused Maven test could not run locally because this checkout requires JDK 17 ([17,18)]; the environment only has JDK 26, where Error Prone fails before the test phase. CI is needed for the full test.

log4j-core-test/src/test/java/org/apache/logging/log4j/core/appender/db/AbstractDatabaseManagerTest.java
```diff
@@ -241,6 +230,61 @@ void testBuffering04() throws Exception {
...
+        doThrow(new DbAppenderLoggingException("Failed to connect"))
+                .doNothing()
+                .when(manager)
+                .connectAndStart();
+
+        assertThrows(DbAppenderLoggingException.class, manager::flush);
+
+        manager.flush();
+
+        // Events must not be retried after a failed connection attempt.
```
ramanathan1504: The test name says this already. Can it go?

```suggestion
```

ramanathan1504 (changes requested):
Three inline comments are still in, one of them new in the connect test, and the changelog description is still three lines. The earlier threads have the detail.

ramanathan1504:
@ppkarwasz  and @vy 
Dropping the batch when `connectAndStart()` fails loses events the database never saw. A short outage used to end with those events written. Should `flush()` cap the buffer instead, so retry stays?

PR #4303, opened 2026-09-05
Fix Pattern Layout's stack trace rendering for non-idempotent `Throwable::getCause`

UsmanEjaz10:
## What & why

Rendering a `Throwable` walks its causal chain in two passes: a metadata pre-computation pass and a render pass. The render pass called `Throwable#getCause()` a *second* time after the metadata pass had keyed the first result in its `IdentityHashMap`. For a `getCause()` override that returns a *different instance on every call* (legal per the `Throwable` API contract), the second lookup misses, and rendering dereferences a null `Metadata`, throwing:

```
java.lang.NullPointerException: Cannot read field "stackLength" because "metadata" is null
    at org.apache.logging.log4j.core.pattern.ThrowableStackTraceRenderer.renderStackTraceElements(...)
```

## Fix

Capture the cause **once** in `Context.Metadata` during the metadata pass and reuse `metadata.cause` at render time — in the base renderer (`%ex`, and `%xEx` which inherits it) and in the inverted renderer (`%rEx`). This is the same treatment the class already applies to `getStackTrace()` (#3940) and `getSuppressed()` (#3929).

`ThrowableExtendedStackTraceRenderer` is intentionally left alone: its independent JAR-enrichment walk null-checks and degrades gracefully, and it is a *legitimate* second `getCause()` caller — so `%xEx` invokes `getCause()` twice by design.

## Tests

New `NonIdempotentGetCauseTest` asserts the fix across all three converters:
- `%ex` / `%rEx`: `getCause()` invoked exactly once per render.
- `%xEx`: invoked exactly twice (its own enrichment walk is the second, legitimate call).
- A non-idempotent `getCause()` renders its cause without throwing.

Kept as a standalone class rather than added to the shared `AbstractStackTraceTest`, which mutates a static `EXCEPTION` concurrently and makes adding test methods there unsafe w.r.t. execution order.

## Verification

- `./mvnw verify` (scoped to `log4j-core`, `log4j-core-test` and upstream modules): **passes** — RAT, Spotless, japicmp all green.
- Tests: 1,398 across the four affected test classes, 0 failures.
- Changelog entry added under `src/changelog/.2.x.x/`.

Fixes #4279

PR #4324, opened 2026-09-17
Keep `Check port label` from leaving stale blocking checks

ramanathan1504:
Every run of `Check port label` leaves its own check on the head commit, and a cancelled or failed one keeps the PR blocked even after a re-run passes.                    
#4303 shows this. On `2575885`:
Re-running the old run cannot recover it, since a re-run replays the original event payload, which carried no labels.


ramanathan1504:
Closed becasued already label `Ci` working as per design

PR #4330, opened 2026-09-17
Use the NPM dependencies from the `gha/v0` branch of `logging-parent`

ppkarwasz:
Companion of apache/logging-parent#506 and apache/logging-parent#507, which centralize the NPM dependencies of all Apache Logging websites on the `gha/v0` branch of `logging-parent`.

- The committed `package.json` shrinks to a single Git dependency on `gha/v0`. Local builds and release website staging keep using it; since `package-lock.json` is not committed, npm resolves the dependency tree on first use.
- Staging and production website deployments pass `npm-dependencies-ref: gha/v0` to `deploy-site-reusable`, so they build with the exact `package.json` and `package-lock.json` from that branch.

No changelog entry, as with previous CI-only changes to this workflow.


package.json
```diff
@@ -1,10 +1,10 @@
 {
+  "//": [
+    "Only local builds and release website staging use this file.",
```
ramanathan1504, line 3:
```suggestion
    "Only local builds, CI site builds and release website staging use this file.",
```

ramanathan1504 (commented):
@ppkarwasz CI builds the site from the new `package.json` and the `apiref` macro still loads.
`src/docgen/apiref-macro.js` requires `handlebars`, which now comes in only through Antora. Can we list it in the `gha/v0` `package.json` next to `fast-xml-parser`?

PR #4335, opened 2026-09-18
Guard %rEx{short.*} against an empty root-cause stack trace

### LINKED ISSUE #4334: `%rEx{short.*}` throws `ArrayIndexOutOfBoundsException` when the root cause has an empty stack trace
`ThrowableInvertedPropertyRendererFactory#extractThrowingMethod` returns `Throwables.getRootCause(throwable).getStackTrace()[0]` without checking the array length, whereas `ThrowablePropertyRendererFactory#extractThrowingMethod` guards against a `null` or empty stack trace and returns `null`. Rendering `%rEx{short.className}`, `%rEx{short.methodName}`, `%rEx{short.lineNumber}` or `%rEx{short.fileName}` for an exception whose root cause has an empty stack trace therefore throws `ArrayIndexOutOfBoundsException`; with the default `ignoreExceptions="true"` the appender drops the event and reports the error to the status logger, with `ignoreExceptions="false"` an `AppenderLoggingException` propagates to the caller. `%ex{short.*}` and the plain `%rEx` render the same event fine.

An empty stack trace is legal per the `Throwable` contract (`setStackTrace(new StackTraceElement[0])`, exceptions created with `writableStackTrace = false`) and is also produced by HotSpot's default `-XX:+OmitStackTraceInFastThrow` for implicit exceptions thrown repeatedly at the same site. The code was introduced with the stack trace rendering rework in 2.25.0 (#2691, #3045) and is identical on `main`. This issue originates from a private security report classified as a bug (**not** a vulnerability).

Fix: mirror the guard of the non-inverted factory in `ThrowableInvertedPropertyRendererFactory`, and add a test rendering all four `short.*` properties with `%rEx` for a wrapped exception whose root cause has an empty stack trace.

kalayciburak:
Fixes #4334.

`%rEx{short.className}` (and `short.methodName` / `short.lineNumber` / `short.fileName`) indexed `getStackTrace()[0]` on the root cause without checking length. An empty stack trace is legal (`setStackTrace(new StackTraceElement[0])`, `writableStackTrace = false`, HotSpot `-XX:+OmitStackTraceInFastThrow`) and threw `ArrayIndexOutOfBoundsException`. `%ex{short.*}` already guards this path.

This change mirrors `ThrowablePropertyRendererFactory`'s empty-stack guard in `ThrowableInvertedPropertyRendererFactory.extractThrowingMethod`.

## Test

Executed:

```
JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 ./mvnw -pl log4j-core-test -am test -Dtest=RootThrowablePatternConverterTest#short_properties_should_not_fail_when_root_cause_has_empty_stack_trace -Dsurefire.failIfNoSpecifiedTests=false
```

RED: 4 errors (`ArrayIndexOutOfBoundsException` at `extractThrowingMethod`). GREEN: 4 tests, 0 failures.

Also executed `RootThrowablePatternConverterTest$PropertyTest` and `ThrowablePatternConverterTest$PropertyTest`: 256 tests, 0 failures.

## Checklist

* Focused tests passed as above
* Changelog entry in `src/changelog/.2.x.x`
* Tests are provided


ppkarwasz (approved):
@kalayciburak,

Looks good to me, thanks!

PR #4336, opened 2026-09-19
Grow DatagramOutputStream buffer and drop oversized UDP payloads

### LINKED ISSUE #4333: `DatagramOutputStream` reallocates and copies its whole buffer on every `write()`
`DatagramOutputStream#copy` allocates a new array of exactly the accumulated length and copies the previous contents on every `write()` call. The encoder writes a formatted event in chunks of `Constants.ENCODER_BYTE_BUFFER_SIZE` (8 KiB), so accumulating an event of n bytes before `flush()` sends the datagram costs O(n²): a benchmark writing a single value in 8 KiB chunks takes 0.15 s for 8 MB, 6 s for 64 MB and 99 s for 256 MB, quadrupling each time the payload doubles.

In practice the effect is bounded: `protocol="UDP"` forces `immediateFlush`, so the accumulator is reset after every event, and `DatagramSocket#send` rejects payloads above 65,507 bytes, so a deliverable event needs at most eight copies. Events larger than that already fail at send time. This is therefore a performance and code-quality bug, not a security issue; it originates from a private security report classified as a bug.

Proposal: back the accumulator with a growable buffer (`ByteArrayOutputStream` or a manually doubled array), and reject an accumulated payload above the maximum datagram size in `flush()` with a `StatusLogger` warning instead of letting `send()` fail.

Reported by @August829

kalayciburak:
Fixes #4333.

`DatagramOutputStream.copy` allocated a new exact-size array and copied the accumulated bytes on every `write()`. The encoder writes events in 8 KiB chunks, so assembling a datagram was O(n²).

This change backs the accumulator with `ByteArrayOutputStream` (amortized growth), stops appending once the payload exceeds 65,507 bytes, and skips `DatagramSocket.send` with a `StatusLogger` warning instead of letting `send` fail.

## Test

Executed:

```
JAVA_HOME=/usr/lib/jvm/java-17-openjdk-amd64 ./mvnw -pl log4j-core-test -am test -Dtest=DatagramOutputStreamTest -Dsurefire.failIfNoSpecifiedTests=false
```

RED: `flushShouldWarnAndDropWhenChunkedPayloadExceedsMaxDatagramSize` failed with `SocketException: Message too long` at `DatagramSocket.send`.

GREEN: 2 tests, 0 failures (also 2/0 on `test-without-jackson-databind` and `test-without-jackson-dataformat-yaml`).

## Checklist

* Focused tests passed as above
* Changelog entry in `src/changelog/.2.x.x`
* Tests are provided


src/changelog/.2.x.x/4333_grow_datagram_output_stream_buffer.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<entry xmlns="https://logging.apache.org/xml/ns"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="
+           https://logging.apache.org/xml/ns
+           https://logging.apache.org/xml/ns/log4j-changelog-0.xsd"
+       type="fixed">
+  <issue id="4333" link="https://github.com/apache/logging-log4j2/issues/4333"/>
```
ramanathan1504, line 8:
```suggestion
  <issue id="4333" link="https://github.com/apache/logging-log4j2/issues/4333"/>
  <issue id="4336" link="https://github.com/apache/logging-log4j2/pull/4336"/>
```

log4j-core/src/main/java/org/apache/logging/log4j/core/net/DatagramOutputStream.java
```diff
@@ -121,12 +134,6 @@ public synchronized void close() throws IOException {
...
     private void copy(final byte[] bytes, final int offset, final int length) {
-        final int index = data == null ? 0 : data.length;
-        final byte[] copy = new byte[length + index];
-        if (data != null) {
-            System.arraycopy(data, 0, copy, 0, data.length);
-        }
-        System.arraycopy(bytes, offset, copy, index, length);
-        data = copy;
+        data.write(bytes, offset, length);
     }
```
ramanathan1504, line 147:
```java
private long droppedLength;

private void copy(final byte[] bytes, final int offset, final int length) {
    if (droppedLength > 0 || data.size() + length > MAX_DATAGRAM_LENGTH) {
        droppedLength = (droppedLength > 0 ? droppedLength : data.size()) + length;
        return;
    }
    data.write(bytes, offset, length);
}
```
and in `flush()`:
```java
if (droppedLength > 0) {
    LOGGER.warn(
            "UDP datagram of {} bytes exceeds the maximum of {} bytes and will be dropped",
            droppedLength,
            MAX_DATAGRAM_LENGTH);
} else {
    datagramSocket.send(new DatagramPacket(data.toByteArray(), data.size(), inetAddress, port));
}
...
} finally {
    data.reset();
    droppedLength = 0;
    ...
```
- The guard in `flush()` becomes `(data.size() > 0 || droppedLength > 0)`.
- The existing test keeps passing as is (73,728 bytes → WARN containing "exceeds", nothing received).

ramanathan1504 (changes requested):
The `ByteArrayOutputStream` swap removes the copy on every write, and the drop test covers the `Message too long` case.
In `DatagramOutputStream`, `copy` still buffers the whole event before `flush` checks the size, and `reset()` keeps that grown buffer until close. Can `copy` stop appending once `data.size()` passes `MAX_DATAGRAM_LENGTH` and let `flush` warn and drop?

src/changelog/.2.x.x/4333_grow_datagram_output_stream_buffer.xml
kalayciburak: added the PR link

log4j-core/src/main/java/org/apache/logging/log4j/core/net/DatagramOutputStream.java
kalayciburak: yep, copy stops once we pass MAX_DATAGRAM_LENGTH

ramanathan1504 (commented):
Removing `droppedLength = 0` from `flush` still passes both tests. Can `DatagramOutputStreamTest` send a small event after a dropped one and check it arrives?

Note: Please Don't do force push. Thanks!

kalayciburak:
added, small event after the drop arrives

ramanathan1504:
@kalayciburak the single-write case is still open: one `write` over 65,507 bytes, close without `flush`, check the WARN.

kalayciburak:
added, one write over 65507 then close without flush

ramanathan1504 (approved):
@kalayciburak LGTM!

